PR #11, opened 2020-06-23
Refactor to use pointer instead of reference in memory instance for calling host functions.

q82419:


hydai:
Shall we print an error message when we want to access a null memory instance in the WASI functions?

PR #54, opened 2020-08-24
[AOT] Use mprotect for memory boundary checking

ibmibmibm:
* Use mmap for memory allocation
* Remove Ctor
* Add dummy check for DivideByZero check
* Add check for floating point truncation
* Mangling names for outstanding import/export names
* Register signal handler before entering compiled function to handle
  errors in compiled function
* Move exra `ctx` argument for all compiled function to static variable
  `This`
* Set global variable in compiled module to extern constant for
  optimization, and change back after optimization done
* Validate module before compiling

### DIFF
--- a/include/common/ast/module.h
+++ b/include/common/ast/module.h
@@ -17,6 +17,8 @@
 #include "section.h"
 
 #include <memory>
+#include <string>
+#include <string_view>
 #include <vector>
 
 namespace SSVM::Interpreter {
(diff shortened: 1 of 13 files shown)


include/common/ast/module.h
```diff
@@ -57,19 +59,66 @@ class Module : public Base {
...
+    using namespace std::literals::string_view_literals;
+    const auto NPos = std::string_view::npos;
+    const auto Esc = "$#@\0"sv;
+    std::string Result(1, '$');
+    size_t Cursor = 0;
+    for (size_t Pos = Name.find_first_of(Esc); Pos != NPos;
+         Cursor = Pos + 1, Pos = Name.find_first_of(Esc, Cursor)) {
+      Result += Name.substr(Cursor, Pos - Cursor);
+      switch (Name[Pos]) {
+      case '$':
```
hydai, line 100: Maybe we should have a table or document to record the mapping relations?

include/interpreter/interpreter.h
```diff
@@ -365,16 +370,18 @@ class Interpreter {
...
-  static void trapProxy(Interpreter *This, uint32_t Status);
-  static void callProxy(Interpreter *This, const uint32_t FuncIndex,
-                        const ValVariant *Args, ValVariant *Rets);
-  static uint32_t memGrowProxy(Interpreter *This, const uint32_t NewSize);
-  static uint32_t memSizeProxy(Interpreter *This);
+  /// Pointer to current object.
+  static Interpreter *This;
+  /// jmp_buf for trap.
+  static sigjmp_buf *TrapJump;
+  static uint32_t trapCodeProxy;
```
hydai, line 380: Should it be `TrapCodeProxy`?

include/runtime/instance/memory.h
```diff
@@ -35,8 +36,15 @@ class MemoryInstance {
   MemoryInstance() = delete;
   MemoryInstance(const AST::Limit &Lim)
       : HasMaxPage(Lim.hasMax()), MinPage(Lim.getMin()), MaxPage(Lim.getMax()),
-        CurrPage(Lim.getMin()), Data(CurrPage * kPageSize) {}
-  virtual ~MemoryInstance() noexcept = default;
+        CurrPage(Lim.getMin()) {
+    uint8_t *const Pages = static_cast<uint8_t *>(
+        mmap(nullptr, UINT64_C(0x200000000), PROT_READ | PROT_WRITE,
```
hydai, line 41: Should we have a constant value to represent `0x200000000`?

hydai (changes requested):

include/common/ast/module.h
ibmibmibm: Done.

include/interpreter/interpreter.h
ibmibmibm: Done.

include/runtime/instance/memory.h
ibmibmibm: Done.

hydai (approved):
LGTM

PR #67, opened 2020-09-23
[Runtime] Allocate on request, don't preserve full 4G memory at once.

ibmibmibm:
On low memory platform, overcommitting memory is decline by kernel.
By parsing smaps, we find a usable memory region and request exact
address and not overcomitting memory.


hydai (approved):
LGTM

PR #81, opened 2020-10-19
Merge program counter into stack and add back the End instruction.

q82419:
1. Merged `InstrProvider` into `StackManager`.
2. Added back the `End` instruction at the end of expressions.
3. Refine the validator workflow.


include/runtime/stackmgr.h
```diff
@@ -98,20 +101,19 @@ class StackManager {
   }
 
   /// Unsafe pop top frame. Return number of popped label.
```
hydai, line 103: Your comment doesn't fit the code.

include/runtime/stackmgr.h
```diff
@@ -127,6 +129,24 @@ class StackManager {
   /// Unsafe leave top label.
   void leaveLabel() { LabelStack.pop_back(); }
 
+  /// Unsafe getter of next instruction to execute.
+  const AST::Instruction *getNextInstr() {
+    if (LabelStack.size() == 0) {
+      return nullptr;
+    }
+    auto &L = LabelStack.back();
+    return (*(L.Curr++)).get();
```
hydai, line 138: You use `++` without checking the boundary.

hydai (changes requested):

include/runtime/stackmgr.h
q82419: This is unsafe version, which bases on there's always an End instruction before the end of instruction vector.

hydai (approved):
LGTM

PR #84, opened 2020-10-20
AOT for reference types and bulk memory operations

ibmibmibm:


q82419:
This may not be WASM 1.1.
The reference types and bulk memory operations proposals are not merged into WASM official spec yet.
I think it will be WASM 1.2 or others in the future.

ibmibmibm:
> 
> 
> This may not be WASM 1.1.
> The reference types and bulk memory operations proposals are not merged into WASM official spec yet.
> I think it will be WASM 1.2 or others in the future.

Ok, I'll change commit messages

lib/interpreter/instantiate/function.cpp
```diff
@@ -23,6 +23,11 @@ Expect<void> Interpreter::instantiate(
     auto NewFuncInst = std::make_unique<Runtime::Instance::FunctionInstance>(
         ModInst.Addr, *FuncType, CodeSegs[I]->getLocals(),
         CodeSegs[I]->getInstrs());
+    NewFuncInst->setSymbol(CodeSegs[I]->getSymbol());
```
hydai, line 26: I have no idea about this line and the following lines.
It looks like 26 & (28-30) are doing the same stuff but the above one doesn't check the Symbol before calling the `setSymbol` function

lib/aot/compiler.cpp
```diff
@@ -163,12 +159,7 @@ struct SSVM::AOT::Compiler::CompileContext {
         Trap(llvm::Function::Create(
             llvm::FunctionType::get(VoidTy, {Int32Ty}, false),
             llvm::Function::PrivateLinkage, "trap", LLModule)),
-        Likely(llvm::MDTuple::getDistinct(
-            LLContext, {llvm::MDString::get(LLContext, "branch_weights"),
-                        llvm::ConstantAsMetadata::get(llvm::ConstantInt::get(
-                            LLContext, llvm::APInt(32, 2000))),
-                        llvm::ConstantAsMetadata::get(llvm::ConstantInt::get(
-                            LLContext, llvm::APInt(32, 0)))})) {
+        Likely(llvm::MDBuilder(LLContext).createBranchWeights(2000, 0)) {
```
hydai, line 162: Why we need to set 2000 here? Is this a magic number?

hydai (changes requested):

lib/aot/compiler.cpp
ibmibmibm: It's a magic number at https://github.com/llvm/llvm-project/blob/991e86156c88dfa64c908fe36da3a1945c3d2e50/llvm/lib/Transforms/Scalar/LowerExpectIntrinsic.cpp#L50
Looks like llvm have expect Intrinsic now, I'll change this to an intrinsic call.

lib/interpreter/instantiate/function.cpp
ibmibmibm: Looks like I forgot to remove this line.

hydai (approved):
LGTM

PR #87, opened 2020-10-21
Refactor support into common.

q82419:
1. Move the `include/support` to `include/common`.
2. Move the `include/ast` to the top include folder.
3. Merge some highly related headers.
4. Fix the comments.


include/common/errcode.h
```diff
@@ -12,8 +12,9 @@
 #pragma once
 
 #include "errinfo.h"
-#include "support/expected.h"
```
hydai, line 15: You should keep the `expected.h`.

hydai (changes requested):

hydai (approved):
LGTM

PR #93, opened 2020-10-24
[AOT] Generate reentrant code by adding an execution context parameter

ibmibmibm:
* Remove unused symbols
* Add linker option to share intrinsics table with compiled object through dynamic linker
* Add trap intrinsic


hydai (approved):
LGTM

PR #98, opened 2020-11-05
[Doc] Update README and Changelog for 0.7.1 release.

q82419:


lib/vm/CMakeLists.txt
```diff
@@ -14,13 +14,6 @@ set(ssvmLibs
   ssvmHostModuleSSVMProcess
 )
 
-if (NOT SSVM_DISABLE_AOT_RUNTIME)
```
hydai, line 17: Why do you remove this?
We still need a way to disable aot build.

hydai (changes requested):

lib/vm/CMakeLists.txt
hydai: Since `vm` package is not related to aot component, we can remove this unused condition.

hydai (approved):
LGTM

PR #101, opened 2020-11-27
[AOT] Add support for WebAssembly SIMD proposal

ibmibmibm:
* Add uint128_t to ValVariant for V128
* Add RefVariant for 64bit-width reference variant
* Add formal check for SIMD
* Add dummy SIMD function for interpreter
* Enable SIMD testcase
* Support v128 type in json test file


hydai (approved):
LGTM

PR #130, opened 2020-12-29
[AOT] Rename DLHandle to SharedLibrary and add support for for cross platform

ibmibmibm:


lib/loader/shared_library.cpp
```diff
@@ -0,0 +1,53 @@
...
+#error Unsupported os!
+#endif
+
+namespace SSVM {
+namespace Loader {
+
+/// Open so file. See "include/loader/shared_library.h".
+Expect<void> SharedLibrary::load(const std::filesystem::path &Path) noexcept {
+#if SSVM_OS_WINDOWS
+  Handle = boost::winapi::load_library_ex(Path.c_str(), 0, mode);
```
hydai, line 20: Where does the `mode` come from?

hydai (changes requested):

hydai (approved):

PR #133, opened 2020-12-31
[AOT] Prevent data race across different VM instance

ibmibmibm:
*VM execution is not fully thread-safe yet*.
* Make static pointer `This` as thread_local.
* Move all other static global to private member.
* Register signal handler on entering a compiled function.
* Use a merge-sort like algorithm for finding a usable memory region.
* Change AOT core test to a multi-thread version with different VM instance.


q82419:
Not need and close.

PR #194, opened 2021-05-07
[Misc] Move mmap calls to separate library

ibmibmibm:


q82419 (changes requested):
I think the `system/allocator.h` can be moved into `common/allocator.h`.
And the `system/config.h.in` can be merged into `common/config.h.in`.
And the `wasmedgeSystem` can be merged into `wasmedgeCommon`.

ibmibmibm:
> 
> 
> I think the `system/allocator.h` can be moved into `common/allocator.h`.
> And the `system/config.h.in` can be merged into `common/config.h.in`.
> And the `wasmedgeSystem` can be merged into `wasmedgeCommon`.
I don't think I should, because these class is for different operating system environment, and the header file is for private use only.

hydai (dismissed):
I think the allocator should have a standalone component `System`. This is not related to the `Common` component.

hydai (approved):

PR #204, opened 2021-05-14
[Misc] Update manylinux* dockerfiles for new gcc and llvm

ibmibmibm:


hydai:
@ibmibmibm Maybe we should add some ENV to export the `PATH`, `CC`, `CXX`, and `CXXFLAGS`?

hydai (approved):

hydai (dismissed):

hydai (changes requested):
I think the `build-manylinux.sh` should not be put in the docker image. Because it will be changed.
So the better way is using the file under `utils/docker/build-manylinux.sh` instead.

hydai (approved):
💯

PR #211, opened 2021-05-19
Minor fixes

ibmibmibm:
* Adjust library dependency
* Check for newer compilers in std::filesystem
* Add missing include files, fix typo
* Avoid helper function to prevent zero and pointer ambiguity
* Use u8path for path creation
* Run clang-format-11
* Remove extra semicolons
* Add unlisted enum entries in switch statements
* Use stof on float, stol on uint32_t

### DIFF
--- a/include/experimental/expected.hpp
+++ b/include/experimental/expected.hpp
@@ -1541,13 +1541,13 @@ public:
   constexpr const_lvalue_reference_type operator*() const & {
     return impl_base::val();
   }
-  constexpr lvalue_reference_type operator*() & { return impl_base::val(); };
+  constexpr lvalue_reference_type operator*() & { return impl_base::val(); }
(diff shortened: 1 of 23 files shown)


include/experimental/expected.hpp
```diff
@@ -1544,10 +1544,10 @@ class expected : public detail::expected_move_assign_base<T, E>,
   constexpr lvalue_reference_type operator*() & { return impl_base::val(); };
```
hydai, line 1544: Looks like this line should also remove the `;` too.

hydai (changes requested):

include/experimental/expected.hpp
ibmibmibm: done

hydai (approved):

q82419 (approved):

PR #277, opened 2021-06-28
[Doc] add DCO tips

alabulei1:


hydai:
@alabulei1 Seems like your branch is not clean? I saw several commits which are not related to the document also in this PR.

alabulei1:
> @alabulei1 Seems like your branch is not clean? I saw several commits which are not related to the document also in this PR.

I'm not sure. I will check out soon.

PR #313, opened 2021-07-26
[Rust] Add wasmedge-sys crate

nhynes:
This PR adds a Rust `-sys` (binding) crate for eventual consumption by an idiomatic Rust lib.

Still TODO:
- [x] more robust search for `libwasmedge_c.so`. Right now the crate assumes that it's part of the source tree and hence cannot be published on crates.io. That's not terrible for now, but it's not ideal, and there's no rush.


.github/workflows/rust.yml
```diff
@@ -0,0 +1,48 @@
+name: rust
+
+concurrency:
+  group: build-${{ github.head_ref }}
```
hydai, line 4: The group should be the same as workflow name. Use `rust-${{ github.head_ref }}` instead.

hydai (changes requested):

.rustfmt.toml
```diff
@@ -0,0 +1,73 @@
+max_width = 100
```
nhynes, line 1:
```patch
diff <(rustfmt --print-config default) .rustfmt.toml
19d18
< license_template_path = ""
21c20
< format_macro_matchers = false
---
> format_macro_matchers = true
29c28
< imports_granularity = "Preserve"
---
> imports_granularity = "Crate"
55c54
< edition = "2015"
---
> edition = "2018"
64d62
< required_version = "1.4.37"
```

wasmedge-sys/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "wasmedge-rs"
```
yanganto, line 2: @nhynes 
Maybe it is good to use `wasmedge-sys`?
The crate `wasmedge-rs` is occupied by trademark squatter, not by us.

yanganto (changes requested):

wasmedge-sys/Cargo.toml
nhynes: it's supposed to be that. good catch.

> occupied by trademark squatter, not by us.
yeah. 👎  to squatsmer

yanganto (dismissed):

hydai (dismissed):

hydai:
Hi @nhynes 
I found this error in wasi-test:

```bash
test 13
      Start 13: wasi-test

13: Test command: /bin/bash "-c" "[ -e "$HOME/.cargo/env" ] && . "$HOME/.cargo/env"; "/tmp/WasmEdge"/utils/wasi-test/run-wasi-test.sh "/t
mp/WasmEdge/build/tools/wasmedge""
13: Test timeout computed to be: 1500
13: path = /tmp/WasmEdge/build/tools/wasmedge
13: Cloning git repo...
13: Cloning into 'wasi-test'...
13: Applying patch...
13: Building wasm files...
13: error: current package believes it's in a workspace when it's not:
13: current:   /tmp/WasmEdge/build/test/host/wasi/wasi-test/Cargo.toml
13: workspace: /tmp/WasmEdge/Cargo.toml
13:
13: this may be fixable by adding `build/test/host/wasi/wasi-test` to the `workspace.members` array of the manifest located at: /tmp/Wasm
Edge/Cargo.toml
13: Alternatively, to keep it out of the workspace, add the package to the `workspace.exclude` array, or add an empty `[workspace]` table
 to the package's manifest.
13: removing git repo
13/18 Test #13: wasi-test ........................***Failed    1.08 sec
```

Seems like we have a Cargo.toml in the root folder will conflict with the wasi-test. The detail command is here: https://github.com/WasmEdge/WasmEdge/blob/master/test/host/wasi/CMakeLists.txt#L19


You can use the following commands to reproduce this error:

```bash
cmake -Bbuild -GNinja -DCMAKE_BUILD_TYPE=Debug -DWASMEDGE_BUILD_TESTS=ON -DWASMEDGE_BUILD_COVERAGE=ON .
cmake --build build
LD_LIBRARY_PATH=$(pwd)/build/lib/api ctest -VV
```

hydai (approved):

hydai:
This is awesome! Thank you @nhynes

PR #352, opened 2021-08-12
Add conf-header.py and fix wasmedge.h.in

### LINKED ISSUE #349: Feat: Create a checker/helper tool to sync the `errcode.h` and `wasmedge.h.in`
## Motivation

See the discussion in #341 , https://github.com/WasmEdge/WasmEdge/pull/341#discussion_r685191280

Currently, the `wasmedge.h.in` is synced with `errcode.h` manually. This is not good, we should avoid relying on human to sync code.

## Expected

Create a checker, a helper tool, or split the `wasmedge.h.in` to multiple and sync from `errcode.h` directly.

chenyukang:
It's a little trivial than I thought. 
Workout a hacky version of Python script to check the enums. 

Mark format with:
```
Gen Based: include/common/errcode.h
```

The script will try to update the enum values according to headers.

close #349

### DIFF
--- a/.github/scripts/conf-header.py
+++ b/.github/scripts/conf-header.py
@@ -1,0 +1,106 @@
+import re
+import fnmatch
+import os
+import glob
+import sys
+
+
+def clean_list(l):
+    l = [x.strip() for x in l]
+    l = [x for x in l if x != ""]
+    return l
+
+def enum_line(line):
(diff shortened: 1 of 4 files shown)


q82419:
This file was refactored and the error code enum definitions are put in the same place.

PR #380, opened 2021-09-06
[Docs] Update install instructions

juntao:
Use the new install script.


juntao:
Will submit a cleaner one.

Issue #382, opened 2021-09-06
[Install script] allow non-root users to install


juntao:
The script currently requires root or sudo to run. If the user does not have root privileges, the script should install everything without running the `ldconfig`

At the end, the script should ask the user to run those commands manually. 

*Even if the user does not do that, `wasmedge` and maybe even `wasmedgec` should still work.*

juntao:
Actually, since we are installing the manylinux build in the script, we do NOT need to install the dependency packages via APT. 

Let's remove that section from the script altogether.

Issue #381, opened 2021-09-06
[Install script] Install into a user local directory by default


juntao:
If the user does NOT specify a `-p` on the command line, the script should install into `$HOME/.wasmedge` and then set the PATH to include `$HOME/.wasmedge/bin`

PR #387, opened 2021-09-07
Install into User's Home by default #381 

SAtacker:
* Checks if script is being run with `root` permissions
* Installs into current users home directory
* When run by `root` Checks if `sudo` is present and installs into current user's home directory
* If `sudo` is absent it does not manipulate $HOME

Ref - #381

### DIFF
--- a/utils/install.sh
+++ b/utils/install.sh
@@ -42,9 +42,21 @@ VERSION_TF_TOOLS=$(git -c 'versionsort.suffix=-' \
     tail --lines=1 |
     cut --delimiter='/' --fields=3)
 
-IPATH="/usr/local"
+IPATH="$HOME/.wasmedge"
 EXT="none"
 VERBOSE=0
+
+ENV="#!/bin/sh
+# wasmedge shell setup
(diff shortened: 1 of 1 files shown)


SAtacker:
Let me pull some changes over this too

SAtacker:
Should now be able to fix #382

juntao:
I am seeing this error when running `./install.sh -e all`

```
Fetching WasmEdge-tensorflow-tools-0.8.2
WasmEdge-tensorflow 100%[===================>]  28.20M  46.0MB/s    in 0.6s
/root/.wasmedge/bin/wasmedge-tensorflow: error while loading shared libraries: libtensorflow.so.2: cannot open shared object file: No such file or directory
./install.sh: line 214: [: =: unary operator expected
version 0.8.2 does not match  for wasmedge-tensorflow-0.8.2
```

SAtacker:
Thanks for this @juntao . I just needed to add `$IPATH/lib` to `ldconfig`

juntao (approved):
I tested and it works now. Thanks.

Issue #390, opened 2021-09-08
[Install] Provide an one-line install script


juntao:
Allow the install script to be invoked in one line. Like Rust's install process.

SAtacker:
Does this mean that I'll have to remove `bashisms` from this script so that it is compatible with `shell`?

juntao:
I guess the one liner could pipe to bash?

curl url | bash

Something like that?

SAtacker:
Yes that would work with bash

SAtacker:
Where will this oneliner be hosted ?

juntao:
On GitHub. You can our current instruction here. We will just combine the 3 lines into one:

https://github.com/WasmEdge/WasmEdge/blob/master/docs/install.md

SAtacker:
So, I'll change it to `wget -qO- https://raw.githubusercontent.com/WasmEdge/WasmEdge/master/utils/install.sh | bash`

Issue #384, opened 2021-09-06
[Install script] Remove APT dependencies


juntao:
Since we are installing the manylinux build in the script, we do NOT need to install the dependency packages via APT.

Let's remove that section from the script altogether.

SAtacker:
In #381 
I have added section like
```bash
_pkg_mgr() {
    if [ $PERM_ROOT == 1 ]; then
        apt "$@"
    fi
}
```
1. Here we can add the script to detect the package manager as well. This will be difficult to maintain as we'll have to look out for every package manager out there
2. Or we can check if `apt` is available and go with it , else do nothing

juntao:
My suggestion is to remove APT installs altogether. The install script uses the manylinux build, which should not have those external dependencies. Please try it. 

This way, we will not depend on ANY linux package manager.

SAtacker:
How does the below flow for the install script sound?
1. If `git` is not available --> default the version strings manually for now, 
(This will need changes on every new release)
2. If `wget` is not available, the script will check for `curl`  and if both are not available it'll display a message to have one of those and exit.

If there is some automation already done which I am unaware of (I have checked the GitHub actions) w.r.t. releases without `git` binary please let me know.

juntao:
Let's assume that the user must have `git` and `wget` or `curl` installed in order to use the script. If they do not, just print a message asking the user to install and quit. Thanks.

juntao:
Let's do #390 as well. Thanks.

PR #391, opened 2021-09-08
[Install Script] Improve Install script

SAtacker:
- [X] Remove `apt`/`deb` deps #384 
  - Checks for `git` and if not available then exits.
  - Checks for `wget`, if not then checks for `curl`, if not then exits.
- [X] #390

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -1,6 +1,7 @@
 # Install WasmEdge
 
 * [Quickstart](#quickstart)
+* [Install for all users](#install-for-all-users)
 * [What's installed](#whats-installed)
 * [Install WasmEdge for Node.js](#install-wasmedge-for-nodejs)
 * [Install a Docker / k8s compatible WasmEdge "container"](#install-a-docker--k8s-compatible-wasmedge-container)
@@ -53,9 +54,6 @@ To uninstall WasmEdge, you can run the following command.
 ```
 $ wget -qO- https://raw.githubusercontent.com/WasmEdge/WasmEdge/master/utils/uninstall.sh | bash
 ```
-
-It will remove WasmEdge files, and also print out dependency Tensorflow and image shared libraries for you to manually remove.
-
 
 ## Install WasmEdge for Node.js
 
--- a/docs/install.md
+++ b/docs/install.md
@@ -10,16 +10,16 @@
 The easiest way to install WasmEdge is to run the following commands.
 
 ```
-$ wget https://raw.githubusercontent.com/WasmEdge/WasmEdge/master/utils/install.sh
-$ chmod +x ./install.sh
-$ sudo ./install.sh -p /usr/local
+$ wget -qO- https://raw.githubusercontent.com/WasmEdge/WasmEdge/master/utils/install.sh | bash
 ```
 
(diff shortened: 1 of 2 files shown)


SAtacker:
For quick testing of my branch you can use
```sh
wget -qO- https://raw.githubusercontent.com/SAtacker/WasmEdge/remove_apt_deps/utils/install.sh | bash -s -- -e all
```
and then `source $HOME/.wasmedge/env`

juntao:
The Rust installer adds the following line to the end of `$HOME/.profile` so that the user does not have to source the `env` every time he logs in. We should do that too.

```
. "$HOME/.cargo/env"
```

SAtacker:
> The Rust installer adds the following line to the end of `$HOME/.profile` so that the user does not have to source the `env` every time he logs in. We should do that too.
> 
> ```
> . "$HOME/.cargo/env"
> ```

Actually, IMO, we should let the user handle this because sometimes they may not need those in their path. For example, in the future, if we have some library in LD_LIBRARY_PATH that is analogous to one which they already have, it'll mess with their program as every time they open a new terminal it'll still be sourced and linked to ours. `rust` might do that as their is no `lib` inside ~/.cargo/
Do we need to do this?

juntao:
Hmm, that is a good point. However, I do not think we can ask the user to remember to source the env every time they open a new terminal.

When we install in `/usr/local`, we do `ldconfig`, which also has the potential to introduce potential conflicts to existing libraries. Is that correct?

I think this is something we may need to explain in the `install.md` docs. But by default, I think we should run `env` for the user in a new terminal.

SAtacker:
> Hmm, that is a good point. However, I do not think we can ask the user to remember to source the env every time they open a new terminal.
> 
> When we install in `/usr/local`, we do `ldconfig`, which also has the potential to introduce potential conflicts to existing libraries. Is that correct?
> 
> I think this is something we may need to explain in the `install.md` docs. But by default, I think we should run `env` for the user in a new terminal.

Yes. Thanks. I'll make the change.

juntao:
I think we still need to ask the user to do `source $HOME/.wasmedge/env` immediately after the install. The install script runs in its own process. when it exits, the terminal does not retain the path settings.

Also, I'd like to put the `. "$HOME/.wasmedge/env"` line in `.profile` as it is a sh script not a bash specific script. Thanks.

SAtacker:
> I think we still need to ask the user to do `source $HOME/.wasmedge/env` immediately after the install. The install script runs in its own process. when it exits, the terminal does not retain the path settings.
> 
> Also, I'd like to put the `. "$HOME/.wasmedge/env"` line in `.profile` as it is a sh script not a bash specific script. Thanks.

Yes, I'll default it to `.profile` and the latter part too. Thanks

juntao:
Can you fix the DCO? 

https://github.com/WasmEdge/WasmEdge/runs/3552835039

SAtacker:
In case somebody wants to know what I did to fix a commit message 6 commits before then
`git rebase -i HEAD~6` and then `reword` the commit that led to DCO to fail. 
And then I rebased to master to update it. Should have done that before.

juntao (approved):
This is great. Thank you!

Issue #389, opened 2021-09-08
[Install] Maintain an uninstall list


juntao:
I think a better uninstall approach is to create an `UNINSTALL.txt` file in the `bin` directory. When we run the `uninstall.sh` script it will

* Locate the `bin` directory by running `which wasmedge`
* Remove all files in the `UNINSTALL.txt` list

PR #402, opened 2021-09-11
[Utils] Uninstall script improvements

SAtacker:
- [X] Detects `wasmedge` as per #389 
- [x] Uninstall.txt

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -49,11 +49,38 @@ After installation, you have the following directories and files. Here we assume
 * The `$HOME/.wasmedge/lib` directory contains WasmEdge shared libraries, as well as dependency libraries. They are useful for WasmEdge SDKs to launch WasmEdge programs and functions from host applications.
 * The `$HOME/.wasmedge/include` directory contains the WasmEdge header files. They are useful for WasmEdge SDKs.
 
+
+## Uninstallation
+
 To uninstall WasmEdge, you can run the following command.
 
+Note: Below are [`process substitutions`](https://unix.stackexchange.com/questions/17107/process-substitution-and-pipe) and not pipes.
+
 ```
-$ wget -qO- https://raw.githubusercontent.com/WasmEdge/WasmEdge/master/utils/uninstall.sh | bash
+bash <(wget -qO- https://raw.githubusercontent.com/WasmEdge/WasmEdge/master/utils/uninstall.sh)
 ```
+
+If `wasmedge` binary is not in `PATH` or if it wasn't installed in default `$HOME/.wasmedge` then you must provide the installation path.
+
+```bash
+bash <(wget -qO- https://raw.githubusercontent.com/WasmEdge/WasmEdge/master/utils/uninstall.sh) -p /path/to/parent/folder
+```
+
+If you wish to uninstall uninteractively, you can pass in the `--quick` or `-q` flag.
+```bash
+bash <(wget -qO- https://raw.githubusercontent.com/WasmEdge/WasmEdge/master/utils/uninstall.sh) -q
+```
+
+<details>
+  <summary>More about Uninstall Script:</summary>
+<li>
+It automatically detects what files were installed post commit 19f44012d989527c7d018090f4e9f779c674e494.
+</li>
+<li>
+If the parent folder contains substring ".wasmedge" the folder will be considered for removal which happens to be same as the default location for wasmedge under $HOME/.wasmedge
+</li>
+</details>
+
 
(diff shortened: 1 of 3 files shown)


SAtacker:
Using `UNINSTALL.txt` for tracking installation files will be a workaround. IMO it's too much to rely on the bin directory and assume there will be an `UNINSTALL.txt` present. And in the case of `/usr/local` installations, it'll be a bit non-standard to do so.
Currently, it locates the `wasmedge` bin as you mentioned but I am feeling a bit odd about creating an `UNINSTALL.txt`

juntao:
I am open to suggestions. :) Can you research how other projects handle uninstall?

SAtacker:
> I am open to suggestions. :) Can you research how other projects handle uninstall?

Homebrew does this by making a list inside `/tmp/uninstall.PID` and then asks if they want to do so as mentioned at https://github.com/WasmEdge/WasmEdge/issues/412#issuecomment-919465690.

Currently, this uninstall script detects if `wasmedge` binaries are found on `PATH` and if not it looks for `$HOME/.wasmedge` if the binary is found there then it detects that as the installation path and if not, exits.
Further, all the files that we need to uninstall as per extensions are hardcoded into the script.

juntao:
The hardcoded uninstall list will not work once we have several versions of WasmEdge. So, I guess we do need to maintain an UNINSTALL list.

SAtacker:
> The hardcoded uninstall list will not work once we have several versions of WasmEdge. So, I guess we do need to maintain an UNINSTALL list. 

Yes, I agree with it. But where do we keep this list? Are you suggesting that we modify the `install.sh` to output a list of whatever files are installed into some standard directory like `$HOME/. wasmedge` and a safe fallback in case such a list is deleted unintentionally by the user?

juntao:
Perhaps we can keep a list of files as comments in the `env` file? You can always find the file at `../env` from the wasmedge executable. The uninstall script can parse the `env` file's comments section and decide which files to remove.

If the `env` file is not available, the uninstall script should print an error message and exit.

SAtacker:
Hi, am back here. Will try to complete this today itself.

SAtacker:
Install
```bash
wget -qO- https://raw.githubusercontent.com/SAtacker/WasmEdge/uninstall_script/utils/install.sh | bash -s -- -e all
```
Uninstall
```bash
wget -qO- https://raw.githubusercontent.com/SAtacker/WasmEdge/uninstall_script/utils/uninstall.sh | bash -s -- -q
```
Without the `-q` it runs infinitely, a bit of an issue right now

SAtacker:
Install
```bash
wget -qO- https://raw.githubusercontent.com/SAtacker/WasmEdge/uninstall_script/utils/install.sh | bash -s -- -e all
```
Uninstall Non-Interactive
```bash
wget -qO- https://raw.githubusercontent.com/SAtacker/WasmEdge/uninstall_script/utils/uninstall.sh | bash -s -- -q
```
Uninstall Interactively
```bash
bash <(wget -qO- https://raw.githubusercontent.com/SAtacker/WasmEdge/uninstall_script/utils/uninstall.sh)
```

juntao:
I think the uninstall script should just remove the files in `env`. There is no need to attempt to remove the parent directories unless they are empty. It is okay to leave empty directories. For example, currently, the script removes other files from `/usr/local` when I install there.

```
Removing /usr/local/env
Removing /usr/local/games
Removing /usr/local/share/man
Removing /usr/local/share/ca-certificates
Removing /usr/local/src
Removing /usr/local/lib
Removing /usr/local/bin
Removing /usr/local/include
Removing /usr/local/etc
```

SAtacker:
How about I do that only for default path `$HOME/.wasmedge` ?

SAtacker:
Could you please review @juntao ?

juntao (dismissed):
Looks good

juntao (dismissed):
I made some changes to the docs

hydai:
@juntao your commit message should have a tag such as `[Docs] ` when you modified the documents. Just like @SAtacker has put the `[Utils] ` in his commit messages.

juntao:
Done

> @juntao your commit message should have a tag such as `[Docs] ` when you modified the documents. Just like @SAtacker has put the `[Utils] ` in his commit messages.

hydai:
CleanShot 2021-09-30 at 11 23 51@2x

Seems like not right. You do not modify your origin one, however, you add two more commits.

juntao:
Hmm, how about now?

Issue #399, opened 2021-09-11
[Install] Support ARM64 based systems


juntao:
## Motivation

Support ARM64 computers

## Details

The `install.sh` script should detect whether the underlying CPU architecture is x86 or ARM64. It should install the following two releases respectively.

* WasmEdge-0.8.2-manylinux2014_x86_64.tar.gz
* WasmEdge-0.8.2-manylinux2014_aarch64.tar.gz

However, the Tensorflow tools do not yet have an aarch64 release. So, if the user wants to do `-e all` on aarch64, the script should give an error and stop.

The script should give an error and stop on all other platforms (e.g., M1) -- until we have release builds for those platforms.

PR #403, opened 2021-09-11
[Utils] Install Script improvement

SAtacker:
Addressed issue #399 

* More modular script
* Currently supported platforms are Linux-x86_64 and Linux-aarch64
  * Don't install extensions for aarch64
  * Exit totally for other arches
* Modified the part that gets the latest versions to use GitHub API

* Minor: 
  * Removed suffix from IPATH while checking if specified install the directory is in the path


SAtacker:
Test using
```bash
wget -qO- https://raw.githubusercontent.com/SAtacker/WasmEdge/script_improvement/utils/install.sh | bash -s -- -e all 
```

juntao (approved):
I tested. This looks good. Thanks!

PR #408, opened 2021-09-13
[Proposal] WASI-crypto Proposal

sonder-joker:
This is a long-term pull request that is still working, and will be constantly repair and commit for review. I think making this process public in early will work better.


utils/wasi-cpp-header/crypto_generate.sh
```diff
@@ -0,0 +1,15 @@
...
+ROOT="$(realpath $(dirname "${BASH_SOURCE[0]}")/../../)"
+cargo build --release
+
+for name in "asymmetric_common" "common" "external_secrets" "kx" "signatures" "symmetric" ;
+    do 
+        curl -s -L --remote-name-all https://raw.githubusercontent.com/WebAssembly/wasi-crypto/8e883a5a173062e1ecf7dbd7124188ab8046425d/witx/proposal_${name}.witx
+        # target/release/wasi-cpp-header generate --output "$ROOT/thirdparty/wasi/crypto/${name}.hpp" proposal_${name}.witx
+    done
+
+target/release/wasi-cpp-header generate --output /home/younger/WasmEdge/thirdparty/wasi/crypto/api.hpp proposal_kx.witx proposal_asymmetric_common.witx  proposal_common.witx proposal_signatures.witx proposal_symmetric.witx proposal_external_secrets.witx
```
hydai, line 14: Don't use your local path in a general script.

hydai:
Also, I think you will need a README file in the `docs/witx/crypto` which mentions that the version of `wasi-crypto`

sonder-joker:
Only left many linux- I need check build script

include/common/enum_configure.h
```diff
@@ -96,15 +96,25 @@ enum WasmEdge_Proposal {
 namespace WasmEdge {
 
 /// Host Module Registration C++ enumeration class.
-enum class HostRegistration : uint8_t { Wasi = 0, WasmEdge_Process, Max };
+enum class HostRegistration : uint8_t {
+  Wasi = 0,
+  WasmEdge_Process,
+#ifdef WASMEDGE_BUILD_WASI_CRYPTO
```
hydai, line 102: The HostRegistration should be a static one. So you should not use `ifdef` here to remove it. Instead, just keep it.

include/common/enum_configure.h
```diff
@@ -96,15 +96,25 @@ enum WasmEdge_Proposal {
...
 
 } // namespace WasmEdge
 #endif
 
 /// Host Module Registration C enumeration.
 enum WasmEdge_HostRegistration {
   WasmEdge_HostRegistration_Wasi = 0,
-  WasmEdge_HostRegistration_WasmEdge_Process
+  WasmEdge_HostRegistration_WasmEdge_Process,
+#ifdef WASMEDGE_BUILD_WASI_CRYPTO
```
hydai, line 115: Ditto.

include/host/wasi_crypto/asymmetric_common/func.h
```diff
@@ -0,0 +1,201 @@
...
+  WasiCryptoContext &Ctx;
+};
+
+class KeypairGenerate : public HostFunction<KeypairGenerate> {
+public:
+  using HostFunction::HostFunction;
+
+  Expect<uint32_t> body(Runtime::Instance::MemoryInstance *MemInst,
+                        uint32_t AlgorithmType, const_uint8_t_ptr AlgorithmPtr,
+                        __wasi_size_t AlgorithmLen, uint32_t Options,
```
hydai, line 29: In the top level, we should use `stdint` like `int32_t`, `int64_t`, `uint32_t`, or `uint64_t`. Then, cast this type inside the function body.

E.g. `const_uint8_t_ptr` -> `uint32_t`; `__wasi_size_t` -> `uint32_t`.

include/host/wasi_crypto/asymmetric_common/func.h
```diff
@@ -0,0 +1,201 @@
...
+                        uint8_t_ptr /* Out */ KeypairPtr);
```
hydai, line 30: Ditto.

lib/host/wasi_crypto/asymmetric_common/func.cpp
```diff
@@ -0,0 +1,573 @@
...
+namespace Host {
+namespace WASICrypto {
+namespace AsymmetricCommon {
+
+Expect<uint32_t>
+KeypairGenerate::body(Runtime::Instance::MemoryInstance *MemInst,
+                      uint32_t AlgorithmType, const_uint8_t_ptr AlgorithmPtr,
+                      __wasi_size_t AlgorithmLen, uint32_t Options,
+                      uint8_t_ptr /* Out */ KeypairPtr) {
+  if (MemInst == nullptr) {
```
hydai, line 16: Please use `unlikely(MemInst == nullptr)`

hydai (changes requested):
There are some places which will need to fix first.

sonder-joker:
Current this branch have a bad git log, so I make a new branch to refactor. This branch remain for reference.

PR #409, opened 2021-09-13
[Rust] Refactoring: wasmedge-sys and wasmedge-rs update to 0.2.0

ZhangHanDong:
##  `wasmedge-sys`  0.2.0

- Improvements
    1. Wrap the unsafe code that deals directly with the C-API into the `wasmedge-sys` library.
    2. Expose the necessary interfaces to the upper layers for building the Rust sdk.

- Features
    1. Turning on the Wasi support.

## `wasmedge-rs` 0.2.0

- Improvements
    1. Change the vm api to a chain call and provide explicit semantics.
    2. Design the api to ensure that `config/module/vm` are created in the same scope so that memory can be reasonably freed.

- Features
    1. Turning on the Wasi support.


yanganto:
Hi @ZhangHanDong thanks for the PR, please make the all Rust test cases passed, thanks.

yanganto:
Hey, @ZhangHanDong, 
Testcases are really important things for a developer to use or to maintain the crate.  
Especially, the `doctest` will be the shown on the doccument. 
May you fix the test case instead of ignoring it?

ZhangHanDong:
@yanganto  You are right, documentation testing is important.

But `ignore` is chosen here because of the following reasons:

1.[`#[cfg_attr(test, test)]`](https://github.com/WasmEdge/WasmEdge/blob/b142ed82d418e75499d7fceacf5af5e776cd3534/bindings/rust/wasmedge-rs/examples/quickstart.rs#L3) has been added to `examples/quickstart.rs` and will be run when the test is run.
2. the comments here, I just want to use it simply as documentation, not as a documentation test. I forgot to add `ignore` before.

bindings/rust/wasmedge-sys/src/vm.rs
```diff
@@ -0,0 +1,97 @@
...
+        let raw_params: Vec<_> = params
+            .as_ref()
+            .iter()
+            .copied()
+            .map(wasmedge::WasmEdge_Value::from)
+            .collect();
+
+        let func_type = unsafe { wasmedge::WasmEdge_VMGetFunctionType(self.ctx, raw_func_name) };
+        if func_type.is_null() {
+            return Err(ErrReport::default());
```
yanganto, line 70: May we use following ErrReport for this error?
```
ErrReport {
  code: None,
  message: "wasmedge get function type not found"
}
```

It is good if the code is `None`, the developer can know there is no error code from wasmedge VM, and the message describe something happened in the `-sys` layer.

yanganto (changes requested):

bindings/rust/wasmedge-sys/src/config.rs
```diff
@@ -0,0 +1,162 @@
...
+// the private encapsulation  here can force the third-party sdk to use the api
+#[derive(Debug)]
+pub struct Config {
+    pub(crate) ctx: *mut wasmedge::WasmEdge_ConfigureContext,
+    _private: (),
+}
+
+impl Drop for Config {
+    fn drop(&mut self) {
+        println!("config drop!");
```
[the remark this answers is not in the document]
ZhangHanDong: I forgot to delete it

bindings/rust/wasmedge-rs/src/vm.rs
```diff
@@ -1,98 +1,92 @@
...
+    pub fn create(self) -> Result<Vm<'a>, anyhow::Error> {
+        let vm = self.inner;
+        if let Some(cfg) = vm.config {
+            let vm_instance = wasmedge::Vm::create(cfg).map_err(VmError::Create)?;
+            let vm_instance = vm_instance
+                .load_wasm_from_ast_module(&vm.module.inner)
+                .map_err(VmError::ModuleLoad)?;
+            let vm_instance = vm_instance.validate().map_err(VmError::Validate)?;
+            let vm_instance = vm_instance.instantiate().map_err(VmError::Instantiate)?;
+            println!("vm_instance {:#?}", vm_instance);
```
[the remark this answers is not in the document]
ZhangHanDong: I forgot to delete the previous debugging

bindings/rust/wasmedge-sys/src/vm.rs
ZhangHanDong: The `ErrReport`  field type is currently numeric and receives the error code from the c side. When the pointer is empty, there is actually no error code on the c side. Maybe I can leave the null pointer out of the error handling and just panic if there is a null pointer

bindings/rust/wasmedge-sys/src/config.rs
```diff
@@ -0,0 +1,162 @@
+use super::wasmedge;
+
+// If there is a third-party sdk based on wasmedge-sys
+// the private encapsulation  here can force the third-party sdk to use the api
+#[derive(Debug)]
+pub struct Config {
+    pub(crate) ctx: *mut wasmedge::WasmEdge_ConfigureContext,
+    _private: (),
```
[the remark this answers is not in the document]
ZhangHanDong: ok

bindings/rust/wasmedge-sys/src/vm.rs
yanganto: I think it is excellent to use `panic!` here if people calling the non-exist function. :+1: 

If the `panic!` is used, do we still need the `Default` trait for `ErrReport`?

It may be good to remove the `Default` implementation and let the `ErrReport` always be created from a `wasmedge::WasmEdge_Result`.

juntao:
We are not going to support the WASI config options in version 0.2. Is that correct?

ZhangHanDong:
@juntao  Yes, the current version does not add wasm, it is just refactored. The roadmap for the follow up needs to be planned. I currently know that version 0.3 will support wasm-bindgen.

bindings/rust/wasmedge-sys/src/raw_result.rs
```diff
@@ -0,0 +1,36 @@
...
+// Since WasmEdge_ErrCode is subject to change on the wasmedge side
+// it does not correspond to enum here
+pub fn decode_result(res: wasmedge::WasmEdge_Result) -> Result<(), ErrReport> {
+    if is_ok(res) {
+        Ok(())
+    } else {
+        Err(ErrReport {
+            code: get_code(res),
+            message: get_message(res),
+        })
```
yanganto, line 34: May we use `impl From<wasmedge::WasmEdge_Result> for ErrReport` here?
Because the methods `get_code` and `get_message` are used the same time, it maybe be better use `From` trait without missing one of the method.

ZhangHanDong: good idea.

ZhangHanDong:
@juntao  I tried to increase `wasi` support today and add the hello example. 

There are some bugs at the moment, and they should be fixed tomorrow. I hope they can be released in 0.2. 

I will temporarily convert the current PR to draft and resume it after the modification is completed.

ZhangHanDong:
Added support for wasi-config.  https://github.com/WasmEdge/WasmEdge/issues/425

ZhangHanDong:
I don't know why the strange commits are mixed in, do I need to resend the pull request ?

alabulei1:
> I don't know why the strange commits are mixed in, do I need to resend the pull request ?

Yes, you should delete the mixed commit.

ZhangHanDong:
@alabulei1 OK. I'll resend PR.

ZhangHanDong:
https://github.com/WasmEdge/WasmEdge/pull/432

Issue #412, opened 2021-09-14
Cannot install WasmEdge


ZHANGWENTAI:
### Current State

Cannot install WasmEdge

## Environment

- Hardware Architecture: x86_64 
- Operating system: Ubuntu 16.04

## Steps to Reproduce
I use command ```wget -qO- https://raw.githubusercontent.com/WasmEdge/WasmEdge/master/utils/install.sh | bash``` to install WasmEdge, the output is: 
```
No root permissions.
Detected Linux-x86_64
No path provided
Installing in /home/myusername/.wasmedge
WasmEdge Installation at /home/myusername/.wasmedge
Fetching WasmEdge-0.8.2

    Please see --help
    If issue perists make a trace using -V and submit it to
https://github.com/WasmEdge/WasmEdge/issues/new?assignees=&labels=&template=bug_report.md
```
but in the directory /home/myusername, there are no WasmEdge installed.

I wonder if there something wrong in the install script?

By the way, the ```rm -f``` opreation in the install script could be dangerous, I have lost my /root directory already on another CentOS server by using the command ```wget -qO- https://raw.githubusercontent.com/WasmEdge/WasmEdge/master/utils/install.sh | bash``` with wrong arguments........

hydai:
Hi @ZHANGWENTAI 

I tried to reproduce your issue, however, I cannot reproduce this error:

```
user@8785a5841606:~$ wget -qO- https://raw.githubusercontent.com/WasmEdge/WasmEdge/master/utils/install.sh | bash
No root permissions.
sudo could not be found
Detected Linux-x86_64
No path provided
Installing in /home/user/.wasmedge
WasmEdge Installation at /home/user/.wasmedge
Fetching WasmEdge-0.8.2
WasmEdge-0.8.2-manylinux2014_x86_64. 100%[======================================================================>]  39.00M  31.3MB/s    in 1.2s
Installing WasmEdge-0.8.2-Linux in /home/user/.wasmedge/include
Installing WasmEdge-0.8.2-Linux in /home/user/.wasmedge/lib
Installing WasmEdge-0.8.2-Linux in /home/user/.wasmedge/bin
No extensions to be installed

source /home/user/.wasmedge/env to use wasmedge binaries
user@8785a5841606:~$ cat /etc/os-release
NAME="Ubuntu"
VERSION="16.04.7 LTS (Xenial Xerus)"
ID=ubuntu
ID_LIKE=debian
[...]

Issue #415, opened 2021-09-14
[Install] support non-login bash


juntao:
## Motivation

We currently put `source env` in `.profile` which supports login shells. But in many use cases, such as Docker and codespaces, we need to support non-login shells.

## Details

We can do the same as Rust install script does. It puts the `source env` command in BOTH `.profile` and `.bashrc` files.

SAtacker:
I guess I'll close this one. Thanks.

PR #416, opened 2021-09-14
[Utils] Install Script Fix

SAtacker:
* Removed dangerous removal of path #412 
* Quoting of variables as per shellcheck 2086
https://github.com/koalaman/shellcheck/wiki/SC2086.
* Apologies to the affected users ;-(

### DIFF
--- a/utils/install.sh
+++ b/utils/install.sh
@@ -39,10 +39,14 @@ fi
 
 if command -v sudo &>/dev/null; then
     if [ $PERM_ROOT == 1 ]; then
-        HOME=$(getent passwd "$SUDO_USER" | cut -d: -f6)
+        __HOME__=$(getent passwd "$SUDO_USER" | cut -d: -f6)
     fi
 else
     echo "${YELLOW}sudo could not be found${NC}"
+fi
+
+if [ "$__HOME__" = "" ]; then
+    __HOME__="$HOME"
 fi
 
 RELEASE_PKG="manylinux2014_x86_64.tar.gz"
@@ -89,7 +93,7 @@ VERSION_TF=$(get_latest_release second-state/WasmEdge-tensorflow)
 VERSION_TF_DEPS=$(get_latest_release second-state/WasmEdge-tensorflow-deps)
 VERSION_TF_TOOLS=$(get_latest_release second-state/WasmEdge-tensorflow-tools)
 
-IPATH="$HOME/.wasmedge"
+IPATH="$__HOME__/.wasmedge"
 EXT="none"
 VERBOSE=0
 
@@ -420,25 +424,25 @@ main() {
     fi
 
     echo "$ENV" >"$IPATH/env"
-    local _source=". \"\$HOME/.wasmedge/env\""
-    local _grep=$(cat "$HOME/.profile" 2>/dev/null | grep "wasmedge")
-    if [ -f "$HOME/.profile" ]; then
+    local _source=". \"\$__HOME__/.wasmedge/env\""
+    local _grep=$(cat "$__HOME__/.profile" 2>/dev/null | grep "wasmedge")
+    if [ -f "$__HOME__/.profile" ]; then
         if [ "$_grep" = "" ]; then
-            echo "$_source" >>"$HOME/.profile"
+            echo "$_source" >>"$__HOME__/.profile"
         fi
     fi
-    if [ -f "$HOME/.bashrc" ]; then
-        local _grep=$(cat "$HOME/.bashrc" | grep "wasmedge")
+    if [ -f "$__HOME__/.bashrc" ]; then
+        local _grep=$(cat "$__HOME__/.bashrc" | grep "wasmedge")
         if [ "$_grep" = "" ]; then
-            echo "$_source" >>"$HOME/.bashrc"
+            echo "$_source" >>"$__HOME__/.bashrc"
         fi
-    elif [ -f "$HOME/.bash_profile" ]; then
-        local _grep=$(cat "$HOME/.bash_profile" | grep "wasmedge")
+    elif [ -f "$__HOME__/.bash_profile" ]; then
+        local _grep=$(cat "$__HOME__/.bash_profile" | grep "wasmedge")
         if [ "$_grep" = "" ]; then
-            echo "$_source" >>"$HOME/.bash_profile"
+            echo "$_source" >>"$__HOME__/.bash_profile"
         fi
     else
-        echo "$_source" >>"$HOME/.bashrc"
+        echo "$_source" >>"$__HOME__/.bashrc"
     fi
 
     trap - EXIT
--- a/utils/install.sh
+++ b/utils/install.sh
@@ -385,18 +385,19 @@ main() {
         mkdir -p "$IPATH"
         echo "$ENV" >"$IPATH/env"
         local _source=". \"\$HOME/.wasmedge/env\""
-        local _grep=$(cat $HOME/.profile 2>/dev/null | grep "wasmedge")
+        local _grep=$(cat "$HOME/.profile" 2>/dev/null | grep "wasmedge")
         if [ -f "$HOME/.profile" ]; then
             if [ "$_grep" = "" ]; then
                 echo "$_source" >>"$HOME/.profile"
             fi
-        elif [ -f "$HOME/.bashrc" ]; then
-            local _grep=$(cat $HOME/.bashrc | grep "wasmedge")
+        fi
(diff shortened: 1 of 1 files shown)


SAtacker:
```bash
wget -qO- https://raw.githubusercontent.com/SAtacker/WasmEdge/fix_install_script/utils/install.sh | bash -s -- -e all
```
I have removed the dangerous part so you can test it. Please let me know.

juntao:
I found that if I have `sudo` installed and install wasmedge into `~/.wasmedge`, the script will NOT write the `source env` command into `.profile` or `.bashrc`

SAtacker:
> I found that if I have `sudo` installed and install wasmedge into `~/.wasmedge`, the script will NOT write the `source env` command into `.profile` or `.bashrc`

If they give the path manually then I don't write an `env` file because then I'll have to populate their custom directory

SAtacker:
A bit sad news `"API rate limit exceeded for 111.125.230.183. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)"`. This means I cannot use GitHub's API anymore, we may get more than 1000 a day and nobody else would be able to install like I am facing right now.

juntao:
> If they give the path manually then I don't write an `env` file because then I'll have to populate their custom directory

In my tests, I did not give a path. It installs into HOME/.wasmedge in both tests. However,

* test 1: the system does not have `sudo` and the install behaves correctly.
* test 2: the system has `sudo` and the script did not add the `source env` commands in bash profile files.

juntao:
> A bit sad news `"API rate limit exceeded for 111.125.230.183. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)"`. This means I cannot use GitHub's API anymore, we may get more than 60 an hour and nobody else would be able to install like I am facing right now.

I think the rate limiting is based on IP address? That means the user will not have this problem since they will each have a different IP address. Correct?

SAtacker:
> > If they give the path manually then I don't write an `env` file because then I'll have to populate their custom directory
> 
> In my tests, I did not give a path. It installs into HOME/.wasmedge in both tests. However,
> 
>     * test 1: the system does not have `sudo` and the install behaves correctly.
> 
>     * test 2: the system has `sudo` and the script did not add the `source env` commands in bash profile files.

Could you please confirm if this was after https://github.com/SAtacker/WasmEdge/commit/3d7cfd863a14ced06a758b0e372637262ae715a6?

SAtacker:
> > A bit sad news `"API rate limit exceeded for 111.125.230.183. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)"`. This means I cannot use GitHub's API anymore, we may get more than 60 an hour and nobody else would be able to install like I am facing right now.
> 
> I think the rate limiting is based on IP address? That means the user will not have this problem since they will each have a different IP address. Correct?

I have added https://github.com/SAtacker/WasmEdge/commit/442bc1e277cce2de0f67f8067975e6829e6a336c which removes the GitHub API dependency. I'll have to check on MacOs once if it works (If all the standard tools are present there as well or not)

SAtacker:
It does not work on macOS

SAtacker:
Summar / Changelog
- [Utils] Minor Fix. [Shreyas Atre]

  * __HOME___ was echoed instead of HOME
  * If .profile does not exists, it is created
- [Utils] Fix for macOS. [Shreyas Atre]

  * Apparently macOS cut & tail don'r have long options same as popular
  linux distros.
- [Utils] Remove github API request. [Shreyas Atre]

  * This removes the limitations of GitHub API request using git and
  standard tools itself.
- [Utils] Don't mess with $HOME value. [Shreyas Atre]

  * Introduced __HOME__ and checked if getent is giving empty values.
  * If it is an empty string __HOME__ will be same as HOME
- [Utils] Add to .bashrc & .profile. [Shreyas Atre]

  * Changed previous behaviour of appending to any one of above to both
  the above.
- [Utils] Add sourcing to .bashrc as well. [Shreyas Atre]

  * Added suggestions as per https://github.com/WasmEdge/WasmEdge/issues/415#issue-996302490
- [Utils] Install Script Fix. [Shreyas Atre]

  * Removed dangerous removal of path
  * Quoting of variables as per shellcheck 2086
  https://github.com/koalaman/shellcheck/wiki/SC2086.

juntao:
If I do 

```
wget -qO- https://raw.githubusercontent.com/SAtacker/WasmEdge/fix_install_script/utils/install.sh | bash -s -- -e all -p /usr/local
```

It will have the following line in `.profile` and `.bashrc`

```
. "/usr/local/env"
```

The content of the `/usr/local/env` is as follows.

```
#!/bin/sh
# wasmedge shell setup
# affix colons on either side of $PATH to simplify matching
case :${PATH}: in
    *:/root/.wasmedge/bin:*)
        ;;
    *)
        # Prepending path in case a system-installed wasmedge needs to be overridden
        export PATH=/root/.wasmedge/bin:$PATH
        ;;
esac
case :${LD_LIBRARY_PATH}: in
    *:/root/.wasmedge/lib:*)
        ;;
    *)
        # Prepending path in case a system-installed wasmedge libs needs to be overridden
        export LD_LIBRARY_PATH=/root/.wasmedge/lib:$LD_LIBRARY_PATH
        ;;
esac
```

I think we need a rule to NOT add the `source env` command when the install path is `/usr/local`.

juntao:
Hmm, how did you test on MacOS? I do not think our release has a MacOS binary build yet. :)

SAtacker:
I meant a dry run. 😀

SAtacker:
> If I do
> 
> ```
> wget -qO- https://raw.githubusercontent.com/SAtacker/WasmEdge/fix_install_script/utils/install.sh | bash -s -- -e all -p /usr/local
> ```
> 
> It will have the following line in `.profile` and `.bashrc`
> 
> ```
> . "/usr/local/env"
> ```
> 
> The content of the `/usr/local/env` is as follows.
> 
> ```
> #!/bin/sh
> # wasmedge shell setup
> # affix colons on either side of $PATH to simplify matching
> case :${PATH}: in
>     *:/root/.wasmedge/bin:*)
>         ;;
>     *)
>         # Prepending path in case a system-installed wasmedge needs to be overridden
>         export PATH=/root/.wasmedge/bin:$PATH
>         ;;
> esac
> case :${LD_LIBRARY_PATH}: in
>     *:/root/.wasmedge/lib:*)
>         ;;
>     *)
>         # Prepending path in case a system-installed wasmedge libs needs to be overridden
>         export LD_LIBRARY_PATH=/root/.wasmedge/lib:$LD_LIBRARY_PATH
>         ;;
> esac
> ```
> 
> I think we need a rule to NOT add the `source env` command when the install path is `/usr/local`.

It checks if the PATH contains `/usr/local` and then populates
```bash
case :${PATH}: in
    *:/root/.wasmedge/bin:*)
        ;;
    *)
        # Prepending path in case a system-installed wasmedge needs to be overridden
        export PATH=/root/.wasmedge/bin:$PATH
        ;;
esac
```
Meaning that you should not see overpopulated `PATH` environment variable

juntao:
hmm, I just thought it is unnecessary for the `/usr/local/env` file even if it does not actually affect the `PATH`. Is there an easy fix?

SAtacker:
> hmm, I just thought it is unnecessary for the `/usr/local/env` file even if it does not actually affect the `PATH`. Is there an easy fix?

We can create an `env` under `XDG_CONFIG_HOME` which is a standard for most of the applications.
Ref: https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

juntao:
Perhaps we could do the following

1 Always leave an env file in the installed directory.

2 The env file content should reference the actual installed directory. That is `/usr/local/env` should not have `/root/.wasmedge` path.

3 The `.profile` and `.bashrc` files should contain the source env command ONLY when the `-p` argument is absent. 

What do you think?

SAtacker:
> Perhaps we could do the following
> 
> 1 Always leave an env file in the installed directory.
> 
> 2 The env file content should reference the actual installed directory. That is `/usr/local/env` should not have `/root/.wasmedge` path.
> 
> 3 The `.profile` and `.bashrc` files should contain the source env command ONLY when the `-p` argument is absent.
> 
> What do you think?

Thank you for this! I agree completely. Initially, I overlooked the `*:/root/.wasmedge/bin:*` part even of `/usr/local` .
Sending a patch for all the mentioned suggestions.

juntao (approved):
I tested. It is read to merge now.

PR #432, opened 2021-09-22
[Rust] Refactoring: wasmedge-sys and wasmedge-rs update to 0.2.0

ZhangHanDong:
##  `wasmedge-sys`  0.2.0

- Improvements
    1. Wrap the unsafe code that deals directly with the C-API into the `wasmedge-sys` library.
    2. Expose the necessary interfaces to the upper layers for building the Rust sdk.

- Features
    1. Turning on the Wasi support.

## `wasmedge-rs` 0.2.0

- Improvements
    1. Change the vm api to a chain call and provide explicit semantics.
    2. Design the api to ensure that `config/module/vm` are created in the same scope so that memory can be reasonably freed.

- Features
    1. Turning on the Wasi support.


bindings/rust/wasmedge-sys/src/vm.rs
```diff
@@ -0,0 +1,174 @@
...
+    wasi,
+};
+use std::os::raw::c_char;
+
+// If there is a third-party sdk based on wasmedge-sys
+// the private encapsulation  here can force the third-party sdk to use the api
+#[derive(Debug)]
+pub struct Vm {
+    pub(crate) ctx: *mut wasmedge::WasmEdge_VMContext,
+    _private: (),
```
yanganto, line 15: Please help not to show the private assets at this moment, because we do not have any. :)

yanganto (changes requested):
One minor issue, as mentioned in the previous PR, others LGTM, thanks for PR.

bindings/rust/wasmedge-sys/src/vm.rs
nhynes: actually, based on the reply to my comment in the original PR, I think what you actually want is `#[non_exhaustive]` on structs with all `pub` members. This one has a `pub(crate)` in it, so it's already not constructible from outside of this crate.

However, this evidences that Clippy isn't being run. I _strongly_ suggest running `cargo clippy --all-features --profile test` on the code before shipping. Clippy is a great tool for learning Rust, too.

bindings/rust/wasmedge-sys/src/vm.rs
```diff
@@ -0,0 +1,174 @@
...
+}
+
+impl Vm {
+    pub fn create(config: &crate::config::Config) -> Result<Self, ErrReport> {
+        let ctx = unsafe {
+            wasmedge::WasmEdge_VMCreate(config.ctx, std::ptr::null_mut() /* store */)
+        };
+        if ctx.is_null() {
+            panic!("WasmEdge VM create failed");
```
nhynes, line 24: `panic` should only be used for irrecoverable errors. Here, it's likely because the C lib doesn't like the config, but that can be fixed by application code. If the only failure mode is because of OOM or something like that, you can disregard this comment.

For libraries, panicking is incredibly difficult to work around when it's not easy to check why the operation would fail (compare to `&str/&[T]` methods that panic when an index is wrong, but that's easy to check).

bindings/rust/wasmedge-sys/src/vm.rs
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    pub fn validate(self) -> Result<Self, ErrReport> {
+        unsafe {
+            decode_result(wasmedge::WasmEdge_VMValidate(self.ctx))?;
+        }
+        Ok(self)
+    }
+
+    pub fn instantiate(self) -> Result<Self, ErrReport> {
```
nhynes, line 48: is there really any case that you'd do these individually and not always apply all of them?

bindings/rust/wasmedge-sys/src/vm.rs
```diff
@@ -0,0 +1,174 @@
...
+        unsafe {
+            decode_result(wasmedge::WasmEdge_VMInstantiate(self.ctx))?;
+        }
+        Ok(self)
+    }
+
+    pub fn init_wasi_obj(
+        &self,
+        args: Option<Vec<&str>>,
```
nhynes, line 57: please use named arguments to provide some additional safety and readability. the typechecker can't help here.

bindings/rust/wasmedge-sys/src/vm.rs
```diff
@@ -0,0 +1,174 @@
...
+        envs: Option<Vec<&str>>,
+        dirs: Option<Vec<&str>>,
+        preopens: Option<Vec<&str>>,
```
nhynes, line 60: why are these all `Option`s, though? An empty `Vec` is as good and gives a cleaner API. `vec![]` doesn't allocate FYI. Also, you might want to do `&[impl AsRef<str>]`  on these in case anyone has a vec of `String` or slice of `str` etc. Going from `Vec<String>` to `Vec<&str>` does require an allocation and is annoying to write. Also, you don't seem to need to take ownership of the args anyway.

bindings/rust/wasmedge-sys/src/vm.rs
```diff
@@ -0,0 +1,174 @@
...
+                })
+                .collect::<Vec<*const c_char>>();
+            import_obj.args_len = args.len() as u32;
+        }
+
+        if let Some(mut envs) = envs {
+            import_obj.envs = envs
+                .iter_mut()
+                .map(|env| {
+                    let raw_env: wasmedge::WasmEdge_String = StringRef::from(*env).into();
```
nhynes, line 86: can you run clippy on this? I don't think you need `mut`

nhynes (commented):
Overall, the PR looks good. Please take care to run `cargo clippy` (and add it to CI!), and also try not to make panicky libraries. It's no fun for people who need systems that don't panic (e.g., blockchain).

bindings/rust/wasmedge-sys/src/vm.rs
ZhangHanDong: clippy I use it a lot and it doesn't give you advice on everything. There is no advice or warning here either.

ZhangHanDong: I'm not trying to be `non_exhaustive` either, the field here was previously `pub` and I forgot to remove `_private`.

ZhangHanDong:
> Overall, the PR looks good. Please take care to run `cargo clippy` (and add it to CI!), and also try not to make panicky libraries. It's no fun for people who need systems that don't panic (e.g., blockchain).

clippy I use it a lot and it doesn't give you advice on everything. There is no advice or warning here either. Isn't there always cargo clippy in ci?

bindings/rust/wasmedge-rs/examples/hello.rs
```diff
@@ -0,0 +1,29 @@
...
+    }
+    let module_path =
+        std::path::PathBuf::from(env!("WASMEDGE_DIR")).join("tools/wasmedge/examples/hello.wasm");
+
+    let config = wedge::Config::with_wasi();
+    let module = wedge::Module::new(&config, &module_path)?;
+
+    let vm = wedge::Vm::load(&module)?.with_config(&config)?.create()?;
+
+    vm.init_wasi_obj(Some(args), None, None, None);
```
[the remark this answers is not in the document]
ZhangHanDong: Good suggestion, I am also considering this problem, this place needs to be redesigned.

bindings/rust/wasmedge-rs/src/config.rs
```diff
@@ -1,118 +1,17 @@
...
-                    pub fn [<has_ $proposal:lower:snake>](self) -> bool {
-                        let prop = ffi::[<WasmEdge_Proposal_ $proposal>];
-                        unsafe { ffi::WasmEdge_ConfigureHasProposal(self.ctx, prop) }
-                    }
-                )+
-            }
-        }
+    pub fn with_wasi() -> wasmedge::Config {
+        let config = Self::default();
+        config.reg_wasi()
```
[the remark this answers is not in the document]
ZhangHanDong: This method is defined by you. I just followed the method you defined. I guess it means register wasi, because the corresponding c-api is `WasmEdge_HostRegistration_Wasi`.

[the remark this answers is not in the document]
ZhangHanDong: `enable_wasi ` is good.

bindings/rust/wasmedge-rs/src/module.rs
```diff
@@ -1,56 +1,30 @@
+use super::wasmedge;
```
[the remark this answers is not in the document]
ZhangHanDong: I thought rustfmt did it automatically, but I looked at rustfmt's current configuration and realized it was your configuration error, it should be `imports_granularity = "StdExternalCrate"`, not `imports_granularity = "Crate"`

bindings/rust/wasmedge-rs/src/vm.rs
```diff
@@ -1,98 +1,104 @@
...
+            let vm_instance = vm_instance.validate().map_err(VmError::Validate)?;
+            let vm_instance = vm_instance.instantiate().map_err(VmError::Instantiate)?;
+
+            Ok(Vm {
+                config: vm.config,
+                module: vm.module,
+                inner: Some(vm_instance),
+            })
+        } else {
+            panic!("Failed! Please specify a reasonable module path and configuration");
```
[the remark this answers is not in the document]
ZhangHanDong: Oh, I forgot to change it

bindings/rust/wasmedge-sys/src/vm.rs
ZhangHanDong: This is only the sys library, the final application in the upper sdk, split out to facilitate the handling of errors at each step

bindings/rust/wasmedge-sys/src/vm.rs
ZhangHanDong: I don't understand your suggestion, can you explain further with code examples? What are named arguments？The parameters are passed in externally and no one knows what name it is going to pass.

bindings/rust/wasmedge-sys/src/vm.rs
ZhangHanDong: clippy is not foolproof, I execute it a many times and it won't prompt this.

bindings/rust/wasmedge-sys/src/vm.rs
ZhangHanDong: This is only a temporary solution, and it needs to be refactored later with specific usage scenarios. This should be the goal of version 0.3 .

bindings/rust/wasmedge-sys/src/vm.rs
ZhangHanDong: The panic here is the irrecoverable error.

ZhangHanDong:
The interface used by wasi configuration has to be modified later according to the actual scenario, so let's put this part of the work in version 0.3.

ZhangHanDong:
@yanganto Wouldn't `rustfmt.toml` be better placed under `bindgen/rust`?

yanganto:
> @yanganto Wouldn't `rustfmt.toml` be better placed under `bindgen/rust`?

Agreed, please help on this and make sure the CI works if you want to move the file in this PR.

yanganto:
@ZhangHanDong 
The suggestion from @nhynes to remove the following `Option`  is really a good idea.

bindings/rust/wasmedge-sys/src/vm.rs
```
        args: Option<Vec<&str>>,
        envs: Option<Vec<&str>>,
        dirs: Option<Vec<&str>>,
        preopens: Option<Vec<&str>>,
```

Besides, there is [is_empty](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.is_empty) function for Vec in Rust, so we can check the empty parameter without `is_none`.

May we accept this suggestion from nhynes?

And it may be good to use the same concept on `args: Option<Vec<&str>>` of  `init_wasi_obj`.
The `args` is provided by user, so the arguments in `args` can not be named, however we can sure the `Option` can be removed here.  

Once everything is done, I think we can merge this PR sooner.  Thanks.

ZhangHanDong:
> @ZhangHanDong The suggestion from @nhynes to remove the following `Option` is really a good idea.
> 
> bindings/rust/wasmedge-sys/src/vm.rs
> 
> ```
>         args: Option<Vec<&str>>,
>         envs: Option<Vec<&str>>,
>         dirs: Option<Vec<&str>>,
>         preopens: Option<Vec<&str>>,
> ```
> 
> Besides, there is [is_empty](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.is_empty) function for Vec in Rust, so we can check the empty parameter without `is_none`.
> 
> May we accept this suggestion from nhynes?
> 
> And it may be good to use the same concept on `args: Option<Vec<&str>>` of `init_wasi_obj`. The `args` is provided by user, so the arguments in `args` can not be named, however the `Option` can be removed here.
> 
> Once everything is done, I think we can merge this PR sooner. Thanks.

OK. I was going to leave it in version 0.3 before changing it.

yanganto:
Hey @ZhangHanDong, 

Most Rust developers have suffered from api breaking changes such as `futures` and `tokio` crates.  If we have a chance to make things better with less api change in WasmEdge, we have to do it.  Also, we can be visionary and kindly programmers, not a coder focus on the number of commits.  Furthermore, one of the good reasons to be an open-source project is taking good advice.

Since this is not a tremendous work effort and you already accept these good suggestions, why do we need to wait for version 0.3.0?  

If there is something I forgot to concern about, please tell me thanks. Besides, if you lack time, please tell me, I will take care and keep handle this PR.

ZhangHanDong:
> Hey @ZhangHanDong,
> 
> Most Rust developers have suffered from api breaking changes such as `futures` and `tokio` crates. If we have a chance to make things better with less api change in WasmEdge, we have to do it. Also, we can be visionary and kindly programmers, not a coder focus on the number of commits. Furthermore, one of the good reasons to be an open-source project is taking good advice.
> 
> Since this is not a tremendous work effort and you already accept these good suggestions, why do we need to wait for version 0.3.0?
> 
> If there is something I forgot to concern about, please tell me thanks. Besides, if you lack time, please tell me, I will take care and keep handle this PR.

Because I think v`ec![]` is not the final solution, I just want to leave my energy to plan it properly in 0.3, I think it's normal for api to change before 1.0. Also because, recently, it is also rather busy.
But if you insist on this change, I can fix it as soon as possible in the next two days.

yanganto:
Nice and I am looking forward to this, thanks.

ZhangHanDong:
@yanganto 

I've modified the API, but instead of changing it to `vec![]`, I've redesigned the interface.

``rust
let vm = wedge::Vm::load(&module)? .with_config(&config)? .create()? ;

 vm.init_wasi_obj().with_args(args).build();
 // vm.init_wasi_obj().with_args(args).with_envs(envs);
```

This is much better, no need to pass empty arrays.

I always thought Option was a must. Because semantically, the outside world passes args, etc., and it's None if it's not passed. Option is used to express this semantics, there is no need to use the traditional language practice of determining whether it is empty or not. Besides, args is never empty, so using vec![] is not appropriate here.

**Comment by ZhangHanDong** [comment]
@yanganto 

Since many configurations in `rustfmt` are only available under nightly, I modified the ci to add the nightly toolchain.

**Comment by ZhangHanDong** [comment]
@yanganto  the `codecov/project` check not successful , how to fix it?

**Comment by hydai** [comment]
> @yanganto the `codecov/project` check not successful , how to fix it?

@ZhangHanDong The `codecov` is only for C++/C side. If you don't change those files, please rebase to the latest master. Then, it should be fine.

**Comment by ZhangHanDong** [comment]
> > @yanganto the `codecov/project` check not successful , how to fix it?
> 
> @ZhangHanDong The `codecov` is only for C++/C side. If you don't change those files, please rebase to the latest master. Then, it should be fine.

Not working

**Comment by hydai** [comment]
hmmm, looks like the coverage is replied wrongly. I will check it later.
If you do not change cpp part, please just ignore it now. Thanks.

**Comment by hydai** [comment]
Before we merge this PR, I would like to request changes on your commits. Please apply `git rebase` to squash those useless commits.
For example, you have several `[Rust] modify CI rustfmt for nightly ` commits in your history. This is very strange.

**Comment by ZhangHanDong** [comment]
> Before we merge this PR, I would like to request changes on your commits. Please apply `git rebase` to squash those useless commits. For example, you have several `[Rust] modify CI rustfmt for nightly ` commits in your history. This is very strange.

I moved `rustfmt.toml` to bindgen/rust, but actually now the configuration in `rustfmt.toml` is only executable under nightly rust, so the ci is wrong. That commit is to fix ci by adding nightly rust support and letting fmt execute `cargo +nightly fmt`.

**Comment by yanganto** [comment]
@ZhangHanDong 
There is something in `bindings/rust/wasmedge-rs/src/config.rs`, I think you forgot this.
Thanks.

**Comment by ZhangHanDong** [comment]
> @ZhangHanDong There is something in `bindings/rust/wasmedge-rs/src/config.rs`, I think you forgot this. Thanks.

I do not quite understand what you mean, this file is normal .

**Comment by yanganto** [comment]
> > @ZhangHanDong There is something in `bindings/rust/wasmedge-rs/src/config.rs`, I think you forgot this. Thanks.
> 
> I do not quite understand what you mean, this file is normal . Both examples compile without problems and both work fine.

I have left a question for you about why the default function without `Default` trait?
You delete the Default trait and keep the default function.
Is there any reason for that?
Thanks.
```
   pub fn default() -> wasmedge::Config {
        wasmedge::Config::default()
    }
```

**Comment by ZhangHanDong** [comment]
> > > @ZhangHanDong There is something in `bindings/rust/wasmedge-rs/src/config.rs`, I think you forgot this. Thanks.
> > 
> > 
> > I do not quite understand what you mean, this file is normal . Both examples compile without problems and both work fine.
> 
> I have left a question for you about why the default function without `Default` trait? You delete the Default trait and keep the default function. Is there any reason for that? Thanks.
> 
> ```
>    pub fn default() -> wasmedge::Config {
>         wasmedge::Config::default()
>     }
> ```

Since the `Default` trait is already imported in `std::prelude`, you can use it directly.

See: [https://doc.rust-lang.org/stable/std/prelude/index.html](https://doc.rust-lang.org/stable/std/prelude/index.html)

**Comment by yanganto** [comment]
> > > > @ZhangHanDong There is something in `bindings/rust/wasmedge-rs/src/config.rs`, I think you forgot this. Thanks.
> > > 
> > > 
> > > I do not quite understand what you mean, this file is normal . Both examples compile without problems and both work fine.
> > 
> > 
> > I have left a question for you about why the default function without `Default` trait? You delete the Default trait and keep the default function. Is there any reason for that? Thanks.
> > ```
> >    pub fn default() -> wasmedge::Config {
> >         wasmedge::Config::default()
> >     }
> > ```
> 
> Since the `Default` trait is already imported in `std::prelude`, you can use it directly.
> 
> See: https://doc.rust-lang.org/stable/std/prelude/index.html

There is slight different if you are not implement default function for `Default` trait, people can call the default function but without `Default` trait there will still some behaviors unexpected.  
Follow code is one of the example can not work without `Default` trait.

```
#[derive(Default)]
struct Wrap {
  i: usize,
  config: wasmedge::Config,
}
let w = Wrap {
  i: 1, 
  ..Default::default(),
}
```

**Comment by ZhangHanDong** [comment]
> > > > > @ZhangHanDong There is something in `bindings/rust/wasmedge-rs/src/config.rs`, I think you forgot this. Thanks.
> > > > 
> > > > 
> > > > I do not quite understand what you mean, this file is normal . Both examples compile without problems and both work fine.
> > > 
> > > 
> > > I have left a question for you about why the default function without `Default` trait? You delete the Default trait and keep the default function. Is there any reason for that? Thanks.
> > > ```
> > >    pub fn default() -> wasmedge::Config {
> > >         wasmedge::Config::default()
> > >     }
> > > ```
> > 
> > 
> > Since the `Default` trait is already imported in `std::prelude`, you can use it directly.
> > See: https://doc.rust-lang.org/stable/std/prelude/index.html
> 
> There is slight different if you are not implement default function for `Default` trait, people can call the default function but without `Default` trait, but there will still some behaviors unexpected. Follow code is one of the example can not work without `Default` trait.
> 
> ```
> #[derive(Default)]
> struct Wrap {
>   i: usize,
>   config: wasmedge::Config,
> }
> let w = Wrap {
>   i: 1, 
>   ..Default::default(),
> };
> ```

I understand what you mean, and you are right.

I wasn't actually trying to implement the `Default` trait here, I was just trying to convey that it is a default configuration value. And `Default` can't be implemented here either, because it doesn't return `Self`.

Maybe a different method name would be better : 
``rust
impl Config {
    pub fn default_config() -> wasmedge::Config {
        wasmedge::Config::default()
    }

    pub fn with_wasi() -> wasmedge::Config {
        let config = Self::default_config();
        config.enable_wasi()
    }

```

What do you think?

yanganto (approved):

ZhangHanDong:
@yanganto `Default` has been implemented for `Config`.

yanganto (dismissed):
Looks good, thanks.

Issue #433, opened 2021-09-23
[Install] Support MacOS


juntao:
## Motivation

Support MacOS in the install script

## Details

1 The script detects that the OS is Mac OS X.

2 It checks whether `brew` is installed. If not, ask the user to install `brew` and quit. https://brew.sh/

3 `brew install llvm`

4 Follow the regular process to install `WasmEdge-0.8.2-darwin_arm64.tar.gz` for M1 Macs and `WasmEdge-0.8.2-darwin_x86_64.tar.gz` for Intel Macs.

NOTE: We do NOT install extensions for `-e all` under Mac OS X. It is the same behavior as Linux aarch64.

SAtacker:
I am closing this for now. Thanks.

PR #452, opened 2021-10-01
[Utils] Scripts:MacOS support

SAtacker:
* Added support for macOS 
* Suggestions given at #433

### DIFF
--- a/utils/install.sh
+++ b/utils/install.sh
@@ -505,6 +505,18 @@ main() {
         _shell_="${SHELL#${SHELL%/*}/}"
         _shell_rc=".""$_shell_""rc"
 
+        if [[ "$_shell_" =~ "zsh" ]]; then
+            local _grep=$(cat "$__HOME__/.zprofile" 2>/dev/null | grep "$IPATH/env")
+            if [ "$_grep" = "" ]; then
+                echo "$_source" >>"$__HOME__/.zprofile"
+            fi
+        elif [[ "$_shell_" =~ "bash" ]]; then
+            local _grep=$(cat "$__HOME__/.bash_profile" 2>/dev/null | grep "$IPATH/env")
+            if [ "$_grep" = "" ]; then
+                echo "$_source" >>"$__HOME__/.bash_profile"
+            fi
+        fi
+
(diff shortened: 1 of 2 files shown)


SAtacker:
Testing
Installation
```bash
wget -qO- https://raw.githubusercontent.com/SAtacker/WasmEdge/macos_support/utils/install.sh | bash
```

```bash
wget -qO- https://raw.githubusercontent.com/SAtacker/WasmEdge/macos_support/utils/install.sh | bash -s -- -e all
```

Uninstallation
```bash
bash <(wget -qO- https://raw.githubusercontent.com/SAtacker/WasmEdge/macos_support/utils/uninstall.sh)
```

```bash
bash <(wget -qO- https://raw.githubusercontent.com/SAtacker/WasmEdge/macos_support/utils/uninstall.sh) -q
```

juntao:
I tried it on a Mac without LLVM. It tried to `brew install llvm`, but once brew is done, the installer exits.

I think we should simply check for LLVM, and it is not available, just ask the user to run `brew install llvm` themselves, and exit.

juntao:
I tested on a Mac. It seems that the Mac needs to put the `env` line into several more places as newer Macs use zsh instead of bash.

* .bash_profile
* .zshrc
* .zprofile

SAtacker:
The latest commit detects the shell and puts it in the right configuration file
Except for fish, others should be working fine right now.
In my VM it gets written to .zshrc and . profile.

juntao:
Thanks. Currently, on Mac, if I start a new shell after install, it will not be able to find `wasmedge`. I think we need to use shell-specific profile files instead of the simple `.profile`. For example:

* `.bash_profile` for bash
* `.zprofile` for zsh

The reason is that if `.bash_profile` is present, the system would not read `.profile`.

See 

https://stackoverflow.com/questions/7780030/how-to-fix-terminal-not-loading-bashrc-on-os-x-lion
https://www.baeldung.com/linux/bashrc-vs-bash-profile-vs-profile

SAtacker:
Thanks, @juntao for this.
Commit on the way

juntao (approved):
I tested and it looks good now.

PR #517, opened 2021-10-21
[Proposal] Wasm-signature Proposal

XingZYu:
This is a an initial commit of Wasm-signature proposal implementation. I've made some changes to commane line interface so that `sign` and `verify` subcommands can be used and a template implementation of `Signature` class is serving as `SignatureEngine` in `vm` module.


hydai:
Since this PR is outdated. Closing this PR for now.

PR #636, opened 2021-11-09
[Loader] Filter and print the error message when load AOT mode WASM from buffer.

q82419:


lib/loader/loader.cpp
```diff
@@ -127,7 +127,18 @@ Loader::parseModule(const std::filesystem::path &FilePath) {
...
 Loader::parseModule(Span<const uint8_t> Code) {
-  auto Mod = std::make_unique<AST::Module>();
+  /// Filter out the MacOS or Linux AOT compiled WASM.
+  std::vector<Byte> ELFMagic = {0x7F, 0x45, 0x4C, 0x46};
+  std::vector<Byte> MacMagic = {0xCF, 0xFA, 0xED, 0xFE};
+  if (Code.size() >= 4 &&
+      (std::equal(ELFMagic.begin(), ELFMagic.end(), Code.begin()) ||
+       std::equal(MacMagic.begin(), MacMagic.end(), Code.begin()))) {
+    spdlog::error(ErrCode::MalformedMagic);
+    spdlog::error("    For the AOT compiled WASM loading, WasmEdge only "
```
hydai, line 137: Since we have universal wasm format (wasm + AOT compiled shared library as a custom section). This error message is not correct.

hydai (changes requested):

lib/loader/loader.cpp
q82419: Do you have any suggestion about the error sentence?

q82419: fixed.

lib/loader/loader.cpp
```diff
@@ -127,7 +127,20 @@ Loader::parseModule(const std::filesystem::path &FilePath) {
...
+  /// Filter out the MacOS or Linux AOT compiled WASM.
+  std::vector<Byte> ELFMagic = {0x7F, 0x45, 0x4C, 0x46};
+  std::vector<Byte> MacMagic = {0xCF, 0xFA, 0xED, 0xFE};
+  if (Code.size() >= 4 &&
+      (std::equal(ELFMagic.begin(), ELFMagic.end(), Code.begin()) ||
+       std::equal(MacMagic.begin(), MacMagic.end(), Code.begin()))) {
+    spdlog::error(ErrCode::MalformedMagic);
+    spdlog::error(
+        "    The AOT compiled WASM shared library is not supported for loading "
+        "from memory. Please use the universal Wasm binary or pure WASM, or "
```
hydai, line 139: `Wasm` or `WASM`?

hydai (changes requested):

hydai (approved):

PR #640, opened 2021-11-11
[Docker] Update to the latest NAPI packages

juntao:
The NAPI install script already calls the standard WasmEdge install script. There is no more need to call the standard script again.

Signed-off-by: Michael Yuan <michael@secondstate.io>


hydai:
This is replaced by #901. So I am closing this now.

PR #641, opened 2021-11-11
[Rust] host function

### LINKED ISSUE #422: [Rust SDK] Support adding host functions
## Motivation

Support adding Rust-based host functions to the runtime

## Details

In the Go and C SDKs, we have demonstrated that it is possible to add host functions to the WasmEdge runtime, which allows WebAssembly programs to call functions written in C or Go. We need to support that in Rust SDK as well

## Appendix

* The Go SDK host function example: https://github.com/second-state/WasmEdge-go-examples/tree/master/go_ExternRef

* The C SDK host function: https://github.com/WasmEdge/WasmEdge/blob/master/docs/c_api.md#Host-Functions

yanganto:
fix #422


bindings/rust/wasmedge-sys/src/io/input2.rs
```diff
@@ -0,0 +1,26 @@
...
+pub struct Input2<A, B> {
+    inner: Vec<Value>,
+    phantom_a: PhantomData<A>,
+    phantom_b: PhantomData<B>,
+}
+
+impl<A, B> WasmFnIO for Input2<A, B> {
+    fn parameters() -> Vec<Value> {
+        match (std::any::type_name::<A>(), std::any::type_name::<B>()) {
+            ("i32", "i32") => vec![Value::I32(0), Value::I32(0)],
```
ZhangHanDong, line 14: Only two input parameters are supported?

yanganto: > Only two input parameters are supported?


First, still in drafting.  
Second, If you want to help, write a macro for this, but the API is still changing so far.
Less but not least,  be mannered because this is a draft and not asking for help so far.

ZhangHanDong: I know this is a draft, this part I had some ambiguities in the previous implementation, just communicating with you some questions.

ZhangHanDong: Is the exchange of ideas for realization not allowed here？

ZhangHanDong: I just want to learn from your realization

ZhangHanDong: I'd better wait until you're done with it. 😑

yanganto: Sorry about this.

I am struggling with the interface, so I am not ready to answer that when the interface is not ready.  
Besides, there is also some issue `Extraref` pointer for a function. 
Everything is still changing in a big, the CI help me exam my code under development in the draft, such that I can develop without fully setting up the local environment.
The part of the interface may be much ready now, the implementation will be `I1` to `I32` for the users and these will be enough for users.

And it is still good for me to discuss it, when this branch is ready to review, just in case we are discussing something interface under iterating and does not work at all, such that we can avoid wasting time and afford not on the draft under iterating.

bindings/rust/wasmedge-sys/src/instance/function.rs
```diff
@@ -0,0 +1,162 @@
...
+            std::slice::from_raw_parts(
+                params,
+                param_len
+                    .try_into()
+                    .expect("len of params should not greater than usize"),
+            )
+        };
+        raw_input.iter().map(|r| (*r).into()).collect()
+    };
+    input.remove(0);
```
hydai, line 31: What does this `0` mean?

bindings/rust/wasmedge-sys/src/instance/function.rs
```diff
@@ -0,0 +1,162 @@
...
+    let return_len = return_len
+        .try_into()
+        .expect("len of returns should not greater than usize");
+    let raw_returns = unsafe { std::slice::from_raw_parts_mut(returns, return_len) };
+
+    HOST_FUNCS.with(|f| {
+        let mut host_functions = f.borrow_mut();
+        let real_fn = host_functions
+            .remove(&key)
+            .expect("host function should there");
```
hydai, line 42: `should be there` or something? I have no idea about this error message.

bindings/rust/wasmedge-sys/src/instance/function.rs
```diff
@@ -0,0 +1,162 @@
...
+#[derive(Debug)]
+pub struct Function {
+    pub(crate) ctx: *mut wasmedge::WasmEdge_FunctionInstanceContext,
+    pub(crate) registed: bool,
+    ty: Option<Type>,
+}
+
+impl Function {
+    /// wasmedge::WasmEdge_FunctionInstanceCreate take C functions
+    /// This may not be implement
```
hydai, line 67: Should we keep a `TODO` for a `may not be implemented` comment?

bindings/rust/wasmedge-sys/src/instance/function.rs
```diff
@@ -0,0 +1,162 @@
...
+    // TODO:
+    // binding errors and restict the error types
+    #[allow(clippy::type_complexity)]
+    /// Binding Rust function (HostFunction) to WasmEdgeFunction
+    pub fn create_bindings<I: WasmFnIO, O: WasmFnIO>(
+        real_fn: Box<dyn Fn(Vec<Value>) -> Result<Vec<Value>, u8>>,
+    ) -> Self {
+        let mut key = 0usize;
+        HOST_FUNCS.with(|f| {
+            let mut rng = rand::thread_rng();
```
hydai, line 83: Do we need a rng number here? Or we can have a local counter?

bindings/rust/wasmedge-sys/src/instance/function.rs
```diff
@@ -0,0 +1,162 @@
...
+            let mut host_functions = f.borrow_mut();
+            assert!(
+                host_functions.len() <= host_functions.capacity(),
```
hydai, line 86: Why do we have this limitation? Does the number of host functions have a maximum limitation?

bindings/rust/wasmedge-sys/src/instance/function.rs
```diff
@@ -0,0 +1,162 @@
...
+
+#[derive(Debug)]
+pub struct Type {
+    pub(crate) ctx: *mut wasmedge::WasmEdge_FunctionTypeContext,
+}
+
+impl Type {
+    pub(crate) fn create(input: Vec<Value>, output: Vec<Value>) -> Self {
+        let raw_input = {
+            let mut _a = vec![wasmedge::WasmEdge_ValType_ExternRef];
```
hydai, line 133: Can we have a meaningful name to replace `_a` and `_b`.

hydai (changes requested):

bindings/rust/wasmedge-sys/src/lib.rs
```diff
@@ -15,13 +20,23 @@ pub mod vm;
...
 pub use value::Value;
 pub use version::{full_version, semv_version};
 pub use vm::Vm;
 
+thread_local! {
+    // TODO: allow modify capacity before running
+    #[allow(clippy::type_complexity)]
+    static HOST_FUNCS:
+      RefCell<
+        HashMap<usize, Box<dyn Fn(Vec<Value>) -> Result<Vec<Value>, u8>>>> = RefCell::new(HashMap::with_capacity(500));
```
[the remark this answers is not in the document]
yanganto: Yes, this is magic number, that means the app allow 500 host functions.

bindings/rust/wasmedge-sys/src/instance/function.rs
yanganto: The `0` one is the function reference itself, and the real input parameters start from `1`.

bindings/rust/wasmedge-sys/src/instance/function.rs
yanganto: Before the VM starting, the host function registry and the key return, we now use the key to get the host function back. So the host function is expected in the Hashmap.

bindings/rust/wasmedge-sys/src/instance/function.rs
yanganto: Yes. 
Because create api only accept `C` function, and the `C1 function is possibly not passed into the Rust method based on language limitation.

.gitignore
```diff
@@ -69,3 +69,4 @@ docs/_build
 
 .direnv
 .envrc
+flake.lock
```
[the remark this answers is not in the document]
yanganto: The answer is possible `yes` in the future.
`flake.lock` help others reproduce the developing environment of the project.
However, the flake is the unstable feature of Nix, and keep changing now.
If one does not keep use nix from now, it might be useless in the future.

bindings/rust/wasmedge-sys/src/instance/function.rs
yanganto: The 500 host functions are global variable and memory determined when program started, there are some limitation based on Rust.  It might be good to figure out this in another issue, and this might turn some rust unstable feature on, so I implement this with host function length limitation.

bindings/rust/wasmedge-sys/src/instance/function.rs
yanganto: Okay, I will change these into `head` and `tail`.

bindings/rust/wasmedge-sys/src/instance/function.rs
yanganto: The counter will become another global variable, once we have more threads or processes, there will be a lock on it.  This will be nice and lock-free in the far future without changing code, so the `rng` is decided to use here.

bindings/rust/wasmedge-sys/src/instance/function.rs
hydai: Should we have a comment to explain this? Because it's hard to figure out this from the source code directly.

bindings/rust/wasmedge-sys/src/instance/function.rs
hydai: Sure. Using `rng` + hashmap is fine.

hydai (approved):

bindings/rust/wasmedge-sys/src/io.rs
```diff
@@ -0,0 +1,2461 @@
...
+}
+
+impl<O1, O2> WasmFnIO for I2<O1, O2> {
+    fn parameters() -> Vec<Value> {
+        vec![
+            match_value(std::any::type_name::<O1>()),
+            match_value(std::any::type_name::<O2>()),
+        ]
+    }
+}
```
ZhangHanDong, line 51: This part should be generated by writing a macro in the future, and when `std::any::type_name` supports `const fn`, it should also be possible to replace it with a constant generic type.

bindings/rust/wasmedge-sys/src/lib.rs
```diff
@@ -15,13 +20,23 @@ pub mod vm;
...
 pub use version::{full_version, semv_version};
 pub use vm::Vm;
 
+thread_local! {
+    // TODO: allow modify capacity before running
+    #[allow(clippy::type_complexity)]
+    static HOST_FUNCS:
+      RefCell<
+        HashMap<usize, Box<dyn Fn(Vec<Value>) -> Result<Vec<Value>, u8>>>> = RefCell::new(HashMap::with_capacity(var("MAX_HOST_FUNC_LENGTH").map(|s| s.parse::<usize>().expect("MAX_HOST_FUNC_LENGTH should be a number")).unwrap_or(500)));
+}
```
ZhangHanDong, line 39: So, no need to consider multi-threading here? This is one of the questions I had before making this feature.

bindings/rust/wasmedge-sys/src/value.rs
```diff
@@ -8,6 +10,30 @@ pub enum Value {
     I64(i64),
     F32(f32),
     F64(f64),
+    V128(u128),
+    /// A reference to a Wasm function.
+    FuncRef(u128),
+    /// A reference to opaque data in the Wasm instance.
+    ExternRef(u128),
```
ZhangHanDong, line 17: I see that go-sdk is relatively complex in adding the implementation of the externref part, and also adding support for linear memory management for webassembly. This is another thing I was confused about while working on this feature, doesn't the Rust side need to be consistent with go-sdk?

bindings/rust/wasmedge-sys/src/lib.rs
hydai: Currently, multi-threading in WasmEdge is not yet supported. I would like to approve this first. And yes, we are going to implement multi-threading support in near future. So I think we will need to consider in the future PR. Thanks.

bindings/rust/wasmedge-sys/src/io.rs
ZhangHanDong: Ok, I can try to modify it

bindings/rust/wasmedge-sys/src/value.rs
ZhangHanDong: I thought the host function depended on the memory management capabilities of externref, but now it looks like it doesn't.

bindings/rust/wasmedge-sys/src/lib.rs
ZhangHanDong: Let me add an issue first

PR #648, opened 2021-11-12
Check file types by header instead of by extension.

q82419:


lib/loader/ldmgr.cpp
```diff
@@ -14,12 +14,9 @@ Expect<void> LDMgr::setPath(const std::filesystem::path &FilePath) {
   }
 
   const auto IntrinsicsTable = getSymbol<const void *>("intrinsics");
-  if (unlikely(!IntrinsicsTable)) {
```
hydai, line 17: I think we should keep this error checker.

hydai (changes requested):

lib/loader/ldmgr.cpp
q82419: This is for the case of AOT-WASM without intrinsics table.

q82419:
@ibmibmibm  Please help to double check that the AOT core tests for universal WASM run in AOT mode, not interpreter mode.

tools/wasmedge/wasmedgec.cpp
```diff
@@ -155,9 +155,9 @@ int main(int Argc, const char *Argv[]) {
     if (ConfGenericBinary.value()) {
       Conf.getCompilerConfigure().setGenericBinary(true);
     }
-    if (OutputPath.extension().u8string() == ".wasm"sv) {
+    if (OutputPath.extension().u8string() == ".so"sv) {
```
hydai, line 158: Looks like this is a breaking change in wasmedgec. We should add it on changelog.

hydai (changes requested):

hydai (approved):

PR #667, opened 2021-11-22
[OHOS] Support musl-libc

O3Ol:
Signed-off-by: CChristal <564429983@qq.com>


CMakeLists.txt
```diff
@@ -1,5 +1,48 @@
 # SPDX-License-Identifier: Apache-2.0
 
+option(BUILD_ON_OHOS OFF)
+
+if(BUILD_ON_OHOS)
+  #指定openharmony路径
+  set(OHOS_DIR_PATH "/home/openharmony/")
```
hydai, line 7: We should never use local path to set the `DIR_PATH`.

CMakeLists.txt
```diff
@@ -1,5 +1,48 @@
...
+
+  #指定交叉编译器路径
+  set(TOOLSCHAIN_PATH "${OHOS_DIR_PATH}prebuilts/clang/ohos/linux-x86_64/llvm/")
```
hydai, line 10: Since this should be specified by users. Maybe we should move the OHOS part to another CMakeLists.txt and then include our original one.

hydai (changes requested):

lib/host/wasi/linux.h
```diff
@@ -27,6 +27,14 @@
 #include <netinet/in.h>
 #include <sys/socket.h>
 
+#include <sys/stat.h>
```
hydai, line 30: Why do we need this extra header?

hydai (changes requested):

lib/host/wasi/linux.h
hydai: Got it.

hydai (approved):

PR #795, opened 2021-11-25
Update the test and doc of externref.

q82419:


docs/externref.md
```diff
@@ -2,6 +2,232 @@
...
+  (memory $memory (export "memory") 1))
+```
+
+Users can convert `wat` to `wasm` through [wat2wasm](https://webassembly.github.io/wabt/demo/wat2wasm/) live tool. Noted that `reference types` checkbox should be checked on this page.
+
+### Implement Host Module and Register into WasmEdge
+
+The host module should be implemented and registered into WasmEdge before executing Wasm. Assume that the following code is saved as `main.c`:
+
+```cpp#include <wasmedge/wasmedge.h>
```
hydai, line 52: Wrong format

hydai (changes requested):

hydai (approved):

PR #843, opened 2021-11-26
Fix WASI API.

q82419:


q82419:
@yanganto Please help to check the rust code. Thanks.

yanganto (approved):
LGTM.  Thanks.

lib/api/wasmedge.cpp
```diff
@@ -2174,15 +2172,9 @@ WASMEDGE_CAPI_EXPORT void WasmEdge_ImportObjectInitWASI(
...
-  if (Dirs) {
-    for (uint32_t I = 0; I < DirLen; I++) {
-      DirVec.emplace_back(Dirs[I]);
-    }
-  }
   if (Preopens) {
     for (uint32_t I = 0; I < PreopenLen; I++) {
-      DirVec.emplace_back(std::string(Preopens[I]) + ":" +
-                          std::string(Preopens[I]));
+      DirVec.emplace_back(Preopens[I]);
```
hydai, line 2177: We should have two different cases:

1. `PATH1:PATH2`: map `PATH1` to `PATH2`.
2. `PATH`: map `PATH` to `PATH`

However, this section cannot present these two behaviour.

hydai (changes requested):

hydai:
You should also update the changelog since this is a API breaking change.

lib/api/wasmedge.cpp
q82419: This is actually handled in WASI:

https://github.com/WasmEdge/WasmEdge/blob/master/lib/host/wasi/environ.cpp#L58

hydai: I don't think so. Looks like the `environ.cpp` only handles the `PATH1:PATH2` part. It doesn't create the mapping of second one.

hydai (approved):

yanganto (approved):

PR #844, opened 2021-11-26
[Rust] Module load from buffer

### LINKED ISSUE #763: [Rust SDK] WasmEdge_LoaderParseFromBuffer

yanganto:
Fix #763

### DIFF
--- a/bindings/rust/wasmedge-sys/src/module.rs
+++ b/bindings/rust/wasmedge-sys/src/module.rs
@@ -1,5 +1,6 @@
 use super::wasmedge;
 use crate::raw_result::{decode_result, ErrReport};
+use ::core::mem::MaybeUninit as MU;
 use std::ffi::CString;
 
 #[derive(Debug)]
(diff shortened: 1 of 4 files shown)


bindings/rust/wasmedge-sys/src/module.rs
```diff
@@ -30,4 +31,36 @@ impl Module {
...
+        config: &crate::config::Config,
+        buffer: Vec<u8>,
+    ) -> Result<Self, ErrReport> {
+        let loader_ctx = unsafe { wasmedge::WasmEdge_LoaderCreate(config.ctx) };
+        let mut ctx: *mut wasmedge::WasmEdge_ASTModuleContext = std::ptr::null_mut();
+
+        assert!(!buffer.is_empty(), "WasmEdge fail to load an empty buffer");
+
+        let ptr = unsafe {
+            let ptr = libc::malloc(buffer.len());
```
hydai, line 45: Do we need to free this memory before exiting this function?

hydai (changes requested):

bindings/rust/wasmedge-sys/src/module.rs
yanganto: Yes, we need to delete it, and there is a test to make sure this buffer can be deleted, and the VM still works.  
Thanks.

hydai (approved):

PR #862, opened 2021-12-06
Update the spec test and turn on SIMD.

q82419:


tools/wasmedge/wasmedger.cpp
```diff
@@ -125,11 +125,11 @@ int main(int Argc, const char *Argv[]) {
     Conf.removeProposal(WasmEdge::Proposal::ReferenceTypes);
   }
   if (PropSIMD.value()) {
-    Conf.addProposal(WasmEdge::Proposal::SIMD);
-  }
-  if (PropAll.value()) {
-    Conf.addProposal(WasmEdge::Proposal::SIMD);
+    Conf.removeProposal(WasmEdge::Proposal::SIMD);
   }
+  /// Left for the future proposals.
```
hydai, line 130: Why do you comment these out?

hydai (changes requested):

tools/wasmedge/wasmedger.cpp
q82419: We've turned on the all proposals we implemented now.

hydai: Why not delete them directly?

q82419: I think we should keep this for the proposals we will implement and not merge into the official spec in the future?

hydai (approved):

PR #867, opened 2021-12-07
[CI] Enable commit message checker

hydai:


hydai:
Because the rule is not flexible. I will abandon this PR and look for a better tool.

PR #876, opened 2021-12-08
[Rust] build wasmedge from source if not exist in system

### LINKED ISSUE #869: [Rust] build wasmedge from source
## Motivation
Rust allow user to use git link as dependency and `wasmedge-sys` did not support this, because the build script use wasmedge already in the system now.

## Details
- Build wasmedge in the build script in `wasmedge-sys` crate

## Appendix
derive from #863

yanganto:
Fix #869 

- If there is no wasmedge in the system, the standalone wasmedge will build for RustSDK.
- If the `standalone` feature enabled, RustSDK will always use standalone build, no matter there is another `wasmedge` in system or not.

Signed-off-by: Antonio Yang <yanganto@gmail.com>


.github/workflows/rust-standalone.yml
```diff
@@ -0,0 +1,49 @@
...
+concurrency:
+  group: rust-standalone-${{ github.head_ref }}
+  cancel-in-progress: true
+
+on:
+  push:
+    branches:
+      - master
+    paths-ignore:
+      - 'doc/**'
```
hydai, line 12: Should be `docs`

.github/workflows/rust-standalone.yml
```diff
@@ -0,0 +1,49 @@
...
+      - '**.md'
+  pull_request:
+    branches:
+      - master
+    paths-ignore:
+      - 'doc/**'
```
hydai, line 18: Ditto.

bindings/rust/wasmedge-sys/build.rs
```diff
@@ -1,13 +1,104 @@
...
+        .expect("fail to checkout the commit");
+
+    Command::new("cmake")
+        .current_dir(&wasmedge_src)
+        .args(&[
+            "-Bbuild",
+            "-GNinja",
+            "-DCMAKE_BUILD_TYPE=Release",
+            "-DWASMEDGE_BUILD_TESTS=ON",
+            "-DWASMEDGE_BUILD_AOT_RUNTIME=OFF",
```
hydai, line 44: If you disabled the AoT runtime, you should also disable the aot related features in rust-sdk

hydai (changes requested):

.github/workflows/rust-standalone.yml
```diff
@@ -0,0 +1,49 @@
+name: rust
```
hydai, line 1: Also, I think this should be merged into the original rust workflow. So it will not have multiple workflow with the same bindinges.

hydai (changes requested):

bindings/rust/wasmedge-sys/build.rs
yanganto: Thanks, AOT feature is added and enabled in default

.github/workflows/rust-standalone.yml
yanganto: Thanks.

.github/workflows/rust.yml
```diff
@@ -9,13 +9,13 @@ on:
     branches:
```
hydai, line 9: Maybe we can also rename this workflow? Say `"Rust bindings"`?

hydai (changes requested):

.github/workflows/bindings-rust.yml
```diff
@@ -1,21 +1,21 @@
-name: rust
+name: binding
```
hydai, line 1: binding-rust

hydai (changes requested):

hydai (approved):

PR #917, opened 2021-12-22
[Rust] Added doc comments for Compiler

apepkuss:
This PR is to add the necessary descriptions for Compiler struct.


bindings/rust/wasmedge-sys/src/compiler.rs
```diff
@@ -24,6 +31,17 @@ impl Compiler {
         }
     }
 
+    /// Compiles the input WASM from the given file path.
+    ///
+    /// # Arguments
+    ///
+    /// - `in_path` specifies the input WASM file path.
+    ///
+    /// - `out_path` specifies the output WASM file path.
```
hydai, line 40: It depends.
The default extension is `.wasm` which represents the universal WASM file.
However, if you specify the `.so` extension, then you will get a native binary directly.

hydai (changes requested):

bindings/rust/wasmedge-sys/src/compiler.rs
apepkuss: For this version of the doc comment, it follows [the descriptions](https://github.com/WasmEdge/WasmEdge/blob/master/include/api/wasmedge/wasmedge.h#L1203-L1215) of the corresponding C APIs. We also noticed the issue and are preparing a list of questions for those ambiguous or unclear concepts, descriptions, and comments in [wasmedge.h](https://github.com/WasmEdge/WasmEdge/blob/master/include/api/wasmedge/wasmedge.h#L1203-L1215). Then, we'll refine the doc comments with the more detailed and clearer descriptions.

hydai: cc @q82419 
I believe we should have the same statement in the wasmedge.h

apepkuss: Hi @hydai @q82419, I'm still confused about the C-API [WasmEdge_CompilerCompile](https://github.com/WasmEdge/WasmEdge/blob/master/include/api/wasmedge/wasmedge.h#L1203-L1216). If the input is a wasm file "input.wasm" and the output's specified as "output.wasm", it should be reasonable according to the doc comments. My question is, they're both ".wasm", what are the differences or changes before and after the "compile" op? Thanks a lot!

apepkuss: @q82419 One more thing I need to confirm is the argument of the C-API `WasmEdge_CompilerCreate`. The doc comments do not cover the argument `ConfCxt`, so I'd like to confirm it's an optional or required argument. Thanks!

hydai: > Hi @hydai @q82419, I'm still confused about the C-API [WasmEdge_CompilerCompile](https://github.com/WasmEdge/WasmEdge/blob/master/include/api/wasmedge/wasmedge.h#L1203-L1216). If the input is a wasm file "input.wasm" and the output's specified as "output.wasm", it should be reasonable according to the doc comments. My question is, they're both ".wasm", what are the differences or changes before and after the "compile" op? Thanks a lot!

We have universal wasm binary format, which will embed native code into a custom section inside the origin wasm file.

hydai: > @q82419 One more thing I need to confirm is the argument of the C-API `WasmEdge_CompilerCreate`. The doc comments do not cover the argument `ConfCxt`, so I'd like to confirm it's an optional or required argument. Thanks!

Required.

hydai (approved):

bindings/rust/wasmedge-sys/src/compiler.rs
apepkuss: @hydai Thank you very much for the clear explanation!

> > Hi @hydai @q82419, I'm still confused about the C-API [WasmEdge_CompilerCompile](https://github.com/WasmEdge/WasmEdge/blob/master/include/api/wasmedge/wasmedge.h#L1203-L1216). If the input is a wasm file "input.wasm" and the output's specified as "output.wasm", it should be reasonable according to the doc comments. My question is, they're both ".wasm", what are the differences or changes before and after the "compile" op? Thanks a lot!
> 
> We have universal wasm binary format, which will embed native code into a custom section inside the origin wasm file.

PR #919, opened 2021-12-22
[Rust] Added doc comments for Statistics

apepkuss:
This PR is to add the necessary descriptions for the Statistics struct.

### DIFF
--- a/bindings/rust/wasmedge-sys/src/statistics.rs
+++ b/bindings/rust/wasmedge-sys/src/statistics.rs
@@ -85,7 +85,6 @@ mod tests {
         let stat = result.unwrap();
 
         // check instruction count
-        let count = stat.instr_count();
         assert_eq!(stat.instr_count(), 0);
 
         // check instruction count per second
--- a/bindings/rust/wasmedge-sys/src/statistics.rs
+++ b/bindings/rust/wasmedge-sys/src/statistics.rs
@@ -32,6 +32,26 @@ impl Statistics {
     }
 
     /// Returns the instruction count per second in execution.
+    ///
+    /// # Notice
+    ///
+    /// For the following cases,
+    /// - [`Statistics`] is not enabled, or
+    /// - the total execution time is 0
+    ///
+    /// the instructions per second is `NaN`, which represents `divided-by-zero`.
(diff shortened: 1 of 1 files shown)


bindings/rust/wasmedge-sys/src/statistics.rs
```diff
@@ -56,3 +72,27 @@ impl Drop for Statistics {
...
+        assert!(result.is_ok());
+        let stat = result.unwrap();
+
+        // check instruction count
+        assert_eq!(stat.instr_count(), 0);
+
+        // check instruction count per second
+        let count = stat.instr_per_sec();
+        assert!(count.is_nan());
+        // assert!(count, 0.0);
```
hydai, line 93: Instruction per second  = Total executed instructions / total execution time.
If the statistics is not enabled or the total execution time is `0.0`. Then, the instruction per second should be `Nan` which represents something is divided by zero.

hydai (changes requested):

bindings/rust/wasmedge-sys/src/statistics.rs
apepkuss: For the statement, "If the statistics is not enabled ...", what does "enabled" mean? does it mean the statistics is enabled in a **VM** or something else? Thanks!

apepkuss: The doc comment of the `instr_per_sec` function has been refactored by adding the description on the cases, in which `NaN` is returned. Thanks!

hydai: You can enabled it in the Configure. And the VM/executor will save information to the Statistics module. However, if the VM/executor is not executed before, it should keep empty.

hydai (approved):

bindings/rust/wasmedge-sys/src/statistics.rs
apepkuss: @hydai You're right, `Config` has an option controlling it. Our motivation for this test code is just to verify a corner case, namely, instantiate an individual `Statistics`, and then verify the return values of the APIs of interest. Many thanks for your detailed explanation.

PR #927, opened 2021-12-23
[Utils] Scripts: Add bashrc for install script

chenyukang:
In current install script, it will generate a new `.bash_profile`, which introduce my bash configuration not take affect. In system like Ubuntu, we only have `.bashrc` by default.

This fix will expect `.bashrc`, if `.bashrc` does not exists, it will then work as before.
And make the code shorter btw.


chenyukang:
In general, I don't think generate new configuration files for user, it may broke something and make user confused in that scenario. 

I's better to add configuration to existing configuration files, if we didn't add it successfully, let user source the env by themself.

Do you think we need to fix this?
@hydai

hydai:
> In general, I don't think generate new configuration files for user, it may broke something and make user confused in that scenario.
> 
> I's better to add configuration to existing configuration files, if we didn't add it successfully, let user source the env by themself.
> 
> Do you think we need to fix this? @hydai

I would like to invite @SAtacker to join the discussion.

Personally, I prefer not to break original expected behaviour of users. For examples, `homebrew`, `nvm`, and `gvm` will try to append the `source command` to the original bashrc/zshrc.

SAtacker:
> > In general, I don't think generate new configuration files for user, it may broke something and make user confused in that scenario.
> > I's better to add configuration to existing configuration files, if we didn't add it successfully, let user source the env by themself.
> > Do you think we need to fix this? @hydai
> 
> I would like to invite @SAtacker to join the discussion.
> 
> Personally, I prefer not to break original expected behaviour of users. For examples, `homebrew`, `nvm`, and `gvm` will try to append the `source command` to the original bashrc/zshrc.

Thanks for notifying me. Initially, it was done so because it was requested that we create a `~/.bash_profile`.
I agree with @chenyukang that we should not create `~/.bash_profile` on our own. If it is not present in the system probably bash is not the login shell. When Bash is invoked as an interactive non-login shell, it reads and executes commands from `~/.bashrc`, if that file exists, and it is readable.
Generally, we put the commands that should run every time a new shell is opened in ~/.bashrc` and those that require to be run only once in `~/.bash_profile`.

utils/install.sh
```diff
@@ -576,12 +576,10 @@ main() {
...
-        if [ "$_grep" = "" ]; then
-            echo "$_source" >>"$__HOME__/.profile"
+    if [ "$_grep" = "" ]; then
+        if [ ! -f "$__HOME__/.profile" ]; then
+            echo "Generating $__HOME__/.profile"
         fi
-    else
-        echo "Generating $__HOME__/.profile"
         echo "$_source" >>"$__HOME__/.profile"
     fi
```
SAtacker, line 584: I think this refactoring does the same job but it's more readable.

SAtacker (approved):

chenyukang:
> > > In general, I don't think generate new configuration files for user, it may broke something and make user confused in that scenario.
> > > I's better to add configuration to existing configuration files, if we didn't add it successfully, let user source the env by themself.
> > > Do you think we need to fix this? @hydai
> > 
> > 
> > I would like to invite @SAtacker to join the discussion.
> > Personally, I prefer not to break original expected behavior of users. For examples, `homebrew`, `nvm`, and `gvm` will try to append the `source command` to the original bashrc/zshrc.
> 
> Thanks for notifying me. Initially, it was done so because it was requested that we create a `~/.bash_profile`. I agree with @chenyukang that we should not create `~/.bash_profile` on our own. If it is not present in the system probably bash is not the login shell. When Bash is invoked as an interactive non-login shell, it reads and executes commands from `~/.bashrc`, if that file exists, and it is readable. Generally, we put the commands that should run every time a new shell is opened in `~/.bashrc` and those that require to be run only once in `~/.bash_profile`.

I will rework the PR to only append to existing configuration files.

SAtacker:
> > > > In general, I don't think generate new configuration files for user, it may broke something and make user confused in that scenario.
> > > > I's better to add configuration to existing configuration files, if we didn't add it successfully, let user source the env by themself.
> > > > Do you think we need to fix this? @hydai
> > > 
> > > 
> > > I would like to invite @SAtacker to join the discussion.
> > > Personally, I prefer not to break original expected behavior of users. For examples, `homebrew`, `nvm`, and `gvm` will try to append the `source command` to the original bashrc/zshrc.
> > 
> > Thanks for notifying me. Initially, it was done so because it was requested that we create a `~/.bash_profile`. I agree with @chenyukang that we should not create `~/.bash_profile` on our own. If it is not present in the system probably bash is not the login shell. When Bash is invoked as an interactive non-login shell, it reads and executes commands from `~/.bashrc`, if that file exists, and it is readable. Generally, we put the commands that should run every time a new shell is opened in `~/.bashrc` and those that require to be run only once in `~/.bash_profile`.
> 
> I will rework the PR to only append to existing configuration files.
> 

I think it does that right now. I tested your branch. It doesn't create additional `.bash_profile`.

utils/install.sh
chenyukang: > I think this refactoring does the same job but it's more readable.

Yes, it's shorter.

chenyukang:
> > > > > In general, I don't think generate new configuration files for user, it may broke something and make user confused in that scenario.
> > > > > I's better to add configuration to existing configuration files, if we didn't add it successfully, let user source the env by themself.
> > > > > Do you think we need to fix this? @hydai
> > > > 
> > > > 
> > > > I would like to invite @SAtacker to join the discussion.
> > > > Personally, I prefer not to break original expected behavior of users. For examples, `homebrew`, `nvm`, and `gvm` will try to append the `source command` to the original bashrc/zshrc.
> > > 
> > > 
> > > Thanks for notifying me. Initially, it was done so because it was requested that we create a `~/.bash_profile`. I agree with @chenyukang that we should not create `~/.bash_profile` on our own. If it is not present in the system probably bash is not the login shell. When Bash is invoked as an interactive non-login shell, it reads and executes commands from `~/.bashrc`, if that file exists, and it is readable. Generally, we put the commands that should run every time a new shell is opened in `~/.bashrc` and those that require to be run only once in `~/.bash_profile`.
> > 
> > 
> > I will rework the PR to only append to existing configuration files.
> 
> I think it does that right now. I tested your branch. It doesn't create additional `.bash_profile`.

Not only `.bash_profile`, the current script will also create other files like:

```bash
echo "$_source" >>"$__HOME__/.zprofile"
...
echo "$_source" >>"$__HOME__/.profile"
...
```
What I mean is, test whether these files exist, if yes, append at the end of them.
Otherwise, don't do anything and print out the path to let user source it.

SAtacker:
I agree with this but apparently, there was an issue. Please have a look at https://github.com/WasmEdge/WasmEdge/pull/452/commits/8ce5d0c37ce9345cd5d3cf810ad4a1f486c0f51d.

chenyukang:
> > What I mean is, test whether these files exist, if yes, append at the end of them. Otherwise, don't do anything and print out the path to let user source it.
> 
> I agree with this but apparently, there was an issue. Please have a look at [8ce5d0c](https://github.com/WasmEdge/WasmEdge/commit/8ce5d0c37ce9345cd5d3cf810ad4a1f486c0f51d) and [#452 (comment)](https://github.com/WasmEdge/WasmEdge/pull/452#issuecomment-933168757).

emm, didn't notice it. 
If so, let's keep it that way.

hydai (approved):

chenyukang:
Seems CI is not stable:
```
unable to access 'https://github.com/second-state/WasmEdge-tensorflow-deps.git/': gnutls_handshake() failed: Error in the pull function.
```

hydai:
> Seems CI is not stable:
> 
> ```
> unable to access 'https://github.com/second-state/WasmEdge-tensorflow-deps.git/': gnutls_handshake() failed: Error in the pull function.
> ```

I am going to merge this first. Seems like that's an issue from GitHub.

hydai (approved):

PR #928, opened 2021-12-23
[Tools] Enable --gas-limit for wasmedge cli

hydai:

### DIFF
--- a/include/common/statistics.h
+++ b/include/common/statistics.h
@@ -11,8 +11,10 @@
 //===----------------------------------------------------------------------===//
 #pragma once
 
+#include "common/configure.h"
 #include "common/enum_ast.h"
 #include "common/errcode.h"
+#include "common/log.h"
 #include "common/span.h"
 #include "common/timer.h"
 
@@ -77,6 +79,7 @@ public:
     CostSum += Cost;
     if (unlikely(CostSum > CostLimit)) {
       CostSum = CostLimit;
+      spdlog::error("Cost exceeded limit. Force terminate the execution.");
       return false;
     }
     return true;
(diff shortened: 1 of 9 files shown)


juntao:
Let's deprecate pages like `docs/run.md` and use the `book` as the single source of truth. Thanks.

https://github.com/WasmEdge/WasmEdge/blob/master/docs/book/en/src/start/cli.md

hydai:
> Let's deprecate pages like `docs/run.md` and use the `book` as the single source of truth. Thanks.
> 
> https://github.com/WasmEdge/WasmEdge/blob/master/docs/book/en/src/start/cli.md

Sure. Should I also redirect these pages to the new one?

juntao:
I do not think GitHub will do http redirect? Perhaps we can remove the content in the md file and put a link to the book's html page there for people to click on?

hydai:
That's what I want. Put a link.
However, I am not sure which one is better: put a link to the webpage or to the new markdown file?

juntao:
> That's what I want. Put a link. However, I am not sure which one is better: put a link to the webpage or to the new markdown file?

I think web page is better since the visitor is typically looking to read this content. The HTML page has an "edit" button that brings the reader directly to the MD source file.

include/common/statistics.h
```diff
@@ -131,6 +134,40 @@ class Statistics {
            TimeRecorder.getRecord(Timer::TimerTag::HostFunc);
   }
 
+  void dumpToLog(const Configure &Conf) const noexcept {
+    auto Nano = [](auto &&Duration) {
+      return std::chrono::nanoseconds(Duration).count();
+    };
+    if (Conf.getStatisticsConfigure().isTimeMeasuring() ||
```
ibmibmibm, line 141: `const auto &StatConf = Conf.getStatisticsConfigure();`

ibmibmibm (changes requested):

ibmibmibm (approved):

PR #936, opened 2021-12-23
[CI] Limit the MDBOOK action to merged book file updates

juntao:
Auto triggers only occur when there is a push event in the docs/book folder

Signed-off-by: Michael Yuan <michael@michaelyuan.com>


hydai (changes requested):
We should keep the `build` process when receiving a PR.
But we can disable the `publish` process with `if: github.event_name != 'pull_request'` in the `publish` job phase.

hydai (approved):

PR #958, opened 2021-12-29
[Utils] Remove wget --show-progress option for supporting wget 1.14

hydai:
Some of our users are using CentOS 7, which only has wget 1.14.
The `--show-progress` is a "new" feature in wget 1.16, which is released in 2014/10. However, CentOS 7 doesn't support such a "new" version.


hydai:
This is replaced by #967

Issue #966, opened 2022-01-02
[Install] wget or curl


juntao:
It seems that many projects are using `curl`, instead of `wget`, for their one-line install script -- possibly because `curl` is available by default on more platforms. 

https://daniel.haxx.se/docs/curl-vs-wget.html

What do we think? Shall we switch to `curl`?

SAtacker:
Currently the script supports both. If you mean we prefer to check first for curl and then for wget it's quite doable. If curl is more preferred then I'll make a PR.
Also including the changes in the installation documentation.
Thanks for this.

PR #967, opened 2022-01-02
[Utils] Install Script: Changed first default and wget options

SAtacker:
* [SAtacker] First checks if curl is installed else checks for wget
* [hydai] If wget has progress-bar option goes with it else it does not
* Suggestions from https://github.com/WasmEdge/WasmEdge/issues/966
and https://github.com/WasmEdge/WasmEdge/pull/958

Co-authored-by: hydai <hydai@secondstate.io>

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -12,13 +12,13 @@
 The easiest way to install WasmEdge is to run the following command. Your system should have `git` and `wget` as prerequisites.
 
 ```bash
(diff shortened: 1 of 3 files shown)


juntao (approved):
I think it is better to default to `curl` since it is available on more systems, and most of our related tools (eg Rust and Node) are installed with `curl` -- so the users are more likely to have it already.

hydai (approved):

PR #980, opened 2022-01-05
[Rust] Fixed a bug in the get_data func

apepkuss:
Two fixes are introduced in this PR:
- The `get_data` of `Table` used an inappropriate way to access data via the low-level API.
- Added a new branch in the `find_wasmedge` function in build.rs for searching wasmedge.h in WasmEdge official docker dev environment.
- Removed the `version` module which was removed in the previous PR, but was introduced again by a recent merge conflict resolution.


bindings/rust/wasmedge-sys/build.rs
```diff
@@ -174,6 +174,16 @@ fn find_wasmedge() -> Option<Paths> {
             lib_dir: PathBuf::from("/usr/lib"),
             inc_dir: PathBuf::from("/usr/include"),
         })
+    } else if PathBuf::from("/root/.wasmedge/include/wasmedge")
```
hydai, line 177: Please use the env variable of `HOME` or set a new env variable `WASMEDGE_DIR` to access the path.
The `/root` is a very specific and non-general value, so we should avoid this.

hydai (changes requested):

bindings/rust/wasmedge-sys/build.rs
apepkuss: Accepted. Thanks a lot!

apepkuss: The issue has been fixed. Thanks!

hydai:
Your test is still failed.

apepkuss:
> Your test is still failed.

Yes. I found that a test case related to `Table` in this PR works correctly with release 0.9.0, but fails in the CI environment. The CI uses the master branch, right?

apepkuss:
Hi @hydai, I tried the test code against the release 0.9.0 and the master branch in my local environment (docker image appdev_aarch64:0.9.0), and got the different results. With the failure of the CI checks, I filed an issue #988. Could you please double check it? Thanks a lot!

apepkuss:
Hi @hydai, the previous issue has been fixed by the merged #982, please help review the latest change. Thanks!

hydai (approved):

PR #992, opened 2022-01-06
Fix the 32-bit platform issues.

q82419:


lib/system/allocator.cpp
```diff
@@ -7,15 +7,8 @@
 #include "common/defines.h"
 #include "common/errcode.h"
 
-#include <cassert>
-#include <cctype>
```
hydai, line 11: Why do you remove lots of headers which are used in this file.

hydai (changes requested):

lib/system/allocator.cpp
q82419: These headers are not used.

hydai (approved):

PR #993, opened 2022-01-06
[CI] Add Windows installer release CI

KernelErr:
Signed-off-by: KernelErr <45716019+KernelErr@users.noreply.github.com>

Actions added would do the following things:

- Use WiX Toolset to create MSI installer
- Upload MSI to GitHub Release

Installer's behaviors:

- Put files to `C:\Program Files\WasmEdge` (Files are the same as those in the zip file)
- Add `C:\Program Files\WasmEdge\bin` to system's PATH

Version Number Rule:

- Nightly: WasmEdge-0.9.0.92-windows.msi - Tag with later commits count
- Release: WasmEdge-0.9.0-windows.msi - Only Tag

After having MSI in GitHub release, I would add WasmEdge to winget packages.


hydai:
Please update the `build.yml`, so we can have the nightly build of the msi file.

REF: https://github.com/WasmEdge/WasmEdge/blob/master/.github/workflows/build.yml#L309-L323

KernelErr:
@hydai Hi, WiX Toolset requires a version string like `0.9.0` rather than `0.9.0-89-gcf4a3ece`, if there any version number we could use nightly build?

```
D:\a\WasmEdge\WasmEdge\.github\scripts\wasmedge.wxs(3) : error CNDL0108 : The Product/@Version attribute's value, '0.9.0-89-gcf4a3ece', is not a valid version.  Legal version values should look like 'x.x.x.x' where x is an integer from 0 to 65534.
```

hydai:
Actually, the version string is from `git describe --match "[0-9].[0-9]*" --tag`. So you can find that `0.9.0-89-gcf4a3ece` is composed with the base tag `0.9.0` and the number of unreleased commits and hash.

Giving an random number is not a good idea. I am going to check other projects to figure out how they work.

Ref: https://github.com/WasmEdge/WasmEdge/blob/master/.github/workflows/build.yml#L27-L41

hydai:
Hi @KernelErr 
After several survey, I prefer to use the diff number from the current version tag: `0.9.0-89-gcf4a3ece` the `89` as its fourth version number.
If you have any idea, please feel free to discuss with us.

KernelErr:
I think that's a nice idea. I've pushed the new CI file.

KernelErr:
Here is a nightly build: https://github.com/WasmEdge/WasmEdge/suites/4857746443/artifacts/138181303
Please check if there is any problem. :) @hydai 
Virus Total: https://www.virustotal.com/gui/file/19616bdcaac6fb7183b93e0623af592c22d82d857f994ae19ba1bf6448d27fc9?nocache=1

hydai:
> Here is a nightly build: https://github.com/WasmEdge/WasmEdge/suites/4857746443/artifacts/138181303 Please check if there is any problem. :) @hydai Virus Total: https://www.virustotal.com/gui/file/19616bdcaac6fb7183b93e0623af592c22d82d857f994ae19ba1bf6448d27fc9?nocache=1

Awesome. Is it possible to upload and retrieve virus report from Virtus Total in the workflow? Or we must submit it manually?

KernelErr:
Yes, it's possible to upload the installer file to Virus Total automatically. However, there are some limits for public API: https://developers.virustotal.com/reference/public-vs-premium-api

.github/workflows/release.yml
```diff
@@ -370,3 +370,19 @@ jobs:
         asset_path: build\\WasmEdge-${{ needs.create_release.outputs.version }}-Windows.zip
         asset_name: WasmEdge-${{ needs.create_release.outputs.version }}-windows.zip
         asset_content_type: application/zip
+
+    - name: Package Windows Installer
+      run: |
+        $Env:product_version = "${{ needs.create_release.outputs.version }}"
+        . "$Env:WIX\bin\candle.exe" -arch x64 -o build\wasmedge.wixobj .github\scripts\wasmedge.wxs
+        . "$Env:WIX\bin\light.exe" -out build\WasmEdge-${{ needs.create_release.outputs.version }}-windows.msi build\wasmedge.wixobj
```
hydai, line 378: Should this version be the same as the `env.product_version` in `build.yml`.

hydai (changes requested):

.github/workflows/release.yml
hydai: In the previous discussion, we find the format is `x.y.z.build`. If we only use the release tag `create_release.outputs.version`, it will be `x.y.z` only. Is this acceptable?

KernelErr: According to the Windows document, only the first three fields of version number would be actually used. So if you installed the `0.9.0` version and then install `0.9.0.1` or other versions, you will see those versions coexist, and later installed versions will replace old files. Only `x.y.z` has been changed, then the installer will remove all the old versions and install the new version. So I think it's acceptable. @hydai

hydai: Great. Then, I am merging this.

hydai (approved):

PR #997, opened 2022-01-07
Fix the memory data pointer getting timing

q82419:
The data pointer of a memory instance may changed when memory growing:
https://github.com/WasmEdge/WasmEdge/blob/master/include/runtime/instance/memory.h#L107-L112

Therefore, it's better to retrieve the data pointer right before entering the compiled functions than in the instantiate phase.


q82419:
This PR will merge with the PR #979 .

PR #1004, opened 2022-01-12
[Rust] Refactor traits I1-32 to macro

KernelErr:
Signed-off-by: KernelErr <45716019+KernelErr@users.noreply.github.com>

The original implementation is too long, we should use macro rules for it.


KernelErr:
`concat_idents` is not stabilized yet, so we use uppercase for variable names.

yanganto (approved):

yanganto:
Good catch.  Thanks.

apepkuss (approved):

Issue #1003, opened 2022-01-12
[Utils] Uninstall script does not remove include dir


KernelErr:
After using uninstall script, empty directory `~/.wasmedge/include` were found left.

![image](https://user-images.githubusercontent.com/45716019/149073862-06afb8e5-77a9-4059-bc13-c86b86b9dcdc.png)

I think we should fix this and may add script tests to CI.

SAtacker:
Thanks for this.

PR #1010, opened 2022-01-13
[Utils] Scripts: Minor improvements

SAtacker:
* Fix for https://github.com/WasmEdge/WasmEdge/issues/1003#issue-1099922443
* Minor Spelling correction

### DIFF
--- a/utils/uninstall.sh
+++ b/utils/uninstall.sh
@@ -110,7 +110,8 @@ remove_parsed() {
         IFS=$'\n'
         ask_remove $(parse_env)
         if [[ "$IPATH" =~ ".wasmedge" ]]; then
-            ask_remove $(find "$IPATH" -type d -empty -print)
+            ask_remove $(find "$IPATH" -depth -type d -empty -print)
+            ask_remove $(find "$IPATH" -depth -type d -empty -print)
             if [ -z "$(ls -A "$IPATH")" ]; then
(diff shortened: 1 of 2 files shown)


utils/uninstall.sh
```diff
@@ -110,7 +110,8 @@ remove_parsed() {
         IFS=$'\n'
         ask_remove $(parse_env)
         if [[ "$IPATH" =~ ".wasmedge" ]]; then
-            ask_remove $(find "$IPATH" -type d -empty -print)
+            ask_remove $(find "$IPATH" -depth -type d -empty -print)
+            ask_remove $(find "$IPATH" -depth -type d -empty -print)
```
hydai, line 114: Do we need to remove it twice?

hydai (changes requested):

utils/uninstall.sh
SAtacker: It looks strange but the first time it searches `~/.wasmedge/include/wasmedge` is contained inside `~/.wasmedge/include` which is why it gets excluded and we need to find for empty directory one more time.

SAtacker: Example:
image

SAtacker: image

hydai: Got it.

hydai (approved):

PR #1022, opened 2022-01-18
Fix the async API and add the unit tests and docs.

q82419:


hydai (approved):

alabulei1:
This doc should be updated on WasmEdge Book firstly.

PR #1038, opened 2022-01-22
[Misc] Remove extra word in version string for packaging with windows wix

ibmibmibm:


.github/workflows/release.yml
```diff
@@ -373,7 +373,7 @@ jobs:
 
     - name: Package Windows Installer
       run: |
-        $Env:product_version = "${{ needs.create_release.outputs.version }}"
+        $Env:product_version = ("${{ needs.get_version.outputs.version }}").split("-")[0]
```
hydai, line 376: This should be `create_release`.

hydai (changes requested):

hydai (approved):

PR #1044, opened 2022-01-23
[Android] Add Android platform support

hangedfish:
# Details
* support Android Bionic-libc, for android 6.0+.
* add Android build script.
* add Android build tutorial.

### DIFF
--- a/docs/book/zh/src/os/android.md
+++ b/docs/book/zh/src/os/android.md
@@ -1,0 +1,61 @@
+# ANDROID
+## 使用 ANDROID NDK 构建和测试 WasmEdge
+
+此教程指导使用 Android NDK 构建 Interpreter 模式的 WasmEdge，并在 Android 设备中测试 wasmedge 命令行工具。
+
+### 环境准备
+* [Android NDK 23b](https://developer.android.com/ndk/downloads)
+    * 下载到的 NDK 工具链需要解压到磁盘上
+* [CMake 3.21](https://cmake.org/download/) 及以上版本
+  * 如果本机已安装 cmake，请检查版本是否符合要求
+* [ADB](https://developer.android.com/studio/releases/platform-tools)
+  * 将下载到的 platform-tools 解压到磁盘上，adb 命令在 bin 目录中
+  * 如果您使用的是 debian/ubuntu 系统，可以通过 apt 安装 adb
+* 一台已经 [启用开发者选项和 USB 调试](https://developer.android.com/studio/debug/dev-options) 的 Android 设备, 最低系统版本为 Android 6.0    
+
+### 构建 WasmEdge
+1. 添加 NDK 目录路径到环境变量 `ANDROID_NDK_HOME=path/to/you/ndk/dir`
+2. 执行 WasmEdge 源码路径下的 utils/build_for_ohos.sh 命令行脚本，将自动执行构建，构建结果在 WasmEdge 目录中的 build 目录
+
+### 测试
+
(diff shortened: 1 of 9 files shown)


docs/book/zh/src/os/android.md
```diff
@@ -0,0 +1,61 @@
...
+    34
+    $ ../wasmedge --reactor factorial.wasm fac 12
+    479001600
+    $
+    $ cd js
+    $ ./../wasmedge --dir .:. qjs.wasm hello.js 1 2 3
+    Hello 1 2 3
+    ```
+
+
```
hydai, line 58: Please remove the redundant empty lines.

hydai (changes requested):

utils/android/build_for_android.sh
```diff
@@ -0,0 +1,28 @@
...
+
+WASMEDGE_ROOT_PATH=$(dirname $(dirname $(dirname $0)))
+
+cd ${WASMEDGE_ROOT_PATH}
+mkdir build
+cd build
+
+echo $(pwd)
+
+if ! cmake .. -DCMAKE_BUILD_TYPE=Release -DWASMEDGE_BUILD_AOT_RUNTIME=OFF -DCMAKE_SYSTEM_NAME=Android -DCMAKE_SYSTEM_VERSION=23 -DCMAKE_ANDROID_ARCH_ABI=arm64-v8a -DCMAKE_ANDROID_NDK=$ANDROID_NDK_HOME -DCMAKE_ANDROID_STL_TYPE=c++_static -DWASMEDGE_BUILD_AOT_RUNTIME=OFF; then
```
O3Ol, line 18: Please delete the `-DWASMEDGE_BUILD_AOT_RUNTIME=OFF` at the end of the line.

Besides, `CMAKE_SYSTEM_VERSION` should be set by user.

O3Ol (commented):
It could work.

utils/android/build_for_android.sh
hangedfish: The minimum Android API level requirement for WasmEdge is 23, only using `seekdir` func in bionic libc. Using a higher API level reduces system compatibility and provides no additional benefits.

hydai:
Hi @hangedfish
Please clean up your commit history, it looks like combine some extra commits which are not created by you.

cmake/Helper.cmake
```diff
@@ -6,7 +6,7 @@ if(CMAKE_BUILD_TYPE STREQUAL Release OR CMAKE_BUILD_TYPE STREQUAL RelWithDebInfo
   if(NOT WASMEDGE_FORCE_DISABLE_LTO)
     set(WASMEDGE_INTERPROCEDURAL_OPTIMIZATION ON)
   endif()
-  if (CMAKE_GENERATOR STREQUAL Ninja)
+  if (CMAKE_GENERATOR STREQUAL Ninja AND NOT ANDROID)
```
hydai, line 9: You don't need to add extra check, because you are not using `ninja` building system.

hydai (changes requested):

hangedfish:
This PR cannot be applied to production environment. Due to Android 10+ SELinux policy constraints, the application's system call exec() does no works on files within the application home directory and `/data/local/temp`. Close.

juntao:
Hmm, I actually think it is still worthwhile. Commercial android smartphones are likely to be locked down and they will not allow an alternative runtime. But Android is used in many IoT or customize devices, I think those are the primary targets for WasmEdge's Android support.

PR #1046, opened 2022-01-24
[CI] Fix Windows installer version

KernelErr:
Only use first three digits for installer version number.

Signed-off-by: Rui Li <me@lirui.tech>


hydai:
Hi @KernelErr 
We just merged #1038 which also modified these two files. Could you please check if the modification is the same as yours? Or you can rebase your commits on the latest master.

KernelErr:
I think it's the same. :)

PR #1048, opened 2022-01-25
[Docs] Document the example of wasmedge-sys APIs

apepkuss:
This PR is to add tutorial files for `wasmedge-sys`. The files include two different formats for different purposes:

- The markdown format is added to WasmEdge docs as a part of the mdbook.
- The evcxr_jupyter notebook is for experimenting with the APIs directly, and added as part of the wasmedge-sys examples.


juntao:
My guess is that the quick start should be in the book?

apepkuss:
> My guess is that the quick start should be in the book?

Hi @juntao, yes, you're right. All the contents in the markdown format will finally be added to the relevant sections of the mdbook when we prepare enough cases for the tutorial. For now, it just has one demo case, so keep it in the docs directory temporarily, in case anyone is interested in it.

juntao:
Yes. I would prefer to have the book as a single source of truth. The issue with having docs in the project is that ppl might bookmark it and we will need to maintain two docs for a long time.

juntao:
Yes, code-related artifacts, such as the notebook, could be located together with the code. You can refer to it from the MDBOOK pages if needed.

hydai (changes requested):
Please use spelling and grammar checking tools on the document. E.g. grammarly.com.

hydai (approved):

PR #1057, opened 2022-01-27
[Utils] Uninstall Script: Fix #1056

### LINKED ISSUE #1056: [Uninstall] After running the uninstall script, the `.bash_profile` is cleared.
### Current State

After [running the uninstall script](https://wasmedge.org/book/en/start/install.html#uninstall):

```bash
bash <(curl -sSf https://raw.githubusercontent.com/WasmEdge/WasmEdge/master/utils/uninstall.sh)
```

The bash file `.bash_profile` on MacOS is cleared.

### Expected

Not to clear the bash profile.

## Environment

- Hardware Architecture: x86_64
- Operating system: MacOS 15

## Steps to Reproduce

Run the install script:

```bash
curl -sSf https://raw.githubusercontent.com/WasmEdge/WasmEdge/master/utils/install.sh | bash
```

And then run the uninstall script:

```bash
bash <(curl -sSf https://raw.githubusercontent.com/WasmEdge/WasmEdge/master/utils/uninstall.sh)
```

Please also check the `.bashrc` on Linux.

SAtacker:
Fixes #1056 

* When grep gives empty results it leads to `sed` removing complete file.
* Added simple check to prevent this

### DIFF
--- a/utils/uninstall.sh
+++ b/utils/uninstall.sh
@@ -253,9 +253,9 @@ main() {
     if [ "$line_num" != "" ]; then
         sed -i.wasmedge_backup -e "${line_num}"'d' "${__HOME__}/${_shell_rc}"
         [[ -f "${__HOME__}/.profile" ]] && line_num="$(grep -n ". \"${IPATH}/env\"" "${__HOME__}/.profile" | cut -d : -f 1)" &&
-            sed -i.wasmedge_backup -e "${line_num}"'d' "${__HOME__}/.profile"
+            [[ "$line_num" != "" ]] && sed -i.wasmedge_backup -e "${line_num}"'d' "${__HOME__}/.profile"
         [[ -f "${__HOME__}/.bash_profile" ]] && line_num="$(grep -n ". \"${IPATH}/env\"" "${__HOME__}/.bash_profile" | cut -d : -f 1)" &&
-            sed -i.wasmedge_backup -e "${line_num}"'d' "${__HOME__}/.bash_profile"
+            [[ "$line_num" != "" ]] && sed -i.wasmedge_backup -e "${line_num}"'d' "${__HOME__}/.bash_profile"
     else
         echo "${YELLOW}Sourcing not found in ${__HOME__}/${_shell_rc} ${NC}"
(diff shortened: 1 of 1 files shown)


SAtacker:
Test:

Uninstall
```bash
bash <(curl -sSf https://raw.githubusercontent.com/SAtacker/WasmEdge/%231056/utils/uninstall.sh) -q
```

q82419:
> 

Tested and it works. Thanks.

q82419 (approved):

SAtacker:
> It seems that there are some additional files on MacOS:
> 
> ```shell
> -rw-r--r--   1 q82419  staff      1523 Jan 27 22:47 .bash_profile.wasmedge_backup
> -rw-r--r--   1 q82419  staff         0 Jan 27 22:47 .bashrc
> -rw-r--r--   1 q82419  staff        32 Jan 27 22:47 .bashrc.wasmedge_backup
> -rw-r--r--   1 q82419  staff         0 Jan 27 22:47 .profile
> -rw-r--r--   1 q82419  staff        32 Jan 27 22:47 .profile.wasmedge_backup
> ```
> 
> It's a tiny issue, can you help to fix and not remain the additional files? Thanks.

It was done so to keep a backup just in case `sed` removes something that is not intended. Should I still remove it?

q82419:
> It was done so to keep a backup just in case `sed` removes something that is not intended. Should I still remove it?

Well, then I think that we should not restore the `.bashrc` and `.profile` with the original size of 0 in the case I attached.
In my case, the MacOS only uses the `.bash_profile` for configuration, but the installer added the `.bashrc` and `.profile`.
Maybe we should fix that in the installer:

1. Added the source command into the existing bash config file.
2. If there's no config files, create one according to OS: `.bash_profile` for MacOS, `.bashrc` for Ubuntu, etc.
3. For uninstaller, only restore and backup the edited config file.

For example, I don't think it's correct to edit the `.bashrc` and `.profile` on MacOS platforms.

q82419:
We should merge this first. I think we can fix in the next PR.

PR #1130, opened 2022-01-31
[AOT] Split callIndirect to two path, compiled function or not

ibmibmibm:


lib/aot/compiler.cpp
```diff
@@ -2961,61 +2965,111 @@ class FunctionCompiler {
...
-      Args = llvm::ConstantPointerNull::get(Builder.getInt8PtrTy());
-    } else {
-      auto *Alloca = Builder.CreateAlloca(Builder.getInt8Ty(),
-                                          Builder.getInt64(ArgSize * kValSize));
-      Alloca->setAlignment(Align(kValSize));
-      Args = Alloca;
+    std::vector<llvm::Value *> ArgsVec(ArgSize + 1, nullptr);
+    ArgsVec[0] = F->arg_begin();
+    for (size_t I = 0; I < ArgSize; ++I) {
+      const size_t J = ArgSize - 1 - I + 1;
```
hydai, line 2971: Not sure why you `-1` and then `+1` here.

hydai (changes requested):

lib/aot/compiler.cpp
ibmibmibm: Removed.

hydai (approved):

hydai:
@ibmibmibm There are some build phase failed. Could you check it

Issue #1058, opened 2022-01-27
[Installer] Refine the environment backup and restoration


q82419:
### Current State

For the MacOS platforms, we only use the `.bash_profile` for the bash configuration.
But after the installation by the installer, the `.bashrc` and `.profile` are generated.
And then after the uninstallation, we will get `.bashrc.wasmedge_backup`, `.profile.wasmedge_backup` and the 0 size of `.bashrc` and `.profile`.

### Expected

Not to leave the trash after the uninstaller execution.
We should not generate the unused config files such as `.bashrc` and `.profile` on MacOS platforms.
As the comment in PR #1057 , I think we can follow the rules:

1. Search the existing bash config file and add the `. "$HOME/.wasmedge/env"` command into it.
2. If there's no config file, create one according to OS: `.bash_profile` for MacOS, `.bashrc` for Ubuntu, etc.
3. For uninstaller, only restore and backup the edited config file.

Further more, I think it's much better to edit the config file to remove the `. "$HOME/.wasmedge/env"` line when uninstalling.
Because the users may add or edit the config file after the WasmEdge installation. If we restore the config file, the settings added after we backup will also be restored. Therefore the users' environment is changed by the uninstaller without their acknowledgement.

## Environment

- Hardware Architecture: x86_64
- Operating system: MacOS 15

## Steps to Reproduce

Use the installer and uninstaller.

SAtacker:
> We should not generate the unused config files such as .bashrc and .profile on MacOS platforms.

For this purpose, there is https://github.com/WasmEdge/WasmEdge/blob/706a42ccaf1b6d79ac8e65bd01e6e5aa5b3ceaaf/utils/install.sh#L590 which detects the shell configuration file.
https://github.com/WasmEdge/WasmEdge/blob/706a42ccaf1b6d79ac8e65bd01e6e5aa5b3ceaaf/utils/install.sh#L610-L616

The below
https://github.com/WasmEdge/WasmEdge/blob/706a42ccaf1b6d79ac8e65bd01e6e5aa5b3ceaaf/utils/install.sh#L593-L609
is added to forcibly make files whether they exist or not.

## TLDR; Remove 593-609 to achieve 2,3
- [X] 1. Search the existing bash config file and add the . "$HOME/.wasmedge/env" command into it.
- [ ] 2. If there's no config file, create one according to OS: .bash_profile for MacOS, .bashrc for Ubuntu, etc.
[...]

PR #1138, opened 2022-02-01
[Utils] Install Script Improvement

SAtacker:
* Behaviour: Creates Shell profiles/configuration files only if these pre-exist.
* Suggestions from https://github.com/WasmEdge/WasmEdge/issues/1058

### DIFF
--- a/.github/workflows/test-install-script.yml
+++ b/.github/workflows/test-install-script.yml
@@ -26,10 +26,10 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - name: CentOS 8
+          - name: CentOS 8 Stream
             host_runner: ubuntu-latest
             package_manager: yum
-            docker_image: centos:8.4.2105
+            docker_image: quay.io/centos/centos:stream8
           - name: CentOS 7
             host_runner: ubuntu-latest
             package_manager: yum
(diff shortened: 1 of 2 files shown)


SAtacker:
image

Cent OS 8: https://www.centos.org/centos-linux-eol/ 
Workarounds: https://stackoverflow.com/questions/70926799/centos-through-vm-no-urls-in-mirrorlist

hydai:
Hi @SAtacker 
I would like to know which one is better?
The centos 8 is a testing environment and we don't use it for building wasmedge.
How about upgrade to centos 8 stream? Or centos 9?

SAtacker:
> Hi @SAtacker 
> I would like to know which one is better?
> The centos 8 is a testing environment and we don't use it for building wasmedge.
> How about upgrade to centos 8 stream? Or centos 9?

As per their announcements they have shifted their focus on Cent Os stream. I think more people would use a supported OS. The Cent Os 8 Stream has 2 Years to EOL.

hydai:
Great. Could you please update the CI Action to use CentOS 8 Stream instead of CentOS 8?

SAtacker:
> Great. Could you please update the CI Action to use CentOS 8 Stream instead of CentOS 8?

Yes, :+1:

SAtacker:
There are no official docker images for Stream

hydai:
https://quay.io/repository/centos/centos?tag=latest&tab=tags

I found that there is no official image on dockerhub, but on the quay.io.

hydai (approved):

PR #1174, opened 2022-02-08
[WASI-Socket] Remove empty sa_data check

KernelErr:
There is a problem with the implementation of this check. As the first
element of sa_data array is sin_port and the first number of low ports
is zero, the function would not copy the sa_data even though the data is
valid in this circumstances.

Signed-off-by: KernelErr <me@lirui.tech>


lib/host/wasi/inode-linux.cpp
```diff
@@ -829,11 +829,9 @@ WasiExpect<void> INode::getAddrinfo(const char *NodeStr, const char *ServiceStr,
       CurSockaddr->sa_data_len = 0;
```
hydai, line 829: I believe we can remove this line. The value will be replaced in line 834

hydai (changes requested):

hydai (approved):

PR #1243, opened 2022-02-15
[WASI] Forge zero-terminated string for `::getaddrinfo`, Fix #1234

### LINKED ISSUE #1234: Failed to correctly handle string parameter ending '\0'
# Description

Following example code:

```rust
#[cfg(test)]
mod tests {
    #[test]
    fn test_dns() {
        use wasmedge_wasi_socket::WasiAddrinfo;
        let node = String::from("127.0.0.1");
        let service = String::from("80");
        let hints: WasiAddrinfo = WasiAddrinfo::default();
        let mut sockaddr = Vec::new();
        let mut sockbuff = Vec::new();
        let mut ai_canonname = Vec::new();
        let addrinfo = WasiAddrinfo::get_addrinfo(
            &node,
            &service,
            &hints,
            10,
            &mut sockaddr,
            &mut sockbuff,
            &mut ai_canonname,
        )
        .unwrap();
        for i in 0..addrinfo.len() {
            if addrinfo[i].ai_addrlen.ne(&0) {
                println!("{:?}", sockbuff[i]);
            }
        }
    }
}
```

Failed to get the correct result, and panicked unexpectedly.

# Why

If you copy the function body of `test_dns` into a common function rather than a test function. It will work with no error.

After investigation, the memory layout of the `node` is:

```
"www.google.com\0" in normal function
"www.google.com" + unknown other data in test function
```

In WasmEdge's code:

https://github.com/WasmEdge/WasmEdge/blob/2e9959f4a8e66f63b55dc153c8958964cf776ab4/lib/host/wasi/wasifunc.cpp#L1908-L1919

We use `NodePtr` and `NodeLen` as parameters and cast that address to `Node`. And then, we pass the node string to syscall `get_addrinfo`. However, the syscall function actually read the string until a '\0' in memory.

# Advice

We should check if **all** string parameter from the host function call is ending with '\0'. If `str[strlen - 1]` is not '\0' and we will pass it to syscall function, we should append one.

ibmibmibm:


lib/host/wasi/inode-linux.cpp
```diff
@@ -769,14 +769,39 @@ WasiExpect<Poller> INode::pollOneoff(__wasi_size_t NSubscriptions) noexcept {
...
+  if (!Service.empty()) {
+    if (const auto pos = Service.find_first_of('\0');
+        pos != std::string_view::npos) {
+      ServiceCStr = Service.data();
+    } else {
+      ServiceStr = Service;
+      ServiceCStr = ServiceStr.c_str();
+    }
+  }
+
```
hydai, line 804: Looks like these two transformations are the same. Should we have a helper function to do this?

hydai (changes requested):

lib/host/wasi/inode-linux.cpp
ibmibmibm: changed to helper function

lib/host/wasi/inode-linux.cpp
```diff
@@ -108,6 +108,23 @@ constexpr int openFlags(__wasi_oflags_t OpenFlags, __wasi_fdflags_t FdFlags,
   return Flags;
 }
 
+std::pair<const char *, std::unique_ptr<char[]>>
+createNullTerminatedString(std::string_view View) noexcept {
+  const char *CStr = nullptr;
+  std::unique_ptr<char[]> Buffer;
+  if (!View.empty()) {
+    if (const auto pos = View.find_first_of('\0');
```
hydai, line 116: `Pos`

lib/host/wasi/inode-linux.cpp
```diff
@@ -108,6 +108,23 @@ constexpr int openFlags(__wasi_oflags_t OpenFlags, __wasi_fdflags_t FdFlags,
...
+        pos != std::string_view::npos) {
+      CStr = View.data();
+    } else {
+      Buffer = std::make_unique<char[]>(View.size() + 1);
```
hydai, line 120: Will this make the last char be `\0`?

lib/host/wasi/inode-macos.cpp
```diff
@@ -89,6 +89,23 @@ constexpr int openFlags(__wasi_oflags_t OpenFlags, __wasi_fdflags_t FdFlags,
   return Flags;
 }
 
+std::pair<const char *, std::unique_ptr<char[]>>
+createNullTerminatedString(std::string_view View) noexcept {
+  const char *CStr = nullptr;
+  std::unique_ptr<char[]> Buffer;
+  if (!View.empty()) {
+    if (const auto pos = View.find_first_of('\0');
```
hydai, line 97: `Pos`

hydai (changes requested):

lib/host/wasi/inode-linux.cpp
ibmibmibm: https://stackoverflow.com/questions/42140212/does-make-unique-value-initializes-char-array

lib/host/wasi/inode-linux.cpp
ibmibmibm: Done.

lib/host/wasi/inode-macos.cpp
ibmibmibm: Done.

hydai (approved):

PR #1250, opened 2022-02-17
[Misc] Replace `std::unordered_map` of enum and string to `DenseMap` or `SpareMap`

ibmibmibm:
Use `std::array` to store mapping for compact and fast lookup


include/common/dense_enum_map.h
```diff
@@ -0,0 +1,119 @@
...
+  using mapped_type = T;
+  using value_type = std::pair<const Key, T>;
+  using size_type = std::size_t;
+  using difference_type = std::ptrdiff_t;
+  using reference = value_type &;
+  using const_reference = const value_type &;
+  using pointer = value_type *;
+  using const_pointer = const value_type *;
+  using iterator = Iterator;
+  using const_iterator = Iterator;
```
hydai, line 39: Should this be `const Iterator`?

hydai (changes requested):

q82419:
The `common/enum_configure.h`, `common/enum_errcode.h`, and `common/enum_types.h` are headers which also exported for API use.
In these case, if we want to use the `DenseMap` or `ShareMap`, we should package the header together, or the building will fail if developers use C++ to link our WasmEdge C API library.

include/common/dense_enum_map.h
ibmibmibm: Changed to `ConstIterator`.

lib/api/CMakeLists.txt
```diff
@@ -32,6 +32,8 @@ set(WASMEDGE_CAPI_HEADERS
   ${CMAKE_CURRENT_BINARY_DIR}/../../include/api/wasmedge/enum_errcode.h
   ${CMAKE_CURRENT_BINARY_DIR}/../../include/api/wasmedge/enum_types.h
   ${CMAKE_CURRENT_SOURCE_DIR}/../../include/api/wasmedge/int128.h
+  ${CMAKE_CURRENT_BINARY_DIR}/../../include/api/wasmedge/dense_enum_map.h.h
+  ${CMAKE_CURRENT_BINARY_DIR}/../../include/api/wasmedge/spare_enum_map.h.h
```
hydai, line 36: Why `.h.h`?

hydai (changes requested):

lib/api/CMakeLists.txt
ibmibmibm: Fixed.

hydai (approved):

PR #1262, opened 2022-02-21
[AOT] Add `GasLimit` support in AOT

ibmibmibm:


lib/aot/compiler.cpp
```diff
@@ -397,9 +399,13 @@ struct WasmEdge::AOT::Compiler::CompileContext {
...
     return Builder.CreateExtractValue(ExecCtx, {4});
   }
+  llvm::Value *getGasLimit(llvm::IRBuilder<> &Builder,
+                           llvm::LoadInst *ExecCtx) {
+    return Builder.CreateExtractValue(ExecCtx, {5});
+  }
   llvm::Value *getStopToken(llvm::IRBuilder<> &Builder,
                             llvm::LoadInst *ExecCtx) {
-    return Builder.CreateExtractValue(ExecCtx, {5});
+    return Builder.CreateExtractValue(ExecCtx, {6});
```
hydai, line 408: Looks like this is a breaking change. Could we keep the origin ordering to avoid the incompatible issue?

hydai (changes requested):

hydai (approved):

PR #1269, opened 2022-02-22
[Rust] Supported Windows platform in the build script

apepkuss:
This PR is to add support for the Windows platform in the build script.

PR #1271, opened 2022-02-23
[CI] Added support for building rust-binding on MacOS.

apepkuss:
This PR is to add support for building rust-binding on MacOS.

- Fixed #1269

### DIFF
--- a/.github/workflows/bindings-rust.yml
+++ b/.github/workflows/bindings-rust.yml
@@ -9,72 +9,147 @@ on:
     branches:
       - master
     paths-ignore:
-      - 'docs/**'
-      - '**.md'
+      - "docs/**"
+      - "**.md"
   pull_request:
     branches:
       - master
     paths-ignore:
-      - 'docs/**'
-      - '**.md'
+      - "docs/**"
+      - "**.md"
 
 jobs:
-  rust:
-    runs-on: ubuntu-latest
+  build_ubuntu:
+    name: Ubuntu 20.04
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-20.04]
     container:
       image: wasmedge/wasmedge:ubuntu-build-clang
 
     steps:
-    - uses: actions/checkout@v2
-      with:
-        fetch-depth: 0
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
 
-    - name: Install Stable Rust
-      uses: actions-rs/toolchain@v1
-      with:
-        toolchain: stable
-        components: rustfmt, clippy
+      - name: Install Stable Rust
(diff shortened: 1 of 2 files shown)


.github/workflows/bindings-rust.yml
```diff
@@ -9,72 +9,147 @@ on:
...
+          cargo test --lib --examples --locked
+          cargo test --doc --locked
+
+  build_macos:
+    name: MacOS
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [macos-11]
+    needs: get_version_v2
```
hydai, line 92: Not sure why do you need this?

hydai (changes requested):

.github/workflows/bindings-rust.yml
apepkuss: According to the description from [GitHub](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners), `macos-latest` or `macos-11` denotes `macOS Big Sur 11` currently. When a new version of MacOS is released in the future, for example, macos-12, `macos-latest` cannot cover macos-11 at that time.

hydai: I mean the `get_version_v2`. This is useless.

apepkuss: By using [macos-11] (or [macos-11, macos-11] in the future),  I think we can have a much finer-grained control on the OS version. Thanks!

hydai: I am fine with the `macos-11` part.
However, looks like there is a misunderstanding. What does this line `needs: get_version_v2` mean?

apepkuss: > 

Sorry, I thought the comment was for line 91. I'll remove it, thanks a lot!

apepkuss:
Hi @hydai, the issue has been fixed, could you please help review the latest change? Thanks a lot!

hydai:
@apepkuss 
The macos-11 is still failed. Is this an expected result?

apepkuss:
@hydai I'll check the issue, thanks!

apepkuss:
@hydai The issues related to `binding-rust/MacOS` have been fixed. But there is still a failure. According to my investigation, the failure happened in the `build_windows` job defined in `build.yml`. This PR does not involve this file, so I'm not sure if this is an expected result or not.

hydai:
The test should pass. However, it's very strange. I will take a look.

hydai:
@apepkuss 
Looks like the environment changed. I have created an new issue #1290 to trace this.
And I am going to merge this PR first

apepkuss:
@hydai Thanks for your help and reminder!

hydai (approved):

PR #1283, opened 2022-02-25
[C-API] Implemented new C-APIs

apepkuss:
In this PR, new four C APIs are added, which are

* `WasmEdge_VMGetExecutorContext`
* `WasmEdge_VMGetValidatorContext`
* `WasmEdge_VMGetLoaderContext`
* `WasmEdge_ImportObjectGetModuleName`

### DIFF
--- a/include/api/wasmedge/wasmedge.h
+++ b/include/api/wasmedge/wasmedge.h
@@ -2697,6 +2697,11 @@ WasmEdge_GlobalInstanceDelete(WasmEdge_GlobalInstanceContext *Cxt);
 WASMEDGE_CAPI_EXPORT extern WasmEdge_ImportObjectContext *
 WasmEdge_ImportObjectCreate(const WasmEdge_String ModuleName);
 
+/// Get the module name.
+///
+/// \param Cxt the target WasmEdge_ImportObjectContext.
+///
+/// \returns the name of the WasmEdge_ImportObjectContext.
 WASMEDGE_CAPI_EXPORT extern WasmEdge_String
 WasmEdge_ImportObjectGetModuleName(const WasmEdge_ImportObjectContext *Cxt);
 
--- a/include/api/wasmedge/wasmedge.h
+++ b/include/api/wasmedge/wasmedge.h
@@ -2697,25 +2697,27 @@ WasmEdge_GlobalInstanceDelete(WasmEdge_GlobalInstanceContext *Cxt);
 WASMEDGE_CAPI_EXPORT extern WasmEdge_ImportObjectContext *
 WasmEdge_ImportObjectCreate(const WasmEdge_String ModuleName);
 
+WASMEDGE_CAPI_EXPORT extern WasmEdge_String
+WasmEdge_ImportObjectGetModuleName(const WasmEdge_ImportObjectContext *Cxt);
+
 /// Creation of the WasmEdge_ImportObjectContext for the WASI specification.
 ///
 /// This function will create a WASI host module that contains the WASI host
 /// functions and initialize it. The caller owns the object and should call
 /// `WasmEdge_ImportObjectDelete` to free it.
 ///
-/// \param Args the command line arguments. The first argument suggests being
-/// the program name. NULL if the length is 0.
-/// \param ArgLen the length of the command line arguments.
-/// \param Envs the environment variables in the format `ENV=VALUE`. NULL if the
-/// length is 0.
-/// \param EnvLen the length of the environment variables.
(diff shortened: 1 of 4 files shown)


apepkuss:
Hi @q82419, could you please help review this PR? Thanks a lot!

cc: @hydai

q82419:
Hi @apepkuss ,

Would you explain why it's needed for getting the `Loader`, `Validator`, and `Executor` from the VM context?
For the developer who want to control the flow detaily, we assume that they should use the `WasmEdge_LoaderContext`, etc. directly.

Thanks.

apepkuss:
Hi @q82419 

A direct motivation of these APIs is to simplify the rust API design and reduce the unnecessary performance overhead. I'll give my explanations below, and use `Loader` as an example in my statements.

- In some scenarios, we hope to get a `Loader` instance owned by `Vm` directly when we interact with `Vm`. Currently, `Vm` provides the interfaces named `WasmEdge_VMLoadWasmFromXXX`, but they do not return the loaded `WasmEdge_ASTModuleContext` instance which is necessary for Rust API design. In fact, according to our investigation, a `Loader` instance owned by `Vm` does the real work under the hood. From the view of the Rust API design, we hope we can get the `Loader` instance so that we can get much more flexibility to use it in the API design, and more importantly to get the loaded `WasmEdge_ASTModuleContext` instance.

- Of course, the C-API also provides an independent Loader object, and we can also use it and the relevant APIs to achieve our API design. However, that makes the design verbose. The following code snippet shows the creation APIs for Loader and Vm, respectively:
   ```rust
     // C-API for creating a Loader instance
    WASMEDGE_CAPI_EXPORT extern WasmEdge_LoaderContext *
    WasmEdge_LoaderCreate(const WasmEdge_ConfigureContext *ConfCxt);

    // C-API for creating a Vm instance
    WASMEDGE_CAPI_EXPORT extern WasmEdge_VMContext *
    WasmEdge_VMCreate(const WasmEdge_ConfigureContext *ConfCxt, WasmEdge_StoreContext *StoreCxt);
   ```
  As shown, Both two C-APIs have an argument of `Config` type. According to our investigation on the C++ implementation of `Vm`, the `Config` arguments used by both `Vm` and its inner `Loader` are the same one. In fact, this already imposes some kind of implicit constraint on the `Config` arguments used in Vm and Loader. The API users have to guarantee to pass the same `Config` instance to `Vm` and `Loader` if they wanna use an independent `Loader` to load modules, and then register the loaded modules into the `Vm`.  That could potentially introduce some negative effects on future maintenance. 

Based on the discussions above, providing the new APIs to access the inner `Loader`, `Validator`, and `Executor` instance of a `Vm` instance would be a better solution. And also, they bring about some additional benefits:

- Simplify the APIs of `Vm`. 

    `Vm` can be seen as a “container” of various kinds of other objects. According to the SRP (Single Responsibility Principle), `Vm` only needs to provide the APIs to get its inner instance, for example `Loader`, and the returned inner `Loader` instance is responsible for loading modules via its APIs.
    
- Break the implicit constraints between the APIs. 
    
    As we mentioned above, an implicit constraint between the `Config` arguments of the creation APIs of `Vm` and `Loader` is present. The new APIs can break such a constraint and make the use case clearer: if developers wanna load a module and use it in `Vm`, then just use the inner `Loader` of the `Vm` instance; if developers wanna load a module and just transfer it on the wire, they just need to create a new `Loader` instance with a specific `Config` argument.
    
- Reduce performance overhead.
    
    In some use cases, if we can use the inner `Loader` instance of `Vm`, we can also avoid creating a new `Loader` instance.

q82419:
Hi @apepkuss ,

Thanks for your explanation.
Can I just recognize that the rust API needs the loaded `WasmEdge_ASTModuleContext` object in the VM?
May we simplify the request, i.e., the VM provides the API to get the context is enough?

Thanks.

apepkuss:
> Hi @apepkuss ,
> 
> Thanks for your explanation. Can I just recognize that the rust API needs the loaded `WasmEdge_ASTModuleContext` object in the VM? May we simplify the request, i.e., the VM provides the API to get the context is enough?
> 
> Thanks.

@q82419 

Not exactly. Actually, the Rust API design needs to manipulate those inner objects in `Vm` (like `loader`, `executor`). For example, the inner `Loader` object instantiated in a `Vm` instance is needed, but not one or more existed C-APIs (`WasmEdge_VMLoadWasmFromXXX`) of `Vm` to load a module so that Rust API can manipulate that inner `loader`. Same to the inner `Validator` and `Executor`.

PR #1328, opened 2022-03-07
[Rust] Added the libwasmedge directory for the docs.rs build environment

apepkuss:
In this PR, the `libwasmedge` directory is added in the `wasmedge-sys` crate to satisfy the requirement of the `docs.rs` official build environment ([no network access](https://docs.rs/about/builds#hitting-resource-limits)). In addition, the version  of the `wasmedge-sys` crate comes to `0.4.2`.

### DIFF
--- a/bindings/rust/wasmedge-sys/Cargo.toml
+++ b/bindings/rust/wasmedge-sys/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "wasmedge-sys"
-version = "0.4.1"
+version = "0.4.2"
 edition = "2021"
 links = "wasmedge_c"
 build = "build.rs"
--- a/bindings/rust/wasmedge-sys/Cargo.toml
+++ b/bindings/rust/wasmedge-sys/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "wasmedge-sys"
-version = "0.4.0"
+version = "0.4.1"
 edition = "2021"
 links = "wasmedge_c"
 build = "build.rs"
--- a/bindings/rust/wasmedge-sys/build.rs
+++ b/bindings/rust/wasmedge-sys/build.rs
@@ -78,22 +78,6 @@ fn main() {
         paths
     };
 
-    // #[cfg(not(feature = "standalone"))]
-    // let Paths {
-    //     header,
-    //     lib_dir,
-    //     inc_dir,
-    // } = find_wasmedge()
-    //     .or_else(build_wasmedge)
-    //     .expect("should be dependency paths");
-
-    // #[cfg(feature = "standalone")]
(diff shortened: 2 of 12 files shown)


apepkuss:

hydai (changes requested):
Adding a built shared libraries into the repo is unacceptable.
I would like to have an alternative way to fix this issue.

apepkuss:
> I would like to have an alternative way to fix this issue

@hydai 

Before using this solution, I tried to add `WasmEdge` as a submodule, but the build process also need to download things, like boost-xxx. That violates the rule: the `docs.rs` build environment does not allow any network access. Could we make `WasmEdge` does not download dependencies? If so, adding it as a submodule should be a feasible solution to the current build failure.

hangedfish:
> > I would like to have an alternative way to fix this issue
> 
> @hydai
> 
> Before using this solution, I tried to add `WasmEdge` as a submodule, but the build process also need to download things, like boost-xxx. That violates the rule: the `docs.rs` build environment does not allow any network access. Could we make `WasmEdge` does not download dependencies? If so, adding it as a submodule should be a feasible solution to the current build failure.

Here's an offline build and submodule demo, I think it might work for you.   [CMakeLists.txt](https://github.com/hangedfish/WasmEdge_AndroidExample/blob/master/app/src/main/cpp/CMakeLists.txt)

apepkuss:
Thanks, @hangedfish. Adding the dependencies as submodules is a feasible way. It can avoid `network access`. 

@hydai could we add the dependencies (like boost) to the `WasmEdge` repo as submodules instead of downloading them while building? If so, I can also add `WasmEdge` as a submodule of `wasmedge-sys`. That could be a feasible way. What do you think?

hydai:
This will be very weird. Please notice that the current file structure is:
```
wasmedge
|- wasmedge-sys, which should depends on the libwasmedge
```

If we make the wasmedge become a submodule to `wasmedge-sys`. Then, it will be a recursive structure.
And the design of language binding should be use the released version of wasmedge. For example, `wasmedge-go` depends on an release version of libwasmedge.

And it becomes an interesting issue. Do we really need to fetch and build the wasmedge for the docs.rs? Is it possible to detect the docs.rs env and skip the redundant steps. Ref: https://docs.rs/about/builds#detecting-docsrs
I found an example, maybe this can help: https://github.com/trevyn/librclone/commit/78fdd5c15afbd86ce86819435e63deae39d1e4d5#diff-b11d019fef6ffd7dc6463723c87bea444fb9429c9a21f9bf2dbee677cdfa76beR10-R15

apepkuss:
Closed. #1334 provides a better solution to publishing rustdoc of `wasmedge-sys`.

PR #1334, opened 2022-03-08
[CI] Added workflow for publishing rustdoc.

apepkuss:
In this PR, a new CI workflow is implemented for publishing the rustdoc of  the `wasmedge-sys` crate to github pages.

To enable the Github Pages feature, it is required to specify the `Source` as shown in the red rectangle below:

image

### References
  - [actions-gh-pages](https://github.com/peaceiris/actions-gh-pages#%EF%B8%8F-deploy-to-external-repository-external_repository)
  - [rust-analyzer rustdoc.yaml](https://github.com/rust-analyzer/rust-analyzer/blob/master/.github/workflows/rustdoc.yaml)
  - [Prepare your Rust API docs for Github Pages](https://dev.to/deciduously/prepare-your-rust-api-docs-for-github-pages-2n5i)

### DIFF
--- a/.github/workflows/wasmedge-sys-docs.yml
+++ b/.github/workflows/wasmedge-sys-docs.yml
@@ -7,7 +7,7 @@ concurrency:
 on:
   push:
     tags:
-      - "wasmedge-sys/[0-9]+.[0-9]+.[0-9]+*"
+      - "rust/[0-9]+.[0-9]+.[0-9]+*"
 
 jobs:
   rustdoc:
--- a/.github/workflows/wasmedge-sys-docs.yml
+++ b/.github/workflows/wasmedge-sys-docs.yml
@@ -1,7 +1,7 @@
 name: wasmedge-sys-docs
 
 concurrency:
-  group: rust-crate-release-${{ github.head_ref }}
+  group: wasmedge-sys-docs-${{ github.head_ref }}
   cancel-in-progress: true
 
 on:
--- a/.github/workflows/wasmedge-sys-docs.yml
+++ b/.github/workflows/wasmedge-sys-docs.yml
@@ -1,0 +1,42 @@
+name: wasmedge-sys-docs
+
+concurrency:
+  group: rust-crate-release-${{ github.head_ref }}
+  cancel-in-progress: true
+
+on:
+  push:
+    tags:
+      - "wasmedge-sys/[0-9]+.[0-9]+.[0-9]+*"
+
+jobs:
+  rustdoc:
+    name: rustdoc
+    runs-on: ubuntu-latest
+    container:
+      image: wasmedge/wasmedge:ubuntu-build-clang
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Install Rust toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: nightly
+          profile: minimal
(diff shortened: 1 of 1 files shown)


.github/workflows/wasmedge-sys-docs.yml
```diff
@@ -0,0 +1,42 @@
+name: wasmedge-sys-docs
+
+concurrency:
+  group: rust-crate-release-${{ github.head_ref }}
+  cancel-in-progress: true
+
+on:
```
hydai, line 7: Maybe we need to test the docs generation on master when the related rust files change. But we don't publish it until emitting the release tag.

hydai (changes requested):

.github/workflows/wasmedge-sys-docs.yml
```diff
@@ -0,0 +1,42 @@
+name: wasmedge-sys-docs
+
+concurrency:
+  group: rust-crate-release-${{ github.head_ref }}
```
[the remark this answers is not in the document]
apepkuss: Accepted. I'll fix the issue. Thanks!

.github/workflows/wasmedge-sys-docs.yml
```diff
@@ -0,0 +1,42 @@
...
+  cancel-in-progress: true
+
+on:
+  push:
+    tags:
+      - "wasmedge-sys/[0-9]+.[0-9]+.[0-9]+*"
```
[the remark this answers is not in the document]
apepkuss: The original tag pattern is only for `wasmedge-sys`. Once `wasmedge-rs` is introduced, the original tag could be ambiguous. Based on this consideration, suggest using new tag names for the existed `wasmedge-sys` and the upcoming `wasmedge-rs`, respectively. `wasmedge-sys/[0-9]+.[0-9]+.[0-9]+*` is used for `wasmedge-sys`.

.github/workflows/wasmedge-sys-docs.yml
apepkuss: You're right. Using this strategy is based on the following consideration.  The current workflow is only triggered when there is a new tag coming, like `wasmedge-sys/0.4.0`. Such a setting can guarantee that the crate published on crates.io and the rustdoc published on the Github pages are built from the same commit. However, if we used `push-on-master` to trigger the workflow, the rustdoc would be built from the latest rust code, but not from the published version, that is, there would be version inconsistency between the published crate and the published rustdoc.

.github/workflows/wasmedge-sys-docs.yml
```diff
@@ -0,0 +1,42 @@
+name: wasmedge-sys-docs
+
+concurrency:
+  group: wasmedge-sys-docs-${{ github.head_ref }}
+  cancel-in-progress: true
+
+on:
+  push:
+    tags:
+      - "wasmedge-sys/[0-9]+.[0-9]+.[0-9]+*"
```
hydai, line 10: Hmmm. The previous tag style will be `rust/version`. Do you want to change it to `wasmedge-sys/version`? If so, then you will need to update the rust related workflows.
截圖 2022-03-10 at 16 03 39@2x

hydai (changes requested):

hydai (approved):

apepkuss:
@hydai Thanks for the review!

Issue #1183, opened 2022-02-09
feat: enable Include What You Use scanning in CI


hydai:
## Motivation

Currently, we use a local environment to run the Include What You Use like #1179 
However, it will be better to create a workflow for the regular scanning.

Reference: https://github.com/include-what-you-use/include-what-you-use

abhinandanudupa:
Can you please assign it to me?

hydai:
Sure ;-)

abhinandanudupa:
Sorry, this is taking longer than what I expected. I have exams going on so it will take some more days.

abhinandanudupa:
Should we run `include-what-you-need` on all Operating Systems - Ubuntu, Fedora, MacOS, Windows?

hydai:
We can have one linux distribution only (Both Ubuntu or Fedora work for me), MacOS, and Windows.

abhinandanudupa:
I am a bit confused. I tried scanning with `include-what-you-use` on **Fedora** and found no issues at all. But when I try the same on an **Arch based distro (Endeavour OS)** (updated to the latest version) the results are different - I get these warnings:
```bash
......
[ 16%] Built target utilBlake3
Warning: include-what-you-use reported diagnostics:

/home/user/test/WasmEdge/build/_deps/spdlog-src/include/spdlog/async.h should add these lines:
#include <stddef.h>                          // for size_t
#include <string>                            // for string
#include <utility>                           // for move, forward
#include "spdlog/details/thread_pool-inl.h"  // for thread_pool::thread_pool
namespace spdlog { class logger; }

/home/user/test/WasmEdge/build/_deps/spdlog-src/include/spdlog/async.h should remove these lines:

The full include-list for /home/user/test/WasmEdge/build/_deps/spdlog-src/include/spdlog/async.h:
#include <spdlog/async_logger.h>             // for async_overflow_policy
#include <spdlog/details/registry.h>         // for registry
#include <spdlog/details/thread_pool.h>      // for thread_pool
#include <stddef.h>                          // for size_t
#include <functional>                        // for function
#include <memory>                            // for make_shared, shared_ptr
#include <mutex>                             // for lock_guard, recursive_mutex
#include <string>                            // for string
#include <utility>                           // for move, forward
#include "spdlog/details/thread_pool-inl.h"  // for thread_pool::thread_pool
namespace spdlog { class logger; }
---

[...]

PR #1363, opened 2022-03-21
[CI] Add workflow to scan with IWYU on Unix #1183

abhinandanudupa:
Use this workflow to scan the source code with IWYU on Unix based
systems - Linux and macOS [Issue: #1183]

[Note that this does not perform the scan on Windows]

Signed-off-by: abhinandanudupa <67597710+abhinandanudupa@users.noreply.github.com>


abhinandanudupa:
I am currently facing many issues trying to compile IWYU for Windows. Unfortunately this is the only way to get IWYU working on Windows as the maintainers lasted released a binary for windows for version 0.8

abhinandanudupa:
Unfortunately I am running into an issue after an issue on windows.

Initially I attempted to compile IWYU as outlined [here](https://github.com/include-what-you-use/include-what-you-use/issues/956#issuecomment-991363322) - tested this on my local machine and it is working as expected (took > 4 hours to compile `LLVM` : ) .

But when I try this in the CI environment I run in to [this issue](https://github.com/abhinandanudupa/WasmEdge/runs/5852205923?check_suite_focus=true#step:5:83) `  Invalid character escape '\a'.` Then tried to install it in `D:` drive but that too [failed](https://github.com/abhinandanudupa/WasmEdge/actions/runs/2103122371/workflow#L121) with this [issue](https://github.com/abhinandanudupa/WasmEdge/runs/5852820518?check_suite_focus=true#step:5:85) - `Error: could not load cache
`. I don't know how to solve this and most of the solutions that I read told me to clean the workspace. But in a CI env where a fresh image is used every time, we should not get this error repeatedly (Check the runs linked at the bottom).

Then tried to use the pre-compiled binary that I had compiled locally (uploaded it to a repo and downloaded it in the CI env from there) but that gave me [this issue](https://github.com/abhinandanudupa/WasmEdge/runs/5853605394?check_suite_focus=true#step:5:25) - 
```
GAC    Version        Location
---    -------        --------
False  v4.0.30319     C:\Program Files\PowerShell\7\WindowsBase.dll
OperationStopped: 
```
 I have not found any mention of a similar issue.

Importantly since I don't have direct access to the CI environment it has become every hard to figure out what is going wrong. Further more it is hard to pin point which command is failing as the GitHub logs do not show the corresponding command executed for the output.

I don't know how I can proceed with the IWYU scanning on windows. 

For now would IWYU scanning on just Unix systems be fine?

You can look at all the runs [here](https://github.com/abhinandanudupa/WasmEdge/actions/workflows/iwyu_scan.yml).

.github/workflows/iwyu_scan.yml
```diff
@@ -0,0 +1,108 @@
+name: Build and scan WasmEdge with IWYU on multiple platforms
+
+concurrency:
+  group: build-${{ github.head_ref }}
```
hydai, line 4: Please use the same name as the workflow file name.

hydai (changes requested):

.github/workflows/IWYU_scan.yml
```diff
@@ -0,0 +1,108 @@
+name: IWYU_scan
+
+concurrency:
+  group: build-${{ github.head_ref }}
```
hydai, line 4: `IWYU_scan-${{ github.head_ref }}`. This group should be an unique value. In your case, it will cancel the `build.yml`.

hydai (changes requested):

.github/workflows/IWYU_scan.yml
abhinandanudupa: Ok, so this is why the build was failing

abhinandanudupa: Fixed it

hydai (approved):

PR #1365, opened 2022-03-23
Refactor the store and the module instance.

q82419:
- [x] Move the ownership of the instances from the store manager to the module instance.
- [x] Merge the import object into the module instance.
- [x] Invoke functions by the function instance instead of by function name in executor.
- [x] Refactor the WasmEdge C API.
- [x] Update the tests.
- [x] Add the `Upgrade to 0.10.0` guideline.
- [x] Update the C API document.


apepkuss:
> Details

@q82419 Thanks for the reminder. I checked the number of the refactored C-APIs, and it would be better to launch a separate PR to migrate rust bindings after merging this PR. Currently, the failed binding-rust workflows can be ignored.

apepkuss:
Is it possible to expose the following two APIs?
- Statistics.clear()
- StoreManager.reset()

In addition, I noticed that in the current design `Executor` provides the APIs for registering `ASTModule` and `ImportObject`. Is it possible to move the registering APIs from `Executor` to `Store`? So that in the scenario, like multiple threading, we can only `move` or `share` a store between threads. I'm not very sure if the idea is appropriate or not in real practice.

q82419:
> Is it possible to expose the following two APIs?
> 
> * Statistics.clear()
> * StoreManager.reset()
> 
> In addition, I noticed that in the current design `Executor` provides the APIs for registering `ASTModule` and `ImportObject`. Is it possible to move the registering APIs from `Executor` to `Store`? So that in the scenario, like multiple threading, we can only `move` or `share` a store between threads. I'm not very sure if the idea is appropriate or not in real practice.

For the statistics: ok, I'll expose later.
For the store manager: in the current PR, the ownership is moved into the module instance, developers should keep and handle the module instance by themselves. The store manager only record the module exporting for linking when instantiation. I think it's not necessary to implement the reset function. By the way, it's better to remove the export info in the store manager when a named module being destroyed. I'll update for this.

It's not appropriate to move the registering APIs from the `Executor` to `Store`. The `Store` means the object to handle the instances for searching and module linking, and the `Executor` is the object to instantiate and register WASM modules.

apepkuss:
> > Is it possible to expose the following two APIs?
> > 
> > * Statistics.clear()
> > * StoreManager.reset()
> > 
> > In addition, I noticed that in the current design `Executor` provides the APIs for registering `ASTModule` and `ImportObject`. Is it possible to move the registering APIs from `Executor` to `Store`? So that in the scenario, like multiple threading, we can only `move` or `share` a store between threads. I'm not very sure if the idea is appropriate or not in real practice.
> 
> For the statistics: ok, I'll expose later. For the store manager: in the current PR, the ownership is moved into the module instance, developers should keep and handle the module instance by themselves. The store manager only record the module exporting for linking when instantiation. I think it's not necessary to implement the reset function. By the way, it's better to remove the export info in the store manager when a named module being destroyed. I'll update for this.
> 
> It's not appropriate to move the registering APIs from the `Executor` to `Store`. The `Store` means the object to handle the instances for searching and module linking, and the `Executor` is the object to instantiate and register WASM modules.

@q82419 Thanks for the explanation.

ibmibmibm (changes requested):

include/runtime/instance/module.h
```diff
@@ -23,294 +24,405 @@
...
   }
 
+  /// \name Data for compiled functions.
+  /// @{
+  std::vector<uint8_t *> MemoryPtrs;
+  std::vector<ValVariant *> GlobalPtrs;
+  /// @}
+
+  friend class Runtime::StoreManager;
+  using StoreCallback = void(StoreManager *Store, const ModuleInstance *Mod);
```
ibmibmibm, line 412: `noexcept`

include/runtime/instance/module.h
```diff
@@ -23,294 +24,405 @@
...
+
 class ModuleInstance {
 public:
   ModuleInstance(std::string_view Name) : ModName(Name) {}
-  ~ModuleInstance() = default;
+  virtual ~ModuleInstance() {
+    // When destroying this module instance, call the callbacks to unlink to the
+    // store managers.
+    for (auto &&Pair : LinkedStore) {
+      Pair.second(Pair.first, this);
```
ibmibmibm, line 72: `assuming(Pair.second);`

ibmibmibm (changes requested):

include/runtime/storemgr.h
```diff
@@ -13,333 +13,92 @@
...
+    std::unique_lock Lock(Mutex);
+    auto Iter = NamedMod.find(ModInst->getModuleName());
+    if (likely(Iter != NamedMod.cend())) {
+      return Unexpect(ErrCode::ModuleNameConflict);
+    }
+    NamedMod.emplace(ModInst->getModuleName(), ModInst);
+    // Link the module instance to this store manager.
+    (const_cast<Instance::ModuleInstance *>(ModInst))
+        ->linkStore(this, [](StoreManager *Store,
+                             const Instance::ModuleInstance *Inst) {
```
ibmibmibm, line 80: `noexcept`

ibmibmibm (changes requested):

include/runtime/storemgr.h
q82419: added.

apepkuss:
Hi @q82419

Here is a scenario that the current design of `Vm` could fail to cover. The scenario is 

1. Create a host function and a `Table` instance; then, store the reference to the host function at some index of the table instance.
2. Create a `ModuleInstance` instance, and add the table instance (created in step 1) into the module instance. 
3. Create a `Vm` instance, then register the module instance (created in step 2).

After the steps above, we can get the internal store of the vm instance, and then get the table instance via the returned store, and then get the function reference to the host function via the table instance, but how to call the host function via the function reference?

apepkuss:
@q82419 Thanks for the reply! I'll append the explanation for the scenario in the API doc. Thanks again!

include/runtime/instance/module.h
```diff
@@ -23,294 +24,407 @@
...
-    FuncTypes.emplace_back(FuncType);
+    unsafeAddHostInstance(
+        Name, OwnedFuncInsts, FuncInsts, ExpFuncs,
+        std::make_unique<Runtime::Instance::FunctionInstance>(std::move(Func)));
   }
-
-  /// Register module owns instances with address in Store.
-  void addFunc(FunctionInstance *Func) {
+  void addHostFunc(std::string_view Name,
+                   std::unique_ptr<Instance::FunctionInstance> &Func) {
```
ibmibmibm, line 91: should use `&&`, and merge with next function.

lib/executor/executor.cpp
```diff
@@ -9,11 +9,13 @@
...
-Expect<void> Executor::instantiateModule(Runtime::StoreManager &StoreMgr,
-                                         const AST::Module &Mod) {
-  InsMode = InstantiateMode::Instantiate;
-  if (auto Res = instantiate(StoreMgr, Mod, ""); !Res) {
+/// Instantiate a WASM Module. See "include/executor/executor.h".
+Expect<std::unique_ptr<Runtime::Instance::ModuleInstance>>
+Executor::instantiateModule(Runtime::StoreManager &StoreMgr,
+                            const AST::Module &Mod) {
+  if (auto Res = instantiate(StoreMgr, Mod)) {
+    return std::move(*Res);
```
ibmibmibm, line 17: `return Res;`

lib/executor/executor.cpp
```diff
@@ -22,51 +24,15 @@ Expect<void> Executor::instantiateModule(Runtime::StoreManager &StoreMgr,
...
-                                      std::string_view Name) {
-  InsMode = InstantiateMode::ImportWasm;
-  if (auto Res = instantiate(StoreMgr, Mod, Name); !Res) {
-    spdlog::error(ErrInfo::InfoRegistering(Name));
+/// Register a named WASM module. See "include/executor/executor.h".
+Expect<std::unique_ptr<Runtime::Instance::ModuleInstance>>
+Executor::registerModule(Runtime::StoreManager &StoreMgr,
+                         const AST::Module &Mod, std::string_view Name) {
+  if (auto Res = instantiate(StoreMgr, Mod, Name)) {
+    return std::move(*Res);
```
ibmibmibm, line 34: `return Res;`

lib/executor/instantiate/global.cpp
```diff
@@ -24,22 +23,19 @@ Expect<void> Executor::instantiate(Runtime::StoreManager &StoreMgr,
...
-    // Insert global instance to store manager.
-    Runtime::Instance::GlobalInstance *GlobInst = nullptr;
-    if (InsMode == InstantiateMode::Instantiate) {
-      GlobInst = StoreMgr.pushGlobal(GlobSeg.getGlobalType());
-    } else {
-      GlobInst = StoreMgr.importGlobal(GlobSeg.getGlobalType());
-    }
+    // Create and add the global instance into the module instance.
+    ModInst.addGlobal(GlobSeg.getGlobalType());
+    Runtime::Instance::GlobalInstance *GlobInst =
```
ibmibmibm, line 31: `const auto Index = ModInst.getGlobalNum() - 1;`

ibmibmibm (changes requested):

lib/executor/instantiate/module.cpp
```diff
@@ -50,39 +52,32 @@ Expect<void> Executor::instantiate(Runtime::StoreManager &StoreMgr,
 
   // Instantiate ImportSection and do import matching. (ImportSec)
   const AST::ImportSection &ImportSec = Mod.getImportSection();
-  if (auto Res = instantiate(StoreMgr, *ModInst, ImportSec); !Res) {
+  if (auto Res = instantiate(StoreMgr, *ModInst.get(), ImportSec); !Res) {
```
ibmibmibm, line 55: why not `*ModInst`?

test/api/APIUnitTest.cpp
```diff
@@ -193,16 +194,16 @@ WasmEdge_ImportObjectContext *createExternModule(std::string_view Name,
...
 
   WasmEdge_FunctionTypeDelete(HostFType);
-  return ImpObj;
+  return HostMod;
 }
 
 // Helper function to load wasm file into AST module.
-WasmEdge_ASTModuleContext *loadModule(const WasmEdge_ConfigureContext *Conf,
-                                      const char *Path) {
+WasmEdge_ASTModuleContext *loadModule
```
ibmibmibm, line 205: `[[maybe_unused]] WasmEdge_ASTModuleContext *loadModule(...)`

ibmibmibm (changes requested):

test/api/APIUnitTest.cpp
q82419: Removed.

lib/executor/instantiate/module.cpp
q82419: Fixed.

lib/executor/executor.cpp
q82419: Fixed.

lib/executor/instantiate/global.cpp
q82419: Updated.

ibmibmibm (approved):

Issue #444, opened 2021-09-27
[Challenge] Create a host function for WasmEdge


juntao:
In this challenge, you will create a "host function" that allows Wasm  bytecode programs to "break the sandbox" and call functions external native shared libraries. Follow the instructions below, and send a link to your work in the comment. 

**Note:** Just fork the WasmEdge repo, and create a host function under `tools/wasmedge/examples /host_functions` in your own fork. Send us a pull request for your work to be showcased as an example!

## Steps:

1 Go through the QuickStart guide to create a C program that embeds WasmEdge.

https://github.com/WasmEdge/WasmEdge/blob/master/docs/c_api_quick_start.md

2 Go through the tutorial to register a C-based “host function” into WasmEdge, and run a Wasm program that calls this embedded function. 

https://github.com/WasmEdge/WasmEdge/blob/master/docs/c_api.md#Host-Functions

3 Create a new host function that calls a shared library on your system. It could be anything — SQLite, ffmpeg, git, anything you have on your system.

4 Create a wasm program that calls this host function. You could hand write a WAT file and then turn it into wasm. 

https://developer.mozilla.org/en-US/docs/WebAssembly/Text_format_to_wasm

https://webassembly.github.io/wabt/doc/wat2wasm.1.html

5 Create a complete C program to run this wasm program in an embedded WasmEdge, like in step #2.

Optional: Send your fork to us as a PR to be showcased as an official example.

juntao:
Hi @JialuGong 

We'd love to see the host function you come up with. Send us a PR. Thanks. 

Cheers
Michael

chinzhiweiblank:
Hi @juntao , I have created a host function under this [PR](https://github.com/chinzhiweiblank/WasmEdge/pull/1). It is very similar to the current example of `Add` because I don't know how to process strings in WasmEdge C's SDK. It is currently working but I would like to ask for help to improve the function.

```

WasmEdge_Result parse_json(void *Data, WasmEdge_MemoryInstanceContext *MemCxt,
                    const WasmEdge_Value *In, WasmEdge_Value *Out) {
    FILE *fp = fopen("test.json","r");
    char buffer[1024];
    fread(buffer, 1024, 1, fp);
    fclose(fp);

    struct json_object *number;
    json_object* parsedJson = json_tokener_parse(buffer);
    // string key = In[0];
    // json_object_object_get_ex(parsedJson, key, &value);
    // Out[0] = value;
    return WasmEdge_Result_Success;
[...]

PR #1387, opened 2022-04-08
Add host function with Shared Library

chinzhiweiblank:
#444 Create a host function for WasmEdge

This example creates a host function using the json-c shared library.

I have handwritten a parse-json.wat and used wat2wasm to create parse-json.wasm. I then use WasmEdge_VMRunWasmFromFile to execute from the file.

For the following JSON file,
```
{
    "testValue": "Success" 
}
```
it extracts the value corresponding to the key and returns it.

`Get the result: "Success"`


alabulei1:
@hydai Could you help review this PR?

hydai:
Hi @chinzhiweiblank 
I am not sure if we would like to put the examples into the `tools` folder. Maybe we should have an external path to archive, say `examples` or `docs/examples`.

To accept this PR, there are something need to fix:
1. There is no documentation or README to describe how to compile and run your application.
2. The coding style should follow `.clang-format` and `.clang-tidy`.
3. You include an external header `<json-c/json_tokener.h>`. However, there is no way to know where is this header from.

chinzhiweiblank:
> Hi @chinzhiweiblank I am not sure if we would like to put the examples into the `tools` folder. Maybe we should have an external path to archive, say `examples` or `docs/examples`.

I have added a folder `examples/host_functions`.

> 1. There is no documentation or README to describe how to compile and run your application.
> 3. You include an external header `<json-c/json_tokener.h>`. However, there is no way to know where is this header from.

Added README.md to install the shared library `libjson`, build and run the executable.

examples/host_functions/README.md
```diff
@@ -0,0 +1,20 @@
+# Examples
+
+The `parse-json` example shows how to use a host function with a shared library `json-c`. The `parseJson` function accepts a string as a key, parses the JSON file `test.json` and returns the corresponding value.
+
+The `parse-json.wat` is a WebAssembly script. It is compiled into WebAssembly in the form of `parse-json.wasm` using the [WABT tool](https://github.com/WebAssembly/wabt).
+
+## Installation
```
hydai, line 7: Could you please also mention your environment? The `libjson-c-dev` on different OS may have different version. It's better to have the detailed version information.

hydai (changes requested):
I suggest that using `examples/capi/host_functions/parse_json` as your folder name.

examples/host_functions/README.md
chinzhiweiblank: Added: The version of `json-c` installed is `0.13.1` in `Ubuntu 20.04.3`. Is that correct?

hydai (approved):
ＬＧＴＭ

PR #1417, opened 2022-05-05
[Proposal] Wasm Function References Proposal

eee4017:


eee4017:
To implement [Function Reference Proposal](https://github.com/WebAssembly/function-references/blob/main/proposals/function-references/Overview.md), we are going to modify the type definition.

In AST, we might need to change the original [enum ValType](https://github.com/WasmEdge/WasmEdge/blob/master/include/common/enum_types.h#L31) into class. I would add `class ValType` in [type.h](https://github.com/WasmEdge/WasmEdge/blob/master/include/ast/type.h) and change original `enum ValType` into `enum ValTypeCode` for binary representation only.

The ValType should include a field about the reference type and should be represented as a node in AST. Thus, the original enum representation is not enough. Please refer to the implementation of [proposal interpreter](https://github.com/WebAssembly/function-references/blob/main/interpreter/binary/decode.ml#L174) and [v8 - Turn ValueType from an enum to a class](https://github.com/v8/v8/commit/f3b4167f8b6226b3b828417543aad99f2d58f5c5).

The [ValType of v8](https://github.com/v8/v8/blob/main/src/wasm/value-type.h#L549) is consist of two fields: a ValueKind and a heap representation (for reference types/rtts). We might want to refer to this implementation for future extension on RTT typing system in GC proposal.

q82419:
Please create PR to `proposal/func-ref`. Thanks

q82419:
Is it more convenient if we rebase the `proposal/func-ref` on the `proposal/threads`?

eee4017:
> Is it more convenient if we rebase the `proposal/func-ref` on the `proposal/threads`?

I think they are independent, so the func-ref branch on master is fine.

q82419:
Close this due to the current implementation of typed-function-reference proposal.

PR #1418, opened 2022-05-05
[Rust]: Add Clone trait to Function and Global.

EmisonLu:
In this PR, the `Clone` trait is added for `Function` and `Global`. Crate `wasmedge-sys` version upgraded from v0.5.0 to v0.5.1.

### DIFF
--- a/bindings/rust/wasmedge-sys/src/instance/function.rs
+++ b/bindings/rust/wasmedge-sys/src/instance/function.rs
@@ -61,7 +61,7 @@ extern "C" fn wraper_fn(
 /// A WasmEdge [Function] defines a host function described by its [FuncType]. A host function is a function defined outside WASM module and passed to it.
 ///
 /// In WasmEdge, developers can create [host functions](crate::Function) and other WasmEdge instances, such as [Memory](crate::Memory), and add them into a WasmEdge [ImportObject](crate::ImportObject) for registering into a WasmEdge [Vm](crate::Vm) or [Store](crate::Store).
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub struct Function {
     pub(crate) inner: InnerFunc,
     pub(crate) registered: bool,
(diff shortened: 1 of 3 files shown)


hydai:
@apepkuss Could you please help me review this issue?

bindings/rust/wasmedge-sys/src/instance/function.rs
```diff
@@ -61,7 +61,7 @@ extern "C" fn wraper_fn(
 /// A WasmEdge [Function] defines a host function described by its [FuncType]. A host function is a function defined outside WASM module and passed to it.
 ///
 /// In WasmEdge, developers can create [host functions](crate::Function) and other WasmEdge instances, such as [Memory](crate::Memory), and add them into a WasmEdge [ImportObject](crate::ImportObject) for registering into a WasmEdge [Vm](crate::Vm) or [Store](crate::Store).
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
apepkuss, line 64: The `Clone` trait for `Function` could cause a potential issue. For example, assume that `A` is a `non-registered` function instance (that is, the `registered` field is `False`), and `B` is a clone of `A`, therefore `B` is also a `non-registered` function instance. In such a case, dropping `A` (or `B`) would have the low-level raw pointer dropped, then the other function instance `B` (or `A`) still holds a dangling low-level raw pointer that points to the deallocated memory region.
If would like a clone of a function instance, a better way is to use `Rc` (for single-thread case) or `Arc` (for multi-thread case) to wrap the function instance, then get a clone from the `Rc` or `Arc` instance.

bindings/rust/wasmedge-sys/src/instance/global.rs
```diff
@@ -65,15 +65,15 @@ impl Drop for GlobalType {
...
+#[derive(Debug, Clone)]
 pub(crate) struct InnerGlobal(pub(crate) *mut ffi::WasmEdge_GlobalInstanceContext);
 unsafe impl Send for InnerGlobal {}
 unsafe impl Sync for InnerGlobal {}
 
 /// Struct of WasmEdge Global.
 ///
 /// A WasmEdge [Global] defines a global variable, which stores a single value of the given [GlobalType].
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
apepkuss, line 76: Same as the comments on `Function`.

EmisonLu:
It will be done in version 0.7.0.

Issue #1434, opened 2022-05-08
Fail to build in macOS Monterey aarch64 


hangedfish:
### Current State
 Fail to build WasmEdge project

## Environment

- Hardware Architecture: aarch64
- Operating system: MacOS 12.2.1

- C++ Compiler version: Apple clang version 13.0.0 (clang-1300.0.29.30); Target: arm64-apple-darwin21.3.0
- CMake version: 3.22.2
- CMake flags: -DCMAKE_BUILD_TYPE=Release -DWASMEDGE_BUILD_AOT_RUNTIME=OFF 
- Boost version: 1.76.0

## Steps to Reproduce

```
➜  WasmEdge git:(master) mkdir build
➜  WasmEdge git:(master) cd build
➜  build git:(master) cmake -DCMAKE_BUILD_TYPE=Release -DWASMEDGE_BUILD_AOT_RUNTIME=OFF ..
-- The C compiler identification is AppleClang 13.0.0.13000029
-- The CXX compiler identification is AppleClang 13.0.0.13000029
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /Library/Developer/CommandLineTools/usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /Library/Developer/CommandLineTools/usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Looking for C++ include filesystem
-- Looking for C++ include filesystem - found
-- Performing Test CXX_FILESYSTEM_NO_LINK_NEEDED
-- Performing Test CXX_FILESYSTEM_NO_LINK_NEEDED - Success
-- Looking for pthread.h
-- Looking for pthread.h - found
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE  
-- Looking for mmap
-- Looking for mmap - found
-- Looking for C++ include pwd.h
-- Looking for C++ include pwd.h - found
-- Build spdlog: 1.9.0
-- Build type: Release
-- Configuring done
-- Generating done
-- Build files have been written to: /Users/admin/WorkSpace/projects/wasmedge-embed/vendor/WasmEdge/build
➜  build git:(master) make -j4
[...]

PR #1436, opened 2022-05-08
Fixed fail to build on macos12

hangedfish:
For #1434 
It is worked for me


hangedfish:
Solved with update clang to Apple clang version 13.1.6 (clang-1316.0.21.2.3).

PR #1441, opened 2022-05-11
[Proposal] WASM Extended-const proposal

q82419:


tools/wasmedge/wasmedgec.cpp
```diff
@@ -120,12 +123,13 @@ int main(int Argc, const char *Argv[]) {
   if (PropMultiMem.value()) {
     Conf.addProposal(WasmEdge::Proposal::MultiMemories);
   }
-  if (PropTailCall.value()) {
```
hydai, line 123: Where does the TailCall go?

hydai (changes requested):

tools/wasmedge/wasmedgec.cpp
q82419: Oops.

hydai (approved):
LGTM

PR #1458, opened 2022-05-17
[Tools] Fix the extension name for shared library format in AOT compiler

q82419:
on different platforms.

### DIFF
--- a/test/api/APIAOTCoreTest.cpp
+++ b/test/api/APIAOTCoreTest.cpp
@@ -14,11 +14,12 @@
 ///
 //===----------------------------------------------------------------------===//
 
+#include "common/defines.h"
+#include "helper.h"
+#include "hostfunc_c.h"
 #include "wasmedge/wasmedge.h"
 
 #include "../spec/spectest.h"
-#include "helper.h"
-#include "hostfunc_c.h"
 
 #include <cstdint>
 #include <functional>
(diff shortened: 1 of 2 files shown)


test/api/APIAOTCoreTest.cpp
```diff
@@ -29,14 +29,11 @@
 #include <utility>
 #include <vector>
 
-#if defined(linux) || defined(__linux) || defined(__linux__) ||                \
-    defined(__gnu_linux__)
+#if WASMEDGE_OS_LINUX
```
hydai, line 32: Do you need to include any header related to these macro variables?

hydai (changes requested):

test/api/APIAOTCoreTest.cpp
ibmibmibm: this identifier is defined in `common/defines.h`

tools/wasmedge/wasmedgec.cpp
```diff
@@ -191,7 +200,7 @@ int main(int Argc, const char *Argv[]) {
     if (ConfGenericBinary.value()) {
       Conf.getCompilerConfigure().setGenericBinary(true);
     }
-    if (OutputPath.extension().u8string() == ".so"sv) {
+    if (OutputPath.extension().u8string().compare(EXTENSION) == 0) {
```
ibmibmibm, line 203: Why to use `compare` here, not `==`?

hydai (approved):

PR #1463, opened 2022-05-18
[CI] Fix windows CI

hydai:


.github/workflows/build.yml
```diff
@@ -286,7 +286,7 @@ jobs:
 
   build_windows:
     name: Windows 10
```
q82419, line 288: I think this can be renamed as `Windows 11` or `Windows server 2022`?
And should we add a matrix for both 2019 and 2022?

hydai: Use `Windows server 2022` instead.

PR #1488, opened 2022-05-24
[Rust] Implemented vm in wasmedge-sdk crate

apepkuss:
This PR contains the following changes:

- Introduced `Vm` struct and implemented the relevant APIs;
- Implemented the unit tests for `Vm`;
- Implemented an example for `Vm`;
- Added necessary doc comments for `Vm` and its APIs;
- Refined the existed doc comments.


bindings/rust/wasmedge-sdk/Cargo.toml
```diff
@@ -14,7 +14,7 @@ version = "0.1.0"
 anyhow = "1.0"
 thiserror = "1.0.30"
 wasmedge-sys = {path = "../wasmedge-sys", version = "0.7"}
-wasmedge-types = "0.1.2"
+wasmedge-types = "0.1"
```
hydai, line 17: Why do you downgrade the version?

bindings/rust/wasmedge-sdk/src/externals/function.rs
```diff
@@ -141,7 +146,7 @@ impl Func {
     }
 }
 
-/// Defines a type builder for creating a [FuncType](wasmedge_types::FuncType).
+/// Defines a type builder for creating a [FuncType](https://wasmedge.github.io/WasmEdge/wasmedge_types/struct.FuncType.html) instance.
```
hydai, line 149: I would like to know when to use the crate link or URL?

hydai (changes requested):

bindings/rust/wasmedge-sdk/Cargo.toml
apepkuss: Not exactly. `wasmedge-types = "0.1"` means that the latest `0.1.x` will be used. In this case, for example, `0.1.3` will be used. This way is much more flexible. Thanks!

bindings/rust/wasmedge-sdk/src/externals/function.rs
apepkuss: If the content is in the current crate, use `crate::xxx`; otherwise, use `URL`. In this PR, for example, `FuncType` is defined in the `wasmedge-types` crate, so have to use `URL`. Thanks!

hydai: Got it.

hydai (approved):

apepkuss:
@hydai Thanks for the review!

Issue #1293, opened 2022-03-02
LFX Workspace: Improving the performance of running rustpython


wenym1:
## Motivation

RustPython will be a practical use case on WasmEdge. Current RustPython has some known issues (such as #996 ) about soundness, robustness and performance when running on WasmEdge. This mentorship program will try to fix some issues about RustPython on WasmEdge and try to improve the performance under both AOT and non-AOT scenario for >7%.

## Milestone

- [x] Compile and build RustPython to `rustpython.wasm`
- [x] Build `wasmedge` and `wasmedgec` locally, use `wasmedgec` to transform `rustpython.wasm` to `rustpython-aot.wasm`, an AOT-optimized RustPython wasm file. Run `rustpython.wasm` and `rustpython-aot.wasm` on WasmEdge respectively.
- [x] Read the source code of WasmEdge and understand how it works.
- [ ] Fix #996 and figure out the reason why the program fails on exit.
- [x] Install `wasmtime` and `wasmer` and run RustPython with basic python code.
- [x] Develop several python codes for benchmark, including both computation-intensive and IO-intensive code
- [x] Develop and benchmark tools to easily run benchmark python codes on different wasm runtime and present the comparison in different dimension.
- [x] Use profiling tools to figure out the bottleneck of WasmEdge and try to improve the performance when RustPython is running on it. (Will add more details when we reach here)

wenym1:
Till now we have compiled and built RustPython to rustpython.wasm, built wasmedge and wasmedgec locally and run rustpython with both AOT-optimized and raw wasm file.

The error on exit problem for AOT-optimized wasm file is reproduced locally. Currently we are working on this problem and try to comprehend how WasmEdge works by the way.

We have figured out that the runtime fails when trying to call a host function (haven't figure out which host function was called) on exit. The failure to call the host function trigger a Fault named `terminated` with error code 0x01 and in the end result to an error.

wenym1:
Just analyze the reason why under AOT mode RustPython cannot exit cleanly in #1378

wenym1:
Submitted an issue about build failure on MacOS with M1 chip. #1485

wenym1:
Ran a simple rustpython benchmark with interpreter mode.  Following are some benchmark setting.
- Environment and steps to build the project is described in #1485.
- The wasm file of rustpython can be built according to the guild in [Github/RustPython](https://github.com/RustPython/RustPython).
[...]

PR #1507, opened 2022-05-28
perf(interpreter): use an more effective memcpy for memory load and store in interpreter

wenym1:
Currently, in interpreter mode, when handling op code related to memory load and store, we use a simple `memcpy` to load and store native memory. However, the length of such memory copy in `WasmEdge` is mostly (even all) the power of 2, like 1, 2, 4, 8 and etc, and when we copy memory of this size, it is more effective to load and store the memory as a whole, which is as a `uint8`, `uint16`, `uint32`, `uint64` and etc. The default memcpy simply read and store the memory byte by byte. In this PR, such improvement is implemented in a function `effective_memcpy`.

Besides, the `Length` of memory to be operated on can also be resolved statically at compile time. Therefore, I move the `Length` and `BitWidth` argument to the template argument so that the compiler can better optimized our code and reduce the number of branch.

The optimization in this PR is mentioned in #1293.

With this PR, the total cost to run a rust python benchmark reduces from 119 seconds to 111 seconds on MacOS M1 chip.


wenym1:
@hydai may also want to review the code.

include/runtime/instance/memory.h
```diff
@@ -241,16 +270,11 @@ class MemoryInstance {
...
-  loadValue(T &Value, uint32_t Offset, uint32_t Length) const noexcept {
+  loadValue(T &Value, uint32_t Offset) const noexcept {
     // Check the data boundary.
-    if (unlikely(Length > sizeof(T))) {
-      spdlog::error(ErrCode::MemoryOutOfBounds);
-      spdlog::error(
-          ErrInfo::InfoBoundary(Offset, Length, Offset + sizeof(T) - 1));
-      return Unexpect(ErrCode::MemoryOutOfBounds);
-    }
+    static_assert(Length <= sizeof(T));
```
q82419, line 277: Please don't remove the boundary checking here.
For the memory load out of boundary here, we should return error code but not assert.

include/runtime/instance/memory.h
```diff
@@ -294,16 +320,11 @@ class MemoryInstance {
...
-  storeValue(const T &Value, uint32_t Offset, uint32_t Length) noexcept {
+  storeValue(const T &Value, uint32_t Offset) noexcept {
     // Check the data boundary.
-    if (unlikely(Length > sizeof(T))) {
-      spdlog::error(ErrCode::MemoryOutOfBounds);
-      spdlog::error(
-          ErrInfo::InfoBoundary(Offset, Length, Offset + sizeof(T) - 1));
-      return Unexpect(ErrCode::MemoryOutOfBounds);
-    }
+    static_assert(Length <= sizeof(T));
```
q82419, line 327: As above.

include/runtime/instance/memory.h
```diff
@@ -261,16 +285,18 @@ class MemoryInstance {
     if (likely(Length > 0)) {
       if constexpr (std::is_floating_point_v<T>) {
         // Floating case. Do the memory copy.
-        std::memcpy(&Value, &DataPtr[Offset], sizeof(T));
+        *(reinterpret_cast<T *>(&Value)) =
+            *(reinterpret_cast<const T *>(&DataPtr[Offset]));
       } else {
         if constexpr (sizeof(T) > 8) {
-          assuming(sizeof(T) == 16);
+          static_assert(sizeof(T) == 16);
```
q82419, line 292: Please use `assuming` here instead of `static_assert`. This will fit for both debug and release versions.

q82419 (changes requested):

include/runtime/instance/memory.h
wenym1: Both the `Length` and `sizeof(T)` can be known at compile time and `static_assert` will do the check at compile time and fail the compile if the condition is not met.

Do we have any cases and any op that will explicitly load and store out of bound memory in our WasmEdge runtime code? If not, it will be beneficial to use `static_assert` here. First it can prevent us from writing incorrect code by accident. Second it avoids an unnecessary comparison at runtime and avoids a potential branch miss.

include/runtime/instance/memory.h
wenym1: Same reason to use `static_assert` as above.

include/runtime/instance/memory.h
wenym1: Same reason to use static_assert as above.

include/runtime/instance/memory.h
q82419: Although the `Length` can be known at compile time, the instructions can only be known at runtime.
For WASM, we can only get the `Length` by different instructions at runtime and output the error message.

include/runtime/instance/memory.h
q82419: As explained above.

include/runtime/instance/memory.h
q82419: Update:

Sorry that I misunderstood the code.
The boundary checking will be done in the following 2 lines.
You're right, this can be checked with `static_assert`. Thanks.

include/runtime/instance/memory.h
```diff
@@ -261,16 +285,18 @@ class MemoryInstance {
     if (likely(Length > 0)) {
       if constexpr (std::is_floating_point_v<T>) {
         // Floating case. Do the memory copy.
-        std::memcpy(&Value, &DataPtr[Offset], sizeof(T));
+        *(reinterpret_cast<T *>(&Value)) =
```
ibmibmibm, line 288: Use `std::memcpy` instead.

include/runtime/instance/memory.h
```diff
@@ -261,16 +285,18 @@ class MemoryInstance {
...
+            *(reinterpret_cast<const T *>(&DataPtr[Offset]));
       } else {
         if constexpr (sizeof(T) > 8) {
-          assuming(sizeof(T) == 16);
+          static_assert(sizeof(T) == 16);
           Value = 0;
-          std::memcpy(&Value, &DataPtr[Offset], Length);
+
+          effective_memcpy<Length>(&Value, &DataPtr[Offset]);
```
ibmibmibm, line 295: Use `std::memcpy` instead.

include/runtime/instance/memory.h
```diff
@@ -294,16 +320,11 @@ class MemoryInstance {
   /// \param Length the store length to data. Need to <= sizeof(T).
   ///
   /// \returns void when success, ErrCode when failed.
-  template <typename T>
+  template <typename T, uint32_t const Length = sizeof(T)>
```
ibmibmibm, line 323: `const` in a template parameter definition is meanless.

ibmibmibm (changes requested):

include/runtime/instance/memory.h
```diff
@@ -31,6 +31,35 @@ namespace WasmEdge {
 namespace Runtime {
 namespace Instance {
 
+template <uint32_t const Length>
+void effective_memcpy(void *dest, const void *src) {
```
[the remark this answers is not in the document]
wenym1: If we use `volatile` keyword for the `reinterpret_cast`, can we possibly ignore the alignment check? In the CI, all unit tests succeed in debug mode but some fail with seg fault in release mode. Probably in release mode some optimizations are performed provided the assumption that all memory address are aligned. If we mark the cast as `volatile`, can we fix it?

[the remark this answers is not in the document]
wenym1: I have added an alignment check in the `effective_memcpy`. When the `src` and `dest` is not properly aligned, it will use memcpy instead. 

Now the previous failed CI can all get passed. I also ran the rustpython benchmark, and the performance is basically the same as the one without the alignment check, which means adding the alignment check can still gain almost the same performance improvement. Therefore I think we can still use the `effective_memcpy` instead.

include/runtime/instance/memory.h
wenym1: I have resumed to use the original `assume` instead. 

Thanks for pointing it out. I didn't know the if constexpr restriction.

include/runtime/instance/memory.h
wenym1: I have removed the `const`

include/runtime/instance/memory.h
```diff
@@ -241,16 +270,11 @@ class MemoryInstance {
   /// \param Length the load length from data. Need to <= sizeof(T).
   ///
   /// \returns void when success, ErrCode when failed.
-  template <typename T>
+  template <typename T, uint32_t const Length = sizeof(T)>
```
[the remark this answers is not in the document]
wenym1: I have removed the `const`

wenym1:
I have updated the code according to some comments. 

For the `memcpy`, I added the alignment check, and in case of any of `dest` and `src` is not aligned, it will use `memcpy` instead. 

Are there any new comments? @q82419 @ibmibmibm

ibmibmibm:
Could you just check that performance is improved by using `effective_memcpy`, instead of `std::memcpy`?
I doubt that matter.

include/runtime/instance/memory.h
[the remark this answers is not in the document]
wenym1: In this PR, we only have two main changes, refactor the code with const template and introduce effective_memcpy. And I just ran my benchmark again under different combination on my macbook with M1 chip.
| time | run1 | run2 | run3 | avg |
| --- | --- | --- | --- |  --- |
|Current | 118.08s | 117.88s | 118.87s | 118.28s |
|Current with effective_memcpy| 116.11s | 115.41s | 115.69s | 115.74s|
| const template with std::memcpy | 111.03s | 111.15s | 111.38s | 111.19s |
| const template with effective_memcpy | 110.87s | 111.10s | 112.04s | 111.34s| 

Seems that the performance is the basically the same for std::memcpy and effective_memcpy when we refactor with const template, while in the original code, introducing effective_memcpy does improve the performance. 

Maybe the compiler will do some optimization on std::memcpy when it knows the `Length` in advance in compile time?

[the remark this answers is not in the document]
wenym1: My guess is correct. 

Just try compile the code on `compiler explorer`. If we know the length of `memcpy` in advance, the compiler will not call `memcpy` and instead generate several load instruction. If the length is provided at runtime passed from the function parameter, the generated assembly code will call `memcpy`.

https://godbolt.org/z/5z1z5WPY7

wenym1:
> Could you just check that performance is improved by using effective_memcpy, instead of std::memcpy?
I doubt that matter.

I have put a benchmark result in the discussion above. It seems that when we know the `Length` in advance, the compiler will do the optimization for us automatically, and therefore refactoring with const template is enough and we don't need the extra `effective_memcpy`. I have removed the new function and use memcpy back.

For the optimization done by compiler you can refer to https://godbolt.org/z/5z1z5WPY7

@ibmibmibm

ibmibmibm (approved):

ibmibmibm:
You need to squash your commit into one, and rebase it to master.

wenym1:
@q82419 Could you revert the change request? It's blocking me from merging this PR.

q82419 (approved):

PR #1520, opened 2022-06-07
[WASI] provide a helper function to get the native handler

hydai:

### DIFF
--- a/lib/host/wasi/inode-win.cpp
+++ b/lib/host/wasi/inode-win.cpp
@@ -176,6 +176,10 @@ WasiExpect<void> INode::fdWrite(Span<Span<const uint8_t>> IOVs,
   return {};
 }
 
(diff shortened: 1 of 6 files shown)


lib/host/wasi/inode-win.cpp
```diff
@@ -176,6 +176,10 @@ WasiExpect<void> INode::fdWrite(Span<Span<const uint8_t>> IOVs,
   return {};
 }
 
+WasiExpect<uint64_t> INode::getNativeHandler() const noexcept {
+  return static_cast<uint64_t>(Handle);
```
ibmibmibm, line 180: `reinterpret_cast`

ibmibmibm (changes requested):

lib/host/wasi/inode-win.cpp
hydai: Fixed.

PR #1528, opened 2022-06-08
Point to the right directoy for wasm examples

yixingjia:
Signed-off-by: yixingjia@gmail.com

Previous PR move the example out of the tools folder which will cause
the app cannot load the wasm file.


hydai (approved):

hydai:
Good catch. Thanks

PR #1537, opened 2022-06-13
[Rust]: Fix errors in examples.

EmisonLu:
Fix errors in `hostfunc.rs` and `hostfunc2.rs`.

Issue #1527, opened 2022-06-08
[Rust] wasmedge-sys: A linking error was raised when building `WasmEdge` on the Windows platform


hydai:
A linking error was raised when building `WasmEdge`. It seems related to plugins. Do you have any idea about how to fix it? As a reference, in case, the entire log of the CI workflow can be found [here](https://github.com/apepkuss/WasmEdge-1/runs/6767211091?check_suite_focus=true). Thanks!

```
[133/163] Linking CXX shared library plugins\\wasmedge_process\\wasmedgePluginWasmEdgeProcess.dll
FAILED: plugins/wasmedge_process/wasmedgePluginWasmEdgeProcess.dll plugins/wasmedge_process/wasmedgePluginWasmEdgeProcess.lib
[...]

PR #1541, opened 2022-06-14
[Rust] Fixed Clippy issues

apepkuss:
Duo to the upgrade of Rust nightly to `v1.63`, the feature `explicit_generic_args_with_impl_trait` is supported as default. This case is blocking #1537.  In this PR, the feature `explicit_generic_args_with_impl_trait` is commented out and kept for users with lower than `v1.63`. 

In addition, due to #1527, `WasmEdge Rust SDK` cannot be deployed successfully on the Windows platform. A necessary statement for this case is added in API documents and README.md.

Along with this PR, `wasmedge-sdk` comes to `v0.1.1`, and `wasmedge-sys` comes to `v0.7.1`.


hydai (approved):

hydai:
Hi @apepkuss 
After #1481 merged, the definition of Limit has changed.
Now, we have a new property called shared, which identifies whether the memory is shared or not. So you may need to update it.
Also, I am not sure if you want to have another PR to fix this or not. If so, I can merge this first.

apepkuss:
> Details

@hydai Thanks for the reminder. I'll support #1481 in another PR. Thanks a lot!

apepkuss:
@hydai Thanks for the review!

PR #1545, opened 2022-06-15
[Misc] Add the necessary components of WasmEdge static library.

q82419:


lib/api/CMakeLists.txt
```diff
@@ -91,16 +91,19 @@ if(WASMEDGE_BUILD_SHARED_LIB)
 
 endif()
 
-if(WASMEDGE_BUILD_STATIC_LIB)
+if(WASMEDGE_BUILD_STATIC_LIB) 
 
+  # FIXME: spdlog::spdlog will fail on Darwin.
```
hydai, line 96: Don't we have CI for building and testing this?

hydai (changes requested):

lib/api/CMakeLists.txt
q82419: Our CI doesn't test for `-DWASMEDGE_BUILD_STATIC_LIB`.
Only the dynamic library are tested, because the static library is not often to be used.

q82419: For the current status, the `proxy-wasm` uses our static linking.

lib/api/CMakeLists.txt
```diff
@@ -91,16 +91,19 @@ if(WASMEDGE_BUILD_SHARED_LIB)
 
 endif()
 
-if(WASMEDGE_BUILD_STATIC_LIB)
+if(WASMEDGE_BUILD_STATIC_LIB)
```
hydai, line 94: Is this a redundant space?

hydai (changes requested):

hydai (approved):

PR #1582, opened 2022-06-22
[Feat] Fix extra branch  in addfunc

eat4toast:
* check the first item and dispatch to different loops

Signed-off-by: eat4toast <shihaohang@outlook.com>


lib/executor/instantiate/function.cpp
```diff
@@ -18,17 +18,24 @@ Expect<void> Executor::instantiate(Runtime::Instance::ModuleInstance &ModInst,
   auto TypeIdxs = FuncSec.getContent();
   auto CodeSegs = CodeSec.getContent();
 
-  // Iterate through the code segments to instantiate function instances.
-  for (uint32_t I = 0; I < CodeSegs.size(); ++I) {
-    // Create and add the function instance into the module instance.
-    auto *FuncType = *ModInst.getFuncType(TypeIdxs[I]);
-    if (auto Symbol = CodeSegs[I].getSymbol()) {
-      ModInst.addFunc(&ModInst, *FuncType, std::move(Symbol));
+  if (CodeSegs.size() != 0) {
```
hydai, line 21: Try to use the short cut.

```
if (CodeSegs.size() == 0) {
  return {};
}
then...
```

hydai (changes requested):

PR #1591, opened 2022-06-26
Fix broken JS example links.

Scuilion:


juntao:
Hi @Scuilion 

Thank you so much for spotting and fixing these!

The WasmEdge project requires all contributors to sign a DCO message in their commit before we can merge. Can you do that? Thanks again. 

https://www.secondstate.io/articles/dco/

Cheers
Michael

PR #1643, opened 2022-07-12
[Proposal] wasi-crypto in plugin

q82419:


.github/workflows/build-extensions.yml
```diff
@@ -56,33 +56,188 @@ jobs:
...
             docker_tag: ubuntu-build-gcc
+            test_path: test/plugins/wasi_nn
+            test_bin: wasiNNTests
+            build_type: Debug
+          - name: WASI-NN OpenVINO
+            compiler: g++
+            docker_tag: ubuntu-build-gcc
+            test_path: test/plugins/wasi_nn
+            test_bin: wasiNNTests
             build_type: Release
```
hydai, line 67: Do we need to check both Release type and Debug type?
I found that most of the WASI-NN part are not using Debug feature.

.github/workflows/build-extensions.yml
```diff
@@ -56,33 +56,188 @@ jobs:
...
+  build_manylinux:
+    strategy:
+      matrix:
+        include:
+          - name: WASI-Crypto
+            system: manylinux2014 x86_64
+            host_runner: ubuntu-latest
+            docker_tag: manylinux2014_x86_64
+            plugin_path: build/plugins/wasi_crypto
+            build_type: Debug
```
hydai, line 159: Ditto.

hydai (changes requested):

hydai (approved):

PR #1648, opened 2022-07-14
[Rust] Introduced new host function storing design for the single-thread scenario

apepkuss:
In this PR, a new global host function storing mechanism is introduced for the single-thread use cases, such as hostfunc call chaining. `wasmedge-sys` and `wasmedge-sdk` get involved in this PR.

### DIFF
--- a/bindings/rust/wasmedge-sys/src/lib.rs
+++ b/bindings/rust/wasmedge-sys/src/lib.rs
@@ -179,7 +179,7 @@ lazy_static! {
             env::var("MAX_HOST_FUNC_LENGTH")
                 .map(|s| s
                     .parse::<usize>()
-                    .expect("MAX_HOST_THREADED_FUNC_LENGTH should be a positive integer."))
+                    .expect("MAX_HOST_FUNC_LENGTH should be a positive integer."))
                 .unwrap_or(500)
         )));
 }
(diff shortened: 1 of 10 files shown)


bindings/rust/wasmedge-sys/src/lib.rs
```diff
@@ -168,20 +170,34 @@ pub use vm::Vm;
...
 
 lazy_static! {
     static ref HOST_FUNCS: Arc<Mutex<HashMap<usize, BoxedFn>>> =
         Arc::new(Mutex::new(HashMap::with_capacity(
-            std::env::var("MAX_HOST_FUNC_LENGTH")
+            env::var("MAX_HOST_FUNC_LENGTH")
                 .map(|s| s
                     .parse::<usize>()
-                    .expect("MAX_HOST_FUNC_LENGTH should be a positive integer."))
+                    .expect("MAX_HOST_THREADED_FUNC_LENGTH should be a positive integer."))
```
hydai, line 182: The variable name is not the same as L179. Could you check which name should be used here?
If the name of environment variable is changed, should it be in the changelog?

hydai (changes requested):

bindings/rust/wasmedge-sys/src/lib.rs
apepkuss: Fixed by using `MAX_HOST_FUNC_LENGTH`.

hydai (approved):

apepkuss:
@hydai Thanks for the review!

PR #1649, opened 2022-07-14
Set CMAKE_PROJECT_VERSION instead of just CPACK_

paravoid:
Set CMAKE_PROJECT_VERSION by passing VERSION to project(). This in turn
sets CMAKE_PROJECT_VERSION_MAJOR/MINOR/etc., as well as
CPACK_PACKAGE_VERSION, automatically.

This automatically sets PROJECT_VERSION as well as _MAJOR/_MINOR/_PATCH,
removing custom code that we have to parse the version string with a
regex and set WASMEDGE_VERSION_*, simplifying our CMakeLists.

While at it, move the "git describe" call inside an else(), to avoid
needlessly calling git if there is a VERSION file present.


CMakeLists.txt
```diff
@@ -82,7 +81,7 @@ include(GNUInstallDirs)
 
 string(TOUPPER "${CMAKE_INSTALL_DEFAULT_COMPONENT_NAME}" CMAKE_INSTALL_DEFAULT_COMPONENT_NAME_UPCASE)
 set(CPACK_PACKAGE_VENDOR Second State LLC)
-set(CPACK_PACKAGE_FILE_NAME "${CMAKE_PROJECT_NAME}-${CPACK_PACKAGE_VERSION}-${CMAKE_SYSTEM_NAME}")
+set(CPACK_PACKAGE_FILE_NAME "${CMAKE_PROJECT_NAME}-${CNAME_PROJECT_VERSION}-${CMAKE_SYSTEM_NAME}")
```
hydai, line 84: Is this a typo? Maybe `CMAKE_PROJECT_VERSION`?

hydai (changes requested):

hydai:
WasmEdge uses `git describe` to compose a version such as `0.10.1-alpha.2-20-g7580f982` as the version of any unreleased binaries and libraries. However, I am not sure if the CMake project version can support this format or not. Or do we also need to change the version format to fit the standard of the CMake version?

paravoid:
> WasmEdge uses `git describe` to compose a version such as `0.10.1-alpha.2-20-g7580f982` as the version of any unreleased binaries and libraries. However, I am not sure if the CMake project version can support this format or not. Or do we also need to change the version format to fit the standard of the CMake version?

You're absolutely right. Sadly, I realized that shortly after preparing this patch :/ I tested this with a local tarball and a git tagged version, but failed to test it with a pre-release.

CMake upstream seems to be discussing this here: https://gitlab.kitware.com/cmake/cmake/-/issues/16716 but even if that would get fixed today, it would take a while until that CMake version becomes ubiquitous enough for WasmEdge to be able to rely on it.

I'm not sure how to proceed at this moment. While working on #1677, I found it natural to be able to use CMAKE_PROJECT_VERSION (and _MAJOR etc.), and the code duplication elimination seemed useful. It certainly seems useful to define these versions rather tha 0.0.0 even in prereleases. However I also realize the usefulness of CMake not pretending that master is at 0.10.0 until a new version is released.

Your guidance is welcome, and I can have a look at updating the PR based on your feedback. Thanks for reviewing this so promptly!

hydai:
Hi @paravoid 
After surveying, I have an idea:
1. We still use `git describe` to get the version: 0.10.1-alpha.2-20-g7580f982
2. And split this version into three parts: `X.Y.Z`-`alpha/beta/rc.*`-`hash`.
3. So we can set the CMAKE_PROJECT_VERSION to `X.Y.Z`.
4. Then, we will have a new flag called `unreleased` or something that can present this is not an official release.
5. `wasmedge --version` will show the `X.Y.Z` pluses `unreleased` and `remaining fields`.

Not sure if this is a better idea. Maybe we can have more examples to figure out which is a better solution for this.

PR #1654, opened 2022-07-18
[WASINN] Support PyTorch Backend

gusye1234:
Enable pytorch backend in the Wasi-NN plugin.

### DIFF
--- a/plugins/wasi_nn/wasinnenv.h
+++ b/plugins/wasi_nn/wasinnenv.h
@@ -5,12 +5,15 @@
 
 #include "plugin/plugin.h"
 
+#include "common/log.h"
 #include <cstdint>
 #include <vector>
 
 #ifdef WASMEDGE_PLUGIN_WASI_NN_BACKEND_OPENVINO
-#include "common/log.h"
 #include <c_api/ie_c_api.h>
+#endif
+#ifdef WASMEDGE_WASINN_BACKEND_TORCH
+#include <torch/script.h>
 #endif
 
 namespace WasmEdge {
@@ -26,16 +29,21 @@ enum class ErrNo : uint32_t {
 
 enum class Backend : uint8_t {
   OpenVINO = 0,
+  PyTorch = 1,
 };
 
 class Graph {
 public:
-#ifdef WASMEDGE_PLUGIN_WASI_NN_BACKEND_OPENVINO
   Graph() = delete;
-  Graph(Backend BE) noexcept
-      : GraphBackend(BE), OpenVINONetwork(nullptr),
-        OpenVINOExecNetwork(nullptr), OpenVINOWeightBlob(nullptr) {}
+  Graph(Backend BE) noexcept : GraphBackend(BE) {
+#ifdef WASMEDGE_PLUGIN_WASI_NN_BACKEND_OPENVINO
+    OpenVINONetwork = nullptr;
+    OpenVINOExecNetwork = nullptr;
+    OpenVINOWeightBlob = nullptr;
+#endif
+  }
   ~Graph() noexcept {
+#ifdef WASMEDGE_PLUGIN_WASI_NN_BACKEND_OPENVINO
     if (OpenVINONetwork) {
       ie_network_free(&OpenVINONetwork);
     }
@@ -55,10 +63,8 @@ public:
         ie_network_name_free(&I);
       }
     }
-  }
-#else
-  Graph() noexcept = default;
 #endif
+  }
 
   Backend GraphBackend;
 #ifdef WASMEDGE_PLUGIN_WASI_NN_BACKEND_OPENVINO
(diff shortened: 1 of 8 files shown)


hydai:
Hi @gusye1234 
If you just want to test the PyTorch part first, you can remove the other workflow files/jobs temporarily. Then add them back once you finish the PyTorch workflow. This can reduce the total workflow execution time and accelerate your iteration.

gusye1234:
> Hi @gusye1234 If you want to test the PyTorch part, you can temporarily remove the other workflow files/jobs. Then add them back once you finish the PyTorch workflow. This can reduce the total workflow execution time and accelerate your iteration. The reason is that our organization only has 20 action runners and we have lots of action jobs. The remaining jobs which may not be related to your testing will execute and take the seats of those runners. And this may make your waiting time an extended period.

Thanks for the reminder!  And sorry for my debug pushings disturbing you guys a lot...
Is there a way to temporarily configure to remove the other jobs? Or I have to remove them in the `.yaml` files?

hydai:
Just create a temp commit which remove those files or non-related jobs. Once you finish, then revert that commit.

plugins/wasi_nn/wasinnfunc.cpp
```diff
@@ -270,24 +341,40 @@ WasiNNInitExecCtx::body(Runtime::Instance::MemoryInstance *MemInst,
...
     return static_cast<uint32_t>(WASINN::ErrNo::Success);
 #else
     spdlog::error("[WASI-NN] OpenVINO backend is not built. define "
                   "-DWASMEDGE_PLUGIN_WASI_NN_BACKEND_OPENVINO "
                   "to build it.");
 #endif
+  } else if (Env.NNGraph[GraphId].GraphBackend == WASINN::Backend::PyTorch) {
+#ifdef WASMEDGE_PLUGIN_WASI_NN_BACKEND_TORCH
+    // Check the return value: Context should be valid.
+    uint32_t *Context = MemInst->getPointer<uint32_t *>(ContextPtr, 1);
```
q82419, line 363: As above. The return value pointer validation can be checked outside.

q82419 (changes requested):
Not too much problems.
You can split out the common checkings of all backends to reduce the redundant codes.

plugins/wasi_nn/wasinnenv.h
```diff
@@ -68,26 +74,51 @@ class Graph {
...
+      IEStatusCode Status = ie_exec_network_create_infer_request(
+          G.OpenVINOExecNetwork, &OpenVINOInferRequest);
+      if (Status != IEStatusCode::OK) {
+        OpenVINOInferRequest = nullptr;
+        spdlog::error("[WASI-NN] Unable to create infer request for OpenVINO");
+      }
+#endif
+    } else if (G.GraphBackend == Backend::PyTorch) {
+#ifdef WASMEDGE_PLUGIN_WASI_NN_BACKEND_TORCH
+      TorchModel = G.TorchModel;
```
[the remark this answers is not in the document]
gusye1234: The [official demo](https://pytorch.org/tutorials/advanced/cpp_export.html#step-3-loading-your-script-module-in-c) was using this way to assign a new model:
```c++
torch::jit::script::Module module;
// torch::jit::load returns a Module
module = torch::jit::load(argv[1]);
```
I looked down at the definition of `Module` and didn't find any overload about the `=` operator. So maybe it's copying? I'm not sure, but in fact, we don't need to copy that model. Only one instance is enough.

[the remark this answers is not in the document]
q82419: The example you gave is to load and transfer the ownership of the model to `module`.
In our case, the model is owned by the `Graph`. The `Context` is to create the infering context of the infering requests.
I think you should use the `Context.GraphRef.TorchModel` directly?

q82419 (changes requested):

test/plugins/wasi_nn/wasi_nn.cpp
```diff
@@ -473,3 +477,375 @@ TEST(WasiNNTest, OpenVINOBackend) {
...
+                             std::initializer_list<WasmEdge::ValVariant>{
+                                 UINT32_C(3), UINT32_C(0), SetInputEntryPtr},
+                             Errno));
+    EXPECT_EQ(Errno[0].get<int32_t>(),
+              static_cast<uint32_t>(ErrNo::InvalidArgument));
+  }
+
+  NNContextTmp.emplace_back(NNGraphTmp[0]);
+
+  //   Abort this test, we can't know the inputs for a torch model
```
q82419, line 719: If this test is not necessary, delete this paragraph.

utils/wasi-nn/download-pytorch-fixtures.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+DOWNLOAD_TO=$1
+FIXTURE=https://github.com/gusye1234/torchscript_fixtures/raw/main/mobilenet
```
q82419, line 4: Is there a official site of the mobilenet pytorch model?

q82419 (changes requested):

utils/wasi-nn/download-pytorch-fixtures.sh
gusye1234: Probably not. I didn't find any place that stores the torch-scripted models.

q82419 (approved):

PR #1672, opened 2022-07-20
[Utils] Don't add current directory to path if the $XX_PATH is empty

spacewander:
In POSIX, path `xx/lib:` means two pathes, one is `xx/lib`, another is
`.` (an empty path after ':' means `.`).
Signed-off-by: spacewander <spacewanderlzx@gmail.com>


SAtacker (approved):
Thanks for this!

hydai (approved):

PR #1682, opened 2022-07-26
[RISC-V] Build WasmEdge on RISC-V 64 arch

O3Ol:
# Motivation

This pull request is to add files which make WasmEdge build for `RISC-V 64` arch.

# Details

- Add the definitions of riscv64.
- Add `build_for_riscv.yml` for CI test.
- Add the doc `build_for_RISC-V64.md` for buiding WasmEdge on RISC-V64.


.github/workflows/build_for_riscv.yml
```diff
@@ -0,0 +1,101 @@
...
+        cmake -Bbuild -GNinja -DCMAKE_BUILD_TYPE=Release -DWASMEDGE_BUILD_PACKAGE="TGZ" \
+              -DWASMEDGE_BUILD_AOT_RUNTIME=OFF -DCMAKE_CROSSCOMPILING=TRUE \
+              -DCMAKE_SYSTEM_NAME=Linux -DCMAKE_SYSROOT="/opt/riscv64/sysroot" \
+              -DCMAKE_C_COMPILER=riscv64-unknown-linux-gnu-gcc \
+              -DCMAKE_CXX_COMPILER=riscv64-unknown-linux-gnu-g++ \
+              -DCMAKE_FIND_ROOT_PATH=/opt/riscv64/riscv64-unknown-linux-gnu \
+              -DCMAKE_FIND_ROOT_PATH_MODE_PROGRAM=NEVER \
+              -DCMAKE_FIND_ROOT_PATH_MODE_LIBRARY=ONLY \
+              -DCMAKE_FIND_ROOT_PATH_MODE_INCLUDE=ONLY \
+              -DCMAKE_FIND_ROOT_PATH_MODE_PACKAGE=ONLY \
```
hydai, line 94: Looks like you have an extra `\` in this line. So it broke the workflow.

hydai (changes requested):

.github/workflows/build_for_riscv.yml
```diff
@@ -0,0 +1,102 @@
...
+              -DCMAKE_FIND_ROOT_PATH_MODE_LIBRARY=ONLY \
+              -DCMAKE_FIND_ROOT_PATH_MODE_INCLUDE=ONLY \
+              -DCMAKE_FIND_ROOT_PATH_MODE_PACKAGE=ONLY
+        cmake --build build
+        cmake --build build --target package
+    - name: Upload artifact (v2)
+      uses: actions/upload-artifact@v2
+      with:
+        name: WasmEdge-${{ needs.get_version_v2.outputs.version }}-linux_RISC-V64.tar.gz
+        path: build/WasmEdge-${{ needs.get_version_v2.outputs.version }}-linux_RISC-V64.tar.gz
```
hydai, line 102: The output path should be `***-Linux.tar.gz`. And the name is fine. You can check the output in your workflow. There is an error.

hydai (changes requested):

include/ast/instruction.h
```diff
@@ -37,7 +37,7 @@ class Instruction {
   /// Constructor assigns the OpCode and the Offset.
   Instruction(OpCode Byte, uint32_t Off = 0) noexcept
       : Offset(Off), Code(Byte) {
-#if defined(__x86_64__) || defined(__aarch64__)
+#if defined(__x86_64__) || defined(__aarch64__) || defined(__riscv)
```
hydai, line 40: Before adding the `__riscv` here, you should make sure that the `riscv` target supports `uint128_t`.
Otherwise, it should be fine to fall back to the `else` case.

include/common/int128.h
```diff
@@ -13,7 +13,7 @@
 //===----------------------------------------------------------------------===//
 #pragma once
 
-#if defined(__x86_64__) || defined(__aarch64__)
+#if defined(__x86_64__) || defined(__aarch64__) || defined(__riscv)
```
hydai, line 16: Ditto.

lib/system/allocator.cpp
```diff
@@ -8,7 +8,7 @@
 #include "common/errcode.h"
 
 #if defined(HAVE_MMAP) && defined(__x86_64__) || defined(__aarch64__) ||       \
-    defined(__arm__)
+    defined(__arm__) || defined(__riscv)
```
hydai, line 11: Please make sure the `riscv` target supports MMAP.

include/api/wasmedge/int128.h
```diff
@@ -15,7 +15,7 @@
 #ifndef WASMEDGE_C_API_INT128_H
 #define WASMEDGE_C_API_INT128_H
 
-#if defined(__x86_64__) || defined(__aarch64__)
+#if defined(__x86_64__) || defined(__aarch64__) || defined(__riscv)
```
hydai, line 18: I am not sure if the `__riscv` means `risc-v 64` only. If not, we should not use this directly.
The 32 bits and 64 bits are totally different cases. And we should separate them into different conditions. Also, you may need to check if `uint128_t` and `int128_t` are supported on these platforms.

hydai (changes requested):

docs/book/en/src/extend/build_for_RISC-V64.md
```diff
@@ -0,0 +1,196 @@
...
+
+### Download and compile linux kernel
+
+Download the tar.gz archive of linux-v5.14 and rename the extracted folder to `linux`.
+
+```bash
+$ wget https://github.com/torvalds/linux/archive/refs/tags/v5.14.tar.gz
+$ tar xzvf v5.14.tar.gz
+$ mv linux-v5.14 linux
+```
```
hydai, line 62: Looks like these are static versions. It should be able to package it into a docker image, so users don't need to do these stuff by themselves.

hydai (changes requested):

include/api/wasmedge/int128.h
O3Ol: In fact, there is a definition in `/opt/riscv64/include/c++/11.1.0/riscv64-unknown-linux-gnu/bits/c++config.h` which is used by `riscv-gnu-toolchain`:
```
/* Define if __int128 is supported on this host. */
#define _GLIBCXX_USE_INT128 1
```
And I write the `1.cpp` for test:
```cpp
#include<iostream>
using namespace std;
int main(){
	#if __SIZEOF_INT128__
	cout<<"yes"<<endl;
	#else
	cout<<"no"<<endl;
	#endif
	return 0;
}
```
Then I test it:
```bash
(base) wcy@wcy-Lenovo-XiaoXinAir-14IML-2019:~/test$ riscv64-unknown-linux-gnu-g++ 1.cpp -o 1.o
(base) wcy@wcy-Lenovo-XiaoXinAir-14IML-2019:~/test$ qemu-riscv64 -L /opt/riscv64/sysroot  1.o
yes
```
I'm not sure if it means that the riscv-64 support the `int128_t` or not?

O3Ol:
I find an error:
```
(base) wcy@wcy-Lenovo-XiaoXinAir-14IML-2019:~/riscv64-build/WasmEdge/_install/bin$ qemu-riscv64 -L /opt/riscv64/sysroot/ wasmedge --dir .:. ../../examples/js/qjs.wasm ../../examples/js/hello.js 1 2 3
Capabilities insufficient (os error 76)
```
And I searched for the relevant error meaning as follows:
```
OS error code  76:  Name not unique on network
```

include/api/wasmedge/int128.h
hydai: Yes. I think so

hydai:
To figure out this error, you may need to build a debug version of wasmedge and dig into the internal source code with gdb to check which part is failed in this case.

O3Ol:
I find the `/opt/riscv64` doesn't include the `strlen.c` and I find it is in the `linux kernel` fold.
```
(gdb) 
__GI_strlen (
    str=<error reading variable: dwarf2_find_location_expression: Corrupted DWARF expression.>)
    at strlen.c:42
42	strlen.c: 没有那个文件或目录.
```
So I try it on the filesystem image which includes the linux kernel, and it runs nomally:
```
~ # wasmedge --dir .:. qjs.wasm hello.js 1 2 3
[   24.645261] random: wasmedge: uninitialized urandom read (4 bytes read)
[   24.653626] random: wasmedge: uninitialized urandom read (4 bytes read)
Hello 1 2 3
```
Obviously, the error may happens because of the lack of the linux kernel or other filesystem's basic components.

So I think about the AOT mode. Does AOT mode require native gcc or clang compilation toolchains, or actual hardware support?

hydai:
> So I think about the AOT mode. Does AOT mode require native gcc or clang compilation toolchains, or actual hardware support?

So let me explain how the ahead-of-time mode works first. The ahead-of-time mode means compiling the wasm into a native binary on its platform. Let's take MacBook M1 as an example, the `wasmedgec` will try to compile wasm into `macOS native format` which will use and link some libraries such as libc on the host environment. Especially, some syscalls are needed when you want to use WASI. If there is no filesystem support on the host environment, then all of the filesystem-related functions will fail when you try to use AOT mode.

O3Ol:
> > So I think about the AOT mode. Does AOT mode require native gcc or clang compilation toolchains, or actual hardware support?
> 
> So let me explain how the ahead-of-time mode works first. The ahead-of-time mode means compiling the wasm into a native binary on its platform. Let's take MacBook M1 as an example, the `wasmedgec` will try to compile wasm into `macOS native format` which will use and link some libraries such as libc on the host environment. Especially, some syscalls are needed when you want to use WASI. If there is no filesystem support on the host environment, then all of the filesystem-related functions will fail when you try to use AOT mode.

Thank you for your reply! Previously I tested AOT on a simple filesystem image including the linux kernel mentioned above, but the error still occurred. So I wondered if the reason for the error was missing some necessary shared libraries that I didn't put into the image when I made the filesystem. I will continue to investigate.

lib/system/allocator.cpp
O3Ol: Yes, there is the following description in `sys/mman.h`.
```
#ifndef __USE_FILE_OFFSET64
extern void *mmap (void *__addr, size_t __len, int __prot,
		   int __flags, int __fd, __off_t __offset) __THROW;
#else
# ifdef __REDIRECT_NTH
extern void * __REDIRECT_NTH (mmap,
			      (void *__addr, size_t __len, int __prot,
			       int __flags, int __fd, __off64_t __offset),
			      mmap64);
# else
#  define mmap mmap64
# endif
#endif
```

hydai: Okay.

PR #1700, opened 2022-07-29
[CI] Refactored workflows related to Rust bindings for releasing docs

apepkuss:
In this PR, the `rust-wasmedge-sys-docs` and `rust-wasmedge-types-docs` workflows are removed. The `rust-wasmedge-sdk-docs` workflow is responsible for docs publishing.


apepkuss:
Hi @hydai, could you please help review this PR? Thanks a lot!

hydai:
The release CI should be refactoring either:
https://github.com/WasmEdge/WasmEdge/runs/7575954698?check_suite_focus=true

It depends on the `sys`, but it doesn't release in order.

apepkuss:
> Another thing is whether the document of `types` and `sys` will be published via `sdk` or not. I found you just remove `types` and `sys`. But there is not uploaded phase in the `sdk` one.

`sdk` depends on `sys` and `types`, so their docs will be published together.

apepkuss:
> 

The correct order of publishing the three crates should be `types` first, then `sys`, and the last one `sdk`. Could you please try to publish them in this order on your side?

PR #1713, opened 2022-08-01
[CI] workflow: make the concurrency group fallback to github.ref if t…

hydai:
…he event is not PR


0yi0 (approved):
Note: Link for the [docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-using-a-fallback-value).

Issue #1692, opened 2022-07-27
No IP address assigned for pods when "hostNetwork: false" in k8s


GreyBalloonYU:
## Description
When I follow the instruction (https://wasmedge.org/book/en/kubernetes/kubernetes/kubernetes-crio.html#a-webassembly-based-http-service) to make a WebAssembly-based HTTP service, I try to assign "false" to "hostnetwork" in yaml, as follows:
```
apiVersion: v1
kind: Pod
metadata:
  name: http-server
  namespace: default
  annotations:
    module.wasm.image/variant: compat
spec:
  hostNetwork: false  #default value
  containers:
  - name: http-server
    image: avengermojo/http_server:with-wasm-annotation
    command: [ "/http_server.wasm" ]
    ports:
    - containerPort: 1234
      protocol: TCP
    livenessProbe:
      tcpSocket:
        port: 1234
      initialDelaySeconds: 3
      periodSeconds: 30
```
But the result is: these is no IP address assigned for the pod:
```
NAME          READY   STATUS    RESTARTS   AGE   IP       NODE        NOMINATED NODE   READINESS GATES
http-server   1/1     Running   0          9s    <none>   127.0.0.1   <none>           <none>
```
Does the pod have IP address only when "hostNetwork: true"? 
Are there any solutions to solve this problem?
Or cannot solve it because the inherent characteristics of WebAssembly?
## Environment

- Hardware Architecture: x86_64
- Operating system:  Ubuntu 20.04

hydai:
Hi @GreyBalloonYU 

AFAIK, the IP means the public IP, if you set the hostnetwork to false, then it should not have the IP. I am not so familiar with k8s, if I said something wrong, please correct me, thanks.

GreyBalloonYU:
@hydai The IP is a virtual IP of the k8s pod. When I deploy a container like busybox in k8s, which "hostnetwork" is "false", k8s will assign a virtual IP for pod:
```
apiVersion: v1
kind: Pod
metadata:
  name: busybox
  namespace: default
spec:
  hostNetwork: false
  containers:
  - name: busybox
    image: library/busybox
    command: [ "sleep", "3600" ]
```
Then the pod will have a virtual IP address:
```
NAME      READY   STATUS    RESTARTS   AGE    IP           NODE        NOMINATED NODE   READINESS GATES
busybox   1/1     Running   0          8m4s   10.85.6.83   127.0.0.1   <none>           <none>
```
[...]

Issue #1338, opened 2022-03-09
[Docs] Add crun "-smart" annotation


0xE282B0:
Crun has just merged a workaround to run containers with different runtimes in the same pod: https://github.com/containers/crun/pull/886
Thanks to @flouthoc

With this change a pod can be annotated with `module.wasm.image/variant: compat-smart` and crun decides on the ending of the entrypoint if the container is WASM payload or an oci container. **That makes it possible to run WASM containers with sidecars.**

It would be great if that behavior would be documented.

0xE282B0:
We could add a chapter to the crun section https://wasmedge.org/book/en/kubernetes/container/crun.html or add a new section "Working with sidecars" under Kubernetes? 

Open for feedback 😊

juntao:
This is excellent. We would really appreciate an example and a chapter in the docs book. :) Maybe you could create an example here?

https://github.com/second-state/wasmedge-containers-examples

Thanks!

PR #1720, opened 2022-08-03
[Docs] use the annotation "module.wasm.image/variant: compat-smart" instead of "module.wasm.image/variant: compat"

GreyBalloonYU:
As mentioned by #1692 and #1338,
in case only wasm apps are used the module.wasm.image/variant: compat-smart annotation should behave exactly like the module.wasm.image/variant: compat annotation. It should only make a difference when standard OCI containers and wasm apps are in the same pod.
So we can update all examples to compat-smart.


GreyBalloonYU:
@juntao @0xE282B0

hydai (approved):

hydai:
Thanks!

PR #1730, opened 2022-08-08
[Common] Extend the error code as a struct to support user-defined

q82419:
errors.


include/common/enum_errcode.h
```diff
@@ -57,20 +57,45 @@ static inline constexpr auto WasmPhaseStr = []() constexpr {
...
+#undef UseErrCategory
+};
+
+#if (defined(__cplusplus) && __cplusplus > 201402L) ||                         \
+    (defined(_MSVC_LANG) && _MSVC_LANG > 201402L)
+namespace WasmEdge {
+
 /// Error code C++ enumeration class.
-enum class ErrCode : uint8_t {
+enum class ErrEnum : uint32_t {
```
hydai, line 86: What's the difference between `ErrEnum` and `ErrCategory`?

hydai (changes requested):

include/common/enum_errcode.h
q82419: `ErrEnum` is the origin `ErrCode` to specify the WASM runtime error codes.
`ErrCategory` is the error code category. Currently `WASM` is the `ErrEnum` collections, and `UserErrorLevel` is the user-defined error codes. In the future, we can extend the `WasiNN` or the `WasiCrypto` into the category to separate the error codes from `ErrEnum`. It means, the category size can be extended into 256 (8-bit) in maximum.

include/api/wasmedge/wasmedge.h
```diff
@@ -51,13 +51,22 @@ typedef struct WasmEdge_String {
   const char *Buf;
 } WasmEdge_String;
 
-/// Opaque struct of WASM execution result.
+/// WasmEdge result struct.
 typedef struct WasmEdge_Result {
-  uint8_t Code;
+  uint8_t Prefix;
+  uint8_t Code1;
```
hydai, line 57: `uint8_t Codes[3]`.

include/api/wasmedge/wasmedge.h
```diff
@@ -51,13 +51,22 @@ typedef struct WasmEdge_String {
   const char *Buf;
 } WasmEdge_String;
 
-/// Opaque struct of WASM execution result.
+/// WasmEdge result struct.
 typedef struct WasmEdge_Result {
-  uint8_t Code;
+  uint8_t Prefix;
```
hydai, line 56: `Category`.

include/common/enum_errcode.h
hydai, line 86: We should use `ErrCodeValue`. The `ErrEnum` is pretty wierd.

include/common/errcode.h
```diff
@@ -32,13 +32,71 @@
 
 namespace WasmEdge {
 
-static inline WasmPhase getErrCodePhase(ErrCode Code) {
-  return static_cast<WasmPhase>((static_cast<uint8_t>(Code) & 0xF0) >> 5);
+class ErrCode {
+public:
+  constexpr ErrCategory getCategory() const noexcept {
+    return static_cast<ErrCategory>(static_cast<uint8_t>(Inner.Num >> 24));
```
hydai, line 38: Considering the `endianess` issue, we should use `array (Your code in C)` or `shift (Your code in C++)` in both C and C++ implementations.

include/api/wasmedge/wasmedge.h
```diff
@@ -51,13 +51,22 @@ typedef struct WasmEdge_String {
...
+  uint8_t Code1;
+  uint8_t Code2;
+  uint8_t Code3;
 } WasmEdge_Result;
```
hydai, line 60: To avoid the `endianess` issue, it will be better to use `uint32_t` directly and provide some helper functions to get and set the internal values.

hydai (changes requested):

include/common/enum_errcode.h
hydai: Maybe we should move this enum class into `class ErrCode` as its member enum directly.

hydai (changes requested):

q82419:
@apepkuss Please help to fix the rust side after this PR merged. Thanks.

apepkuss:
@q82419 Thanks for the message. I'm on vacation so I'll refactor `wasmedge-sys` to support the new feature later. Thanks again!

include/common/enum_errcode.hpp
```diff
@@ -0,0 +1,134 @@
...
+#define Line(NAME, VALUE) NAME = VALUE,
+#include "enum.inc"
+#undef Line
+#undef UseErrCategory
+};
+
+class ErrCode {
+public:
+  /// Error code C++ enumeration class.
+  enum class Value : uint32_t {
```
hydai, line 59: Use `enum` directly, so the `errors` can be exposed at the same layer.
Such as `ErrCode::XXXError`. It makes no sense to create a new layer called `Value`.

hydai (changes requested):

include/common/enum_errcode.hpp
q82419: This is not supported in dense_map.

hydai (approved):

include/common/errcode.h
```diff
@@ -32,13 +32,16 @@
 
 namespace WasmEdge {
 
-static inline WasmPhase getErrCodePhase(ErrCode Code) {
-  return static_cast<WasmPhase>((static_cast<uint8_t>(Code) & 0xF0) >> 5);
+static inline WasmPhase getErrCodePhase(const ErrCode &Code) {
+  return static_cast<WasmPhase>(Code.getCategory() == ErrCategory::WASM
+                                    ? ((Code.getCode() & 0xF0U) >> 5)
+                                    : 0x08U);
```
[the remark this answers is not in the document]
q82419: It should be `WasmPhase::UserDefined`. Fixed.

hydai (approved):
LGTM.

hydai (approved):

hydai (approved):

PR #1733, opened 2022-08-08
[Rust] Refactored Memory::write method in `wasmedge-sdk`

### LINKED ISSUE #1728: [Rust SDK]function `wasmedge_sdk::memory::write`
Currently, the `memory` type of wasmedge-sdk, has the method [`write()`](https://github.com/WasmEdge/WasmEdge/blob/master/bindings/rust/wasmedge-sdk/src/externals/memory.rs#L94), which is able to safely write contents of a buffer to this `memory` at the given offset. However, the method [`write()`](https://github.com/WasmEdge/WasmEdge/blob/master/bindings/rust/wasmedge-sdk/src/externals/memory.rs#L94) requires taking ownership of the input parameter `data`, which is undesirable in some cases.

For example, if `data` is of type `Arc<Vec<...>>`, we might need the following way to call this function: `memory.write(unsafe { &*Arc::into_raw(data)}.clone(), offset)` (perhaps there is a much better way to implement this).

These cases can be avoided if the parameter `data` passed in by [`write()`](https://github.com/WasmEdge/WasmEdge/blob/master/bindings/rust/wasmedge-sdk/src/externals/memory.rs#L94) is a reference.

When using wasmedge-sys, since there is no `write()` API, I use the following as an alternative.

```rust
fn write_memory(memory: &mut Memory, offset: usize, buffer: &[u8]) -> Result<(), anyhow::Error> {
    let base_ptr_mut: *mut u8 = memory
        .data_pointer_mut(0, 1)
        .expect("failed to returns the mut data pointer to the memory.");

    unsafe { std::slice::from_raw_parts_mut(base_ptr_mut, (memory.size() * 64 * 1024) as usize) }
        .get_mut(offset..)
        .and_then(|s| s.get_mut(..buffer.len()))
        .unwrap()
        .copy_from_slice(buffer);

    Ok(())
}
```

@apepkuss Whether the argument passed in can be a `&[]`? Please take a look at this question, thank you very much!

apepkuss:
This PR is to refactor `Memory::write` method in `wasmedge-sdk` and `Memory::set_data` method in `wasmedge-sys`.

- Fixed #1728

### DIFF
--- a/bindings/rust/wasmedge-sdk/src/externals/memory.rs
+++ b/bindings/rust/wasmedge-sdk/src/externals/memory.rs
@@ -91,7 +91,7 @@ impl Memory {
     /// # Error
     ///
     /// If fail to write to the memory, then an error is returned.
-    pub fn write(&mut self, data: impl IntoIterator<Item = u8>, offset: u32) -> WasmEdgeResult<()> {
+    pub fn write(&mut self, data: impl AsRef<[u8]>, offset: u32) -> WasmEdgeResult<()> {
         self.inner.set_data(data, offset)?;
         Ok(())
     }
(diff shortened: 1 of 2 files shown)


apepkuss:

hydai (approved):

PR #1746, opened 2022-08-12
[Loader] Fix for the binary format change in multi-memory proposal.

### LINKED ISSUE #1725: [Proposal] breaking change in multi-memory spec. The order of load/store immediates should be switched
## Description

There is a breaking change in multi-memory proposal. We need to change the behavior of decoding multi-memory instruction.

Reference:
* https://github.com/WebAssembly/multi-memory/pull/29
* https://github.com/WebAssembly/multi-memory/issues/28

### Expected

The order of load/store immediates should be switched in binary format.

q82419:
Close #1725

### DIFF
--- a/lib/loader/ast/instruction.cpp
+++ b/lib/loader/ast/instruction.cpp
@@ -856,6 +856,7 @@ Expect<void> Loader::loadInstruction(AST::Instruction &Instr) {
   case OpCode::Atomic__fence:
     return {};
 
+  // Atomic Memory Instructions.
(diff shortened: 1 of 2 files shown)


hydai:
This PR should be merged after https://github.com/WebAssembly/multi-memory/pull/33 gets merged.

hydai:
We can merge this before the spec updates. I think it's very clear that the ordering between `memidx` and `offset` must be swapped.

hydai (approved):

PR #1778, opened 2022-08-23
README-zh-TW: get rid of Simplified Chinese

yan12125:
"语言" is a Simplified Chinese term. I changed it to the corresponding Traditional Chinese term.

Signed-off-by: Chih-Hsuan Yen <yan12125@gmail.com>


hydai (approved):
Nice catch

yan12125:
Thanks!

PR #1783, opened 2022-08-23
Enable SONAME and SOVERSION for WasmEdge C API

hydai:
In this PR,

* Change the name of C API library from `libwasmedge_c.so` to `libwasmedge.so`
* The SONAME is set to `libwasmedge.so.0`.
* The SOVERSION is set to `libwasmedge.so.0.0.0`.
* NOTICE: All the library users should change the linker option from `lwasmedge_c` to `lwasmedge`.
* Also, all language bindings will need to update.

@ibmibmibm @q82419 
Please take a look, all of us should have a consensus before this is getting merged.


hydai:
Seems like we have more option to set on macOS: https://cmake.org/cmake/help/latest/prop_tgt/SOVERSION.html

lib/api/CMakeLists.txt
```diff
@@ -137,26 +139,28 @@ if(WASMEDGE_BUILD_STATIC_LIB)
...
+  add_dependencies(wasmedge_static wasmedge_static_target)
 
-  set_target_properties(wasmedge_c_static
+  set_target_properties(wasmedge_static
     PROPERTIES
-    IMPORTED_LOCATION "${CMAKE_CURRENT_BINARY_DIR}/libwasmedge_c.a"
+    IMPORTED_LOCATION "${CMAKE_CURRENT_BINARY_DIR}/libwasmedge.a"
     INTERFACE_INCLUDE_DIRECTORIES ${PROJECT_BINARY_DIR}/include/api
+    VERSION "${WASMEDGE_CAPI_VERSION}"
+    SOVERSION "${WASMEDGE_CAPI_SOVERSION}"
```
hydai, line 160: Static library doesn't have `SOVERSION`.

hydai (commented):
OpenWRT should be updated with the new soname.

hydai:
Hi @q82419 @ibmibmibm 
This PR is ready to review.

ibmibmibm (approved):

PR #1791, opened 2022-08-24
[Misc] add print support for uint128

### LINKED ISSUE #1704: [CLI tool] `v128` return value is not handled
## Description
When executing the `func1` function in main.wasm, wasmer print `79228162514264337593543950336`, but wasmedge print nothing.
The main.wasm is uploaded as main.txt since it is not supported to upload a file end with .wasm.

[main.txt](https://github.com/WasmEdge/WasmEdge/files/9228709/main.txt)

### Current State
print nothing 

### Expected

print `79228162514264337593543950336`

## Environment

- Hardware Architecture: ( x86\_64)
- Operating system: (Ubuntu 20.04 and macOS 10) the problem occur both in ubuntu and macOS 

## Steps to Reproduce
In wasmer:  `wasmer --enable-all main.wasm -i func1`
In wasmedge : `wasmedge --reactor --enable-all main.wasm func1` or `wasmedgec main.wasm main-edge.wasm; wasmedge --reactor --enable-all main-edge.wasm func1`

LFsWang:
Fix #1704 

Signed-off-by: LFsWang <tnst92002@gmail.com>


lib/driver/runtimeTool.cpp
```diff
@@ -22,6 +22,21 @@
 #include <vector>
 
 namespace WasmEdge {
+std::ostream &operator<<(std::ostream &OS, uint128_t Value) {
+  std::string S;
+  if (Value < (uint64_t)-1) {
```
q82419, line 27: Use the `std::numeric_limits<uint64_t>::max`

q82419: `<=`, not `<`

lib/driver/runtimeTool.cpp
```diff
@@ -22,6 +22,21 @@
...
+    OS << uint64_t(Value);
+    return OS;
+  }
+
+  while (Value != 0) {
+    S.push_back('0' + Value % 10);
```
q82419, line 33: It's proper to separate the `uint128_t` into 2 uint64_t to print, rather than construct them into a string.

q82419 (changes requested):

lib/driver/runtimeTool.cpp
LFsWang: fixed

include/common/int128.h
```diff
@@ -14,17 +14,20 @@
 #pragma once
 
 #if defined(__x86_64__) || defined(__aarch64__)
+#include <ostream>
```
hydai, line 17: Looks like the `<ostream> is contained in both `if` and `else` conditions. Maybe we can move this include out?

lib/common/int128.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+#include <algorithm>
+#include <cstdint>
+#include <iomanip>
+#include <limits>
+#include <ostream>
+#include <sstream>
+
+namespace WasmEdge {
+static inline uint128_t Mod(uint128_t LHS, uint128_t RHS) noexcept;
```
hydai, line 14: The function name should be `lower cases`.

hydai (changes requested):

hydai (approved):

q82419 (approved):

PR #1794, opened 2022-08-25
DOCs-zh-TW:Translate /docs/book/en/src/os.md into zh-TW

Miseratio:
I translated /docs/book/en/src/os.md into zh-TW. Please check.
Signed-off-by: Cheng-En Lee <chengenl@usc.edu>


hydai (approved):
Thanks!

Issue #1680, opened 2022-07-26
[Example] Add a WASI-crypto example


q82419:
## Motivation

In current state, we only have unit tests for wasi-crypto plugin in WasmEdge.
We need a tutorial for creating a WASM program which uses wasi-crypto.

## Details

1. A rust source to use wasi-crypto and compile into the wasm32-wasi target.
2. Use WasmEdge with wasi-crypto plugin to run the WASM.

MediosZ:
Hi @q82419, I am trying to add a example of using `wasi-crypto` in rust.
But I fail to import host function like this:
```rust
#[link(wasm_import_module = "wasi_ephemeral_crypto_signatures")]
extern "C" {
    pub fn signature_import(
        AlgPtr: *const c_char,
        AlgLen: u32,
        EncodedPtr: *const u8,
        EncodedLen: u32,
        Encoding: u32,
        SigHandlePtr: *mut u32,
    );
}
```
I build the wasmedge from source with enabling the wasi-crypto plugin.
How can I confirm if wasi-crypto plugin is loaded?

Azanul:
@hydai Is this completed or are you looking for more examples?

hydai:
Fixed:
https://github.com/WasmEdge/WasmEdge/tree/master/examples/plugin/wasi-crypto-signature

PR #1817, opened 2022-09-03
[Examples]: Add wasi-crypto example.

MediosZ:
Solve #1680 
Signed-off-by: Tricster <mediosrity@gmail.com>


hydai (changes requested):
This PR should have a README to tell how to build, how to execute, and what's the expected output of this example.

MediosZ:
Ok, I will try to add one.

examples/plugin/wasi-crypto/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Run the example
+
+We can run this example with `wasmedge` like
+
+```sh
+wasmedge target/wasm32-wasi/debug/wasi-crypto-example.wasm
+```
+
+This example should run successfully and print out the signatures.
```
hydai, line 49: Is the output the same every time? If so, please post the execution output for users to check if the result is exactly matched.

hydai (changes requested):
Because this example only showed some signature-related functions, I suggest that the name should change to wasi-crypto-signature for better meaning.

MediosZ:
I change the name of this example to `wasi-crypto-signature`. This is ok cause I only have signature related examples.
I believe we can have a different repo for `wasi-crypto` rust bindings like `wasmedge_wasi_socket`, this will be useful for developers who want to use `wasi-crypto` in rust and WasmEdge.

hydai (approved):
LGTM

Issue #1815, opened 2022-09-02
[Tools] Support the compiler optimization level in AOT tool


q82419:
## Motivation

By default, the AOT compiler will compile WASM in `O2` optimization level.
Actually, WasmEdge supports `O0` to `Oz` levels.
The AOT compiler tool should support assigning optimization levels.

## Details

Support assigning optimization level in command line argument, and set it into the compiler configuration.

PR #1819, opened 2022-09-03
[AOT]: Allow assigning optimization levels.

MediosZ:
Solve #1815 
Signed-off-by: Tricster <mediosrity@gmail.com>


lib/driver/compilerTool.cpp
```diff
@@ -100,6 +102,7 @@ int Compiler([[maybe_unused]] int Argc,
            .add_option("enable-extended-const"sv, PropExtendConst)
            .add_option("enable-threads"sv, PropThreads)
            .add_option("enable-all"sv, PropAll)
+           .add_option("optimization-level"sv, OptimizationLevel)
```
q82419, line 105: @hydai ,
Do you think we should use `-O0`, `-O1`, ...etc. like the `gcc` and `clang` compilers?

lib/driver/compilerTool.cpp
```diff
@@ -76,7 +76,9 @@ int Compiler([[maybe_unused]] int Argc,
   PO::Option<PO::Toggle> PropThreads(
       PO::Description("Enable Threads proposal"sv));
   PO::Option<PO::Toggle> PropAll(PO::Description("Enable all features"sv));
-
+  PO::Option<std::string> OptimizationLevel(
+      PO::Description("Optimization level, can be one of O1, O2, O3, Os, Oz"sv),
```
q82419, line 80: Should be `O0` to `Oz`

lib/driver/compilerTool.cpp
```diff
@@ -76,7 +76,9 @@ int Compiler([[maybe_unused]] int Argc,
...
+      PO::MetaVar("OPTIM_LEVEL"sv), PO::DefaultValue(std::string("Oz")));
```
q82419, line 81: Default value should be `O2`

q82419 (changes requested):

lib/driver/compilerTool.cpp
MediosZ: In that case, we will have multiple options in the help text.
As in gcc, options on optimizations are not shown in the help text.

lib/driver/compilerTool.cpp
hydai, line 81: I prefer to use:

```
--optimize=[0,1,2,3,s,z], and the default value is `2`
```
Because this option is not usually used in typical cases, we don't need to set the `MetaVar`. Also, the `MetaVar` is not very simple.

lib/driver/compilerTool.cpp
hydai, line 80: The candidates can be `[0, 1, 2, 3, s, z]`

lib/driver/compilerTool.cpp
```diff
@@ -149,6 +152,27 @@ int Compiler([[maybe_unused]] int Argc,
...
+    Conf.getCompilerConfigure().setOptimizationLevel(
+        WasmEdge::CompilerConfigure::OptimizationLevel::O2);
+  } else if (Optimization == "O3") {
+    Conf.getCompilerConfigure().setOptimizationLevel(
+        WasmEdge::CompilerConfigure::OptimizationLevel::O3);
+  } else if (Optimization == "Os") {
+    Conf.getCompilerConfigure().setOptimizationLevel(
+        WasmEdge::CompilerConfigure::OptimizationLevel::Os);
+  } else {
+    Conf.getCompilerConfigure().setOptimizationLevel(
```
hydai, line 172: The default condition must be `O2`, which is the current default value.

lib/driver/compilerTool.cpp
hydai: We can use the `--optimize=[0,1,2,3,s,z]`. This is not for the common users, and we don't need to use the same option name as `gcc` and `clang` does.

hydai (changes requested):

hydai (approved):
LGTM

hydai (dismissed):
LGTM

lib/driver/compilerTool.cpp
```diff
@@ -76,7 +76,9 @@ int Compiler([[maybe_unused]] int Argc,
   PO::Option<PO::Toggle> PropThreads(
       PO::Description("Enable Threads proposal"sv));
   PO::Option<PO::Toggle> PropAll(PO::Description("Enable all features"sv));
-
+  PO::Option<std::string> OptimizationLevel(
```
q82419, line 79: Please follow the coding style: use `PropOptimizationLevel`

lib/driver/compilerTool.cpp
```diff
@@ -149,6 +152,27 @@ int Compiler([[maybe_unused]] int Argc,
     Conf.addProposal(Proposal::Threads);
   }
 
+  auto Optimization = OptimizationLevel.value();
+  if (Optimization == "0") {
```
q82419, line 156: `if (PropOptimizationLevel.value() == "0")`

q82419 (changes requested):

hydai:
@q82419 PTAL

q82419 (approved):

q82419:
LGTM

PR #1822, opened 2022-09-05
[Fix] Use meta information from component to control module loader

dannypsnl:
### Background
Traditionally, the module is the top-level form of WASM. In the component model, it isn't, however. Our previous implementation will try to read all data via an infinite loop, based on the old assumption.

### Solution
To resolve it, I propose two possible ways.
1. Modify file manager interface, supports sub-range in it.
2. Provides base offset and expected size that can get from component binary.

I prefer the first one, but I don't think it's a good time and place to do that, after some experiments. For example, a `loader` has one `FMgr` is a fact that happens in our code base everywhere now. This means stacking sub-range will bring a huge impact immediately, maybe this refactoring should base on the `master` branch.

The second one is a hack for now, the origin interface `loadModule()` calls `loadModule(0, std::nullopt)`, and component section will call `loadModule(Base, make_optional(*Size)` instead. It works, with minimal effort, however.


hydai (approved):
LGTM. @q82419 PTAL

q82419 (changes requested):
In `include/common/configure.h`:

In the constructor of the `Configure` class, we only add the proposals which are already merged into the WASM spec. Please remove the component-model proposal here.

dannypsnl:
@q82419 I would remove it in `proposal/component-model` branch

q82419 (approved):

PR #1825, opened 2022-09-07
[Rust] Fixed the issue 1816

### LINKED ISSUE #1816: [Rust] `CallingFrame` in wasmedge-sdk
When I use the `CallingFrame` in `wasmedge-sdk`, for example with the method `module_instance()`, the return type is `wasmedge_sys::Instance`, rather than `wasmedge_sdk::Instance`. Such inconsistencies will undoubtedly cause problems and be detrimental to the use of the `wasmedge-sdk`.

@apepkuss Please take a look at this question, thank you very much!

apepkuss:
This PR is to fix #1816. Duo to `cargo nightly clippy` updated to `v0.1.65` in CI, in addition, some new clippy issues are also fixed in this PR.

### DIFF
--- a/bindings/rust/wasmedge-macro/src/lib.rs
+++ b/bindings/rust/wasmedge-macro/src/lib.rs
@@ -2,6 +2,7 @@ use proc_macro::TokenStream;
 use quote::quote;
 use syn::{parse_macro_input, spanned::Spanned, FnArg, Item, Pat, PatType};
 
+/// Expand a synchronous host function defined with `wasmedge-sdk` crate.
 #[proc_macro_attribute]
 pub fn host_function(_attr: TokenStream, item: TokenStream) -> TokenStream {
     let body_ast = parse_macro_input!(item as Item);
--- a/bindings/rust/wasmedge-macro/src/lib.rs
+++ b/bindings/rust/wasmedge-macro/src/lib.rs
@@ -18,6 +18,44 @@ pub fn host_function(_attr: TokenStream, item: TokenStream) -> TokenStream {
 fn expand_host_func(item_fn: &syn::ItemFn) -> syn::Result<proc_macro2::TokenStream> {
     let outer_fn_name_ident = &item_fn.sig.ident;
     let outer_fn_name_literal = outer_fn_name_ident.to_string();
+    let outer_fn_return = &item_fn.sig.output;
+
+    let inner_fn_name_literal = format!("inner_{}", outer_fn_name_literal);
+    let inner_fn_name_ident = syn::Ident::new(&inner_fn_name_literal, item_fn.sig.span());
+    let inner_fn_inputs = &item_fn.sig.inputs;
+    let inner_fn_return = &item_fn.sig.output;
+    let inner_fn_block = &item_fn.block;
+
+    let ret = quote!(
+        fn #outer_fn_name_ident (frame: &wasmedge_sys::CallingFrame, args: Vec<wasmedge_sdk::WasmValue>) #outer_fn_return {
+            fn #inner_fn_name_ident (#inner_fn_inputs) #inner_fn_return {
+                #inner_fn_block
+            }
+
+            let caller = Caller::new(frame);
+            #inner_fn_name_ident(&caller, args)
+        }
+    );
+
+    Ok(ret)
+}
+
+#[proc_macro_attribute]
+pub fn async_host_function(_attr: TokenStream, item: TokenStream) -> TokenStream {
+    let body_ast = parse_macro_input!(item as Item);
+    if let Item::Fn(item_fn) = body_ast {
+        match expand_async_host_func(&item_fn) {
+            Ok(token_stream) => token_stream.into(),
+            Err(err) => err.to_compile_error().into(),
+        }
+    } else {
+        TokenStream::new()
+    }
+}
+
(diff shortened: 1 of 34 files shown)


apepkuss:
@hydai Could you please help review this PR? Please ignore the failed release workflows, as the dependencies are not published onto `crates.io`. Thanks a lot!

hydai:
Hi @apepkuss 
Just want to check if we do need these failed workflows to be executed in PR. It's highly related to the published version. However, there is no way to use it before publishing it. Is there any solution or practice to make this smooth?

apepkuss:
> Hi @apepkuss Just want to check if we do need these failed workflows to be executed in PR. It's highly related to the published version. However, there is no way to use it before publishing it. Is there any solution or practice to make this smooth?

My idea is to remove the `Dry run` jobs from the release workflows, as the `Publish` jobs have a `dry run` step that can check the code before publishing. What do you think about it?

hydai:
IIRC, we failed the publish because there is no dry-run CI before we tagged.
And we can also find that the versions have dependencies. Maybe we should bump the version per PR? The current rust PR will do too many things inside a single PR. If we can bump the version separately, then it can be tested sequentially. Just like how we do in releasing process.

apepkuss:
> IIRC, we failed the publish because there is no dry-run CI before we tagged. And we can also find that the versions have dependencies. Maybe we should bump the version per PR? The current rust PR will do too many things inside a single PR. If we can bump the version separately, then it can be tested sequentially. Just like how we do in releasing process.

Let me explain more about the version bump and the `dry-run` test:

- The reason for bumping versions in this PR is that we introduced some new objects in `sdk` and `sys` crates; meanwhile, due to the upgrade of `cargo::clippy` used in the CI workflows, the new lint named "result-large-err" requires us to refactor the original design of `wasmedge_types::WasmEdgeResult` that forces the APIs in Rust bindings to support returning new error types. And more importantly, the new result type has already broken the APIs. Therefore, we have to bump the versions in this PR so that we can distinguish between the original design and the new one.

- For the `dry-run` test, you're right, we also run into the same failure caused by the incorrect publishing order. In fact, the real reason for the failure is also the publish operation did not find the correct dependencies on `crates.io`, as we did not publish the dependencies before publishing the high-level crate. Back to the failures in this PR, even if we did not bump the versions, that is, we still used the original versions, the `cargo publish --dry-run` step would also fail, as the dependencies the cargo find out on `crates.io` had no the APIs we need. Therefore, the `dry-run` step will always fail if any API or any design is changed in the low-level dependencies.

hydai:
Okay, so this is the expected behavior of the current workflows.
Just another thing to check, once this PR got merged, do I need to publish something to crates.io? And I didn't find any release or build workflows which is related to the `wasmedge-macro` crates.

bindings/rust/wasmedge-macro/src/lib.rs
```diff
@@ -0,0 +1,258 @@
...
+        match sys_async_expand_host_func(&item_fn) {
+            Ok(token_stream) => token_stream.into(),
+            Err(err) => err.to_compile_error().into(),
+        }
+    } else {
+        TokenStream::new()
+    }
+}
+
+fn sys_async_expand_host_func(item_fn: &syn::ItemFn) -> syn::Result<proc_macro2::TokenStream> {
```
hydai, line 165: What's the naming rule?
I found `expand_async_host_func2` and `async_host_func2`, however, these are `sys_async_host_function` and `sys_async_host_function`.

hydai (changes requested):

apepkuss:
> Okay, so this is the expected behavior of the current workflows. Just another thing to check, once this PR got merged, do I need to publish something to crates.io? And I didn't find any release or build workflows which is related to the `wasmedge-macro` crates.

For now, I don't think we need to publish `wasmedge-macro`, as it's still under development. In addition, this PR only uses the `host_function` macro, while others are defined for async cases that could be launched in future PRs. According to the plan, the `wasmedge-macro` crate will be published as part of the next Rust bindings release.

hydai:
> For now, I don't think we need to publish `wasmedge-macro`, as it's still under development. In addition, this PR only uses the `host_function` macro, while others are defined for async cases that could be launched in future PRs. According to the plan, the `wasmedge-macro` crate will be published as part of the next Rust bindings release.

Thanks for clarifying this.

bindings/rust/wasmedge-macro/src/lib.rs
apepkuss: The `async_host_func2` macro is defined for declaring async host function when using `sdk`, while `sys_async_host_function` and `sys_async_host_function2` are two different macro types and used for declaring async host function when using `sys`. These three macros are used in another branch. As for `expand_async_host_func2`, it is just a private function used by the `async_host_func2` macro, and not exposed to the outside. Thanks!

apepkuss:
> > For now, I don't think we need to publish `wasmedge-macro`, as it's still under development. In addition, this PR only uses the `host_function` macro, while others are defined for async cases that could be launched in future PRs. According to the plan, the `wasmedge-macro` crate will be published as part of the next Rust bindings release.
> 
> Thanks for clarifying this.

My pleasure :)

hydai (approved):

apepkuss:
@hydai Thanks for the review!

Issue #1477, opened 2022-05-22
GSoC Workspace: Implement `WASI` and `wasmedge_process` host functions on the Windows platform


abhinandanudupa:
# Motivation

Currently, the `WASI` and `wasmedge_process` components in WasmEdge are only implemented on *nix platforms (mac OS and Linux) and most of the host functions are not yet supported on the Windows platform. Since there are a lot of windows users would like to use `WasmEdge`, it will be important to implement these host functions and satisfy their requirements.

# Milestones

### Community bonding period [20th May, 2022 to June 12th, 2022]
* [x] Setup the working environment (Windows SDK v`19041`).
* [x] Understand the existing code base and the requirements
* [x] List out the milestones to be achieved - you know once you start

### Coding Period [13th June, 2022 to 12th September, 2022]
* [x] ~Implement `DirHolder` using the `Win32` API~ -  not necessary as `HandleHolder` should be sufficient
* [x] Implement a function to map file constants/flags from `win32 API` to `WASI`/`*nix` 
    - `attributeFlags()` - maps file attributes to `WASI`, flags to manage file attributes/types
    - `accessFlags()` - maps access flags to `WASI`, flags to define permissions on the file for the user/process
    - `shareFlags()` - maps share flags to `WASI`, flags to allow other processes to read or to write to the open file
    - `creationDisposition()` - maps creation disposition flags to `WASI`, flags to truncate or to create a new file
* [x] Implement a function to map error codes from `win32 API` to `WASI` [`fromWinError()`]
* [x] Implement a function to convert a signed 64 bit integer to a `LARGE_INTEGER` [`toLargeIntegerFromUnsigned()`]
* [x] Implement a function to convert a unsigned 64 bit integer to a `LARGE_INTEGER` [`toLargeIntegerFromSigned()`]
* [x] Implement a function to map file types to `WASI` [`fromFileType()`]
* [x] Implement a function to define the relatively position of the file pointer during a read or write operation [`fromWhence()`]
* [x] Implement host functions as defined and described in [`include/host/wasi/indoe.h`](https://github.com/WasmEdge/WasmEdge/blob/master/include/host/wasi/inode.h#L182)
* [x] Implement the host functions!

# How I think I would implement the functions

I think it will be a good thing to learn the Linux implementations, find equivalent functions and then implement with `win32 API`. I would not have to go build from scratch (hopefully) and reduce the time taken.

### A. WASI component
[...]

PR #1826, opened 2022-09-07
[Windows] Implement host functions on the Windows platform

### LINKED ISSUE #1227: feat: Implement WASI and wasmedge_process host functions on Windows platform
## Motivation

Currently, the WASI and wasmedge_process components in WasmEdge are only implemented on macOS and Linux platforms. Most of the host functions are not supported on the Windows platform.
Since we have lots of windows users, it is necessary to finish the implementations.

## Details

To finish the remaining host functions, you may need to do the following working items:

A. WASI component 
1. Check the `*win.*` files in [lib/host/wasi](https://github.com/WasmEdge/WasmEdge/tree/master/lib/host/wasi). 
2. Implement those unimplemented functions. E.g. those functions contains `return WasiUnexpect(__WASI_ERRNO_NOSYS);` or `return false;`
3. Make sure all of the WASI unit tests are passed.

B. WASMEDGE_PROCESS component
1. Check the files in [lib/host/wasmedge_process](https://github.com/WasmEdge/WasmEdge/blob/master/lib/host/wasmedge_process).
2. You will find that we [use libc to create and execute a new application process](https://github.com/WasmEdge/WasmEdge/blob/master/lib/host/wasmedge_process/processfunc.cpp#L175-L183). You will need to split the files into the corresponding OS headers like WASI, and you will need to use windows API to finish this feature.

## Recommended Skills
1. Familiar with C++ programming language.
2. Familiar with Windows API.

abhinandanudupa:
### Host Functions: Function-wise Summary

| Function | Status | Comment |
| --------------------- | --------------------- | ------------------------------------------------------------------------- |
| `open` | complete | some flags may not have an equivalent |
| `fdAdvise` | no equivalent | have to find an solution |
| `fdAllocate` | complete | None |
| `fdDatasync` | complete | documentation is not clear on whether metadata is also flushed, refer [here](https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-flushfilebuffers#remarks) |
| `fdFdstatGet` | complete | depends on a partially complete function - `fromFileType` (this function has been implemented partially in linux), find appropriate functions to query the equivalent flags and fill the other fields (the implementation for linux has not filled these fields) |
| `fdFdstatSetFlags` | complete | depends on a partially complete function - `fromFileType` and an equivalent for device ID needs to be found which may be related to the file index |
| `fdFilestatSetSize` | complete | None |
| `fdFilestatSetTimes` | complete | None |
| `fdPread` | complete | there maybe issues due to casting |
| `fdPwrite` | complete | there maybe issues due to casting |
| `fdRead` | complete | had already been implemented |
| `fdWrite` | complete | had already been implemented |
| `fdReaddir` | complete | Need to optimise the function and it depends on a partially implemented function - `fromFileType` |
| `fdSeek` | complete | None |
| `fdSync` | complete | works when the file has been opened with the flags `FILE_FLAG_NO_BUFFERING` and `FILE_FLAG_WRITE_THROUGH` which I suspect is the desired behaviour, refer [here](https://devblogs.microsoft.com/oldnewthing/20210729-00/?p=105494) |
| `fdTell` | complete | None |
| `getNativeHandler` | complete | had already been implemented |
| `pathCreateDirectory` | complete | None |
| `pathFilestatGet` | complete | similar to `stat` which uses absolute paths |
| `pathFilestatSetTimes` | complete | None |
| `pathLink` | complete | None |
| `pathOpen` | complete | None |
| `pathReadlink` | complete | None |
| `pathRemoveDirectory` | complete | had been already implemented |
| `pathRename` | complete | None |
| `pathSymlink` | complete | None |
| `pathUnlinkFile` | complete | None |
| `pollOneoff` | incomplete | could not find a similar concept on windows |
| `sockGetPeerAddr` | incomplete | behaviour is unspecified |
| `unsafeFiletype`  | partially complete | need to find equivalent flags for three file types |
| `filetype` | partially complete | need to find equivalent flags for three file types |
| `isDirectory` | complete | None |
| `isSymlink` | complete | None |
| `filesize` | complete | None |
| `canBrowse` | incomplete | need to find appropriate functions |
| `Poller::clock` | incomplete | could not find a similar concept on windows |
| `Poller::read` | incomplete | could not find a similar concept on windows |
| `Poller::write` | incomplete | could not find a similar concept on windows |
| `Poller::wait` | incomplete | could not find a similar concept on windows |

### Descriptions
- `open()`: to use the flag `__WASI_FDFLAGS_NONBLOCK` we have to used an `OVERLAPPED` structure to query the status of the read/write operation. But in the API we do not describe an a function which queries the status  of the operation. Hence files opened with this flags do not have a way to query that status of the operation.
- `fdAdvise()`: we do not have a direct equivalent for this function in the Win32 API but we can use certain file flags [`FILE_FLAG_RANDOM_ACCESS`, `FILE_FLAG_SEQUENTIAL_SCAN`, `OPEN_EXISTING`]with the `CreateFile()` and functions like [`PrefetchVirtualMemory()`](https://docs.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-prefetchvirtualmemory?redirectedfrom=MSDN) and use certain flags in the read/write functions - `ReadFile` and `WriteFile`.
- `fdAllocate()`: signed bit has to be removed as the parameter `Offset` and `Len` are unsigned.
- `fdDatasync()`: it seems that meta data will be [flushed only when the `Handle` is opened with certain attributes](https://devblogs.microsoft.com/oldnewthing/20210729-00/?p=105494).
- `fdFdstatGet()`: need to find appropriate flags for the fields - `fs_rights_base` and `fs_rights_inheriting` and file types for `fs_filetype`. The linux implementation does not fill the `*rights*` fields.
- `fdFdstatSetFlags()`: the `__WASI_FDFLAGS_APPEND` flag can be safely ignored.
- `fdFilestatGet`: need to find equivalent file types, three are missing. And there is no equivalent for device ID and INode number.
- `fdPread`: does not support asynchronous operation as the the function is similar to `preadv` which is subject to the `O_NONBLOCK` flag as according to the [source](https://man7.org/linux/man-pages/man2/open.2.html)
```
              Note that this flag has no effect for regular files and
              block devices; that is, I/O operations will (briefly)
              block when device activity is required, regardless of
              whether O_NONBLOCK is set.  Since O_NONBLOCK semantics
              might eventually be implemented, applications should not
              depend upon blocking behavior when specifying this flag
              for regular files and block devices.
```
- `fdReaddir()`: might require some optimisation.
- `fdSync()`: same as that for `fdDatasync`.
- `pathFilestatGet()`: there is no equivalent for device number and the INode number.
- `filetype()`: need to find equivalent flags for the file types.
- `unsafeFiletype()`: same as above - need to find equivalent flags for the file types.

> I have made `TODO` and `FIXME`comments where ever functions have been partially implemented or where issues maybe present.

Resolves #1227 and #1477


lib/host/wasi/inode-win.cpp
```diff
@@ -9,9 +9,19 @@
...
 #include "win.h"
+#include <algorithm>
+#include <boost/align/aligned_allocator.hpp>
+#include <new>
+#include <vector>
 #include <winsock2.h>
 #include <ws2tcpip.h>
 
+#define NANOSECONDS_PER_TICK 100ULL
+// #define NANOSECONDS_PER_SECOND 1000000000ULL 0000000
```
hydai, line 20: Do we need this?

lib/host/wasi/inode-win.cpp
```diff
@@ -57,6 +69,445 @@ createNullTerminatedString(std::string_view View) noexcept {
...
 }
+
+inline LARGE_INTEGER toLargeIntegerFromUnsigned(unsigned long long value) {
+  LARGE_INTEGER Result;
+
+  // Does the compiler natively support 64-bit integers?
+#ifdef INT64_MAX
+  Result.QuadPart = static_cast<int64_t>(value);
+#else
+  Result.high_part = value & 0xFFFFFFFF00000000;
```
hydai, line 80: Missing shift after masking.

lib/host/wasi/inode-win.cpp
```diff
@@ -57,6 +69,445 @@ createNullTerminatedString(std::string_view View) noexcept {
...
+  return Result;
+}
+
+inline LARGE_INTEGER toLargeIntegerFromSigned(long long value) {
+  LARGE_INTEGER Result;
+
+#ifdef INT64_MAX
+  Result.QuadPart = static_cast<int>(value);
+#else
+  Result.high_part = value & 0xFFFFFFFF00000000;
```
hydai, line 92: Missing shift after masking.

lib/host/wasi/inode-win.cpp
```diff
@@ -57,6 +69,445 @@ createNullTerminatedString(std::string_view View) noexcept {
...
+#ifdef INT64_MAX
+  Result.QuadPart = static_cast<int64_t>(value);
+#else
+  Result.high_part = value & 0xFFFFFFFF00000000;
+  Result.low_part = value & 0xFFFFFFFF;
+#endif
+  return Result;
+}
+
+inline LARGE_INTEGER toLargeIntegerFromSigned(long long value) {
```
hydai, line 86: `value` -> `Value`.

lib/host/wasi/inode-win.cpp
```diff
@@ -57,6 +69,445 @@ createNullTerminatedString(std::string_view View) noexcept {
...
+#ifdef INT64_MAX
+  Result.QuadPart = static_cast<int>(value);
+#else
+  Result.high_part = value & 0xFFFFFFFF00000000;
+  Result.low_part = value & 0xFFFFFFFF;
+#endif
+  return Result;
+}
+
+inline constexpr __wasi_errno_t fromWinError(DWORD winerr) {
```
hydai, line 98: `winerr` -> `Winerr`

lib/host/wasi/inode-win.cpp
```diff
@@ -57,6 +69,445 @@ createNullTerminatedString(std::string_view View) noexcept {
...
+  __wasi_errno_t error = __WASI_ERRNO_NOSYS;
```
hydai, line 99: `error` -> `Error`.

lib/host/wasi/inode-win.cpp
```diff
@@ -57,6 +69,445 @@ createNullTerminatedString(std::string_view View) noexcept {
...
+  switch (winerr) {
+  case ERROR_ACCESS_DENIED:
+    error = __WASI_ERRNO_ACCES;
+    break;
+  case ERROR_ACCOUNT_DISABLED:
+    error = __WASI_ERRNO_ACCES;
+    break;
+  case ERROR_ACCOUNT_RESTRICTION:
+    error = __WASI_ERRNO_ACCES;
```
hydai, line 108: If there are several cases return the same error number, do the following stuff:

```

  case ERROR_ACCESS_DENIED:
  case ERROR_ACCOUNT_DISABLED:
  case ERROR_ACCOUNT_RESTRICTION:
    error = __WASI_ERRNO_ACCES;
```

lib/host/wasi/inode-win.cpp
```diff
@@ -82,53 +533,308 @@ INode INode::stdErr() noexcept {
...
+
+  if (static_cast<int64_t>((Offset + Len) & 0x7FFFFFFFFFFFFFFF) >
+      FileSize.QuadPart) {
+
+    FILE_STANDARD_INFO StandardInfo;
+    FILE_ALLOCATION_INFO AllocationInfo;
+
+    if (unlikely(GetFileInformationByHandleEx(Handle, FileStandardInfo,
+                                              &StandardInfo,
+                                              sizeof(StandardInfo))) == FALSE)
```
hydai, line 582: Make sure there always be a `{}` for the `if` statement.

lib/host/wasi/inode-win.cpp
```diff
@@ -82,53 +533,308 @@ INode INode::stdErr() noexcept {
...
+      return WasiUnexpect(fromWinError(GetLastError()));
+
+    // Update the size attribute
+    // TODO: Is it ok to do this?
```
hydai, line 586: Sorry, what's the question?

lib/host/wasi/inode-win.cpp
```diff
@@ -82,53 +533,308 @@ INode INode::stdErr() noexcept {
...
+  // this
+  // https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-openfilebyid?redirectedfrom=MSDN
+  // and this
+  // https://docs.microsoft.com/en-us/windows/win32/api/winbase/ns-winbase-file_id_info
+  FileStat.ino = 0;
+  // TODO: Find an equivalent for device ID in windows
+  FileStat.dev = 0;
+  FileStat.nlink = (*FileInfo).nNumberOfLinks;
+  FileStat.size = static_cast<uint64_t>((*FileInfo).nFileSizeLow) +
+                  (static_cast<uint64_t>((*FileInfo).nFileSizeHigh) << 32);
```
hydai, line 693: What's the type of `nFileSizeHigh`? Should we cast it first, then do the shift?

lib/host/wasi/inode-win.cpp
```diff
@@ -145,20 +851,196 @@ WasiExpect<void> INode::fdRead(Span<Span<uint8_t>> IOVs,
...
+
+  switch (CombineResult) {
+  case S_OK:
+    break;
+  case E_INVALIDARG:
+    return WasiUnexpect(__WASI_ERRNO_INVAL);
+  case E_OUTOFMEMORY:
+    return WasiUnexpect(__WASI_ERRNO_OVERFLOW);
+  default:
+    return WasiUnexpect(__WASI_ERRNO_NAMETOOLONG);
```
hydai, line 891: Just want to know why the default case of CombineResult is `Name too long`?

lib/host/wasi/inode-win.cpp
```diff
@@ -145,20 +851,196 @@ WasiExpect<void> INode::fdRead(Span<Span<uint8_t>> IOVs,
...
+      return WasiUnexpect(fromWinError(GetLastError()));
+    }
+    Seek++;
+  }
+
+  uint32_t NumberOfBytesRead = 0;
+  BOOL FindNextResult = FALSE;
+
+  do {
+
```
hydai, line 912: Redundant empty line.

lib/host/wasi/inode-win.cpp
```diff
@@ -180,51 +1062,628 @@ WasiExpect<uint64_t> INode::getNativeHandler() const noexcept {
...
+  }
+
+  // Fill the Buffer with the contents of the link
+  HANDLE LocalFileHandle;
+
+  LocalFileHandle = CreateFileW(FullPathW, GENERIC_READ, FILE_SHARE_READ,
+                                nullptr, OPEN_EXISTING, 0, nullptr);
+
+  if (likely(LocalFileHandle != INVALID_HANDLE_VALUE)) {
+    // MAX_PATH or Buffer.size() in the function call?
```
hydai, line 1450: According to the document:
```
If the function succeeds, the return value is the length of the string received by lpszFilePath, in TCHARs. This value does not include the size of the terminating null character.
```

So the size should be the same as the buffer size.

lib/host/wasi/inode-win.cpp
```diff
@@ -687,18 +2146,102 @@ WasiExpect<void> INode::sockGetPeerAddr(uint8_t *, uint32_t *,
 }
 
 __wasi_filetype_t INode::unsafeFiletype() const noexcept {
+
+  // TODO: Find equivalents to the other file types
+  // To be completed along with other similar functions
+
+  if (unlikely(GetFileInformationByHandle(Handle, &(*FileInfo))) == FALSE) {
```
hydai, line 2153: Looks like this function can use `fromFileType` instead.

lib/host/wasi/inode-win.cpp
```diff
@@ -687,18 +2146,102 @@ WasiExpect<void> INode::sockGetPeerAddr(uint8_t *, uint32_t *,
...
+  switch (GetFileType(Handle)) {
+  case FILE_TYPE_CHAR:
+    return __WASI_FILETYPE_CHARACTER_DEVICE;
+  }
+  return __WASI_FILETYPE_UNKNOWN;
+
   return __WASI_FILETYPE_UNKNOWN;
 }
 
 WasiExpect<__wasi_filetype_t> INode::filetype() const noexcept {
```
hydai, line 2175: This function can use `unsafeFileType` to get the value.

lib/host/wasi/inode-win.cpp
```diff
@@ -687,18 +2146,102 @@ WasiExpect<void> INode::sockGetPeerAddr(uint8_t *, uint32_t *,
...
+  switch (GetFileType(Handle)) {
+  case FILE_TYPE_CHAR:
+    return __WASI_FILETYPE_CHARACTER_DEVICE;
+  }
+  return __WASI_FILETYPE_UNKNOWN;
 }
 
-bool INode::isDirectory() const noexcept { return false; }
+bool INode::isDirectory() const noexcept {
+  if (FileInfo.has_value()) {
```
hydai, line 2201: Looks like this section can be a function.
`isSymlink` has the same pattern.
```
if (!FileInfo.has_value()) {}
plus
  if (Handle != INVALID_HANDLE_VALUE &&
      unlikely(GetFileInformationByHandle(Handle, &(*FileInfo))) == FALSE) {
    return false;
  }
```

This pattern is what we do in the `inode-linux.cpp: updateStat()`

lib/host/wasi/inode-win.cpp
```diff
@@ -687,18 +2146,102 @@ WasiExpect<void> INode::sockGetPeerAddr(uint8_t *, uint32_t *,
...
+      unlikely(GetFileInformationByHandle(Handle, &(*FileInfo))) == FALSE) {
+    return false;
+  }
+
+  return (*FileInfo).dwFileAttributes == FILE_ATTRIBUTE_REPARSE_POINT;
+}
+
+WasiExpect<__wasi_filesize_t> INode::filesize() const noexcept {
+  if (FileInfo.has_value()) {
+    return (*FileInfo).dwFileAttributes == FILE_ATTRIBUTE_REPARSE_POINT;
```
hydai, line 2232: This function is `filesize`. Why do we return a condition check here?

hydai (changes requested):

lib/host/wasi/inode-win.cpp
abhinandanudupa: Will resolve this

lib/host/wasi/inode-win.cpp
abhinandanudupa: Must have copied the code from the function that checks for a symlink.

lib/host/wasi/inode-win.cpp
abhinandanudupa:
```
// Update the size attribute
// TODO: Is it ok to do this?
AllocationInfo.AllocationSize.QuadPart =
    static_cast<int64_t>((Offset + Len) & 0x7FFFFFFFFFFFFFFF);
```
We are converting a signed value into a unsigned value and here I am removing the sign bit. So the range of `QuadPart` is now the range of `uint64_t` (`Offset + Len`) / 2 which is lesser. So is this reduced range acceptable?

lib/host/wasi/inode-win.cpp
abhinandanudupa: `nFileSizeHigh` is a `DWORD` a 32 bit signed integer. Yes, we have to cast it first. By casting it first we are converting it into a 64 bit integer which occupies the less significant half of the integer. So we have to shift it to the left by 32 bits.

lib/host/wasi/inode-win.cpp
abhinandanudupa: No there are four possible return values. `PATHCCH_E_FILENAME_TOO_LONG` is the case the default is representing. But this constant is not defined in `Pathcch.h` and I could not find where it was defined. So I made it the default case.
Reference: https://docs.microsoft.com/en-us/windows/win32/api/pathcch/nf-pathcch-pathcchcombine#return-value

lib/host/wasi/inode-win.cpp
abhinandanudupa: If you mean the commented part then no.

lib/host/wasi/inode-win.cpp
abhinandanudupa: But how to generalize the `if (!FileInfo.has_value()) {...}` part? The other functions have different return values.

lib/host/wasi/inode-win.cpp
abhinandanudupa: I have deleted the comment.

lib/host/wasi/inode-win.cpp
abhinandanudupa: I have defined a new function to update the `FileInfo` for this purpose.

abhinandanudupa:
Should I try to pass all the checks by pushing a new commit again? The workflows have failed not because of the PR related changes but some internal issues.

hydai:
> Should I try to pass all the checks by pushing a new commit again? The workflows have failed not because of the PR related changes but some internal issues.

The fedora platform failed because the LLVM upstream just released the 15.0.0 version. We will have another PR to fix it.

abhinandanudupa:
> The fedora platform failed because the LLVM upstream just released the 15.0.0 version. We will have another PR to fix it.

Ok : )

lib/host/wasi/inode-win.cpp
```diff
@@ -57,6 +68,347 @@ createNullTerminatedString(std::string_view View) noexcept {
...
+  }
+
+  if ((FdFlags & __WASI_FDFLAGS_APPEND) != 0) {
+    Flags |= FILE_APPEND_DATA;
+  }
+
+  return Flags;
+}
+/*
+  constexpr DWORD shareFlags(__wasi_oflags_t OpenFlags) {
```
hydai, line 361: Do we need to keep this?

lib/host/wasi/inode-win.cpp
```diff
@@ -82,53 +434,310 @@ INode INode::stdErr() noexcept {
...
-WasiExpect<INode> INode::open(std::string, __wasi_oflags_t, __wasi_fdflags_t,
-                              uint8_t) noexcept {
-  return WasiUnexpect(__WASI_ERRNO_NOSYS);
+WasiExpect<INode> INode::open(std::string Path, __wasi_oflags_t OpenFlags,
+                              __wasi_fdflags_t FdFlags,
+                              uint8_t VFSFlags) noexcept {
+
+  DWORD AttributeFlags = attributeFlags(OpenFlags, FdFlags);
+  DWORD AccessFlags = accessFlags(FdFlags, VFSFlags);
+  // DWORD ShareFlags = shareFlags(OpenFlags);
```
hydai, line 443: Do we still need this?

lib/host/wasi/inode-win.cpp
hydai: I think you have to explain this in the comment.

lib/host/wasi/inode-win.cpp
```diff
@@ -82,53 +434,310 @@ INode INode::stdErr() noexcept {
...
+                          __wasi_fstflags_t FstFlags) const noexcept {
+
+  // Let FileTime be initialized to zero if the times need not be changed
+  FILETIME AFileTime = {0, 0};
+  FILETIME MFileTime = {0, 0};
+
+  // For setting access time
+  if (FstFlags & __WASI_FSTFLAGS_ATIM) {
+    uint64_t Aticks = ATim / NANOSECONDS_PER_TICK + TICKS_TO_UNIX_EPOCH;
+    AFileTime.dwLowDateTime = static_cast<DWORD>(Aticks % 0x100000000ULL);
```
hydai, line 652: `0x100000000ULL` is in too many place, use a const instead.

lib/host/wasi/inode-win.cpp
```diff
@@ -82,53 +434,310 @@ INode INode::stdErr() noexcept {
...
+  uint64_t LocalOffset = Offset;
+
+  for (auto IOV : IOVs) {
+    DWORD NumberOfBytesRead = 0;
+    OVERLAPPED Result;
+
+    Result.Offset = static_cast<uint32_t>(LocalOffset);
+    Result.OffsetHigh = static_cast<uint32_t>(LocalOffset >> 32);
+
+    // There maybe issues due to casting IOV.size() to unit32_t
```
hydai, line 687: What's the issue?

lib/host/wasi/inode-win.cpp
```diff
@@ -82,53 +434,310 @@ INode INode::stdErr() noexcept {
...
+  uint64_t LocalOffset = Offset;
+
+  for (auto IOV : IOVs) {
+    DWORD NumberOfBytesWritten = 0;
+    OVERLAPPED Result;
+
+    Result.Offset = static_cast<uint32_t>(LocalOffset);
+    Result.OffsetHigh = static_cast<uint32_t>(LocalOffset >> 32);
+
+    // There maybe issues due to casting IOV.size() to unit32_t
```
hydai, line 721: What's the issue?

lib/host/wasi/inode-win.cpp
```diff
@@ -180,51 +964,629 @@ WasiExpect<uint64_t> INode::getNativeHandler() const noexcept {
...
+      return WasiUnexpect(__WASI_ERRNO_NOTDIR);
+    }
+
+    wchar_t PathW[MAX_PATH];
+
+    // Convert the path from char_t to wchar_t
+    mbstowcs(PathW, Path.c_str(), MAX_PATH);
+
+    // Append the paths together
+
```
hydai, line 1271: Redundant empty line.

lib/host/wasi/inode-win.cpp
```diff
@@ -180,51 +964,629 @@ WasiExpect<uint64_t> INode::getNativeHandler() const noexcept {
...
+    wchar_t HandleFullPathW[MAX_PATH];
+
+    // First get the paths of the handles
+    if (unlikely(GetFinalPathNameByHandleW(Handle, HandleFullPathW, MAX_PATH,
+                                           FILE_NAME_NORMALIZED)) == FALSE) {
+      return WasiUnexpect(fromWinError(GetLastError()));
+    }
+
+    if (!PathIsDirectoryW(HandleFullPathW)) {
+
```
hydai, line 1261: Redundant empty line.

lib/host/wasi/inode-win.cpp
```diff
@@ -180,51 +964,629 @@ WasiExpect<uint64_t> INode::getNativeHandler() const noexcept {
...
+      return WasiUnexpect(__WASI_ERRNO_NAMETOOLONG);
+    }
+
+  } else {
+    mbstowcs(FullPathW, Path.c_str(), MAX_PATH);
+  }
+
+  DWORD AttributeFlags = attributeFlags(OpenFlags, FdFlags);
+  DWORD AccessFlags = accessFlags(FdFlags, VFSFlags);
+  // DWORD ShareFlags = shareFlags(OpenFlags);
```
hydai, line 1292: Do we need this?

hydai (changes requested):

lib/host/wasi/inode-win.cpp
abhinandanudupa: It is fine if multiple processes can access ,read and write potentially at the same time? Currently all functions allow external processes to read/write a file opened by WasmEdge. If yes then we should remove it else we may have to modify it is some way.

lib/host/wasi/inode-win.cpp
abhinandanudupa: Same as the above comment above.

lib/host/wasi/inode-win.cpp
abhinandanudupa: Will be using one of the time constants.

lib/host/wasi/inode-win.cpp
abhinandanudupa: `IOV.size()` returns a 64 bit integer but the underling windows API uses a 32 bit integer. So can there be any issues due to the narrowing of the range?

lib/host/wasi/inode-win.cpp
abhinandanudupa: Same as above.
> `IOV.size()` returns a 64 bit integer but the underling windows API uses a 32 bit integer. So can there be any issues due to the narrowing of the range?

lib/host/wasi/inode-win.cpp
abhinandanudupa: Related to this issue:
> Is it fine if multiple processes can access ,read and write potentially at the same time? Currently all functions allow external processes to read/write a file opened by WasmEdge. If yes then we should remove it else we may have to modify it is some way.

lib/host/wasi/inode-win.cpp
hydai: Well, just add a check here, if the range is larger than the limit of windows API, report an error.

abhinandanudupa: Okay

lib/host/wasi/inode-win.cpp
hydai: Yes, IIRC, we don't limit this. So it should be fine if there is any external process to access the file.

abhinandanudupa: Okay

lib/host/wasi/inode-win.cpp
abhinandanudupa: Instead should I divide the larger number into two parts - `INT_MAX` and `IOV.size() - INT_MAX` and then perform two operations with these?

lib/host/wasi/inode-win.cpp
abhinandanudupa: Resolved in [here](https://github.com/WasmEdge/WasmEdge/pull/1826#discussion_r981949078)

lib/host/wasi/inode-win.cpp
abhinandanudupa: Resolved in [here](https://github.com/WasmEdge/WasmEdge/pull/1826#discussion_r981951772)

lib/host/wasi/inode-win.cpp
abhinandanudupa: Done.

hydai (approved):
LGTM.

Issue #1290, opened 2022-03-01
bug: Windows 10 action failed


hydai:
## Description

The building CI is failed from last week. Not sure which part of the environment is changed.

```bash
ninja: error: 'C:/Program Files (x86)/Microsoft Visual Studio/2019/Enterprise/DIA SDK/lib/amd64/diaguids.lib', needed by 'test/aot/wasmedgeAOTCoreTests.exe', missing and no known rule to make it
```

### Expected

The building process should be fine.

LFsWang:
I am not sure why there is a hardcode path. replace it with local version will fix the build.

https://github.com/WasmEdge/llvm-windows release
```LLVM-13.0.0-win64\lib\cmake\llvm\LLVMExports.cmake``` L385

It is fixed for me.
 ```C:/Program Files/Microsoft Visual Studio/2022/Community/DIA SDK/lib/amd64/diaguids.lib```

hydai:
Fixed in #1330

abhinandanudupa:
I think this needs to be reopened as I am getting the same error on the latest build (and the solution by @LFsWang works).

PR #1832, opened 2022-09-09
[Docs] Rearrange and rephrase buld instructions

abhinandanudupa:
Just made a few rearrangements and added a link.

### DIFF
--- a/docs/book/en/src/extend/build.md
+++ b/docs/book/en/src/extend/build.md
@@ -4,7 +4,7 @@ Please follow this guide to build and test WasmEdge from the source code.
 
 The following guide is for Linux distributions. For MacOS, please refer to [Build for macOS](build_on_mac.md). For Windows, please refer to [Build for Windows](build_on_windows.md). For Android, please refer to [Build for Android](build_for_android.md).
 
-> If you just want the latest builds from the `HEAD` of the `master` branch, and do not want to build it yourself, you can download the release package directly from our Github Action's CI artifact. [Here is an example](https://github.com/WasmEdge/WasmEdge/actions/runs/1521549504#artifacts).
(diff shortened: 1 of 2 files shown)


abhinandanudupa:
```powershell
# Download our pre-built LLVM 13 binary
$llvm = "LLVM-13.0.1-win64.zip"
curl -sLO https://github.com/WasmEdge/llvm-windows/releases/download/llvmorg-13.0.1/LLVM-13.0.1-win64.zip -o $llvm
Expand-Archive -Path $llvm
```
https://wasmedge.org/book/en/extend/build_on_windows.html#prepare-the-environment
@hydai We should be removing the part where we are downloading `clang`. Downloading the pre-built `LLVM` is not ideal as users will have to fix an issue related to #1290. Let the users download LLVM and clang with Visual Studio itself. Should I remove this part and add instructions to install from Visual Studio?

hydai:
Sure, the previous steps are using in our CI. 
It's fine to use the VS one.

abhinandanudupa:
Ok, will do it then.

abhinandanudupa:
I raised this PR with the intention of updating it to use `LLVM 14` assuming that the `LLVM` from Visual Studio is being used in the CI and locally. But it turns out that that is not the case as not specifying the directory to the downloaded `LLVM 13` will result in the build system asking for a path to `LLVM` - as tested locally.  Hence it is not useful to continue with the PR - until we use `LLVM 14`.

hydai:
Hi @abhinandanudupa 

It's totally fine to use any version of LLVM including 12, 13, 14, and 15. You can use the latest version from the VS's one.

abhinandanudupa:
Ok, then I will find a way to use the built-in version from Visual Studio.

PR #1843, opened 2022-09-16
[Proposal] component model: Instantiate & Execution

dannypsnl:


dannypsnl:
See #1877 for the reason we stop this for a while.

PR #1847, opened 2022-09-19
[Fix] component section should follow the same trick as module

dannypsnl:
The old implementation is because I misunderstood the description, component and module will all be referred by de Bruijn indices.

The test didn't find the error because it treats same level component as nested component will be for loading, only gets wrong semantic. Hence, I also test how many components we got.

NOTE: de Bruijn indices
λx.λy.(y x) can be λ.λ.(0 1)
by translating variable to binder carefully


dm4 (approved):
LGTM

PR #1850, opened 2022-09-19
[Proposal] component model: validator

dm4:


dannypsnl:
As discussed, the **wasm-tools** already validates the code, we don't have to do it again.

PR #1858, opened 2022-09-21
[Rust] Add some lost functions for integration with rust containerd-shim-wasm

CaptainVincent:
Usage (use case) could reference to this [repo](https://github.com/second-state/runwasi) and [demo](https://drive.google.com/file/d/15_EsPANwQVNr9GapQ77ZqD21yOUu8V01/view)


CaptainVincent:
@hydai Could you please help review this PR or assign Sam help this? Thanks a lot!

bindings/rust/wasmedge-sdk/src/vm.rs
```diff
@@ -248,6 +254,28 @@ impl Vm {
         Ok(returns)
     }
 
+    pub fn run_func_async(
```
apepkuss, line 257: Notice that the `wasmedge_sys::run_registered_function_async` method is an experimental API as it cannot return a Rust standard `future` as result. Therefore, this wrapper function can only serve as a temporary API. Once the async design is done, this wrapper and the existing low-level async APIs defined in `wasmedge_sys::Vm` will be replaced.

CaptainVincent:
Fix CI tests done.

bindings/rust/wasmedge-sdk/src/vm.rs
```diff
@@ -468,6 +496,15 @@ impl Engine for Vm {
...
 }
+impl Clone for Vm {
+    fn clone(&self) -> Vm {
+        let config = self
+            .config_snapshot
+            .as_ref()
+            .map(|config| Config::copy_from(config).unwrap());
+        Vm::new(config).unwrap()
+    }
+}
```
apepkuss, line 507: The `config_snapshot` field would introduce additional memory consumption if `clone` is never or rarely used in upstream apps; According to the code in `clone` function, in addition, it intends to create a new vm instance with a "copy" of `config`, so it does not exactly match the semantics of "Clone" trait. Based on these two reasons, it is recommended to use the following sample code to realize the idea of "cloning" a vm instead of implementing `Clone` trait for `Vm`:
```rust
// Assume that a config instance is created and named "config" already, then
let config_copy = Config::copy_from(&config)?;
// create a new vm instance with config_copy
let new_vm = Vm::new(Some(config_copy))?;
...
```

CaptainVincent: I am afraid this is not possible. This behavior (clone and pass engine) is needed by  wasm-shim. I suggest you could help review whole runwasi project to design a reasonable clone function fit this use case. I think the code of shim is closed for modification at this moment.

bindings/rust/wasmedge-sdk/src/vm.rs
CaptainVincent: Hmm ~ I think experimental API is ok to be merged. If your design ready later you can notify me, I will happily replace it in runwasi future.

apepkuss: Sure! I'll let you know if the async design is ready for publishing.

bindings/rust/wasmedge-sdk/src/vm.rs
apepkuss: Could you please specify the relevant code snippets in `runwasi` project? I'll reference it to consider how to design an API for "clone". Thanks!

CaptainVincent: https://github.com/second-state/runwasi/blob/408e4a009efde080b6ee4fef234632d4d6740448/crates/containerd-shim-wasm/src/sandbox/shim.rs#L923
I think there exist more use cases in other places. You could just mark clone function as comment and recompile it will see more detail error message.

run `make build` or `cargo test -- --nocapture` on top layer

apepkuss: Ok, I see now. The `Local` struct requires the `E` must implement the `Clone` trait. Let's keep the clone implementation. Thanks!

apepkuss (approved):

apepkuss:
@CaptainVincent Could you please rebase this PR onto the latest master? as the `wasmedge-sdk` crate introduced some signification changes. Thanks!

PR #1859, opened 2022-09-22
[AOT] Support LLVM 15

hydai:
Signed-off-by: hydai <z54981220@gmail.com>

### DIFF
--- a/lib/aot/compiler.cpp
+++ b/lib/aot/compiler.cpp
@@ -17,6 +17,7 @@
 #include <limits>
 #include <lld/Common/Driver.h>
 #include <llvm/Analysis/TargetLibraryInfo.h>
+#include <llvm/Analysis/TargetTransformInfo.h>
 #include <llvm/IR/IRBuilder.h>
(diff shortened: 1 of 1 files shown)


lib/aot/compiler.cpp
```diff
@@ -40,6 +40,10 @@
 #include <llvm/Object/COFF.h>
 #endif
 
+#if LLVM_VERSION_MAJOR >= 15
```
ibmibmibm, line 43: No needs to check version here, the header was presented before llvm 9

ibmibmibm (changes requested):

lib/aot/compiler.cpp
hydai: @ibmibmibm 
We can use your PR instead #1863

hydai: Oops, looks like you already deleted the branch, let me update this now.

ibmibmibm (approved):

PR #1867, opened 2022-09-23
[Rust] Fixed the potential deadlock issue.

### LINKED ISSUE #1813: [Rust] Deadlocks will occur in some cases
Under multi-threading, the global variable [`HOST_FUNCS`](https://github.com/WasmEdge/WasmEdge/blob/master/bindings/rust/wasmedge-sys/src/lib.rs#L231) (of type `Arc<RwLock<HashMap<usize, BoxedFn>>`) may deadlock in some cases.

For example, creating a host function in a host function causes a situation where the write operation is performed while there is a read lock held.

But what needs to be confirmed is whether runtime core supports the act of creating a host function within a host function?

apepkuss:
In this PR, we refactor the original design of the host function storage so that some potential deadlock issues caused by the read/write contention can be solved entirely.

- Fixed #1813

### DIFF
--- a/.github/workflows/inventory.yml
+++ b/.github/workflows/inventory.yml
@@ -49,6 +49,7 @@ jobs:
           WasmEdge_VMRunWasmFromASTModule
           WasmEdge_VMRunWasmFromBuffer
           WasmEdge_VMRunWasmFromFile
+          WasmEdge_StringWrap
           EOF
           echo "===Allowlist==="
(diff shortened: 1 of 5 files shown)


apepkuss:
@hydai could you please help review this PR? Thanks a lot!

hydai (approved):

apepkuss:
@hydai Thanks for the review!

PR #1888, opened 2022-10-01
[MISC] add qualified name std for move and forward

harry900831:
This PR format the include/experimental/expected.hpp by clang-format and solve the error due to the new Warning -Wunqualified-std-cast-call added by clang-15.


harry900831:
Link to the warning: https://clang.llvm.org/docs/DiagnosticsReference.html#wunqualified-std-cast-call

q82419:
@ibmibmibm 
This file is for the compiler with c++14 standard before.
Do you think we should fix it or drop it with using std::expected in c++17 instead?

hydai:
> @ibmibmibm This file is for the compiler with c++14 standard before. Do you think we should fix it or drop it with using std::expected in c++17 instead?

We already used the c++17 standard.
https://github.com/WasmEdge/WasmEdge/blob/master/cmake/Helper.cmake#L81

hydai (approved):

PR #1902, opened 2022-10-05
[Installer] add new options that user could specify os and arch for installer

CaptainVincent:
The reason is we need different WasmEdge with system environment os-arch version when docker cross compile image


CaptainVincent:
@hydai I could ask who help review installer, thx.

hydai (approved):

PR #1966, opened 2022-10-14
[Misc] Add off log system api

CaptainVincent:
avoid some error log confuse users when migrate wasmedge into other production

> I think the error (ex. interrupt) msg should be able hide thus avoid scare user (?

Signed-off-by: vincent <vincent@secondstate.io>

### DIFF
--- a/.github/workflows/inventory.yml
+++ b/.github/workflows/inventory.yml
@@ -50,6 +50,7 @@ jobs:
           WasmEdge_VMRunWasmFromBuffer
           WasmEdge_VMRunWasmFromFile
           WasmEdge_StringWrap
+          WasmEdge_Plugin_GetDescriptor
           EOF
           echo "===Allowlist==="
           cat allowlist.txt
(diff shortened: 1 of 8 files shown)


hydai (changes requested):
Because you are trying to add new C API, please bump the SOVERSION from `0.0.0` to `0.0.1`.

.github/workflows/rust-wasmedge-sys-docs.yml
```diff
@@ -21,6 +21,9 @@ on:
       - ".github/workflows/rust-wasmedge-sys-docs.yml"
       - "bindings/rust/wasmedge-sys/**"
 
+permissions:
+  pull-requests: write
```
hydai, line 25: The PR should not have the right to write anything on our branch.

hydai (changes requested):

hydai (approved):

hydai (approved):

q82419 (approved):

PR #1980, opened 2022-10-17
[Utils] Installer CI: Change debug option

SAtacker:

### DIFF
--- a/.github/workflows/test-install-script.yml
+++ b/.github/workflows/test-install-script.yml
@@ -96,25 +96,25 @@ jobs:
         ${{ matrix.package_manager }} install -y ${{ matrix.python_package }}
     - name: Install WasmEdge latest release
       run: |
-        bash utils/install.sh -V
+        bash utils/install.sh -D
     - name: Uninstall WasmEdge
       run: |
         bash utils/uninstall.sh -q -V
     - name: Install WasmEdge latest release with all extensions
       run: |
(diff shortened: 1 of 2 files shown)


hydai (dismissed):

SAtacker:
`which` does not exist on Cent OS

SAtacker:
And GitHub has blocked access to /usr/lib
```
 ls: cannot access '/usr/lib/wasmedge/': No such file or directory
Fail: Plugins not found
Error: Process completed with exit code 1.
```

hydai:
Sounds like the `which` command is an issue.

hydai:
> And GitHub has blocked access to /usr/lib
> 
> ```
>  ls: cannot access '/usr/lib/wasmedge/': No such file or directory
> Fail: Plugins not found
> Error: Process completed with exit code 1.
> ```

Do you know any documents about this rule? If we use a docker image why the GitHub runner can block this access?

hydai:
Also, we may need to have a check for the `which` command. If the environment is not found, we should warn users.

https://stackoverflow.com/questions/70417033/bash-script-error-which-command-not-found

SAtacker:
> Also, we may need to have a check for the `which` command. If the environment is not found, we should warn users.
> 
> https://stackoverflow.com/questions/70417033/bash-script-error-which-command-not-found

Yes, on it

hydai (approved):

Issue #1909, opened 2022-10-06
Test failed on Apple M1 


harry900831:
## Description
The test failed on Apple M1.

### Current State
The ctest failed on the following tests:
```
	  8 - wasmedgeExecutorCoreTests (Failed)
	 11 - wasmedgeAPIVMCoreTests (Failed)
	 12 - wasmedgeAPIStepsCoreTests (Failed)
	 15 - wasiSocketTests (Failed)
	 16 - wasiTests (Failed)
```
They should pass on [macOS 12(x86_64)](https://github.com/WasmEdge/WasmEdge/actions/runs/3191861325/jobs/5208678873#step:6:56) except for the `wasiSocketTests` which is already a known issue.

The first three failed at `TestUnit/CoreTest.TestSuites/80, where GetParam() = “core simd_f64x2_rounding”`.
```
[2022-10-06 20:51:58.654] [info] core simd_f64x2_rounding
/Users/hchiang/WasmEdge/test/spec/spectest.cpp:500: Failure
Value of: compares(Returns, *Res)
  Actual: false
Expected: true
/Users/hchiang/WasmEdge/test/spec/spectest.cpp:500: Failure
Value of: compares(Returns, *Res)
  Actual: false
Expected: true
/Users/hchiang/WasmEdge/test/spec/spectest.cpp:500: Failure
Value of: compares(Returns, *Res)
  Actual: false
Expected: true
/Users/hchiang/WasmEdge/test/spec/spectest.cpp:500: Failure
Value of: compares(Returns, *Res)
  Actual: false
Expected: true
/Users/hchiang/WasmEdge/test/spec/spectest.cpp:500: Failure
Value of: compares(Returns, *Res)
  Actual: false
Expected: true
/Users/hchiang/WasmEdge/test/spec/spectest.cpp:500: Failure
Value of: compares(Returns, *Res)
  Actual: false
Expected: true
/Users/hchiang/WasmEdge/test/spec/spectest.cpp:500: Failure
Value of: compares(Returns, *Res)
  Actual: false
Expected: true
...
```
And the `wasiTests` failed with the following log:
```
[ RUN      ] WasiTest.PollOneoffSocket
/Users/hchiang/WasmEdge/test/host/wasi/wasi.cpp:1042: Failure
Expected equality of these values:
  NEvents
    Which is: 1
  2
/Users/hchiang/WasmEdge/test/host/wasi/wasi.cpp:1044: Failure
Expected equality of these values:
[...]

PR #1989, opened 2022-10-18
Fix the test failed on M1

harry900831:
#1909 
> A strict rule is: Never ever write to an input operand. ...
> By http://www.ethernut.de/en/documents/arm-inline-asm.html

So I take [llvm's code](https://github.com/llvm/llvm-project/blob/main/libc/src/__support/FPUtil/aarch64/nearest_integer.h) as an example to change it. I can pass the test failed by "simd_f64x2_rounding" on M1 chip after this commit.

I'm also wondering whether we should add __volatile__ to prevent from compiler optimization.

### DIFF
--- a/include/common/roundeven.h
+++ b/include/common/roundeven.h
@@ -56,7 +56,7 @@ inline float roundevenf(float Value) {
   return Ret;
 #elif defined(__aarch64__)
   float Ret;
-  __asm__("frintn %s0, %s0" : "=w"(Ret) : "w"(Value));
+  __asm__("frintn %s0, %s1" : "=w"(Ret) : "w"(Value));
   return Ret;
 #else
   assuming(fegetround() == FE_TONEAREST);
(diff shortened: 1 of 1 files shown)


q82419:
@ibmibmibm Can you help to check this is available or not on aarch64 so that we can merge this before alpha next week? Thanks.

include/common/roundeven.h
```diff
@@ -56,7 +56,7 @@ inline float roundevenf(float Value) {
   return Ret;
 #elif defined(__aarch64__)
   float Ret;
-  __asm__("frintn %s0, %s0" : "=w"(Ret) : "w"(Value));
+  __asm__("frintn %s0, %s1" : "=w"(Ret) : "w"(Value));
```
hydai, line 59: Is this failed only on the macOS m1, or on any aarch64 platform?
If it's only related to macOS, then you must identify that in a `defined` condition.

hydai (changes requested):

ibmibmibm (approved):

include/common/roundeven.h
hydai: Ah, should apply to all aarch64. Thanks.

hydai (approved):

PR #1995, opened 2022-10-22
[Rust] Changes for supporting youki

apepkuss:
In this PR, the following changes are introduced for supporting [youki runtime](https://github.com/containers/youki),

* Improved the code for loading plugin in `wasmedge_sdk::Vm`
* Improved the search mechanism in the build script


apepkuss:

bindings/rust/wasmedge-sdk/src/vm.rs
```diff
@@ -108,7 +108,9 @@ impl Vm {
     /// If fail to create, then an error is returned.
     pub fn new(config: Option<Config>) -> WasmEdgeResult<Self> {
         // load wasmedge_process plugins
-        sys::utils::load_plugin_from_default_paths();
+        if std::env::var_os("WASMEDGE_PLUGIN_PATH").is_some() {
+            sys::utils::load_plugin_from_default_paths();
```
hydai, line 112: This function will also try to load the plugins under the system-wise path and the default user-level installed path.

hydai (changes requested):

bindings/rust/wasmedge-sdk/src/vm.rs
apepkuss: When we tried to integrate WasmEdge with [youki runtime](https://github.com/containers/youki) recently, we run into an issue with the call to `sys::utils::load_plugin_from_default_paths` which simply calls the C-API `WasmEdge_PluginLoadWithDefaultPaths` inside. It just returned without any error message or anything else while running the call to the function. When we commented it out, the issue disappeared. This is the first time of observing such an issue, and the explanation on the `WasmEdge_PluginLoadWithDefaultPaths` C API is limited, so we decided to handle it based on the environment variable `WASMEDGE_PLUGIN_PATH`. According to your review comments, I think it would be better to remove the plugin loading logic from `Vm::new` and let the users decide if they wanna load plugins or not in their scenario. Thanks!

apepkuss: BTW, could `WasmEdge_PluginLoadWithDefaultPaths` return a boolean value to indicate if the plugins are loaded successfully or not? Just an advice, if it's infeasible, please ignore it :) Thanks!

hydai (approved):

apepkuss:
@hydai Thanks for the review!

PR #2010, opened 2022-10-25
Add document for integrating WasmEdge with Knative

zhlhahaha:
Recently, we are trying to integrate Serverless with WASM on Arm64 platform. It works fine, and I find there is no document about it, so I write down this in case some people may have an interest in this topic. Would you like to take a look when you have time? Thanks!


q82419:
If you want to add a new page, you should add the page path to the `SUMMARY.md`.

PR #2022, opened 2022-10-27
[Docs] Update WASI-NN TFLite supporting docs.

q82419:


docs/book/en/src/write_wasm/rust/wasinn.md
```diff
@@ -178,7 +200,7 @@ We can find that the outputted WASM files import the necessary WASI-NN functions
   ...
 ```
 
-### Using WASI-NN with OpenVINO™ Backend in Rust
```
hydai, line 181: You should keep the `TM`.

hydai (changes requested):

docs/book/en/src/write_wasm/rust/wasinn.md
q82419: This will cause the link failed.

hydai (approved):

Issue #2029, opened 2022-10-30
Remove None from ValType and support reading S33


little-willy:
## Motivation

According to the [wasm spec](https://webassembly.github.io/spec/core/binary/types.html#binary-valtype), None is not one of the `ValType`. However, in our current code, None is one of `ValType` with code `0x40`. I think the reason for including it in `ValType` is for the condition block type of instruction `block`, `loop` and `if` as in [block-type](https://webassembly.github.io/spec/core/binary/instructions.html#binary-blocktype), where after we include `None` in `ValType`, we seems to be able to make the instruction loader code easier. However, first, it makes our data structure inconsistent to the one in the spec, and second, it makes our code more complicated and unclean, since we have to check for the `None` case every time we use `ValType`.

Besides, in the [wasm spec](https://webassembly.github.io/spec/core/binary/types.html#binary-valtype), the value for type index is `s33`, but in our code, the type index is read as `s32`, which narrows the index range, and gets inconsistent to the spec.

## Details

We can do the following refactor:
* In `BlockType`, we may store `std::optional<ValType>` to indicate the none case, and modify the instruction loader accordingly.
* Remove `None` from the `ValType`
* Remove all the usage of `ValType::None`
* Remove the `AcceptNone` from the parameter list of `checkValTypeProposals`
* Support parsing `s33`.

PR #2030, opened 2022-10-30
feat: support read sn and read s33

little-willy:
In our current code of file manager, we have two method, `readS32` and `readS64`, which both read a variable length integer from the file. Actually, they seems to share the same read logic as described in [sN](https://webassembly.github.io/spec/core/binary/values.html#binary-int). Besides, as proposed in #2029, we may support reading `s33`, which also share the same read logic. 

Therefore, in this PR, I support reading `sN` as described in [sN](https://webassembly.github.io/spec/core/binary/values.html#binary-int), where N becomes a static constant template variable, and enables static usage check to ensure correctness. `readS32`, `readS64` and `readS33` are simply specialization of `readSN`.


juntao:
Hi @little-willy 

Thank you for your contribution!

CNCF requires its projects to have DCO signatures for all contributions. Do you mind fixing the DCO on your commits? Thank you!

https://www.secondstate.io/articles/dco/

lib/loader/filemgr.cpp
```diff
@@ -275,8 +287,8 @@ Expect<std::string> FileMgr::readName() {
   if (unlikely(Status != ErrCode::Value::Success)) {
     return Unexpect(Status);
   }
-  // If UTF-8 validation or readU32() or readBytes() failed, the last succeeded
-  // reading offset will be at the start of `Name`.
+  // If UTF-8 validation or readU32() or readBytes() failed, the last
+  // succeeded reading offset will be at the start of `Name`.
```
q82419, line 291: This line is not necessary to be changed.

lib/loader/filemgr.cpp
```diff
@@ -411,7 +423,8 @@ Expect<void> FileMgr::jumpContent() {
   return {};
 }
 
-// Helper function for reading number of bytes. See "include/loader/filemgr.h".
+// Helper function for reading number of bytes. See
+// "include/loader/filemgr.h".
```
q82419, line 427: This line is not necessary to be changed.

q82419 (changes requested):
Can you help to fix the loading of block type [here](https://github.com/WasmEdge/WasmEdge/blob/master/lib/loader/ast/instruction.cpp#L188)?

Thanks!

lib/loader/filemgr.cpp
```diff
@@ -144,84 +144,96 @@ Expect<uint64_t> FileMgr::readU64() {
...
-  uint64_t Offset = 0;
-  Byte Byte = 0x80;
-  while (Byte & 0x80) {
-    if (unlikely(Offset >= 64)) {
-      Status = ErrCode::Value::IntegerTooLong;
-      return Unexpect(Status);
-    }
-    if (auto Res = testRead(1); unlikely(!Res)) {
+    WasmEdge::Byte Byte;
+    if (auto Res = readByte(); unlikely(!Res)) {
```
q82419, line 173: Please use the `testRead(1)` and `Byte = Data[Pos++]` mechanism as the origin code.
This is for printing the correct error message of the loading offset when error occurs.

q82419 (changes requested):

little-willy:
@q82419 Thanks for reviewing. I have fixed the comment and added unit tests for `readS33`. PTAL.

> Can you help to fix the loading of block type [here](https://github.com/WasmEdge/WasmEdge/blob/master/lib/loader/ast/instruction.cpp#L188)?

Can I include this fix in a later PR that will fix #2029 totally and merge this PR first?

q82419 (approved):

q82419:
> @q82419 Thanks for reviewing. I have fixed the comment and added unit tests for `readS33`. PTAL.
> 
> > Can you help to fix the loading of block type [here](https://github.com/WasmEdge/WasmEdge/blob/master/lib/loader/ast/instruction.cpp#L188)?
> 
> Can I include this fix in a later PR that will fix #2029 totally and merge this PR first?

Sure.
Thanks for your contribution!

PR #2036, opened 2022-11-01
[Plugin] Handle the case of WasmEdge installation path in `/usr`.

q82419:


lib/plugin/plugin.cpp
```diff
@@ -8,6 +8,7 @@
 #include <type_traits>
 #include <variant>
 
+#include <iostream>
```
hydai, line 11: Redundant line.

hydai (changes requested):

hydai (approved):

PR #2042, opened 2022-11-02
[MISC] Fix default plugin paths

harry900831:
Signed-off-by: Harry Chiang <harry900831@gmail.com>

### DIFF
--- a/lib/plugin/plugin.cpp
+++ b/lib/plugin/plugin.cpp
@@ -272,9 +272,9 @@ std::vector<std::filesystem::path> Plugin::getDefaultPluginPaths() noexcept {
       Result.push_back(LibPath / std::filesystem::u8path("wasmedge"sv));
     } else {
       // The installation path of the WasmEdge library is not under "/usr", such
-      // as "$HOME/.wasmedge". Plug-in path will be in "LIB_PATH/../plugin".
-      Result.push_back(LibPath / std::filesystem::u8path(".."sv) /
-                       std::filesystem::u8path("plugin"sv));
+      // as "$HOME/.wasmedge". Plug-in path will be in "LIB_PATH/../../plugins".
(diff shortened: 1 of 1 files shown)


harry900831:
I follow the Build WasmEdge from https://wasmedge.org/book/en/contribute/build_from_src/macos.html, the tree is as follow:

build/
|-- lib
|   |-- api
|   |  |-- libwasmedge.so -> libwasmedge.so.0
|   |  |-- libwasmedge.so.0 -> libwasmedge.so.0.0.1
`-- plugins
    |-- wasmedge_wasiNN
    |  |-- libwasmedgePluginWasiNN.so
    |-- wasmedge_process
    |  |-- libwasmedgeprocess.so

q82419:
> I follow the Build WasmEdge from https://wasmedge.org/book/en/contribute/build_from_src/macos.html, the tree is as follow:
> 
> build/ |-- lib | |-- api | | |-- libwasmedge.so -> libwasmedge.so.0 | | |-- libwasmedge.so.0 -> libwasmedge.so.0.0.1 `-- plugins |-- wasmedge_wasiNN | |-- libwasmedgePluginWasiNN.so |-- wasmedge_process | |-- libwasmedgeprocess.so

Its the build tree, you should install it.
If you modify the code as this PR, the installer case will fail.

q82419:
You can refer to the build command here:

https://wasmedge.org/book/en/contribute/build_from_src/plugin_wasi_nn.html

You should install the build to put the plug-in into the correct path, or set the `WASMEDGE_PLUGIN_PATH`.

harry900831:
Aha, I found my mistake...
I've tried cmake --install after cmake --build before, and the path still wrong... that's because I'm still using the build/tools/wasmedge/wasmedge so I thought this part is incorrect.

Thanks!

PR #2058, opened 2022-11-07
refactor: unify the logic of loading vector

little-willy:


little-willy:
@hydai @q82419 I have fixed the CI. Now except the `OpenWrt` that also fails in other PRs, other CI have all passed. Could you take a look at this PR?

And I have a question. In the current master branch, after we read the vector size, we will return error on `VecCnt / 2 > Fmgr.getRemainSize()`. What is the reason for adding a `/ 2` in the check? Originally I think checking `VecCnt > Fmgr.getRemainSize()` should be enough, but then I failed in some CI such as `Test WasmEdge Core / build_on_ubuntu_20_04 / Build on ubuntu-20.04-coverage with g++` and `Test WasmEdge Core / build_on_fedora / Fedora 35 (pull_request)`. Is there any magic in this `/ 2`?

q82419:
> @hydai @q82419 I have fixed the CI. Now except the `OpenWrt` that also fails in other PRs, other CI have all passed. Could you take a look at this PR?
> 
> And I have a question. In the current master branch, after we read the vector size, we will return error on `VecCnt / 2 > Fmgr.getRemainSize()`. What is the reason for adding a `/ 2` in the check? Originally I think checking `VecCnt > Fmgr.getRemainSize()` should be enough, but then I failed in some CI such as `Test WasmEdge Core / build_on_ubuntu_20_04 / Build on ubuntu-20.04-coverage with g++` and `Test WasmEdge Core / build_on_fedora / Fedora 35 (pull_request)`. Is there any magic in this `/ 2`?

Thanks for helping.

Because in the WASM spec tests, there are tests to detect the errors in loading phase.
The concept is that the WASM binary will give a vector in a section, and not give enough content size (usually lack of 1 byte), so the error occurs.
In WasmEdge, we restrictly compare the error message when error occurs to pass the spec tests. If we apply the `VecCnt > Fmgr.getRemainSize()`, the error code will usually not match the expected error code in WASM spec tests.
But for the too-large-vector-size case, we cannot wait for reading until the end -- suppose that there may be `2^31` of vector count, but only 2 bytes of data, the `std::vector` will resize to `2^31` first.
So.. Yes, you can seem it as a magic number.

little-willy:
@q82419 Any comment on this PR? Could we get this PR merged?

include/loader/loader.h
```diff
@@ -152,30 +152,50 @@ class Loader {
...
+  }
+
+  Expect<uint32_t> loadVecCnt() {
     // Read the vector size.
     if (auto Res = FMgr.readU32()) {
-      VecCnt = *Res;
-      if (VecCnt / 2 > FMgr.getRemainSize()) {
-        return logLoadError(ErrCode::Value::IntegerTooLong,
-                            FMgr.getLastOffset(), NodeAttrFromAST<T>());
+      // Question: why we used to divide by 2 here?
```
hydai, line 165: According to the definition, there should not be larger than 2*remainsize. Otherwise, we should show an error about the length being too long.

lib/loader/ast/instruction.cpp
```diff
@@ -234,21 +234,23 @@ Expect<void> Loader::loadInstruction(AST::Instruction &Instr) {
 
   case OpCode::Br_table: {
     uint32_t VecCnt = 0;
-    // Read the vector of labels.
-    if (auto Res = readU32(VecCnt); unlikely(!Res)) {
-      return Unexpect(Res);
-    }
-    if (VecCnt / 2 > FMgr.getRemainSize()) {
-      // Too many label for Br_table.
```
hydai, line 242: You also deleted the reason why we have to check the value here.

hydai (changes requested):
@q82419 PTAL

q82419:
> @q82419 PTAL

I think the changes are OK. Please help to fix the comments. Thanks.

q82419:
Hi @little-willy ,

Would you help to update this first? Thanks!

include/loader/loader.h
little-willy: I have added these explanation to where the commented code is.

lib/loader/ast/instruction.cpp
little-willy: I think this comparison with `VecCnt / 2` with `FMgr.getRemainSize()` is a general error, and we don't have to specially include this comment here in the code.

little-willy:
@q82419 @hydai I have addressed the comment by adding the explanation for divide by a magic number 2 in the comment.

I think this PR is not related to supporting func-ref and GC, and does not involve breaking change. So I just switched it into merging into the master branch. PTAL, thanks.

include/loader/loader.h
dannypsnl: I might use
```
// NOTE: why ...?
//
// According to ...
```
to avoid confusion here for reviewer

dannypsnl:
Do we still plan to merge this PR?

q82419:
> Do we still plan to merge this PR?

This is refactored in the PRs in component model and GC.
Close this PR.

PR #2086, opened 2022-11-14
[Rust] Support wat files

apepkuss:
In this PR, some new APIs are introduced into `Vm` and `Module` of `wasmedge-sdk` to support `wat` files. In addition, improve the implementation of some APIs in `Vm` to support `AOT` files.


apepkuss:
@hydai Thanks for the review!

PR #2095, opened 2022-11-18
Fix the implementation of threads proposal.

q82419:


hydai (approved):
LGTM. @ibmibmibm PTAL

PR #2099, opened 2022-11-20
refactor: turn ValType into a structured FullValType

### LINKED ISSUE #2081: feat: support ValType as a structured class with variant
## Motivation

In our current support for WASM, we store all `ValType`s as uint_8, since in the current official WASM spec, all type op codes are only one byte. However, as we are supporting function-reference and GC proposal, the type specifier of `ValType` may occupy multiple bytes for typed reference.

image

Therefore, to enable the best flexibility, we may support ValType as a structured class.

## Details

We may replace the current ValType enum class definition with a structured class, which initially may only act as a wrapper for the current enum class. In the future as we are support function-reference and GC proposal, we may add more information in the structured class.

Note: since in our java and rust SDK, we also expose such ValType, the binding interface may not be compatible. Not sure whether this is acceptable.

little-willy:
As proposed in https://github.com/WasmEdge/WasmEdge/issues/2081, we may support a structured `ValType` so that we can support the extended `ValType` definition in function-reference and GC proposal. After I examine the repo, it seems that  the current `ValType` is greatly coupled with the uint8_t definition. Therefore, in this PR, we introduce a new class `FullValType` with the following structure definition. `FullRefType` is introduced for the current `RefType` in the same manner.
```
union WasmEdge_ValTypeExt {};

class FullValType {
  ValType TypeCode;
  WasmEdge_ValTypeExt Ext;
};
```

With the new structured ValType definition, we now store the `FullValType` and `FullRefType` in the param and result type in FunctionType, and also `GlobalType` and `TableType`. Other code are refactored accordingly.

This PR is after #2097, and will close #2081


hydai:
Since this PR will break the current ABI, I would like to know if the current implementation already covers all proposals related to the Value Type.

little-willy:
> Since this PR will break the current ABI, I would like to know if the current implementation already covers all proposals related to the Value Type.

The C API and the interface of java and rust SDK is unchanged. You may refer to wasmedge.h, which involves little code diff in this PR. The only changed to the ABI will be the change on the struct of `WasmEdge_Value`, which stores a different ValType struct. If some users are doing dynamic linking, their code may not work normally due to the ABI change. If they compile their code together with the new version that involves this PR, their code should compile with no error.

little-willy:
This PR is after https://github.com/WasmEdge/WasmEdge/pull/2097, which is an refactor on the VType on the validation checker. Before reviewing this PR, you can review this previous PR and gets it merged first. The previous PR has nothing to do with the C ABI. @hydai

hydai:
After this PR gets merged, we have to jump the SOVERSION to notify our downstream about the ABI changes such as crun(using `dlopen`) and other projects(using `dynamic link`).

ABI changes will affect lots of people. In case we have to extend this again, I just want to make sure the new implementation covers the remaining active wasm proposals.

little-willy:
@hydai If this is related to the ABI change, we may have to pay attention to the size of the `union WasmEdge_ValTypeExt {};`, since it may increases as we add more cases on it.

Besides, in the the implementation in this PR, we still store the `ValType` in the `FullValType`. However, in my next PR that really supports loading the types in the function reference proposal, I will introduce a new enum class `ValTypeCode`, and store this new enum in the `FullValType`. The original `ValType` may be unchanged and marked deprecated for backward compatibility. Do you think I should involve this `ValTypeCode` change in this PR? For your information, the definition of `ValTypeCode` will be the same as `ValType`, but in the PR that supports function reference proposal, it will turn the `ValType::FuncRef` and `ValType::ExternRef` into a specialization of `ValTypeCode::Ref` and `ValTypeCode::RefNull`.

hydai:
> @hydai If this is related to the ABI change, we may have to pay attention to the size of the `union WasmEdge_ValTypeExt {};`, since it may increases as we add more cases on it.

Agreed. The next release will be version 0.12.0 with the new SOVERSION 1.0.0 at the end of the Dec. I think we still have some time to check the size of `WasmEdge_ValTypeExt`.

> Besides, in the the implementation in this PR, we still store the `ValType` in the `FullValType`. However, in my next PR that really supports loading the types in the function reference proposal, I will introduce a new enum class `ValTypeCode`, and store this new enum in the `FullValType`. The original `ValType` may be unchanged and marked deprecated for backward compatibility. Do you think I should involve this `ValTypeCode` change in this PR? For your information, the definition of `ValTypeCode` will be the same as `ValType`, but in the PR that supports function reference proposal, it will turn the `ValType::FuncRef` and `ValType::ExternRef` into a specialization of `ValTypeCode::Ref` and `ValTypeCode::RefNull`.

It's fine to have another PR. I will suggest that there is a universal issue to record these PRs and we can have a `breaking change` label and a checklist for making sure everything is done before we create the release.

little-willy:
@hydai @ibmibmibm @q82419 The preliminary PR https://github.com/WasmEdge/WasmEdge/pull/2097 is merged. Could you take a look at this PR?

bindings/rust/wasmedge-sys/build.rs
```diff
@@ -41,6 +41,8 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
         .clang_arg(format!("-I{}", inc_dir.as_path().display()))
         .prepend_enum_name(false) // The API already prepends the name.
         .dynamic_link_require_all(true)
+        .derive_debug(true)
```
hydai, line 44: @apepkuss 
Could you please help check this part? Will this change any behavior of our current wasmedge-sys crate?

include/api/wasmedge/wasmedge.h
```diff
@@ -48,7 +48,7 @@ typedef struct WasmEdge_Value {
   // functions. Developers should use the corresponding `WasmEdge_ValueGen`
   // functions to generate this struct, and the `WasmEdge_ValueGet` functions to
   // retrieve the value from this struct.
-  enum WasmEdge_ValType Type;
+  WasmEdge_FullValType Type;
```
hydai, line 51: This is a C-style header. Should we keep the `enum` keyword here?

hydai: Another thing is that you changed the enum/class name without modifying the corresponding comments.

include/api/wasmedge/wasmedge.h
```diff
@@ -1166,6 +1166,14 @@ WasmEdge_GlobalTypeCreate(const enum WasmEdge_ValType ValType,
 WASMEDGE_CAPI_EXPORT extern enum WasmEdge_ValType
 WasmEdge_GlobalTypeGetValType(const WasmEdge_GlobalTypeContext *Cxt);
 
+/// Get the full extended value type from a global type.
+///
+/// \param Cxt the WasmEdge_GlobalTypeContext.
+///
+/// \returns the full extended value type of the global type.
+WASMEDGE_CAPI_EXPORT WasmEdge_FullValType
+WasmEdge_GlobalTypeGetFullValType(const WasmEdge_GlobalTypeContext *Cxt);
```
hydai, line 1175: You are introducing a new public API; however, there is no detailed explanation for the difference between the `ValType` and `FullValType.`

hydai (changes requested):

include/api/wasmedge/wasmedge.h
little-willy: The `WasmEdge_FullValType` is a struct rather than a `enum`. The `struct` keyword before the type name is eliminated by the `typedef` in the code that defines the struct, so I think there is no need to add a keyword before it.

include/api/wasmedge/wasmedge.h
little-willy: This function is not compulsory for this PR. Do you think we should remove this function for this PR and implemented a list of new public API for this refactor in a separate PR. @hydai

include/api/wasmedge/wasmedge.h
hydai: Ah, I misunderstood this line.

include/api/wasmedge/wasmedge.h
hydai: It would be better if we collect all changes of the public API in another PR. So we can make sure the document and APIs are all synced and updated.

include/common/enum_types.h
```diff
@@ -24,6 +24,15 @@ enum WasmEdge_ValType {
 #undef UseValType
 };
 
+union WasmEdge_ValTypeExt {
+  uint32_t TypeIdx;
```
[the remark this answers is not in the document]
little-willy: @apepkuss 

1. Actually, in this PR, the field is not used anywhere. This is just a temporary placeholder for the union in this PR, since the CI on Windows fails on a warning that the size of empty union type differs between C and C++. The final struct will be introduced in https://github.com/WasmEdge/WasmEdge/pull/2107. You may refer to the code changes in `enum.inc` and `enum_types.h` for that PR.

For you reference, the struct will be like
```
typedef struct WasmEdge_HeapType {
   enum WasmEdge_HeapTypeCode HeapTypeCode;
   uint32_t DefinedTypeIdx;
 } WasmEdge_HeapType;

 union WasmEdge_ValTypeExt {
   WasmEdge_HeapType HeapType;
 };

 typedef struct WasmEdge_FullValType {
   enum WasmEdge_ValTypeCode TypeCode;
   union WasmEdge_ValTypeExt Ext;
 } WasmEdge_FullValType;
```

The `HeapType` in `ValTypeExt` is only meaningful when the `TypeCode` is `Ref` or `RefNull`.


2. The union is defined as a C style code, so all the field types are `Copy` types. Besides, though the field types are `Copy`, dereferencing the value in the union is unsafe. Personally I will recommend defining the ValType in the Rust SDK as 
```
enum ValType {
    I32,
    I64,
    F32,
    F64,
    V128,
    RefNull(HeapType),
    Ref(HeapType),
}

enum HeapType {
    Func,
    Extern,
    Defined(u32),
}
```
, and then you can turned the struct defined in the generated rust binding code into this self defined enum by first checking the `ValTypeCode`, and then extract the specific fields in the union for different `ValTypeCode` so that the SDK user will not have to handle unsafe code.

3. For `NumType`, the `Ext` field will be meaningless, so if it is required, you can fill in any case as you like. For `RefType`, the `Ext` field will be meaningful, and you simply fill in HeapType in the union as it requires.

include/api/wasmedge/wasmedge.h
little-willy: I see. I will remove this function.

include/common/enum_types.h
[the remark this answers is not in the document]
apepkuss: Thanks for the detailed explanation. As for 3, I am still not sure what value of the `Ext` field is returned from the low-level runtime back to Rust SDK for a `NumType`. I'll check it further. Thanks!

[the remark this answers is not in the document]
little-willy: If a value is of `NumType`, it does not have any extension in the type definition, so there will not be `Ext` set for it, in other word, you don't have to interpret the `Ext` field if the value is of `NumType`. 

The generated struct for `FullValType` is `Sized`. Therefore, even if the `NumType` does not have `Ext`, because it shares the same struct definition with `RefType`, which has something set in the `Ext`, the `NumType` of `FullValType` still occupies some extra unused bytes of `Ext` fields in its memory layout. Since `NumType` does not use the `Ext` field, the bytes for the `Ext` fields can be any random value. And we don't have to pay attention to the `Ext` field for `NumType`, because it is not used. For `NumType`, you can still interpret the `Ext` field as `HeapType`, which is originally for `RefType`, and it can return a `HeapType` struct represented by those random bytes, and such interpretation is safe in rust, but just meaningless.

little-willy:
@hydai I have removed the added C API function. 

Is there anything else I need to modify for this PR?

bindings/rust/wasmedge-sys/build.rs
little-willy: @apepkuss Does the rust side change look good to you?

hydai:
Hi @apepkuss 
I found the rust standalone tests all failed, could you please explain why this workflow will fail in this case?

hydai:
@q82419 PTAL

apepkuss:
> Hi @apepkuss I found the rust standalone tests all failed, could you please explain why this workflow will fail in this case?

@hydai Please ignore the standalone workflows as it only works with `libwasmedge 0.11.2`.

little-willy:
> Please ignore the standalone workflows as it only works with libwasmedge 0.11.2

Can we possibly run the standalone test on the current branch instead of a released version?

apepkuss:
> > Please ignore the standalone workflows as it only works with libwasmedge 0.11.2
> 
> Can we possibly run the standalone test on the current branch instead of a released version?

Of course, yes, but the standalone workflow was originally introduced for `Youki` integration task, so it is just for release versions currently. Maybe we can make it more general in the future. Thanks!

little-willy:
> Use the ValType, RefType, NumType as the class naming.

At the very beginning I was trying to do such change. But our current code is greatly coupled with the `uint8_t` definition on `ValType`, and if we turn the `ValType` into a class, there will be a lot of code change, and more importantly, the C API, which may break the backward compatibility of the user that use the C library. This is the reason why I introduce the new `FullValType` class and the new `ValTypeCode` enum.

little-willy:
@hydai  @q82419 BTW, though this PR is for `function-references` proposal, the great changes in the PR will impact the whole `WasmEdge`. Do you think we should directly merge this PR into `master`, or we merge into `proposal/function-references`, and get the changes in this PR merged to `master` together with `proposal/function-references`?

q82419:
> @hydai @q82419 BTW, though this PR is for `function-references` proposal, the great changes in the PR will impact the whole `WasmEdge`. Do you think we should directly merge this PR into `master`, or we merge into `proposal/function-references`, and get the changes in this PR merged to `master` together with `proposal/function-references`?

Due to impacting the whole WasmEdge, I think it's better to merge it into the `proposal/function-reference` first.
I'll keep updating the branch up to the master, and we'll merge the proposal into master after completion.
Although the refactoring impacts a lot, but I think the other currently developing proposals not depend on this proposal.
After all, working on the proposal branch can let us review the code changes more conveniently when applying this whole proposal.

q82419:
Merge into the proposal branch and fix in the future.

PR #2101, opened 2022-11-21
[Rust] change to Rust-stable

apepkuss:
In this PR the unstable Rust-lang feature is removed. The entire Rust SDK will use Rust-lang in the stable channel.


apepkuss:

hydai (changes requested):
Just want to know, does this PR also remove the ability to pass additional data into the host function?

apepkuss:
> Just want to know, does this PR also remove the ability to pass additional data into the host function?

@hydai Yes, it is equivalent to passing nothing to the `Data` argument of the C-API:

```c
WASMEDGE_CAPI_EXPORT extern WasmEdge_FunctionInstanceContext *
WasmEdge_FunctionInstanceCreateBinding(const WasmEdge_FunctionTypeContext *Type,
                                       WasmEdge_WrapFunc_t WrapFunc,
                                       void *Binding, void *Data,
                                       const uint64_t Cost);
```
If we tried to expose the argument to users, a raw pointer is exposed in the Rust interface; Or, have to use an unstable Rust-lang feature. However, these two cases are not acceptable. So, have to "hide" it.

hydai:
> If we tried to expose the argument to users, a raw pointer is exposed in the Rust interface; Or, have to use an unstable Rust-lang feature. However, these two cases are not acceptable. So, have to "hide" it.

Do you mean the current Rust SDK cannot pass the data into a host function?

apepkuss:
> > If we tried to expose the argument to users, a raw pointer is exposed in the Rust interface; Or, have to use an unstable Rust-lang feature. However, these two cases are not acceptable. So, have to "hide" it.
> 
> Do you mean the current Rust SDK cannot pass the data into a host function?

@hydai Still can pass data arguments to a host function, but not via that `Data` argument in C-API.

apepkuss:
> If I missed something, please correct me. Do we have any example for this(`pass data arguments to a host function`)?

We had two examples for demonstrating the usage of the `Data` argument, but remove them along with this PR. They are:

- `data_object_for_host_func.rs` in `wasmedge-sdk`
- `data_object_to_host_func.rs` in `wasmedge-sys`

hydai (dismissed):

hydai (approved):

apepkuss:
@hydai Thanks for the review!

Issue #1919, opened 2022-10-12
[Windows] WASI Host function - `open`


abhinandanudupa:

Berrysoft:
https://github.com/WasmEdge/WasmEdge/blob/b16d9d3e80b84e17974a911c2de94887194ada58/lib/host/wasi/inode-win.cpp#L493-L495

* `CreateFileA` uses current code page, which may not be UTF-8. At least Rust APIs is passing UTF8 path directly to C APIs. The strings should be converted to UTF-16 and passed to `CreateFileW`.
* When preopening directories for WASI, `INode::open` is also used. `FILE_FLAG_BACKUP_SEMANTICS` should be passed to `dwFlagsAndAttributes` to open the directory. See the documentation [here](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilew).

Berrysoft:
I haven't successfully built WasmEdge on Windows, and I don't use it currently neither. It's just a bug I found when trying it. The fix seems not that easy, and it's up to you all to fix it:)

And by the way, it is a hard time to map a Windows path to WASI with WasmEdge, because the colon after the drive letter will be separated here:

https://github.com/WasmEdge/WasmEdge/blob/e8cf55f25eb88fa9eebbdb5e191d036dbc5982d4/lib/host/wasi/environ.cpp#L59

Luckily, UNC path works...

PR #2120, opened 2022-11-25
[Bug] Fix `INode::open` on windows

abhinandanudupa:
Not ready to be merged yet; more work needs to be done; using CI as I am having build issues on my local system.

Resolves: #1919


abhinandanudupa:
@hydai @ibmibmibm It should be ready now.

lib/host/wasi/inode-win.cpp
```diff
@@ -489,8 +489,24 @@ WasiExpect<INode> INode::open(std::string Path, __wasi_oflags_t OpenFlags,
...
   DWORD CreationDisposition = creationDisposition(OpenFlags);
 
+  int NumCharacters =
+      MultiByteToWideChar(CP_ACP, MB_PRECOMPOSED, Path.c_str(), -1, nullptr, 0);
+
+  if (unlikely(NumCharacters <= 0)) {
+    return WasiUnexpect(fromWinError(GetLastError()));
+  }
+
+  WCHAR PathBuffer[2 * NumCharacters];
```
hydai, line 499: Don't use VLA in C++. Ref: https://stackoverflow.com/questions/1887097/why-arent-variable-length-arrays-part-of-the-c-standard

hydai (changes requested):

lib/host/wasi/inode-win.cpp
abhinandanudupa: I will use `MAX_PATH` instead and also will remove the multiplication by 2 - leftover from the previous attempt :)

hydai: AFAIK the limitation of `MAX_PATH` can be removed after Windows 10 build 1607 by using the following commands:

```PowerShell
New-ItemProperty -Path "HKLM:\SYSTEM\CurrentControlSet\Control\FileSystem" `
-Name "LongPathsEnabled" -Value 1 -PropertyType DWORD -Force
```

Maybe we will need to check the `NumCharacters` if it's larger than the `MAX_PATH.`

Ref: https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry

abhinandanudupa: @hydai Soory for the late reply.

> Maybe we will need to check the `NumCharacters` if it's larger than the `MAX_PATH.`

I think that you are suggesting this: We should check if the `Path` is longer than `MAX_PATH` and if it is then we should query the registry checking if long paths are enabled and if they are enabled then return an error else continue with the normal execution.

I think that this approach is too cumbersome and I suggest that we should do this instead:
Define a constant with the value `32,767` which will be the size of the buffer- `PathBuffer`. So we don't have to worry about the path lengths as `CreateFileW` allows for long paths by default.

hydai (approved):
LGTM

hydai:
I am merging this once the CI is green. Sorry for the delay.

hydai:
Thanks @abhinandanudupa

abhinandanudupa:
@hydai I am very sorry to be replying late!

It's ok :)
And thank you for merging.

PR #2122, opened 2022-11-27
feat: add command line option to enable function reference type

little-willy:
In this PR, we add a new command line option to enable function reference proposal. The option is only added to `wasmedge` command. After this PR, the user can add `--enable-function-reference` to enable the proposal. Though the user can specify enabling the proposal, the support for the proposal is not implemented yet.


q82419 (changes requested):
Hi @little-willy ,

As issue #2161
Thanks.

q82419:
Merge into the proposal branch and fix in the future.

PR #2107, opened 2022-11-23
refactor: introduce heap type and turn reftype into referencing heaptype

little-willy:

PR #2124, opened 2022-11-28
feat: support loading the type extended from function references proposal

little-willy:
after #2107


include/common/types.h
```diff
@@ -157,6 +157,24 @@ class ValTypeBase {
...
+  // Constructor for the heap types (func and extern).
+  HeapType(HeapTypeCode HT) noexcept
+      : ValTypeBase(ValTypeCode::RefNull, HT, 0) {
+    assuming((Inner.Data.Code == ValTypeCode::RefNull));
+    assuming((Inner.Data.HTCode != HeapTypeCode::TypeIndex) &&
+             (Inner.Data.HTCode != HeapTypeCode::NotHeapType));
+  }
+  // Constructor for the heap types (type index).
+  HeapType(uint32_t I) noexcept
+      : ValTypeBase(ValTypeCode::RefNull, HeapTypeCode::TypeIndex, I) {
```
q82419, line 173: I prefer to reserve the originial design here (hidden the `HeapType` and use `RefType` directly).
Because the `HeapType` doesn't describe the `RefType`, the reference type may be `Ref` or `RefNull`, not always `RefNull`.

lib/loader/ast/type.cpp
```diff
@@ -9,22 +9,21 @@ namespace WasmEdge {
 namespace Loader {
 
 // Load binary decode HeapType. See "include/loader/loader.h".
-Expect<RefType> Loader::loadHeapType(RefTypeCode Code, ASTNodeAttr From) {
+Expect<HeapType> Loader::loadHeapType(ASTNodeAttr From) {
```
q82419, line 12: As mentioned above, your `HeapType` implementation here will always create a value type with `RefNull`.
It losts the information of the reference type. The value type may be `Ref` or `RefNull`.

q82419 (changes requested):
Maybe you can choose to drop this PR?
I think the goal of this PR is almost done on the branch already.

lib/loader/ast/instruction.cpp
```diff
@@ -283,13 +283,7 @@ Expect<void> Loader::loadInstruction(AST::Instruction &Instr) {
...
   // Reference Instructions.
   case OpCode::Ref__null: {
-    Expect<RefType> Res;
-    if (Conf.hasProposal(Proposal::FunctionReferences)) {
-      // The func-ref proposal use the heap type to replace ref type here.
-      Res = loadHeapType(RefTypeCode::RefNull, ASTNodeAttr::Instruction);
-    } else {
-      Res = loadRefType(ASTNodeAttr::Instruction);
-    }
+    auto Res = loadRefType(ASTNodeAttr::Instruction);
```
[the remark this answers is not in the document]
little-willy: I will change to use `loadHeapType` in the next PR https://github.com/WasmEdge/WasmEdge/pull/2125. Because in the next PR I will switch to store `HeapType` instead of `RefType` in the `ref.null` instruction.

If we look at the binary code, the `loadHeapType` after supporting is a superset of `loadRefType` in the current stable spec. After we switch to use `loadHeapType`, we call add a check that if func-ref proposal is not enabled, the loaded heap type can only be `func` and `extern`.

include/common/types.h
q82419: Alright. I didn't realize the GC proposal. Thanks.

lib/loader/ast/type.cpp
```diff
@@ -69,7 +68,12 @@ Expect<ValType> Loader::loadValType(ASTNodeAttr From) {
                                Proposal::FunctionReferences,
                                FMgr.getLastOffset(), From);
       }
-      return loadHeapType(static_cast<RefTypeCode>(Code), From);
+      if (auto LoadRes = loadHeapType(From)) {
+        return RefType(static_cast<RefTypeCode>(Code), *LoadRes);
+
```
q82419, line 73: Remove the additional line.

lib/loader/ast/type.cpp
```diff
@@ -104,7 +108,13 @@ Expect<RefType> Loader::loadRefType(ASTNodeAttr From) {
         return logNeedProposal(FailCode, Proposal::FunctionReferences,
                                FMgr.getLastOffset(), From);
       }
-      return loadHeapType(Code, From);
+      if (auto LoadRes = loadHeapType(From)) {
+        return RefType(Code, *LoadRes);
+
```
q82419, line 113: Remove the additional line.

q82419 (changes requested):

q82419:
Don't care about the LLVM related error of the Fedora CI.
This is fixed and merged into master.

little-willy:
@q82419 I have addressed the comment. PTAL

q82419 (approved):

PR #2125, opened 2022-11-28
feat: support loading the instructions extended from the function references proposal

little-willy:
after #2124


lib/loader/ast/instruction.cpp
```diff
@@ -998,7 +997,8 @@ Expect<void> Loader::checkInstrProposals(OpCode Code,
                              Offset, ASTNodeAttr::Instruction);
     }
   } else if (Code == OpCode::Return_call ||
-             Code == OpCode::Return_call_indirect) {
+             Code == OpCode::Return_call_indirect ||
+             Code == OpCode::Return_call_ref) {
```
q82419, line 1001: Not to add the `Return_call_ref` in this condition.
The `Return_call_ref` case will be blocked in this statement and always not run into the condition below.

lib/loader/ast/instruction.cpp
```diff
@@ -1011,6 +1011,14 @@ Expect<void> Loader::checkInstrProposals(OpCode Code,
...
                              Offset, ASTNodeAttr::Instruction);
     }
+  } else if (Code == OpCode::Call_ref || Code == OpCode::Return_call_ref ||
+             Code == OpCode::Ref__as_non_null || Code == OpCode::Br_on_null ||
+             Code == OpCode::Br_on_non_null) {
+    if (!Conf.hasProposal(Proposal::FunctionReferences)) {
+      return logNeedProposal(ErrCode::Value::IllegalOpCode,
+                             Proposal::FunctionReferences, Offset,
+                             ASTNodeAttr::Instruction);
+    }
```
q82419, line 1021: You can add the `tail-call` condition for the `Return_call_ref` OpCode.

q82419 (changes requested):

q82419:
Hi @little-willy ,
Can you help to update this PR? Thanks.

little-willy:
@q82419 Comments are addressed. PTAL

q82419:
> @q82419 Comments are addressed. PTAL

The Loader test failed. Please fix it. Thanks.

little-willy:
> The Loader test failed. Please fix it. Thanks.

@q82419 It seems that the test fails in some platforms while succeeds in others. Do you have any idea what causes this failure?

little-willy:
@q82419 I have fixed the unit test. It is caused by not checking the reference type proposal when we load heap type. But this bug seems to be independent to platforms and should fail in all platforms. Do you know why the CI can succeed in some platforms?

q82419:
> @q82419 I have fixed the unit test. It is caused by not checking the reference type proposal when we load heap type. But this bug seems to be independent to platforms and should fail in all platforms. Do you know why the CI can succeed in some platforms?

The Windows and MacOS CI are set as continue-on-failed for tests because the WASI will fail on these platforms.
For the previous error, you can find that the same tests failed on these platforms with the log.
Therefore we should guarantee the linux platforms should pass the CI.

q82419 (approved):

PR #2131, opened 2022-11-29
[CI] Refactor the release workflows

apepkuss:
In this PR, the release workflows for `wasmedge-macro` and `wasmedge-types` are updated.

### DIFF
--- a/.github/workflows/rust-wasmedge-macro-release.yml
+++ b/.github/workflows/rust-wasmedge-macro-release.yml
@@ -22,10 +22,10 @@ jobs:
       - name: Set up build environment
         run: |
           sudo apt-get update
-          sudo apt-get install -y software-properties-common libboost-all-dev ninja-build
+          sudo apt-get install -y software-properties-common libboost-all-dev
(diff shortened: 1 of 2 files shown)


hydai (dismissed):

hydai:
Hi @apepkuss 
Why do you roll back the cargo-smart-release? IIRC, we change it just a few weeks ago.

apepkuss:
> Hi @apepkuss Why do you roll back the cargo-smart-release? IIRC, we change it just a few weeks ago.

@hydai The motivation for using `cargo-smart-release` is that it can (1) make tags automatically and (2) automate the changelog generation. However, in the recent two releases, we found that (1) it is not allowed to tag a release automatically, (2) the tag cannot be customized, and (3) the contents of the CHANGELOG generated do not satisfy our requirements. In addition, `cargo publish` is documented better and much more lightweight than `cargo-smart-release`. Based on the reasons above we believe that `cargo publish` is a better choice for our current requirements. Thanks!

hydai (approved):
LGTM

apepkuss:
@hydai Thanks for the review!

PR #2141, opened 2022-12-01
[CI] Refactor rust workflows

apepkuss:
In this PR, refactor the Rust SDK related workflows to use `dtolnay/rust-toolchain@stable`.

### DIFF
--- a/.github/workflows/bindings-rust.yml
+++ b/.github/workflows/bindings-rust.yml
@@ -39,7 +39,7 @@ jobs:
           fetch-depth: 0
 
       - name: Install Rust v1.65
-        uses: dtolnay/rust-toolchain@1.65
+        uses: dtolnay/rust-toolchain@stable
         with:
           toolchain: 1.65
           components: rustfmt, clippy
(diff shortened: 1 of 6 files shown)


.github/workflows/bindings-rust.yml
```diff
@@ -39,7 +39,7 @@ jobs:
           fetch-depth: 0
 
       - name: Install Rust v1.65
-        uses: dtolnay/rust-toolchain@1.65
+        uses: dtolnay/rust-toolchain@stable
         with:
           toolchain: 1.65
```
hydai, line 44: Should we also use the latest rust toolchain instead of `1.65`?

hydai (changes requested):

.github/workflows/bindings-rust.yml
apepkuss: From the perspective of Rust SDK, it's fine to use the latest stable version, but I'm not sure if any dependency in the dependency chain is compatible with the latest stable version. Of course, `1.65` is the latest stable version and no issues on this version. In my view, it could be better for us to use the recently published two or three stable versions in our CI workflows. If the latest version caused any compatibility failure, the others can help us.

hydai: So do we need to bump the version now? Or we can keep the `1.65` for a while.

hydai (approved):
LGTM

apepkuss:
@hydai Thanks for the review!

PR #2151, opened 2022-12-05
feat: support loading table extended from function reference

little-willy:
After #2125.


alabulei1:
flows summarize

alabulei1:

------

Overall, the patches involve a mix of changes related to refactoring, formatting, bug fixes, support for new types, and additions of new functions. The potential issues and errors include: the introduction of a new optional type that could introduce new issues if not handled properly, the possibility of breaking compatibility with existing code depending on the changes being made, and the removal of helper functions that could cause issues down the line. 

The most important findings include: the introduction of heap types and the deprecation of previous type definitions, which will require thorough testing and may impact external interfaces or dependencies, and the addition of a new function without a clear description or testing. Additionally, it is important to thoroughly review the changes in each patch, even for those with minor modifications, to ensure that they do not introduce any regressions or conflicts.

## Details

### Commit 1
The patch includes refactoring of the FormChecker class in the Validator namespace to use an optional ValType for validator VType. The key changes are the replacement of the `enum class VType` with `typedef std::optional<ValType> VType` and the update of the respective functions to use this new type: `ASTToVType`, `isNumType`, `isRefType`, `FormChecker`, `VTypeToAST`, `checkInstr`, `popType`, `popTypes`, `pushType`, and `StackTrans`. 

An important finding is that `VType` is now an optional, and this could introduce new issues if it's not handled properly. Also, the patch is quite large, with over 300 deletions and 370 insertions, so it is essential to review the changes carefully to ensure that the code still works as intended.

### Commit 2
The patch changes the formatting of the code to conform to clang-format. It appears to make changes to whitespace and line breaks. There are no changes in the functionality of the code. 

The changes appear to be safe and should not cause any problems.

### Commit 3
This patch contains a number of changes throughout the codebase, centered around the `ValType` type which has been refactored into `FullValType`. 

One potential issue is that this patch contains changes to a significant number of files, so the scope of the changes should be carefully considered. Additionally, changes to types could result in knock-on effects in other parts of the codebase, so a thorough review of any dependent code should be undertaken. 

Finally, there are some changes to the format of the code, such as the addition of whitespace and indentation, which could result in conflicts when merging.

### Commit 4
This patch seems to be modifying six different source code files. The changes include adding line breaks, indentation, and spaces to improve the code's readability. The patch also adds a few `noexcept` annotations along with some parameters indentation. 

There don't seem to be any potential problems with these changes. However, as a reviewer it's always important to check the first and last lines of the patch to determine if there are any conflicts or errors with the file modifications.

### Commit 5
This patch contains two changes in two different files:

In bindings/rust/wasmedge-sys/src/types.rs:
- The Debug implementation for WasmValue now uses Rust's block formatting syntax instead of method chaining for better readability.

In lib/api/wasmedge.cpp:
- A new field is added to the returned WasmEdge_FullValType struct, which is initialized with an empty WasmEdge_ValTypeExt struct.

There do not seem to be any potential problems in this patch.

### Commit 6
This patch adds a new CAPI, "WasmEdge_GlobalTypeGetFullValType", to the "wasmedge.h" header file. This CAPI retrieves the full extended value type from a global type. The patch does not mention any potential problems.

### Commit 7
The patch introduces changes to multiple files in the project that replace the use of `ValType` with `FullValType` and `RefType` with `FullRefType`. `FullValType` includes an additional field `WasmEdge_ValTypeExt`, while `FullRefType` extends `RefType` with an `Ext` field. The `ValType` and `RefType` fields are now accessed through functions. 

One potential problem is that the changes may break compatibility with existing code that relies on the previous versions of `ValType` and `RefType`. Additionally, the patch does not include any unit tests to ensure that the updated code functions as expected.

### Commit 8
This patch contains changes to two files: "enum_types.h" and "helper.cpp". The purpose of the changes is to fix a problem with the Windows CI build. 

In "enum_types.h", the `WasmEdge_ValTypeExt` union is modified to include a new field `TypeIdx`. This change could potentially introduce bugs if the new field is not used correctly in the codebase.

In "helper.cpp", there is a minor change of using `Types[I].asCStruct()` instead of `static_cast<WasmEdge_ValType>(Types[I])`. This change seems trivial and should not cause any issues.

Overall, the changes in this patch seem relatively small and targeted towards fixing a known issue. However, it is important to test the changes thoroughly to ensure that they do not introduce any new problems or regressions.

### Commit 9
Key Changes:
- The function `loadFullValType` now has an additional input parameter `TypeCode`.
- The two helper functions `checkValTypeProposals` and `checkRefTypeProposals` have been removed.
- The function `loadFullRefType` now performs checks for valid reference types.

Potential Problems:
- The removal of `checkValTypeProposals` and `checkRefTypeProposals` may cause issues down the line when loading values and references.
- The changes made to `loadFullValType` could potentially introduce new bugs due to the addition of the `TypeCode` parameter.
- It is unclear whether any other parts of the code rely on the removed helper functions, so their removal may cause unforeseen issues.

### Commit 10
The patch aims to make the Windows CI (Continuous Integration) happy by making some changes to the `lib/loader/ast/instruction.cpp` file. The changes consist of modifying the way the `Expect<void>` function is called, which is related to handling errors during the code execution. Specifically, the `loadFullValType()` function call was modified to use a variable `TypeRes` to store the result which is then used in the `setBlockType()` method call. Additionally, some formatting changes were made in the `loadInstruction()` method.

The potential problem with this patch is that it is not clear what issue the Windows CI was facing, and whether these changes are the correct solution or not. Moreover, it is unclear if other platforms might be affected by these changes, and if the proposed code modifications will cause any negative impacts elsewhere. Additionally, there should be some explanation of why this change is necessary, and how it impacts the overall system.

### Commit 11
The patch introduces changes in several files that add support for new types to the WasmEdge virtual machine. The most significant changes involve the introduction of heap types and the deprecation of the previous RefType and ValType definitions. The patch also includes changes to the enum_types.h header file and related files to implement these changes. 

It is difficult to identify potential problems without more context on the requirements/specifications of the system, but it is important to ensure that these changes are thoroughly tested and do not introduce any backwards compatibility issues. Additionally, it may be necessary to update any external interfaces or dependencies that rely on the previous ValType and RefType definitions.

### Commit 12
This patch changes a comment in the `FormChecker` class in `formchecker.cpp`. The comment is modified to correctly describe a check being performed and clarify the intent of the code. No potential problems are identified in this patch.

### Commit 13
This patch makes a very minor change of deleting one line of code in the file `include/common/enum_types.h` that relates to an enum called `UseRefType`. The purpose of this change is to make the formatting tool `clang-fmt` happy. 

One potential problem with this change is that the enum `UseRefType` is no longer included in the `WasmEdge_RefTypeCode` enum. If this enum is used somewhere in the code, it may cause an issue due to this change. However, since this is a minor change with limited impact, it's unlikely to cause any significant issues.

### Commit 14
Key changes:
- Added `derive_debug(true)` and `impl_debug(true)` to `bindgen` in `build.rs`.
- Removed `impl Debug for ffi::WasmEdge_Value` from `types.rs`

Potential problems:
- None apparent. These changes appear to be simple updates to the Rust bindings for the `wasmedge-sys` library, enabling generation of debug information for the `WasmEdge_Value` struct. The code builds successfully and there are no reported errors.

### Commit 15
The patch is for a Rust test file in the wasmedge-sys library. It fixes a compilation issue that was causing a panic on a certain condition. The change includes modifying a panic message in the code. There are no potential problems identified in the patch.

### Commit 16
Key changes:
- A new function (WasmEdge_GlobalTypeGetFullValType) has been added to the Rust inventory.

Potential problems:
- There is no description of what the new function does or why it is needed. This may make it difficult for other developers to understand its purpose and how it fits into the larger project.
- It is unclear if the new function has been thoroughly tested and whether it has any potential performance or stability issues.

### Commit 17
The patch adds a single line of code to the `wasmedge.cpp` file, which is a break statement inside a switch statement. The purpose is to make the linter happy.

One potential problem is that we don't see the context in which this code is being added, so it's hard to tell if this change could introduce any regressions or side-effects. It would be helpful to have more information about what the linter error was and why the `break` statement resolved the issue. Additionally, the code comments indicate the presence of a TODO which may or may not be related to this patch, so it's unclear what other changes will be needed to address any issues the linter found.

### Commit 18
Summary:
- The patch modifies the definition of classes and enumerations in several files.
- The RefType class is split into RefTypeExt and FullRefType classes.
- The FullValType class is modified to include support for RefTypeExt definition.
- The patch includes changes in both header files and implementation files.

Potential Problems:
- The changes to class and enumeration definitions may affect other parts of the codebase and cause conflicts.
- The changed code has not been tested, and the modifications may result in unexpected behavior. 
- The patch may introduce new bugs or regressions in the codebase.

### Commit 19
The patch adds a new command line option to enable the "Function Reference" proposal. It checks if a prerequisite proposal "Reference Type" is also enabled and prints an error message and exits if it is not. Otherwise, it adds the "Function Reference" proposal to the list of enabled proposals. The change file affected is `runtimeTool.cpp`.

A potential problem is that the error message printed in case of missing prerequisite proposal may not be visible unless the command is executed with appropriate options to print to the terminal.

### Commit 20
This patch adds a new `HeapTypeCode` enum and replaces the `WasmEdge_RefTypeExt` struct with the `WasmEdge_HeapType` struct, which contains a `HeapTypeCode` and a `DefinedTypeIdx`. It also changes the `RefTypeExt` class to `HeapType`. This seems to be a refactoring to split `RefType` into `RefTypeCode` and `HeapType`. The code changes seem reasonable overall, but there may be issues with backward compatibility or performance that should be considered.

### Commit 21
The patch adds a new command line option to enable the function reference proposal. The code looks good and there are no syntax errors. However, there is one potential problem with the code. If the function reference proposal is enabled while the reference type proposal is not enabled, the code will print an error message and exit. The patch does not clarify why this restriction exists or whether it is necessary. Nevertheless, it should be investigated further to ensure that everything works as intended.

### Commit 22
This patch adds support for loading the type defined in a function reference proposal. Specifically, it adds a new function `loadHeapType()` to `Loader` class and also modifies the `loadFullValType()` and `loadFullRefType()` functions to handle function references. 

The potential problem with this patch is that it relies on a proposal, which means that it might be unstable and subject to change. Also, the changes to the `loadFullValType()` and `loadFullRefType()` functions might affect the existing code that relies on these functions. Lastly, without seeing the entire codebase, it's hard to assess the impact of this patch on the application's performance or functionality.

### Commit 23
The patch introduces support for load function reference instructions. It adds two new opcodes, `Call_ref` and `Return_call_ref`, and modifies the `Instruction` class definition by replacing the `ReferenceType` data member with `HType` (a heap type) to represent function references. Additionally, the `Loader` class was modified to read and write these references from and to Wasm. Finally, a check was added to `runtimeTool.cpp` to detect if function references are being used.

One potential problem is that the check in `runtimeTool.cpp` only outputs an error message if `Proposal::FunctionReferences` is present, it does not actually prevent the loading of such code. Therefore, erroneous code may still be loaded and executed. This should be addressed.

### Commit 24
The patch adds a proposal check for the instruction loader in the AST for several OpCode types. Specifically, it adds a check for `OpCode::Return_call_ref`, `OpCode::Call_ref`, `OpCode::Ref__as_non_null`, `OpCode::Br_on_null`, and `OpCode::br_on_non_null`. If the `Conf` does not have the `Proposal::FunctionReferences` proposal, an error message will be logged. This change intends to ensure that the Function References Proposal is respected for these OpCode types.

Based on the patch, the potential problems may arise if the Function References Proposal is not implemented properly, or there is a problem with the `Conf` configuration. Additionally, this patch is dependent on the context of the codebase, so a review of the entire project source code will be necessary to determine if it fits into the codebase seamlessly.

### Commit 25
The patch introduces support for loading table extended from function reference. It creates a new Table class containing a TableType and an InitExpr. The TableSection class is updated to use the new Table class. The Loader class now has a loadType() function for tables. The Validator class also has a new validate() function for validating tables. The TableInstance class is updated to support the new Table class.

One potential problem is the need to check whether the ref type of table type is nullable. Another problem is the lack of initialization of the table instance by the InitExpr. 

Furthermore, the error handling in testReadByte() function is incomplete. If it fails, it returns an Unexpect(Status) instead of logging an error. 

Finally, more context would be helpful to understand the changes fully, such as what kind of code the changed files are related to and whether there are any unit tests included with the patch.

### Commit 26
This GitHub patch removes a C++ API function named "WasmEdge_GlobalTypeGetFullValType" for "global type" operation in the header file "wasmedge.h" and implementation file "wasmedge.cpp". The change is made by the author named "Little Willy" and the reason for the change is not mentioned. 

This change may cause problems for consumers who are using the removed function. Since the function is removed without any note regarding its impact on the rest of the codebase, it might lead to a build or runtime error(if someone was using this function). Therefore, the author should provide additional information on why this function is being removed and how it affects the overall codebase.

### Commit 27
The patch adds support for loading a reference as non-null to the Loader class in instruction.cpp. The only change is the addition of a new case in the switch statement for OpCode::Ref__as_non_null. The patch does not appear to have any potential problems. However, without seeing the entire codebase or more context it is difficult to be certain.

### Commit 28
This patch fixes loading predefined heap type in the `Loader::loadHeapType()` function. The `-` sign is removed from `HTypeCode` initialization and instead, it is set to `0x80 + *Res`. Additionally, an error check is added to ensure that `*-Res` is not greater than or equal to 0x80. 

The potential problem with this patch is related to backward compatibility. The change increases the potential for compatibility issues since it modifies the heap type loading mechanism. Furthermore, it's unclear from the patch if any tests have been created to prevent regressions. Additionally, it's unclear if this change could have unwanted side effects elsewhere in the codebase.

### Commit 29
The patch simply adds an `End` instruction at the end of a default table value. This is a very small change and does not appear to have any potential problems.

### Commit 30
This patch fixes a typo in the enum definition `br_on_non_null` to `Br_on_non_null`. It also changes its usage in `instruction.cpp` to the correct casing. This change seems straightforward and valid.

The potential problem that could arise is if there are other parts of the codebase that rely on the previous typo `br_on_non_null`. However, since the typo is clearly a mistake, it is unlikely that this is the case. Additionally, this change only affects the names of the enum and the code that uses it, so it's unlikely that there will be any negative effects on the software's functionality.

### Commit 31
The patch adds a new structure called `FullValType` that is used to replace `ValType` in the `WasmEdge_Value` struct. The change seems to affect many files, with a lot of `ValType` being replaced by `FullValType`. The change may extend the codebase's functionalities or improve its performance. However, there are also several changes to existing code that may cause potential bugs or issues. 

Firstly, a rename of `ValType` to `FullValType` may cause build errors since it affects many files, and not all files in the repository may have been updated with the correct changes. Secondly, the definition of the new `FullValType` structure is not included in the patch, so it's challenging to tell how it's being implemented or how it benefits the codebase. Finally, many of the previously existing functions and methods seem to be impacted by the change, with new methods being added and old ones being modified. This suggests that significant refactoring may have been carried out to accommodate the new `FullValType`. Consequently, more rigorous testing and regression testing should be performed to ensure that no bugs have been introduced.

### Commit 32
This patch introduces a new command line option ("enable-function-reference") that enables the Function Reference proposal. The Function Reference proposal is now being added to the list of possible configurations that the user can enable. 

A potential problem is identified in the code. Enabling the Function References proposal is based on enabling the Reference Type proposal which is checked by the "PropRefTypes" option. If "PropRefTypes" is not enabled, then the code will return an error. However, the "PropRefTypes" option is not being defined anywhere in this patch. Therefore, it is unclear what the behavior would be if the user tries to enable the Function Reference proposal without the Reference Type proposal.

Another potential problem is that there is no description provided for what the Function Reference proposal does. Without this explanation, it may be confusing for users to know what they are enabling by using this command line option. 

Finally, it would be good to add some testing to ensure that enabling the Function Reference proposal works as expected in various scenarios.

### Commit 33
This patch introduces a new class called `HeapType` and refactors existing code to use the new class instead of the old `RefType` class. It also replaces a couple of magic numbers with `HeapTypeCode`s. 

The most important finding is that the code that loads `HeapType`s may return an error, but the error case isn't handled properly in a couple of places in `loadValType` and `loadRefType`. If an error occurs while loading a `HeapType`, the caller will receive a malformed `ValType` or `RefType`. This could cause problems down the line. 

Furthermore, it's unclear what problem this patch is solving. There's no explanation or justification for why the new `HeapType` class is better than the old `RefType` class, and it's not clear from the code itself what the benefits are.

### Commit 34
The patch adds a new constructor to the `HeapType` class and a new member function `getHeapType()` to the `RefType` class. 

One potential problem is that there is no explanation for why this change is necessary. Additionally, without more context about the project and the classes involved, it's difficult to determine if these changes will cause any issues or if they are the best solution to a given problem.

### Commit 35
This patch fixes an assertion bug in the `HeapType` class constructor by changing the argument of one of the `assuming` statements. Specifically, the condition `(Inner.Data.HTCode == HeapTypeCode::TypeIndex)` has been changed to `(Inner.Data.HTCode != HeapTypeCode::TypeIndex)`. The change ensures that the code inside the constructor is executed correctly.

Potential problems with this patch could arise if the change in the `assuming` statement alters the behavior of the `HeapType` class in some unexpected way. Additionally, there could be other parts of the codebase that rely on the previous behavior of this constructor, which means this patch could cause a regression. Therefore, it would be a good idea to review the test suite that covers this class to see if any relevant tests might have been affected.

include/ast/type.h
```diff
@@ -173,6 +174,19 @@ class TableType {
   /// @}
 };
 
+class Table {
```
q82419, line 177: For this object which contains the expression and type, you should add it in the `Segment`, such as the `GlobalSegment` and `ElementSegment`.

q82419 (changes requested):
I think the design of the AST should like this:

The section of `Table` : vector of `TableSegment`.
The content of `TableSegment`: `TableType` + `Expression` (can be empty).
The `Expression` of `TableSegment` provide the initialization of the table instance when instantiation.
With this design, the `TableInstance` will not necessary to be changed and capable with the old definitions.

little-willy:
@q82419 I have addressed the comments and turn the original `Table` into `TableSegment`. PTAL.

q82419 (approved):

PR #2152, opened 2022-12-05
feat: add support for function reference proposal for validation.

little-willy:
After #2151

PR #2153, opened 2022-12-05
refactor: unify all reference value to RefVariant

little-willy:
After #2152.

### DIFF
--- a/include/common/types.h
+++ b/include/common/types.h
@@ -315,42 +315,31 @@ private:
 
 // >>>>>>>> Value definitions >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
 
-/// UnknownRef definition.
-struct UnknownRef {
-  uint64_t Value = 0;
-  UnknownRef() = default;
-};
-
 /// FuncRef definition.
 namespace Runtime::Instance {
 class FunctionInstance;
 }
-struct FuncRef {
-#if __INTPTR_WIDTH__ == 32
-  const uint32_t Padding = -1;
-#endif
-  const Runtime::Instance::FunctionInstance *Ptr = nullptr;
-  FuncRef() = default;
-  FuncRef(const Runtime::Instance::FunctionInstance *P) : Ptr(P) {}
-};
 
-/// ExternRef definition.
-struct ExternRef {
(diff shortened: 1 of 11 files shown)


q82419:
@little-willy  can you help to update?

q82419 (changes requested):
The tests from clang compiler failed.
Maybe you used something which are undefined behavior.

little-willy:
> The tests from clang compiler failed.
Maybe you used something which are undefined behavior.

@q82419 While you are reviewing my PR, did you get any idea on where is the undefined behavior?

q82419:
Maybe you can use the sentilizer (#2527) to test.

little-willy:
@q82419 Did you recently upgraded the clang version from 12 to 15 in the CI? 

I just tried running the test that failed in this PR on the master branch, and it also fails. It is run on the image `wasmedge/wasmedge:latest`, which currently uses clang-15. If I ran it with clang-12, it will not fail. Besides, from logs of the CI of the previous PR, I noticed it was ran with clang-12. So I wonder whether the failed CI is caused by the upgrade from clang 12 to 15 in the CI image?

little-willy:
@q82419 I just tried resetting the current PR to be the same as the `proposal/func-ref` and triggered the CI. The CI failed in the same way. So I think the current PR may not have any problem. The difference of the current CI and the CI of the previous PR is the difference clang version.

Could you rebase the `proposal/func-ref` to the latest master and see whether it can pass the CI?

little-willy:
@q82419 I just tried rebase on the current PR to the latest proposal/func-ref, but the wasi tests still failed in the CI. I noticed that you did not rebase the the proposal/func-ref on the latest master branch, but on a commit `[WASI] Fix WASI ci test` which is about 3 weeks ago. Good news is that the CI that failed previously get passed now.

Could you try rebasing more commits to see whether we can pass the CI?

little-willy:
> But after your rebase, there's no changes?

I temporarily reset the PR to the current `proposal/func-ref` to trigger the CI on it to verify whether the CI can pass. I will recover the change back.

little-willy:
@q82419 I have updated the PR. Only the `wasiTests` failed, which is the same failure as before this PR.

q82419 (approved):

PR #2157, opened 2022-12-10
[Docs] fix various typo

RRobot-lm:
Signed-off-by: RRobot-lm <RRobot-lm@users.noreply.github.com>


RRobot-lm:
Perhaps we can set up a linter like codespell or misspell to catch the typo automatically?

hydai (approved):
LGTM

hydai:
> Perhaps we can set up a linter like codespell or misspell to catch the typo automatically?

Sounds good. Do you have any recommendations for these two tools? Or any other choices?

RRobot-lm:
I would suggest using codespell. Although it has some false positives, it can be suppressed via `--ignore-words` file. The ability of misspell is limited.

Anyway, both tools have the option to fix typos automatically.

PR #2158, opened 2022-12-12
feat: support execution of function references

little-willy:


little-willy:
@q82419 I have rebased the PR. You can take a look.

BTW, I see that you will download the test suites of some spec proposal. How can I add the test suites of function reference to the test?

q82419:
@little-willy 
Please use `wasm-dev` branch [here](https://github.com/WasmEdge/WasmEdge/blob/master/test/spec/CMakeLists.txt#L10).
And you should add the `function-references` [here](https://github.com/WasmEdge/WasmEdge/blob/master/test/spec/spectest.cpp#L220-L224).

little-willy:
@q82419 It seems that the pre-built wasm file in the downloaded test suite is inconsistent to the wasm built from the interpreter in the function reference spec repo.

For example, for test `br_table/br_table.0.wasm`, the spec interpreter fails to validate, while the spec interpreter succeeds to validate the wat.

When further looking into it, for wasm compiled from wat `(table $t (ref null $t) (elem $tf))`, the element case code is 2 in `br_table/br_table.0.wasm`, while it is 6 in the wasm built from the interpreter.

q82419:
Merge for rebasing.

PR #2167, opened 2022-12-19
[Rust] Refactor call method

apepkuss:
In this PR, the following changes are introduced:

- Refactored the signature of `call` methods of `Func` and `FuncRef` structs, respectively
- Updated the `rust-standalone` workflow
- Updated the `bindings-rust` workflow to support Rust `1.64`, `1.65`, and `1.66`


hydai (approved):

apepkuss:
@hydai Thanks for the review!

PR #2170, opened 2022-12-20
[CI] Add the static library artifact

q82419:


CMakeLists.txt
```diff
@@ -91,8 +91,10 @@ if(WASMEDGE_BUILD_TOOLS)
     # Static tools will link LLVM statically.
     set(WASMEDGE_LINK_LLVM_STATIC ON)
     # Tools will forcefully turn on the static library building.
-    set(WASMEDGE_LINK_TOOLS_STATIC ON)
```
hydai, line 94: Don't we need to set the tools to static now?

cmake/Helper.cmake
```diff
@@ -199,7 +199,6 @@ if((WASMEDGE_LINK_LLVM_STATIC OR WASMEDGE_BUILD_STATIC_LIB) AND WASMEDGE_BUILD_A
     else()
       # If not build static lib, dynamic link libz and libtinfo.
       list(APPEND WASMEDGE_LLVM_LINK_SHARED_COMPONENTS
-        rt
```
hydai, line 202: Is this no longer needed?

hydai (changes requested):

CMakeLists.txt
q82419: This was a mistake typo. The modified one is correct.

cmake/Helper.cmake
q82419: The `rt` is added above in the previous `list` command.

hydai (approved):

PR #1823, opened 2022-09-06
[Misc] Fix fdReaddir on large numbers of entries

LFsWang:
Fix #1711

The definition of arguments for `fdReaddir` are too ambiguous, I am not sure the excepted behavior.

some discussion: https://github.com/WebAssembly/wasi-filesystem/issues/3
definition: https://docs.rs/wasi/0.9.0+wasi-snapshot-preview1/wasi/fn.fd_readdir.html

Signed-off-by: LFsWang <tnst92002@gmail.com>

PR #2182, opened 2022-12-23
Fix fdReaddir on large numbers of entries

harry900831:
This is based on #1823.

According to the [definition of fdReaddir](https://docs.rs/wasi/0.9.0+wasi-snapshot-preview1/wasi/fn.fd_readdir.html), the return value is equal to the size of the read buffer whenever the buffer size is not enough, so it may need additional `fdReaddir` to get the rest of the files, potentially truncating the last directory entry. 

If `fdReaddir` is returned due to not enough size of buffer, then we may not erase the last value on `Dir.buffer` so that it can be copy to the buffer by the next `fdReaddir` call. Also when `Buffer.size` is less than the `Dir.buffer.size()`, the original implementation may copy an incomplete data to the `Buffer` then erase the whole `Dir.buffer` which may leads to crashing.

@LFsWang @hydai Please review the change, thanks. 

I'm working on adding the unit test to this situation.

### DIFF
--- a/include/host/wasi/inode.h
+++ b/include/host/wasi/inode.h
@@ -107,8 +107,9 @@ struct DirHolder {
     Dir = NewDir;
   }
 
+  const static uint64_t DIRCOOKIE_START = 0;
   DIR *Dir = nullptr;
-  uint64_t Cookie = 0;
+  uint64_t Cookie = DIRCOOKIE_START;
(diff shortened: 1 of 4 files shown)


harry900831:
Ooops...
I spot that on linux, this may output duplicate file. I believe this is the reason so that the CI is failed by failing wasiTest... Let me check it.

Issue #2123, opened 2022-11-28
WASM WASI has wrong behavior about file system on Mac OS


CaptainVincent:
## Description

Use WalkDir visit folder on Mac has wrong behavior.

Ex.
Run [lsr-wasm-docker](https://github.com/assambar/lsr-wasm-docker) test a folder with large amount files.

### Current State
1. lost files
2. ~~show some files but actually not exist~~
3. ~~garbled name~~

### Expected
Walk all folders and subfolders normally.

## Environment
- Wasmedge
install Wasmedge (version 0.11.2) by installer script
(curl -sSf https://raw.githubusercontent.com/WasmEdge/WasmEdge/master/utils/install.sh | bash)
- Rust
  - cargo 1.68.2 (6feb7c9cf 2023-03-26)
  - rustc 1.68.2 (9eb3afe9e 2023-03-27)
  - rustup 1.25.2 (fae52a197 2023-02-01)

> Use the same version of Wasmer and Rust toolchain  both platforms

- Hardware Architecture: (e.g. x86\_64 or aarch64)
Macbook pro 2019

- Operating system: (e.g. Ubuntu 20.04 or CentOS 7.6)
macOS Monterey 12.5.1

## Steps to Reproduce

```bash
// Install wasmedge and rust
curl -sSf https://raw.githubusercontent.com/WasmEdge/WasmEdge/master/utils/install.sh | bash
source "$HOME/.wasmedge/env"
curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
source "$HOME/.cargo/env"
rustup default stable
rustup target add wasm32-wasi

// build test suite
git clone https://github.com/assambar/lsr-wasm-docker
cd lsr-wasm-docker
cargo build --release --target=wasm32-wasi
export WALKDIR=$(realpath target/wasm32-wasi/release/lsr.wasm)

// run inside lsr-wasm-docker folder
wasmedge --dir /:$(pwd) $WALKDIR
```

Mac log
```
Starting from root_dir "/"
/
/Cargo.toml
/LICENSE
```

Ubuntu log
```
Starting from root_dir "/"
/
/Makefile
/README.md
/LICENSE
/target
/target/release
/target/release/deps
/target/release/examples
/target/release/build
/target/release/.cargo-lock
/target/release/.fingerprint
/target/release/incremental
/target/CACHEDIR.TAG
/target/.rustc_info.json
/target/wasm32-wasi
[...]

PR #2194, opened 2022-12-27
[MISC] Fix fd_readdir setting incorrect d_next on macOS

harry900831:
According to the man page, the d_seekoff for a dirent on macOS is optional, so it's value is always zero for this case. To fix this, we use telldir to get the offset.

Signed-off-by: Harry Chiang <harry900831@gmail.com>


harry900831:
I've tested it locally and this may fix #2123. But the implementation for fd_readdir is still under discussion. So I'm not sure whether we should close #2123.

hydai (approved):
LGTM. @ibmibmibm @q82419  PTAL

PR #2195, opened 2022-12-27
[Rust] Fix issue 2189

### LINKED ISSUE #2189: WASM file with no file extension should be supported
## Motivation

In some use cases we couldn't guarantee user will give us a file name include extension.

## Details

Maybe we should not use file extension as rule for recognize target platform.
Or we just don't blocking run file with filename no extension.

apepkuss:
This PR is to fix #2189.


hydai (commented):
@CaptainVincent PTAL

hydai (approved):

apepkuss:
@hydai Thanks for the review!

PR #2200, opened 2022-12-29
[Rust] Support for WasmEdge static library

apepkuss:
In this PR, the build script provides support for the bindings to WasmEdge static library.


.github/workflows/rust-static-lib.yml
```diff
@@ -0,0 +1,73 @@
+name: binding-rust-static-lib
```
hydai, line 1: Hi @apepkuss 
IMO, we should collect `rust*` workflows into one file. We can reuse many similar parts instead of creating more and more rust-related workflow files, and this will increase the maintenance cost.

hydai (changes requested):

.github/workflows/rust-static-lib.yml
apepkuss: This new workflow can be seen as a temp workflow as it could be changed (maybe a lot). Therefore, use a separate workflow for now. But, I agree with the point to move all `rust*` workflows into one file. I believe creating another PR to do this job would be better. Thanks!

hydai (approved):

apepkuss:
@hydai Thanks for the review!

apepkuss:
@hydai Thanks for the review! Happy New Year!

PR #2215, opened 2023-01-09
[WASI-NN]: add Tensorflow backend

gusye1234:
- [ ] implment wasi-nn interface for Tensorflow
- [ ] Demos
- [ ] CI
- [ ] tests


plugins/wasi_nn/wasinnfunc.cpp
```diff
@@ -44,6 +59,52 @@ namespace {
   return DeviceName;
 }
 
+#ifdef WASMEDGE_PLUGIN_WASI_NN_BACKEND_TF
+tensorflow::DataType TensorflowTypeMap(const WASINN::TensorType InType) {
+  uint32_t TFInType;
+  switch (InType) {
+  case WASINN::TensorType::F32:
+    TFInType = 1; // TF_FLOAT
```
q82419, line 67: I prefer to use `return static_cast<tensorflow::DataType>(TF_DataType::TF_FLOAT)` here.

plugins/wasi_nn/wasinnfunc.cpp
```diff
@@ -44,6 +59,52 @@ namespace {
...
+    break;
+  case WASINN::TensorType::F16:
+    TFInType = 19; // TF_HALF
```
q82419, line 70: As above.

plugins/wasi_nn/wasinnfunc.cpp
```diff
@@ -44,6 +59,52 @@ namespace {
...
+    break;
+  case WASINN::TensorType::U8:
+    TFInType = 4; // TF_UINT8
```
q82419, line 73: As above.

plugins/wasi_nn/wasinnfunc.cpp
```diff
@@ -44,6 +59,52 @@ namespace {
...
+    break;
+  case WASINN::TensorType::I32:
+    TFInType = 3; // TF_INT32
```
q82419, line 76: As above.

plugins/wasi_nn/wasinnfunc.cpp
```diff
@@ -44,6 +59,52 @@ namespace {
...
+    break;
+  }
+  return static_cast<tensorflow::DataType>(TFInType);
```
q82419, line 79: For the other default unsupported cases, please handle and return error.

plugins/wasi_nn/wasinnenv.h
```diff
@@ -90,6 +109,14 @@ class Graph {
 #ifdef WASMEDGE_PLUGIN_WASI_NN_BACKEND_TORCH
   torch::jit::Module TorchModel;
 #endif
+#ifdef WASMEDGE_PLUGIN_WASI_NN_BACKEND_TF
+  TF_Buffer *TFBuffer = nullptr;
+  TF_Graph *TFGraph = nullptr;
+
+  std::filesystem::path TFSavedModelPath;
```
q82419, line 116: Is it possible to set load the saved model in `wasi_nn::load` function and delete the temp dir, so that the saved model path is no longer to be saved here?

q82419 (changes requested):

plugins/wasi_nn/wasinnenv.h
gusye1234: I plan to place the deletetion of temp dir in the graph class's deconstruction function.

PR #2246, opened 2023-02-01
[Rust] Implement Clone trait for external instances

apepkuss:
In this PR, the `Clone` trait is implemented for `Func`, `Memory`, `Global` and `Table` external instances.

### DIFF
--- a/bindings/rust/wasmedge-sys/src/instance/memory.rs
+++ b/bindings/rust/wasmedge-sys/src/instance/memory.rs
@@ -242,6 +242,14 @@ impl Drop for Memory {
         }
     }
 }
+impl Clone for Memory {
+    fn clone(&self) -> Self {
+        Self {
+            inner: self.inner.clone(),
+            registered: self.registered,
+        }
+    }
+}
 
 #[derive(Debug)]
 pub(crate) struct InnerMemory(pub(crate) *mut ffi::WasmEdge_MemoryInstanceContext);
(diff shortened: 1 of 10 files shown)


apepkuss:

bindings/rust/wasmedge-sys/src/instance/table.rs
```diff
@@ -153,10 +154,16 @@ impl Table {
...
-            }
+        if !self.registered && Arc::strong_count(&self.inner) == 1 && !self.inner.0.is_null() {
+            unsafe { ffi::WasmEdge_TableInstanceDelete(self.inner.0) };
+        }
+    }
+}
+impl Clone for Table {
+    fn clone(&self) -> Self {
+        Table {
+            inner: self.inner.clone(),
```
hydai, line 165: It looks like the `clone` will only copy the pointer. Is this a expected behavior?

bindings/rust/wasmedge-sys/src/instance/memory.rs
```diff
@@ -248,11 +237,19 @@ impl Memory {
...
-        if !self.registered && !self.inner.0.is_null() {
+        if !self.registered && Arc::strong_count(&self.inner) == 1 && !self.inner.0.is_null() {
             unsafe { ffi::WasmEdge_MemoryInstanceDelete(self.inner.0) };
         }
     }
 }
+impl Clone for Memory {
+    fn clone(&self) -> Self {
+        Self {
+            inner: self.inner.clone(),
```
hydai, line 248: I am not familiar with this one(`Arc`). But is it fine to share the pointer(C-level memory pointer) between multiple duplicated memory? Will it cause any data lock issue?

hydai (changes requested):

bindings/rust/wasmedge-sys/src/instance/table.rs
apepkuss: Actually, the `clone` operation does not copy the pointer. The type of `self.inner` is `Arc<InnerTable>`, and `Arc<T>` is a reference type and uses an inner reference counter to record how many references it has at some time. The `clone` method of `Arc<T>` just increments the counter, but does not copy the low-level raw pointer. Thanks!

hydai: Good to know. Thanks.

hydai (approved):

bindings/rust/wasmedge-sys/src/instance/memory.rs
apepkuss: There is no data lock issue, as `Arc` is just an atomic reference counter and has no locking mechanism. Thanks a lot!

apepkuss:
@hydai Thanks for the review!

PR #2267, opened 2023-02-09
[WASI] Set filestat of special fd to 0

harry900831:
In order to pass this [test](https://github.com/WebAssembly/wasi-testsuite/blob/prod/testsuite-base/tests/rust/src/bin/fd_filestat_get.rs).

### DIFF
--- a/lib/host/wasi/inode-linux.cpp
+++ b/lib/host/wasi/inode-linux.cpp
@@ -273,14 +273,16 @@ INode::fdFilestatGet(__wasi_filestat_t &Filestat) const noexcept {
     return WasiUnexpect(Res);
   }
 
-  Filestat.dev = Stat->st_dev;
-  Filestat.ino = Stat->st_ino;
+  // Zeroing out these values to prevent leaking information about the host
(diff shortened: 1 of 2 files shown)


lib/host/wasi/inode-linux.cpp
```diff
@@ -273,14 +273,14 @@ INode::fdFilestatGet(__wasi_filestat_t &Filestat) const noexcept {
     return WasiUnexpect(Res);
   }
 
-  Filestat.dev = Stat->st_dev;
-  Filestat.ino = Stat->st_ino;
+  Filestat.dev = isSpecialFd(Fd) ? 0 : Stat->st_dev;
```
hydai, line 276: We should have a comment section explaining why the special fds need this handling way.

hydai (changes requested):

hydai:
To have a better way of tracking the status of the wasi test suite, we should have a standalone issue talking about this and link it to the corresponding PR.

harry900831:
Close #2269

hydai (approved):

Issue #2224, opened 2023-01-17
[Misc] Tests can not be built on clang 15.0.1 (Win MSVC 17.4.4)


LFsWang:
## Description

* Build with test on Visual Studio 2022 Developer PowerShell v17.4.4
```
$Env:CC = "clang-cl"
$Env:CXX = "clang-cl"
 
cmake -Bbuild -GNinja -DCMAKE_SYSTEM_VERSION="10.0.22000.0" -DCMAKE_MSVC_RUNTIME_LIBRARY=MultiThreadedDLL "-DLLVM_DIR=$llvm_dir" -DWASMEDGE_BUILD_TESTS=ON -DWASMEDGE_BUILD_PACKAGE="ZIP" .
 
cmake --build build
```

note
```
clang-cl --version

clang version 15.0.1
Target: x86_64-pc-windows-msvc
Thread model: posix
InstalledDir: C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\Llvm\x64\bin

```
#### gtest
 
```
[...]

PR #2278, opened 2023-02-10
[Windows] Fix building tests on Windows

abhinandanudupa:
- Add compiler option to ignore out-of-line static constexpr data member declarations
- Initialise un-initialised variables

Resolves: #2224


q82419 (approved):
LGTM!

PR #2286, opened 2023-02-14
[RISCV64] Support WasmEdge on RISC-V64 arch.

O3Ol:
# Motivation

This pull request is to add files which make WasmEdge build for `RISC-V 64` arch.

# Details

- Add the definitions of riscv64.
- Add `build_for_riscv.yml` for CI test.
- Add the doc `riscv64.md` for buiding WasmEdge on RISC-V64.


hydai:
Hi @O3Ol 
Please fix the failing CI. Thanks.

lib/aot/compiler.cpp
```diff
@@ -5146,6 +5149,9 @@ Expect<void> Compiler::compile(Span<const Byte> Data, const AST::Module &Module,
     if (!Conf.getCompilerConfigure().isGenericBinary()) {
       CPUName = llvm::sys::getHostCPUName();
     }
+#if defined(__riscv) && __riscv_xlen == 64
+    CPUName = "generic-rv64";
```
hydai, line 5153: Just want to know if `llvm::sys::getHostCPUName()` will return `generic-rv64` or not.
I think this should be moved to line 5145. And you can create a constant string `"generic"` or `"generic-rv64`, depends on the platform.

.github/workflows/build_for_riscv.yml
```diff
@@ -0,0 +1,77 @@
...
+    needs: lint
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
+    - uses: uraimo/run-on-arch-action@v2
+      name: Build WasmEdge
+      id: build
+      with:
+        arch: none
```
hydai, line 58: Do we need to specify the `rv64` here?

hydai (changes requested):

lib/aot/compiler.cpp
O3Ol: In my environment `llvm::sys::getHostCPUName()` did not return `generic-rv64`, so I defined it. I would move it to line 5145 as you said.

.github/workflows/build_for_riscv.yml
O3Ol: No, I will remove unnecessary content.

hydai (approved):

hydai:
@O3Ol Thanks for your contribution!

PR #2290, opened 2023-02-15
[WASI] Fix environ_get and args_get

### LINKED ISSUE #2270: Unable to get the environment variable with special character
## Description
WasmEdge failed on this [test](https://github.com/WebAssembly/wasi-testsuite/blob/44a77decc39a0dec4908b4fa1b55f2c7d8784ca7/tests/assemblyscript/testsuite/environ_get-multiple-variables.ts), which is getting the environment variable with `a=text`, `b=escap " ing` and `c=new\nline`.

harry900831:
Fix #2270.

The problem isn't come from the special characters. 

There's no need to add additional 1 to  `ArgvSize` and `EnvSize`. They should be same as what `args_sizes_get` and `environ_sizes_get` return.


lib/host/wasi/wasifunc.cpp
```diff
@@ -33,10 +33,18 @@ inline __wasi_size_t calculateBufferSize(const Container &Array) noexcept {
...
-template <> struct WasiRawType<uint64_t> { using Type = uint64_t; };
+template <> struct WasiRawType<uint8_t> {
+  using Type = uint8_t;
+};
+template <> struct WasiRawType<uint16_t> {
+  using Type = uint16_t;
+};
+template <> struct WasiRawType<uint32_t> {
+  using Type = uint32_t;
+};
```
harry900831, line 44: It seems that this part doesn't pass the CI. But according to my own clang-format this may be formatted like this.

My clang-format version is 15.0.7.

hydai: I will check this tmr.

hydai: This will be fixed by #2299

hydai:
It looks like the CI failed. Could you please check this? @harry900831

hydai:
Hi @harry900831 
Could you please check if this is still available or not?

harry900831:
This has been resolved by #2500.

PR #2291, opened 2023-02-15
[MISC] Set exit code of execution failed to 128 + SIGABRT

### LINKED ISSUE #2268: feat: semantic exit codes for traps in wasm code execution
## Description

Status for [abort.wasm](https://github.com/caspervonb/wasi-test-suite/blob/main/libc/abort.c) and [assert-false.wasm](https://github.com/caspervonb/wasi-test-suite/blob/main/libc/assert-false.c) should be 134, status for wasmedge running abort.wasm is 1.

According to the wat, It would call the function that run into a trap. 
```
  (func $abort (type $t1)
    (unreachable)
    (unreachable))
```

And I check the [implementation of wasmtime](https://github.com/bytecodealliance/wasmtime/blob/43022c862a1d497f4b8c7c2fce47148775503add/src/commands/run.rs#L208-L251), it seems that it return 134 whenever encountering a trap.

So changing the following three parts to return 128 + SIGABRT is a work around:

- https://github.com/WasmEdge/WasmEdge/blob/4f1059b4aa934c4de3c3edff4055c28e6f0b9311/lib/driver/runtimeTool.cpp#L295
- https://github.com/WasmEdge/WasmEdge/blob/4f1059b4aa934c4de3c3edff4055c28e6f0b9311/lib/driver/runtimeTool.cpp#L411
- https://github.com/WasmEdge/WasmEdge/blob/4f1059b4aa934c4de3c3edff4055c28e6f0b9311/lib/driver/runtimeTool.cpp#L328

harry900831:
Resolve #2268.


q82419 (dismissed):

harry900831:
I've updated the commit message and add a comment above the line I change.

hydai (approved):

Issue #1790, opened 2022-08-24
[CI] Reduce the building time of OpenWrt


hydai:
## Motivation

The GitHub runners have only 20 slots. However, we found that the workflow of building WasmEdge on the OpenWrt will take almost two hours, which is totally insane. This occupies the most time in our whole workflow. The reason is that the workflow builds not only WasmEdge itself but also the whole OpenWrt. Since the base commit is exactly the same in each build, we should figure out a way to create a cache or something to reduce the total building time.

## Appendix

The workflow file for OpenWrt: https://github.com/WasmEdge/WasmEdge/blob/master/.github/workflows/build_for_openwrt.yml
An example run: https://github.com/WasmEdge/WasmEdge/actions/runs/2915634679

Sush-Dew:
@hydai Could you assign this issue to me? I am a beginner and this issue seems a good first issue to resolve.

RahulSurwade08:
Hey, I would love to work on this issue. How can i get started?

hydai:
> Hey, I would love to work on this issue. How can i get started?

Hi @RahulSurwade08,
You can refer to this workflow: https://github.com/WasmEdge/WasmEdge/blob/master/.github/workflows/build_for_openwrt.yml
And then try to figure out which part can be done inside a docker image, so we can use that image instead of building the whole openwrt again and again to reduce the building time.

hydai:
> @hydai Could you assign this issue to me? I am a beginner and this issue seems a good first issue to resolve.

Hi @Sush-Dew 
Are you still on this? If not, I would like to transfer the assignee to @RahulSurwade08

Sush-Dew:
Hi! @hydai I am working on this issue. please let me do it.

dracoooooo:
Hi @hydai, can you assign this to me? I would like to use [cache action](https://github.com/actions/cache) to save OpenWrt artifact to reduce the building time.

PR #2309, opened 2023-02-24
[CI] Use cache to reduce the building time of OpenWrt

dracoooooo:
#1790
This update will make the workflow cache the first built OpenWrt on each branch and use it for subsequent builds.
Screenshot_20230224_125901


hydai (approved):
LGTM. Thanks!

Issue #2249, opened 2023-02-02
Tests that fail on several wasi testsuite


harry900831:
## Description

I've test several wasi test suites, and there're several issues with the wasi implementation in WasmEdge.

For the [caspervonb/wasi-test-suite](https://github.com/caspervonb/wasi-test-suite.git):
- #2268
- The error number for [fopen-directory-parent-directory.wasm](https://github.com/caspervonb/wasi-test-suite/blob/main/libc/fopen-directory-parent-directory.c) should be ENOTCAPABLE, error message for running wasmedge is NO SUCH FILE OR DIRECTORY. This may not be a problem currently since which error it should be is not strictly defined, and we can see that in WebAssembly/wasi-testsuite, it sometimes accepts multiple error numbers on same behavior.

For [WebAssembly/wasi-testsuite](https://github.com/WebAssembly/wasi-testsuite/tree/prod/testsuite-base):
- #2269
- #2270
- #2271
- #2272
- #2292 
- #2294
- #2295
- Maybe we could consider adding this test into our CI since it seems to be an official test.
    - #2328

This is the discussion for several issues: https://github.com/WebAssembly/WASI/issues/514

## Environment
- Hardware Architecture: M1 2020
- Operating system: macOS 13.1

## Steps to Reproduce

I can provide my test script if someone needs it, since all the above testsuites don't contain script for WasmEdge.

harry900831:
Update for the problem of exit code for  `abort.wasm` and `assert-false.wasm`. 

According to the wat, It would call the function that run into a trap. 
```
  (func $abort (type $t1)
    (unreachable)
    (unreachable))
```

And I check the [implementation of wasmtime](https://github.com/bytecodealliance/wasmtime/blob/43022c862a1d497f4b8c7c2fce47148775503add/src/commands/run.rs#L208-L251), it seems that it return 134 whenever encountering a trap.

So it seems that we can change the following three parts to return 128 + SIGABRT:


- https://github.com/WasmEdge/WasmEdge/blob/4f1059b4aa934c4de3c3edff4055c28e6f0b9311/lib/driver/runtimeTool.cpp#L295
- https://github.com/WasmEdge/WasmEdge/blob/4f1059b4aa934c4de3c3edff4055c28e6f0b9311/lib/driver/runtimeTool.cpp#L411
[...]

PR #2328, opened 2023-03-01
Add wasi-testsuite to CI

harry900831:
This should be merge after fixing all the issues inside #2249.

I'm new to GitHub Action. If there's any suggestion, please let me know.

Thanks!

### DIFF
--- a/.github/workflows/wasi-testsuite.yml
+++ b/.github/workflows/wasi-testsuite.yml
@@ -1,0 +1,74 @@
+name: Test Wasi Testsuite
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
+  cancel-in-progress: true
+
+on:
+  push:
+    paths:
+      - ".github/workflows/wasi-testsuite.yml"
+      - "lib/host/wasi/**"
+      - "include/host/wasi/**"
+      - "thirdparty/wasi/**"
+  pull_request:
+    paths:
+      - ".github/workflows/wasi-testsuite.yml"
+      - "lib/host/wasi/**"
+      - "include/host/wasi/**"
+      - "thirdparty/wasi/**"
+
+jobs:
+  test:
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest]
+    runs-on: ${{ matrix.os }}
+    steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/wasi-testsuite.yml
```diff
@@ -0,0 +1,102 @@
...
+  test_on_macos:
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          path: WasmEdge
+
+      - name: Install dependencies for WasmEdge
+        run: |
+          brew install llvm@14 ninja boost cmake
```
hydai, line 71: Ditto. You may also try `llvm` instead of `llvm@14`.

hydai (changes requested):

hydai (dismissed):

.github/workflows/wasi-testsuite.yml
```diff
@@ -0,0 +1,102 @@
...
+      - "thirdparty/wasi/**"
+  pull_request:
+    paths:
+      - ".github/workflows/wasi-testsuite.yml"
+      - "lib/host/wasi/**"
+      - "include/host/wasi/**"
+      - "thirdparty/wasi/**"
+
+jobs:
+  test_on_ubuntu:
```
loganek, line 22: `test_on_ubuntu` and `test_on_macos` look very similar so I wonder if it makes sense to use [matrix strategies](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs) and [if expressions](https://docs.github.com/en/actions/learn-github-actions/expressions) to remove duplications in the job definitions?

loganek:
Hi, is there any plan to merge this PR? Currently the daily runtime validation job fails for WasmEdge due to lack of `.github/scripts/wasi-testsuite-adapter.py` file: https://github.com/WebAssembly/wasi-testsuite/actions/runs/4642264910/jobs/8215987344. I can temporarily disable WasmEdge job but if the PR is ready (or almost ready) to be merged, we should probably do that instead.

.github/workflows/wasi-testsuite.yml
harry900831: Thanks for your advise!

hydai (approved):

PR #2336, opened 2023-03-04
[Docs] update LLVM_DIR for macOS

xxchan:
According to https://docs.brew.sh/Installation, `/usr/local` for macOS Intel, `/opt/homebrew` for Apple Silicon, so use `brew --prefix` to unify them.


hydai (approved):
Nice catch.

PR #2348, opened 2023-03-09
[WASI] Handle NaN in runMaxOp() and runMinOp()

### LINKED ISSUE #2098: Unexpected value on the stack
### Current State
0x0 0x0 0x0 0x2f 0x0 0x0 0xf0 0xff

### Expected
........ 0xf8 0xff (The most significant bit of payload is 1)

Raise an exception indicating stack overflow.

## Environment
- Code Version: git commit id b16d9d3e80b84e17974a911c2de94887194ada58
- Hardware Architecture: x86\_64 
- Operating system: Ubuntu 20.04

- C++ Compiler version: clang-14
- CMake version: 3.21.6
- CMake flags:  `-DCMAKE_BUILD_TYPE=Release -DWASMEDGE_BUILD_AOT_RUNTIME=ON`
- Boost version: libboost1.71-dev:amd64

- build commands
  ```
  export CC=/usr/bin/clang-14
  export CXX=/usr/bin/clang++-14
  rm -rf build ; mkdir build ; 
  cd build ; cmake -DCMAKE_BUILD_TYPE=Release -DWASMEDGE_BUILD_AOT_RUNTIME=ON .. ; make -j ; cd ..
  ```

## Steps to Reproduce
[wasmedge_nan.zip](https://github.com/WasmEdge/WasmEdge/files/10047792/wasmedge_nan.zip)

1. build the wasmedge
2. run the test case by `build/tools/wasmedge/wasmedge --reactor <test_case_name> to_test `

dracoooooo:
This PR fixes #2098. 

According to [the documentation](https://webassembly.github.io/spec/core/exec/numerics.html#aux-nans), when processing instructions such as f32.max, if NaN is encountered, the most significant bit of the payload of the result needs to be set to 1.

Canonical and arithmetic NaNs are handled.


alabulei1:
@dracoooooo  Thanks for your contribution. Could you please fix the DCO issues? https://www.secondstate.io/articles/dco/

dracoooooo:
> @dracoooooo Thanks for your contribution. Could you please fix the DCO issues? https://www.secondstate.io/articles/dco/

Fixed.

include/executor/engine/binary_numeric.ipp
```diff
@@ -157,9 +157,22 @@ TypeF<T> Executor::runMinOp(ValVariant &Val1, const ValVariant &Val2) const {
...
+  if (std::isnan(Z1) || std::isnan(Z2)) {
+    if (std::isnan(Z2)) {
+      Z1 = Z2;
+    }
+    // Set the most significant bit of the payload to 1.
+    if (sizeof(T) == sizeof(uint32_t)) {
+      uint32_t I32;
+      std::memcpy(&I32, &Z1, sizeof(T));
+      I32 |= (uint32_t)0x1 << 22;
+      std::memcpy(&Z1, &I32, sizeof(T));
```
q82419, line 169: `reinterpret_cast<uint32_t &>(Z1) |= (static_cast<uint32_t>(0x01U) << 22)`;

include/executor/engine/binary_numeric.ipp
```diff
@@ -157,9 +157,22 @@ TypeF<T> Executor::runMinOp(ValVariant &Val1, const ValVariant &Val2) const {
...
+    } else if (sizeof(T) == sizeof(uint64_t)) {
+      uint64_t I64;
+      std::memcpy(&I64, &Z1, sizeof(T));
+      I64 |= (uint64_t)0x1 << 51;
+      std::memcpy(&Z1, &I64, sizeof(T));
```
q82419, line 174: `reinterpret_cast<uint64_t &>(Z1) |= (static_cast<uint64_t>(0x01U) << 51)`;

include/executor/engine/binary_numeric.ipp
```diff
@@ -176,9 +189,22 @@ TypeF<T> Executor::runMaxOp(ValVariant &Val1, const ValVariant &Val2) const {
...
+  if (std::isnan(Z1) || std::isnan(Z2)) {
+    if (std::isnan(Z2)) {
+      Z1 = Z2;
+    }
+    // Set the most significant bit of the payload to 1.
+    if (sizeof(T) == sizeof(uint32_t)) {
+      uint32_t I32;
+      std::memcpy(&I32, &Z1, sizeof(T));
+      I32 |= (uint32_t)0x1 << 22;
+      std::memcpy(&Z1, &I32, sizeof(T));
```
q82419, line 201: `reinterpret_cast<uint32_t &>(Z1) |= (static_cast<uint32_t>(0x01U) << 22)`;

include/executor/engine/binary_numeric.ipp
```diff
@@ -176,9 +189,22 @@ TypeF<T> Executor::runMaxOp(ValVariant &Val1, const ValVariant &Val2) const {
...
+    } else if (sizeof(T) == sizeof(uint64_t)) {
+      uint64_t I64;
+      std::memcpy(&I64, &Z1, sizeof(T));
+      I64 |= (uint64_t)0x1 << 51;
+      std::memcpy(&Z1, &I64, sizeof(T));
```
q82419, line 206: `reinterpret_cast<uint64_t &>(Z1) |= (static_cast<uint64_t>(0x01U) << 51)`;

q82419 (changes requested):
Thanks for your help.
If available, do you interest in the same issue of the [spec test here](https://github.com/WasmEdge/WasmEdge/blob/master/test/spec/spectest.cpp#L269)?
Thanks!

dracoooooo:
> Thanks for your help. If available, do you interest in the same issue of the [spec test here](https://github.com/WasmEdge/WasmEdge/blob/master/test/spec/spectest.cpp#L269)? Thanks!

Sure. I'm willing to do this.

include/executor/engine/binary_numeric.ipp
dracoooooo: It looks like some CI tests failed because of `reinterpret_cast`. Do you have any suggestions?

dracoooooo: I have tried `*reinterpret_cast<uint32_t *>`, but it's the same.

q82419:
Hi @dracoooooo ,

Can you help to update with the comments? Thanks.

dracoooooo:
> Hi @dracoooooo ,
> 
> Can you help to update with the comments? Thanks.

Hi, @q82419 

After making the modifications you requested, I discovered a strange bug that only appears in the release version and prevents some test cases from passing. When testing locally, I found that running the following WASM program produces different results in debug and release modes - debug returns 'NaN' while release returns '1.76324e-38'(0xc00000). It seems that the highest payload position of '0x1p-126'(0x800000) has been set to 1. I'm not sure what the cause is.

My test case is: 
```
(module
  (func $test (export "test") (result f32)
    (f32.min
      (f32.const 0x1p-126)
      (f32.const -nan:0x200000))))
```

q82419:
> > Hi @dracoooooo ,
> > Can you help to update with the comments? Thanks.
> 
> Hi, @q82419
> 
> After making the modifications you requested, I discovered a strange bug that only appears in the release version and prevents some test cases from passing. When testing locally, I found that running the following WASM program produces different results in debug and release modes - debug returns 'NaN' while release returns '1.76324e-38'(0xc00000). It seems that the highest payload position of '0x1p-126'(0x800000) has been set to 1. The error may be caused by undefined behavior from `reinterpret_cast<uint32_t &>(Z1)`.
> 
> My test case is:
> 
> ```
> (module
>   (func $test (export "test") (result f32)
>     (f32.min
>       (f32.const 0x1p-126)
>       (f32.const -nan:0x200000))))
> ```

Hi @dracoooooo ,

Thanks for your test.
Can you revert to your original implementation with the `memcpy`? Thanks.

q82419:
LGTM

q82419 (changes requested):

q82419:
LGTM

q82419 (approved):

PR #2360, opened 2023-03-15
[WASI-NN] tensorflow lite backend - fix set_input and get_output bugs when checking

yanghaku:
1. Update errno for wasi-nn plugin.    [wasi-nn spec main branch](https://github.com/WebAssembly/wasi-nn/blob/main/wasi-nn.wit.md)
2. ```TfLiteTensorCopyToBuffer``` requires that ```output_data_size == TfLiteTensorByteSize(tensor)```. when getting output and ```OutBuffMaxSize < TfLiteTensorByteSize(tensor)```, TfLite will return ```kTfLiteError``` and wasi-nn plugin will return ```Success```. In this situation, the wasi-nn plugin may return the ```InvalidArgument``` to tell the user ```get_output``` fail.
3. In ```SetInput```, variable ```DimensionBuf``` is used before checking it if is ```NULL```, and variable ```TFDimension``` has never been used, so I delete it.  And I add the dimension check and tensor bytes check to ensure ```input_data_size == TfLiteTensorByteSize(tensor)```.

BTW, in tflite API:
> The caller retains ownership of the `model_data` buffer and should ensure that the lifetime of the `model_data` buffer must be at least as long as the lifetime of the `TfLiteModel` and of any `TfLiteInterpreter` objects created from that `TfLiteModel`, and furthermore the contents of the `model_data` buffer must not be modified during that time."

WASI-NN tensorflow lite plugin should ensure that the ```model_data``` cannot be released or modified during ```TfLiteModel``` and ```TfLiteInterpreter``` lifetime to avoid segment fault.

TODO: solve the problem above.


yanghaku:
Add GPU target support for PyTorch.

Test in my environment:
* cc (GCC) 11.3.0
* cuda: 12.1
* libtorch: 2.0.0

cmake flags: 
```shell
cmake ../.. -G "Ninja" -DCMAKE_BUILD_TYPE=Release -DWASMEDGE_PLUGIN_WASI_NN_BACKEND="PyTorch" -DCMAKE_PREFIX_PATH=/path/to/pytorch/torch -DCMAKE_INSTALL_PREFIX="./install"
```

gusye1234:
Sure

gusye1234:
Hi @yanghaku, I reviewed your code change on TFLite and it looks good to me.
Can you extract the commits involving PyTorch backend from this PR, and form a new PR?

yanghaku:
Of course, but how to do it? May I delete the PyTorch-related code and generate a new tflite only commit in this PR?

gusye1234:
> Of course, but how to do it? May I delete the PyTorch-related code and generate a new tflite only commit in this PR?

Sure. just fork a new branch from this one for the PyTorch PR. And you can delete the PyTorch commits in this pr and then force-push it.

yanghaku:
Hello, @gusye1234  I don't know how to delete the commit in middle, so I just delete the code and generate a new commit.

Do you have any idea to solve the model_data lifetime problem I mentioned above?
in wasm user code (pseudocode) :
```code
let model_data: Vec<u8>;
let graph = wasi_nn::load();  // Create a TfLiteModel
drop(model_data); // Free the model data
let context = wasi_nn::init_execution_context(graph); // Create a TfLiteInterpreter but segment fault because model data has been freed.
```

gusye1234:
> Hello, @gusye1234 I don't know how to delete the commit in middle, so I just delete the code and generate a new commit.
> 
> Do you have any idea to solve the model_data lifetime problem I mentioned above? in wasm user code (pseudocode) :
> 
> ```
> let model_data: Vec<u8>;
> let graph = wasi_nn::load(&model_data);  // Create a TfLiteModel
> drop(model_data); // Free the model data
> let context = wasi_nn::init_execution_context(graph); // Create a TfLiteInterpreter but segment fault because model data has been freed.
> ```

For this demo and TFLite, simply copy the graph builders in `load`?  It avoids the external free of the model data.

yanghaku:
Thanks, it can solve it surely. But I wonder if more memory overhead can be tolerated. In some cases, the model data is more than 1GB (e.g. bert) and the users do not free model data in wasm, which will cause more than 1GB of memory overhead.


> > Hello, @gusye1234 I don't know how to delete the commit in middle, so I just delete the code and generate a new commit.
> > Do you have any idea to solve the model_data lifetime problem I mentioned above? in wasm user code (pseudocode) :
> > ```
> > let model_data: Vec<u8>;
> > let graph = wasi_nn::load(&model_data);  // Create a TfLiteModel
> > drop(model_data); // Free the model data
> > let context = wasi_nn::init_execution_context(graph); // Create a TfLiteInterpreter but segment fault because model data has been freed.
> > ```
> 
> For this demo and TFLite, simply copy the graph builders in `load`? It avoids the external free of the model data.

gusye1234:
@q82419 , gentle ping.
Is there a way for WASI to claim ownership of the buffer?

q82419:
> @q82419 , gentle ping. Is there a way for WASI to claim ownership of the buffer?

I don't think it has any relationship to WASI. The data allocation/deallocation are done by the compiler.
If the `load` copies the buffer data, then of course the `drop(model_data)` can be right after the `load`.
So maybe we should copy and keep the buffer in the host functions.

gusye1234:
Hi,  @yanghaku. Maybe adding the buffer copying in `load` is a workaround for your concern?

> > @q82419 , gentle ping. Is there a way for WASI to claim ownership of the buffer?
> 
> I don't think it has any relationship to WASI. The data allocation/deallocation are done by the compiler. If the `load` copies the buffer data, then of course the `drop(model_data)` can be right after the `load`. So maybe we should copy and keep the buffer in the host functions.

yanghaku:
Hi, @gusye1234 

Copying buffer data directly in the load function may lead to an overhead issue. If this operation is performed, it may cause significant unnecessary overhead.
For instance, a TfLite model file may bundle many data, including associated files. If the model data is only 1MB, but the associated files are 100MB, copying the entire buffer of 101MB would result in an unnecessary 100MB overhead. 
To avoid memory overhead, one possible approach is to parse the buffer data and then copy only the required model data. However, this approach may cause a time overhead.  
It's therefore essential to consider the tradeoff between memory and time overhead when selecting the approach for data copying.

Given the potential issues and tradeoffs, I think that copying buffer data is a temporary but not perfect solution (I will not implement it in this PR).

q82419:
Hi @yanghaku ,

Can you help to resolve the conflict? I think this PR can be merged. Thanks!

q82419:
It seems like the TFLite backend failed.

yanghaku:
I will fix it as quickly as possible.

yanghaku:
the test model  `lite-model_aiy_vision_classifier_birds_V1_3.tflite`  has input tensor shape ```[1,224,224,3]```, but in test code, input shape is ```[1, 3, 224, 224]```, [code link](https://github.com/WasmEdge/WasmEdge/blob/master/test/plugins/wasi_nn/wasi_nn.cpp#L876).  I add the input shape check in the TfLite backend, so the unit test fails.

Now I change the shape to the model's input shape, the test will pass.

q82419 (approved):

PR #2368, opened 2023-03-19
draft: Cpp sdk for WasmEdge

Subhra264:
# Changes
Added a `wasmedge.hh` file in `include/api/wasmedge/` folder, it contains the declarations of methods and classes the C++ SDK is expected to expose. For now, I didn't focus on setting up the SDK, rather want to set a clear goal of classes and methods we want to expose. Although any suggestions on the setup part are welcome too :)

I will keep modifiying this description as new changes are added.


hangedfish:
Optimization of certain APIs can be achieved by using `std::string_view` appropriately to reduce memory copying.

Subhra264:
> Optimization of certain APIs can be achieved by using `std::string_view` appropriately to reduce memory copying.

Thanks for the feedback, will fix it.

Subhra264:
@hangedfish I checked `std::string_view` as per your suggestion, but since it is a C++17 feature and we also want to target the C++11 audience, I think the use of `string_view` can be discarded for public API declarations. For the next couple of days, I will be revamping the code base and will dive deeper into performance enhancement and how `string_view` can be used internally.  Your feedbacks and reviews are greatly welcome.

q82419:
> Optimization of certain APIs can be achieved by using `std::string_view` appropriately to reduce memory copying.

As the comment above, I think it's better to use `std::string` here to have the maximum compatibility for C++11.

include/api/wasmedge/wasmedge.hh
```diff
@@ -0,0 +1,707 @@
...
+
+namespace WasmEdge {
+  enum WASMEDGE_CPP_API_EXPORT ErrCategory {
+    WASM = 0x00,
+    UserLevelError = 0x01
+  };
+
+  enum WASMEDGE_CPP_API_EXPORT HostRegistration {
+    Wasi,
+    WasmEdge_Process,
```
q82419, line 48: The following configurations are removed in master.

include/api/wasmedge/wasmedge.hh
```diff
@@ -0,0 +1,707 @@
...
+    Var = 0x01
+  };
+
+  // TODO: ValType is an extension of RefType
+  enum WASMEDGE_CPP_API_EXPORT RefType {
+    FuncRef = 0x70,
+    ExternRef = 0x6F
+  };
+
+  // TODO: should Version get its own namespace?
```
q82419, line 128: Maybe you can make a `Version` class for it?

include/api/wasmedge/wasmedge.hh
```diff
@@ -0,0 +1,707 @@
...
+  WASMEDGE_CPP_API_EXPORT std::string VersionGet();
+  WASMEDGE_CPP_API_EXPORT uint32_t VersionGetMajor();
+  WASMEDGE_CPP_API_EXPORT uint32_t VersionGetMinor();
+  WASMEDGE_CPP_API_EXPORT uint32_t VersionGetPatch();
+
+  // TODO: Should Log get its own namespace?
```
q82419, line 134: Also, maybe add a `Log` class?

include/api/wasmedge/wasmedge.hh
```diff
@@ -0,0 +1,707 @@
...
+  WASMEDGE_CPP_API_EXPORT void LogSetDebugLevel();
+  WASMEDGE_CPP_API_EXPORT void LogOff();
+
+  class WASMEDGE_CPP_API_EXPORT Value {
+  public:
+    // TODO
+  };
+
+  class WASMEDGE_CPP_API_EXPORT String {
+  public:
```
q82419, line 145: I don't think the `String` class is necessary.
In C API, we needs the buffer and length data, so we created the `WasmEdge_String` struct.
But in C++, we can use `std::string` directly.

include/api/wasmedge/wasmedge.hh
```diff
@@ -0,0 +1,707 @@
...
+      Min(Min),
+      Max(Max) {};
+    ~Limit() = default;
+
+    bool HasMax { false };
+    bool Shared { false };
+    uint32_t Min { 0 };
+    uint32_t Max { 0 };
+
+    bool IsEqual(const Limit Lim);
```
q82419, line 195: Use the `operator==` overloading is better.

include/api/wasmedge/wasmedge.hh
```diff
@@ -0,0 +1,707 @@
...
+  class WASMEDGE_CPP_API_EXPORT FunctionType {
+  public:
+    FunctionType(const std::vector<ValType> &ParamList,
+                const uint32_t ParamLen,
+                const std::vector<ValType> &ReturnList,
+                const uint32_t ReturnLen);
+    ~FunctionType() = default;
+
+    uint32_t GetParametersLength();
+    std::vector<ValType> GetParameters();
```
q82419, line 207: Return `const std::vector<ValType> &`.

include/api/wasmedge/wasmedge.hh
```diff
@@ -0,0 +1,707 @@
...
+    // TODO
+  };
+
+  class WASMEDGE_CPP_API_EXPORT TableType {
+  public:
+    TableType(const RefType RefType, const Limit Limit);
+    ~TableType() = default;
+
+    RefType GetRefType();
+    Limit GetLimit();
```
q82419, line 221: Return `const Limit &`.

include/api/wasmedge/wasmedge.hh
```diff
@@ -0,0 +1,707 @@
...
+  private:
+    // TODO
+  };
+
+  class WASMEDGE_CPP_API_EXPORT MemoryType {
+  public:
+    MemoryType(const Limit Limit);
+    ~MemoryType() = default;
+
+    Limit GetLimit();
```
q82419, line 231: Return `const Limit &`.

include/api/wasmedge/wasmedge.hh
```diff
@@ -0,0 +1,707 @@
...
+
+    Result LoadWasmFromFile(const std::string &Path);
+    Result LoadWasmFromBuffer(const std::vector<uint8_t> &Buf,
+                  const uint32_t BufLen);
+    Result LoadWasmFromASTModule(const ASTModule &ASTCxt);
+
+    Result Validate();
+    Result Instantiate();
+
+    Result Execute(const String FuncName, const std::vector<Value> &Params,
```
q82419, line 418: The `ParamLen` for C API is because the C API doesn't have `vector`.
With `std::vector`, the length argument is not needed.

include/api/wasmedge/wasmedge.hh
```diff
@@ -0,0 +1,707 @@
...
+                  const uint32_t ParamLen, const std::vector<Value> &Returns,
+                  const uint32_t ReturnLen);
+    Result ExecuteRegistered(const String ModuleName, const String FuncName,
```
q82419, line 421: You can overload this function instead of making a different name.

include/api/wasmedge/wasmedge.hh
```diff
@@ -0,0 +1,707 @@
...
+                  const std::vector<Value> &Params, const uint32_t ParamLen,
+                  std::vector<Value> &Returns, const uint32_t ReturnLen);
+
+    std::unique_ptr<Async> AsyncExecute(const String FuncName,
+                  const std::vector<Value> &Params, const uint32_t ParamLen);
+    std::unique_ptr<Async> AsyncExecuteRegistered(const String ModuleName,
+                  const String FuncName, const std::vector<Value> &Params,
+                  const uint32_t ParamLen);
+
+    std::unique_ptr<const FunctionType> GetFunctionType(const String FuncName);
```
q82419, line 431: Why `unique_ptr`?
You can return a `const FunctionType &`.

include/api/wasmedge/wasmedge.hh
```diff
@@ -0,0 +1,707 @@
...
+                            const uint32_t CmdsLen,
+                            const bool AllowAll);
+    String GetModuleName();
+
+    FunctionInstance &FindFunction(const String Name);
+    TableInstance &FindTable(const String Name);
+    MemoryInstance &FindMemory(const String Name);
+    GlobalInstance &FindGlobal(const String Name);
+
+    uint32_t ListFunctionLength();
```
q82419, line 589: The length getter is not needed.

include/api/wasmedge/wasmedge.hh
```diff
@@ -0,0 +1,707 @@
...
+    uint32_t ListTableLenth();
+    uint32_t ListTable(std::vector<std::string> &Names, const uint32_t Len);
+
+    uint32_t ListMemoryLength();
+    uint32_t ListMemory(std::vector<std::string> &Names, const uint32_t Len);
+
+    uint32_t ListGlobalLength();
+    uint32_t ListGlobal(std::vector<std::string> &Names, const uint32_t Len);
+
+    void AddFunction(const String Name,
```
q82419, line 601: For the signature, the `(const std::string &, FunctionInstance &&)` is better.

q82419 (changes requested):
Please refine your API designs with the following concepts:

1. The `length` in C API is because the C doesn't have the vector container. With STL in C++, your API designs shouldn't reserve this.
2. Think about the ownership of your objects. Some function returns can transfor the ownership such as list the names, but some function returns the reference to the object which is not suitable to be copied.
3. Notice the `const` decorations of your functions and arguments.

include/api/wasmedge/wasmedge.hh
Subhra264: Sure, but in that case, the constructor needs to be private since there is no need to instantiate it. Also, all the member functions must be static. Is there any specific reason for using a class instead of a namespace?

Subhra264:
I have addressed the reviews in the last two commits and also added function signatures for `Value`. Would appreciate any feedback on the recent commits.

include/api/wasmedge/wasmedge.hh
```diff
@@ -0,0 +1,684 @@
...
+  class WASMEDGE_CPP_API_EXPORT Async {
+    Async();
+    ~Async() = default;
+
+  public:
+    void Wait();
+    bool WaitFor(uint64_t Milliseconds);
+    void Cancel();
+    uint32_t GetReturnsLength();
+    Result Get(std::vector<Value> &Returns);
```
q82419, line 302: As mentioned. The length getter is not necessary.

include/api/wasmedge/wasmedge.hh
```diff
@@ -0,0 +1,684 @@
...
+                  const std::string FuncName, const std::vector<Value> &Params);
+
+    Result LoadWasm(const std::string &Path);
+    Result LoadWasm(const std::vector<uint8_t> &Buf);
+    Result LoadWasm(const ASTModule &ASTCxt);
+
+    Result Validate();
+    Result Instantiate();
+
+    Result Execute(const std::string FuncName, const std::vector<Value> &Params,
```
q82419, line 427: The `returns` is the output and cannot be `const`.

include/api/wasmedge/wasmedge.hh
```diff
@@ -0,0 +1,684 @@
...
+                  const std::vector<Value> &Returns);
+    Result Execute(const std::string ModuleName,
```
q82419, line 429: Use `const std::string &`.

include/api/wasmedge/wasmedge.hh
```diff
@@ -0,0 +1,684 @@
...
+
+    void Cleanup();
+    uint32_t GetFunctionList(std::vector<std::string> &Names,
+                  const std::vector<FunctionType &> &FuncTypes);
+
+    ModuleInstance &GetImportModuleContext(const HostRegistration Reg);
+    const ModuleInstance& GetActiveModule();
+    const ModuleInstance &GetRegisteredModule(const std::string ModuleName);
+
+    uint32_t ListRegisteredModule(std::vector<std::string> &Names);
```
q82419, line 450: Can use the `std::vector<std::string>` as return here.

include/api/wasmedge/wasmedge.hh
```diff
@@ -0,0 +1,684 @@
...
+    // TODO
+  };
+
+  class WASMEDGE_CPP_API_EXPORT Store {
+  public:
+    Store();
+    ~Store() = default;
+
+    const ModuleInstance &FindModule(const std::string Name);
+    uint32_t ListModule(std::vector<std::string> &Names);
```
q82419, line 544: Can return `std::vector<std::string>` here.

include/api/wasmedge/wasmedge.hh
```diff
@@ -0,0 +1,684 @@
...
+    void InitWasmEdgeProcess(const std::vector<const std::string> &AllowedCmds,
+                            const bool AllowAll);
+    std::string GetModuleName();
+
+    FunctionInstance &FindFunction(const std::string Name);
+    TableInstance &FindTable(const std::string Name);
+    MemoryInstance &FindMemory(const std::string Name);
+    GlobalInstance &FindGlobal(const std::string Name);
+
+    uint32_t ListFunction(std::vector<std::string> &Names);
```
q82419, line 577: Can return `std::vector<std::string>` here.

q82419 (changes requested):

juntao:
flows summarize

alabulei1:

------

In summary, there are several potential issues and errors present in some of the source code patches. Firstly, the `WASMEDGE_CPP_API_EXPORT` macro is defined differently for Windows and Unix, which could cause compilation issues on different platforms. Secondly, some of the newly added classes such as `ImportType` and `ExportType` have not been fully implemented, and some methods such as `TODO` comments in `StatisticsContext` class need further work. Additionally, the lack of documentation with regard to the usage of some newly introduced classes and methods can make it difficult to understand how to use them. 

On the other hand, some patches include significant changes, such as the new "VM" class, and modifications to the API of the WasmEdge library, that may require additional review and testing to ensure that they do not introduce bugs or unintended behavior. 

Overall, careful attention needs to be paid to potential issues and errors, and thorough testing needs to be conducted to ensure that the changes are stable and do not compromise the overall functionality of the code.

## Details

### Commit 1
This patch introduces the `wasmedge.hh` header file which contains the initial declarations for the WasmEdge namespace, `String`, `Result`, and `Limit` classes. 

One potential issue could be with the `WASMEDGE_CPP_API_EXPORT` macro which is defined differently for Windows and Unix. This could affect compilation on different platforms if not handled properly. 

Another issue could be related to the implementation of the `String` class. It is not clear from the code how the `const std::string Buf` member variable is intended to be used - whether it is to be modified or not. This needs further clarification. 

Lastly, the `Limit` class seems straightforward, but some more context on how and where it is used would be helpful in terms of review.

### Commit 2
Key Changes:
- A new class ConfigureContext is added.
- Enumerations HostRegistration, Proposal, CompilerOutputFormat and ErrCategory are added.
- Methods are added to ConfigureContext class for handling different configurations such as Proposals, Host registration, MaxMemoryPage, Compiler Config, and Statistics Config. 

Potential Problems:
- None identified. 
- However, more details are required about the configuration settings in the ConfigureContext class and how they are relevant to the WasmEdge library.

### Commit 3
Key Changes:
- Added `Version` and `Log` functions
- Declared `StatisticsContext` class
- Added `CompilerOptimizationLevel` enum
- Added a few functions to `Configure` class to support the new enum

Potential Problems:
- The `TODO` comments in `StatisticsContext` class may mean that it's not fully implemented yet, so it may not work as intended or at all. 
- It's not clear what the input parameters for `LogSetErrorLevel`, `LogSetDebugLevel`, `LogOff`, `SetCostTable`, and `SetCostLimit` should be, so the documentation may need to be updated for these functions.
- Unless there is a specific reason, it may be better for `Version` and `Log` functions to be in their own namespace instead of `WasmEdge` to avoid potential conflicts with similar functions in other libraries.

### Commit 4
Key changes:
- Four classes, namely Loader, Executor, Validator, and Store, have been added.
- Each class has its constructor and destructor methods defined.
- The Validator and Executor classes accept additional parameters in their constructors, namely ConfigContext and Statistics Context, respectively.

Potential problems:
- It is unclear whether the Loader class can parse a file correctly.
- There is no implementation of the TODO comments for the Loader, Validator, Executor, and Store classes, which suggests that the patch may be incomplete.
- The lack of information in the comments about what each method does can make it difficult to understand how to use these classes.

### Commit 5
The patch adds several new data structures and methods to the code, specifically `ASTModule`, `FunctionType`, `ValType`, `ImportType`, `ExportType`, `Value`, `Limit`, and `String`. Additionally, there are modifications to the `Loader` and `Executor` classes, and a new `ASTModule` class is introduced. 

One potential issue is that there is lack of documentation with regard to the usage of some newly introduced classes and methods such as the `Value` class, which was not defined or commented in the patch. Also, some of the newly added classes such as `ImportType` and `ExportType` have not been fully implemented and hence need further work. Lastly, the `SetCostTable` signature in `StatisticsContext` needs to be corrected to take a vector instead of a pointer to array.

### Commit 6
Key Changes:
- Enum types RefType, ExternalType, and Mutability are added.
- New classes TableType, MemoryType, GlobalType, and ASTModule are added.
- Method names have been updated to be more descriptive.
- New method implementations have been added for initialization, finding, and listing of Function, Table, Memory, and Global.

Potential Problems:
- There are a lot of TODOs left in the code.
- The purpose of the changes and their impact on the overall project cannot be determined without more context.

### Commit 7
This patch adds two significant changes: a new "Async" class, and a "VM" class which replaces the existing loader, validator, and executor classes. 

The Async class is not fully implemented, as there are some functions without definitions. This should be addressed before merging this patch.

There is also a new dependency on the "functional" library, which needs to be added to the project.

The new VM class is a significant change to the architecture of the project, and it should be reviewed carefully to ensure that it does not introduce bugs or unintended behavior.

### Commit 8
The patch includes several changes to the `wasmedge.hh` header file. Key changes include:
- `Version`, `Log`, and `Value` classes have been modified.
- `String` class has been removed.
- Constructors of `FunctionType`, `TableType`, and `MemoryType` have been modified.

There is a potential problem with the `registerModule()` functions as all three have return types of `Result&` instead of `Result`. This may cause confusion for developers trying to use the return value from these functions. It's also unclear why the change was made to the constructors of `FunctionType`, `TableType`, and `MemoryType`.

### Commit 9
The patch modifies the API of the WasmEdge library. It fixes the argument and return types of some functions and changes the type of the "Ref" argument in the "Value" class to a shared pointer. 

The potential problem is that this change may require changes in the code that uses this library. The change in the "Value" class may also affect the memory management of the library, so careful testing is necessary to ensure that there are no memory leaks or other issues. 

Additionally, since the patch contains only changes to the API, it is difficult to assess the impact of the changes on the functionality of the library. Therefore, it is important to review the corresponding code changes and tests, which are not included in this patch.

### Commit 10
This patch seems to address the second round of reviews. It modifies a single file, `wasmedge.hh`, by adding two header file inclusions at the top of the file and removing the `int128_t` and `uint128_t` typedefs of the `Value` class. It also removes the `uint32_t GetReturnsLength()` method of the `VM` class and changes the signature of several methods by adding the `const` qualifier to some arguments.

There don't appear to be any potential problems introduced by this patch.

q82419 (approved):
Thanks for help.
Merge into the dev branch and complete them in the future.

PR #2380, opened 2023-03-25
[VM] Allow two VM instance create in one thread

ibmibmibm:
* Add test for multiple VM instance

### DIFF
--- a/include/vm/async.h
+++ b/include/vm/async.h
@@ -34,7 +34,6 @@ public:
         std::thread([FPtr, P = std::move(Promise),
                      Tuple = std::tuple(
                          &TargetVM, std::forward<ArgsT>(Args)...)]() mutable {
-          std::get<0>(Tuple)->newThread();
           P.set_value(std::apply(FPtr, Tuple));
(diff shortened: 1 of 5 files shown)


include/vm/async.h
```diff
@@ -34,7 +34,6 @@ template <typename T> class Async {
         std::thread([FPtr, P = std::move(Promise),
                      Tuple = std::tuple(
                          &TargetVM, std::forward<ArgsT>(Args)...)]() mutable {
-          std::get<0>(Tuple)->newThread();
```
hydai, line 37: Do we need to call prepare here?
It looks like the `newThread` is replaced by `prepare`.

hydai (changes requested):

include/vm/async.h
ibmibmibm: No more `prepare` required here.

hydai (dismissed):
LGTM

hydai (approved):

juntao:
Flows review

PR #2385, opened 2023-03-29
[CI] update for unified tool

michael1017:
Add wasmedge compile in CI

### DIFF
--- a/.github/workflows/build_for_riscv.yml
+++ b/.github/workflows/build_for_riscv.yml
@@ -68,9 +68,12 @@ jobs:
           wasmedge -v
           wasmedgec -v
           wasmedge compile -h
+          wasmedge run -h
           cd ../examples/wasm
           wasmedgec fibonacci.wasm fibonacci_aot_c.wasm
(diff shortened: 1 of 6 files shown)


hydai (approved):

.github/workflows/build_for_riscv.yml
```diff
@@ -67,6 +67,10 @@ jobs:
           cmake -DCMAKE_BUILD_TYPE=Release .. && make -j && make install
           wasmedge -v
           wasmedgec -v
+          wasmedge compile -h
```
hydai, line 70: The CI Failed. Does the `wasmedge compile -h` return non-zero?

hydai (changes requested):

.github/workflows/build_for_riscv.yml
michael1017: Yes you are right, `wasmedge compile -h` return non-zero. Let me fix this out.

michael1017: @hydai It should be fixed now

hydai (approved):

PR #2394, opened 2023-04-01
[Rust] Optimize host function

apepkuss:
In this PR, the host function design is improved to support removing host functions from the lazy static `HOST_FUNCS`.

### DIFF
--- a/.github/workflows/rust-static-lib.yml
+++ b/.github/workflows/rust-static-lib.yml
@@ -30,7 +30,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-20.04, ubuntu-22.04]
-        rust: [1.64, 1.65, 1.66]
+        rust: [1.66, 1.67, 1.68]
     container:
       image: wasmedge/wasmedge:latest
 
@@ -69,5 +69,5 @@ jobs:
         run: |
           export WASMEDGE_DIR="$(pwd)/../../"
           export WASMEDGE_BUILD_DIR="$(pwd)/../../build"
-          cargo test --workspace --features static --locked
-          cargo test --examples --features static --locked
+          cargo test --workspace --features static --locked -- --test-threads=1
+          cargo test --examples --features static --locked -- --test-threads=1
--- a/.github/workflows/bindings-rust.yml
+++ b/.github/workflows/bindings-rust.yml
@@ -247,5 +247,5 @@ jobs:
           export WASMEDGE_BUILD_DIR="$(pwd)/../../build"
           export WASMEDGE_PLUGIN_PATH="$(pwd)/../../build/plugins/wasmedge_process"
           export LD_LIBRARY_PATH="$(pwd)/../../build/lib/api"
-          cargo test --workspace --locked --features aot,async,ffi
-          cargo test --examples --locked --features aot,async,ffi
+          cargo test --workspace --locked --features aot,async,ffi -- --test-threads=1
+          cargo test --examples --locked --features aot,async,ffi -- --test-threads=1
--- a/.github/workflows/bindings-rust.yml
+++ b/.github/workflows/bindings-rust.yml
@@ -72,8 +72,8 @@ jobs:
           export WASMEDGE_BUILD_DIR="$(pwd)/../../build"
           export WASMEDGE_PLUGIN_PATH="$(pwd)/../../build/plugins/wasmedge_process"
           export LD_LIBRARY_PATH="$(pwd)/../../build/lib/api"
-          cargo test --workspace --locked --features aot,async,ffi
-          cargo test --examples --locked --features aot,async,ffi
+          cargo test --workspace --locked --features aot,async,ffi -- --test-threads=1
+          cargo test --examples --locked --features aot,async,ffi -- --test-threads=1
 
   build_macos:
     name: MacOS
@@ -127,8 +127,8 @@ jobs:
           export WASMEDGE_BUILD_DIR="$(pwd)/../../build"
           export WASMEDGE_PLUGIN_PATH="$(pwd)/../../build/plugins/wasmedge_process"
           export DYLD_LIBRARY_PATH="$(pwd)/../../build/lib/api"
-          cargo test --workspace --locked --features aot,async,ffi
-          cargo test --examples --locked --features aot,async,ffi
+          cargo test --workspace --locked --features aot,async,ffi -- --test-threads=1
+          cargo test --examples --locked --features aot,async,ffi -- --test-threads=1
 
   build_windows:
     name: Windows
@@ -197,8 +197,8 @@ jobs:
           Enter-VsDevShell -VsInstallPath $vsPath -SkipAutomaticLocation -DevCmdArguments "-arch=x64 -host_arch=x64 -winsdk=10.0.19041.0"
           $env:Path="$env:Path;C:\Users\runneradmin\.rustup\toolchains\stable-x86_64-pc-windows-msvc\bin;D:\a\WasmEdge\WasmEdge\build\lib\api"
           cd bindings\rust
-          cargo test --workspace --features aot,async,ffi --locked
-          cargo test --examples --features aot,async,ffi --locked
+          cargo test --workspace --features aot,async,ffi --locked -- --test-threads=1
+          cargo test --examples --features aot,async,ffi --locked -- --test-threads=1
 
   build_fedora:
     name: Fedora 37
--- a/bindings/rust/wasmedge-sys/src/instance/function.rs
+++ b/bindings/rust/wasmedge-sys/src/instance/function.rs
@@ -1178,8 +1178,9 @@ mod tests {
         );
         println!("len of HOST_FUNCS: {}", HOST_FUNCS.read().len());
 
-        assert!(
-            HOST_FUNCS.read().len() - start_num >= funcs.len(),
+        assert_eq!(
+            HOST_FUNCS.read().len() - start_num,
+            funcs.len(),
             "expected: {}, actual: {}, start_num: {}",
             funcs.len(),
             HOST_FUNCS.read().len() - start_num,
--- a/bindings/rust/wasmedge-sys/src/instance/function.rs
+++ b/bindings/rust/wasmedge-sys/src/instance/function.rs
@@ -1178,9 +1178,8 @@ mod tests {
         );
         println!("len of HOST_FUNCS: {}", HOST_FUNCS.read().len());
 
-        assert_eq!(
-            HOST_FUNCS.read().len() - start_num,
-            funcs.len(),
+        assert!(
+            HOST_FUNCS.read().len() - start_num >= funcs.len(),
             "expected: {}, actual: {}, start_num: {}",
             funcs.len(),
             HOST_FUNCS.read().len() - start_num,
--- a/bindings/rust/wasmedge-sys/src/instance/function.rs
+++ b/bindings/rust/wasmedge-sys/src/instance/function.rs
@@ -176,14 +176,6 @@ impl Function {
         cost: u64,
     ) -> WasmEdgeResult<Self> {
         let mut map_host_func = HOST_FUNCS.write();
-        if map_host_func.len() >= map_host_func.capacity() {
-            return Err(Box::new(WasmEdgeError::Func(FuncError::CreateBinding(
-                format!(
-                    "The number of the host functions reaches the upper bound: {}",
-                    map_host_func.capacity()
-                ),
-            ))));
-        }
 
         // generate key for the coming host function
         let mut rng = rand::thread_rng();
@@ -738,7 +730,7 @@ unsafe impl Sync for InnerFuncRef {}
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::{types::WasmValue, Executor};
+    use crate::{types::WasmValue, Executor, HOST_FUNCS};
     use std::{
         sync::{Arc, Mutex},
         thread,
@@ -1117,5 +1109,57 @@ mod tests {
         assert!(result.is_ok());
         let returns = result.unwrap();
         assert_eq!(returns[0].to_i32(), 3);
+    }
+
+    #[test]
+    fn test_crazy_function_creation() {
+        #[sys_host_function]
+        fn real_add(
+            _frame: CallingFrame,
+            input: Vec<WasmValue>,
+        ) -> Result<Vec<WasmValue>, HostFuncError> {
+            println!("Rust: Entering Rust function real_add");
+
+            if input.len() != 2 {
+                return Err(HostFuncError::User(1));
+            }
+
+            let a = if input[0].ty() == ValType::I32 {
+                input[0].to_i32()
+            } else {
+                return Err(HostFuncError::User(2));
+            };
+
+            let b = if input[1].ty() == ValType::I32 {
+                input[1].to_i32()
+            } else {
+                return Err(HostFuncError::User(3));
+            };
+
+            let c = a + b;
+            println!("Rust: calcuating in real_add c: {c:?}");
+
+            println!("Rust: Leaving Rust function real_add");
+            Ok(vec![WasmValue::from_i32(c)])
+        }
+
+        let mut funcs = vec![];
+        for _ in 1..=1_000_000 {
+            // create a FuncType
+            let result = FuncType::create(vec![ValType::I32; 2], vec![ValType::I32]);
+            assert!(result.is_ok());
+            let func_ty = result.unwrap();
+            // create a host function
+            let result = Function::create(&func_ty, Box::new(real_add), 0);
+            assert!(result.is_ok());
+            let host_func = result.unwrap();
+            funcs.push(host_func);
+        }
+
+        assert_eq!(HOST_FUNCS.read().len(), funcs.len());
+        println!(
+            "The number of the entries in HOST_FUNCS: {}",
+            HOST_FUNCS.read().len()
+        )
     }
 }
--- a/bindings/rust/wasmedge-sys/src/instance/function.rs
+++ b/bindings/rust/wasmedge-sys/src/instance/function.rs
@@ -3,6 +3,7 @@
 use crate::{
     error::{FuncError, HostFuncError, WasmEdgeError},
     ffi, BoxedFn, CallingFrame, Engine, WasmEdgeResult, WasmValue, HOST_FUNCS,
+    HOST_FUNC_FOOTPRINTS,
 };
 #[cfg(feature = "async")]
 use crate::{r#async::FiberFuture, ASYNC_STATE};
@@ -193,6 +194,11 @@ impl Function {
             data,
             cost,
         );
+
+        // create a footprint for the host function
+        let footprint = ctx as usize;
+        let mut footprint_to_id = HOST_FUNC_FOOTPRINTS.lock();
+        footprint_to_id.insert(footprint, key);
 
         match ctx.is_null() {
             true => Err(Box::new(WasmEdgeError::Func(FuncError::Create))),
@@ -504,6 +510,13 @@ impl Function {
 impl Drop for Function {
     fn drop(&mut self) {
         if !self.registered && Arc::strong_count(&self.inner) == 1 && !self.inner.0.is_null() {
+            // remove the real_func from HOST_FUNCS
+            let footprint = self.inner.0 as usize;
+            if let Some(key) = HOST_FUNC_FOOTPRINTS.lock().remove(&footprint) {
+                let mut map_host_func = HOST_FUNCS.write();
+                map_host_func.remove(&key);
+            }
+            // delete the function instance
             unsafe { ffi::WasmEdge_FunctionInstanceDelete(self.inner.0) };
         }
     }
@@ -1143,8 +1156,11 @@ mod tests {
             Ok(vec![WasmValue::from_i32(c)])
         }
 
+        let start_num = HOST_FUNCS.read().len();
+        println!("start_num: {}", start_num);
+
         let mut funcs = vec![];
-        for _ in 1..=1_000_000 {
+        for _ in 1..=10 {
             // create a FuncType
             let result = FuncType::create(vec![ValType::I32; 2], vec![ValType::I32]);
             assert!(result.is_ok());
@@ -1156,10 +1172,19 @@ mod tests {
             funcs.push(host_func);
         }
 
-        assert_eq!(HOST_FUNCS.read().len(), funcs.len());
         println!(
-            "The number of the entries in HOST_FUNCS: {}",
-            HOST_FUNCS.read().len()
-        )
+            "len of HOST_FUNC_FOOTPRINTS: {}",
+            HOST_FUNC_FOOTPRINTS.lock().len()
+        );
+        println!("len of HOST_FUNCS: {}", HOST_FUNCS.read().len());
+
+        assert_eq!(
+            HOST_FUNCS.read().len() - start_num,
+            funcs.len(),
+            "expected: {}, actual: {}, start_num: {}",
+            funcs.len(),
+            HOST_FUNCS.read().len() - start_num,
+            start_num
+        );
     }
 }
--- a/bindings/rust/wasmedge-sys/src/instance/function.rs
+++ b/bindings/rust/wasmedge-sys/src/instance/function.rs
@@ -1160,7 +1160,7 @@ mod tests {
         println!("start_num: {}", start_num);
 
         let mut funcs = vec![];
-        for _ in 1..=10 {
+        for _ in 1..=1_000_000 {
             // create a FuncType
             let result = FuncType::create(vec![ValType::I32; 2], vec![ValType::I32]);
             assert!(result.is_ok());
--- a/bindings/rust/wasmedge-sys/src/lib.rs
+++ b/bindings/rust/wasmedge-sys/src/lib.rs
@@ -43,7 +43,7 @@
 extern crate lazy_static;
 
 use parking_lot::{Mutex, RwLock};
-use std::{collections::HashMap, env, sync::Arc};
+use std::{collections::HashMap, sync::Arc};
 
 #[allow(warnings)]
 /// Foreign function interfaces generated from WasmEdge C-API.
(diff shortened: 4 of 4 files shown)


apepkuss:

.github/workflows/rust-static-lib.yml
```diff
@@ -69,5 +69,5 @@ jobs:
         run: |
           export WASMEDGE_DIR="$(pwd)/../../"
           export WASMEDGE_BUILD_DIR="$(pwd)/../../build"
-          cargo test --workspace --features static --locked
-          cargo test --examples --features static --locked
+          cargo test --workspace --features static --locked -- --test-threads=1
```
hydai, line 72: Why do we need to add this limit? Is there any new requirement changed in this PR?

alabulei1:

------

Based on the set of reviews provided, I see that there are several patches that aim to improve code efficiency, test reliability, and workflow stability. The changes themselves look straightforward, simple, and well-documented, with some potential problems or limitations specified. The reviewers also raised some concerns about the commit message's clarity and the lack of information about the codebase's context. Overall, the patches seem reasonable and unlikely to cause major issues, but more testing and validation may be necessary for some changes.

## Details

### Commit 1
The patch removes some code that appears to be no longer needed from the `wasmedge-sys` Rust crate. The change also includes a new unit test that creates a large number of new host functions, which could potentially cause performance problems or hit resource limits on the system. However, the test could be useful for testing the code's ability to handle large numbers of host functions effectively. Overall, the changes themselves look straightforward and simple.

One thing it's worth noting is that the commit message could be more informative, as the subject line "refactor(rust-sys): remove the inappropriate code" is somewhat vague and does not provide substantial detail.

### Commit 2
This patch adds support for removing real function in Rust-Sys of Wasmedge. The key changes include:
- Create a `footprint` for each host function in `Function::from_raw_fn_ptr()`.
- Remove the real function from `HOST_FUNCS` in `Function::drop()`.
- Create a `HOST_FUNC_FOOTPRINTS` hash map to store the mapping between the address of each host function pointer to the key of `HOST_FUNCS`.

This patch looks reasonable and well-documented to me. However, there is a potential problem in a loop of unit testing (lines 518-556). The declaration of `start_num` before the loop does not reflect the initial length of `HOST_FUNCS`, which is a problem for the assertion on line 559. Instead, the test should store the initial length of `HOST_FUNCS` before the loop starts to compare it to the total number of functions added to `HOST_FUNCS` at the end of the loop.

### Commit 3
This patch is a minor change to the test code for `bindings/rust/wasmedge-sys/src/instance/function.rs`. The key change is replacing a specific assertion that checks for an exact match of the length of `funcs` with a conditional assertion that checks if `funcs` is smaller or equal to the length of `HOST_FUNCS`. 

There don't seem to be any potential problems with this patch as it appears to be a fairly simple change which updates the existing test code to be more robust. However, it may be helpful to know more about the context before accepting this patch.

### Commit 4
This patch updates the GitHub Actions workflow for the Rust bindings in a specific GitHub repository. Specifically, it modifies the cargo test commands by adding the flag `--test-threads=1` to set the number of test threads to 1. The purpose of this change is to prevent unexpected test failures due to race conditions when multiple tests are running concurrently.

One potential problem that I see is that setting the number of test threads to 1 may slow down the test execution, especially if there are a large number of tests. This may increase the overall build and test time, which could reduce the efficiency of the workflow.

Overall, this patch seems like a reasonable change to improve the reliability of the tests in the Rust bindings.

### Commit 5
This patch is updating test code for a function in the rust-sys bindings of the Wasmedge project. Specifically, it's changing an assertion from a comparison to an equality check. It also includes a commit message and signed-off-by tag. 

One potential issue that stands out is that the commit message doesn't explain why the change is being made or what problem it is resolving. Additionally, without more context about the codebase itself and the purpose of this test function, it's difficult to confidently assess whether this change will have any unforeseen effects.

### Commit 6
This patch updates the GitHub Actions workflow for a Rust static library project to test with newer versions of Rust (1.66, 1.67, 1.68) and runs the tests using a single thread (--test-threads=1). The change exports environment variables for the Wasmedge directory and builds directory. These changes are expected to improve the stability and performance of the project.

There are no potential problems in this patch.

### Commit 7
This patch updates the test code in the `function.rs` file by changing the loop limit from 10 to 1,000,000. It seems to be a performance optimization for testing purposes.

Potential problems could arise if this change impacts the memory or resource usage of the code. It's also possible that this change could cause the test to run much longer, so testing the test suite before and after the change may be necessary to validate the effect of this patch.

hydai (changes requested):

.github/workflows/rust-static-lib.yml
apepkuss: In this PR we added a new test case `test_crazy_function_creation`. This test function is used to verify the global static variable `HOST_FUNCS` which is used to store the host functions. However, other test cases also access the global variable, therefore, use `-- --test-threads=1` to avoid the case of other test cases modifying `HOST_FUNCS` while `test_crazy_function_creation` is running. In short, the option turns off the execution of all test cases in parallel.

hydai (dismissed):

alabulei1:

------

Here is a summary of all the reviews:

1. The patch removes inappropriate code, defines new criteria for host function creation, and adds a new test case that creates a large number of host functions that could raise concerns about code design.
2. This patch adds support for removing real functions from `HOST_FUNCS` and adds `HOST_FUNC_FOOTPRINTS` to store the mapping from host function pointers to `HOST_FUNCS`. There are no potential issues.
3. This patch modifies the assertion statement in the `mod tests` block for `function.rs` to add an inequality check. Potential issues are difficult to identify without knowing the context.
4. The patch modifies the `bindings-rust.yml` file by adding the `--test-threads=1` flag to two `cargo test` commands. Decreased concurrency could increase execution times for certain hardware configurations.
5. The patch changes the `assert` statement in the test module to use `assert_eq`. There do not appear to be any issues.
6. The patch updates the `rust-static-lib.yml`. The changes involve replacing the version of Rust, which could have implications for the project's dependencies.
7. The patch updates the test code by testing a larger number of cases. The potential problem could be a slower or resource-intensive test.
8. The patch modifies the test command across the GitHub workflow, which could slow down the testing process.

## Details

### Commit 1
Based on the given patch, the key changes made to the code are:

- Removed an inappropriate code on line 7-13. 
- Redefined a criteria for creating a new host function on line 14-24.
- Added a new test case on line 27-56.

As for potential problems, it is hard to tell without a full understanding of the project and codebase. However, I can see this patch adds a new test case that creates a large number of host functions, which might cause performance issues or raise concerns on the code design.

### Commit 2
This is a patch for a single commit with the following changes:

- In `bindings/rust/wasmedge-sys/src/instance/function.rs`:
  - Added support for removing a real function from `HOST_FUNCS`.
  - Created a footprint for the host function in `HOST_FUNC_FOOTPRINTS`.
- Updated the test case to have only ten host functions and added a check to ensure the correct number is present in `HOST_FUNCS`.

The main changes are the support for removing real functions from `HOST_FUNCS` and the addition of `HOST_FUNC_FOOTPRINTS` to store the mapping from the address of each host function pointer to the key of the `HOST_FUNCS`. However, there are no potential issues that appear to be present in this patch as the changes are straightforward and do not involve any significant or risky code changes. Overall, this patch seems valid and should be reviewed further.

### Commit 3
This GitHub patch updates the test code in the `function.rs` module of the Rust WasmEdge Sys bindings. Specifically, it modifies the assertion statement in the `mod tests` block to add an inequality check to compare `HOST_FUNCS.read().len() - start_num` to `funcs.len()` instead of an equality check. This is to ensure that the test passes when more host functions are added. 

Potential issues:
- It's difficult to identify potential issues without knowing the context of the codebase and the purpose of the changes.

### Commit 4
This patch updates the CI workflow for Rust bindings in a GitHub repository. The change modifies the `bindings-rust.yml` file by adding the `--test-threads=1` flag to two `cargo test` commands in three different sections of the `jobs:` matrix.

This is a minor change that seems unlikely to introduce new problems. That being said, decreased concurrency could lead to an increase in execution times for certain hardware configurations. Additionally, further changes to the CI pipeline may benefit from manual testing before being merged, in order to ensure that the changes have the desired effect.

### Commit 5
The patch updates the test code in the `function.rs` file of the Rust sys library for the Wasmedge runtime. Specifically, it changes the `assert` statement in the test module to use `assert_eq` instead and confirms that the length of `HOST_FUNCS` is equal to `funcs.len()`. There do not appear to be any issues with this patch, but it's important to verify that the changes made in this patch are not negatively impacting any other areas of the codebase.

### Commit 6
Here is the summary of changes:

- The patch updates a GitHub workflow file `rust-static-lib.yml`
- The line `rust: [1.64, 1.65, 1.66]` is replaced with `rust: [1.66, 1.67, 1.68]`
- The `--test-threads=1` flag is added to two `cargo test` commands

Potential problems:

- The patch should have a detailed explanation of why the changes were made to the workflow file.
- It looks like the changes are made for testing purposes. Still, any downstream implications should be considered, like the compatibility of these Rust versions with the project's dependencies.

### Commit 7
Based on the patch, it seems that the key change involves updating the test code in `wasmedge-sys/src/instance/function.rs`. Specifically, the loop iteration previously went up to 10, but now it goes up to 1,000,000. This suggests the test code is being made more robust by testing a larger number of cases.

One potential problem that could arise from this change is if the larger number of iterations causes the test to become too slow or resource-intensive. However, without knowledge of the context and test environment, I cannot determine whether or not this is a potential issue.

### Commit 8
This patch updates a GitHub workflow for a Rust bindings project. Specifically, it modifies the test command to include the "--test-threads=1" option for cargo test. This sets the number of threads used for running tests to 1, which could potentially slow down the testing process.

Overall, this patch seems straightforward and well-documented. However, if the project relies heavily on multithreaded code or has large test suites, limiting the number of threads could result in longer testing times.

apepkuss:
@hydai I added the option `--test-threads=1` for Fedora in the `bindings-rust` workflow. Sorry, I forgot it in the previous commits. Please help review it again. Thanks a lot!

hydai (approved):

apepkuss:
@hydai Thanks for the review!

juntao:
flows summarize again

alabulei1:

------

The reviews include patches with various updates and modifications to Rust libraries and workflows. The changes include adding new unit tests, modifying test assertions, updating CI workflows with flags and environment variables, and updating Rust versions. Overall, the patches seem reasonable and do not show any potential problems. However, reviewers recommend further testing to gauge the impact on performance, especially for large iteration counts.

## Details

### Commit 1
This patch makes some changes to the `function.rs` file of `wasmedge-sys`. The changes include removing some inappropriate code and adding a new unit test `test_crazy_function_creation()`.
 
Overall, the changes seem to be reasonable, and the new test is helpful to ensure that the system handles the creation of a large number of host functions well. However, reviewers might want to examine the test function more thoroughly. It creates one million host functions, which may be too large and slow down the system. It may be more appropriate to reduce this value to a smaller number or to use a test case using a randomly generated number of hosts.

### Commit 2
This pull request adds support for removing a real function, and also introduces a HashMap to store a mapping between the address of each host function pointer, and the key of the `HOST_FUNCS` HashMap. Potential issues to consider include: 

- There are no test cases included for the new removal feature. Tests should be inclusive of any new changes. 
- There could potential performance issues with the `HOST_FUNC_FOOTPRINTS` HashMap. Further testing may be required to gauge the impact on performance.

### Commit 3
This patch updates the test code for the Rust sys bindings in the `function.rs` file. Specifically, it modifies an assertion to check that the number of host functions registered in a struct is greater than or equal to the number of functions requested to be registered. As a reviewer, I don't see any potential problems in this change.

### Commit 4
This patch updates the bindings-rust GitHub workflows for testing and building the project to include the `--test-threads=1` flag to ensure thread safety. It also includes various environment variable exports to point to the correct directories. No potential problems were identified in this patch.

### Commit 5
This is a patch that updates the test code for a Rust library. Specifically, it modifies the function `tests()` in the `function.rs` file. The change updates the assertion so that `assert_eq!` is used instead of `assert!`. This is the only change in the patch, which inserts three lines of code and removes two lines of code in `function.rs`.

There are no identified potential problems in this patch. The change appears to be straightforward and should not cause any issues.

### Commit 6
This patch updates the CI workflow for building a Rust static library. The patch specifically updates the Rust version to 1.66, 1.67, and 1.68, and adds the `--test-threads=1` flag to the `cargo test` command.

There do not appear to be any potential problems with this patch. However, it is always a good idea to double-check that the new Rust versions are supported by the project's dependencies and that the test results remain consistent.

### Commit 7
This is a small fix that updates the function.rs file under the `src/instance` folder to modify the test code. Specifically, it is changing the number of iterations in a loop from 10 to 1,000,000. The purpose of this change is not clearly stated in the commit message. However, it looks like the author is testing performance improvements. 

There seem to be no potential problems with this change. However, it would be good to verify that the increased number of iterations does not cause any significant issues with the test.

### Commit 8
This patch updates the CI workflow for Rust bindings, specifically the `cargo test` command. It adds the `--test-threads=1` option to limit the number of threads used for testing. This is likely done to reduce resource utilization.

There do not appear to be any potential problems with this patch.

juntao:
flows summarize with GPT4

alabulei1:

------

Overall Summary:

The patches consist of modifications to the `wasmedge-sys` library, tests, and GitHub Actions workflow for Rust bindings. Key changes include the removal of code that checks the maximum capacity of host functions, the addition of a test case that creates a large number of host functions, updates to the test conditions and assertions, memory management improvements, and updates to the Rust version matrix and test execution in the CI workflow.

The potential issues and errors associated with these changes may include:

1. Unbounded scaling of host functions, which could lead to memory or storage issues.
2. A test case that consumes significant system resources due to the creation of numerous host functions under certain conditions.
3. Possibility of memory leaks related to real function removal and Mutex<HashMap> implementation.
4. Incompatibility with the new Rust versions in the version matrix.
5. Slower test performance due to running tests sequentially with `--test-threads=1`.
6. Changes in assertion conditions that may produce false positives or not catch issues and bugs.
7. Increased test execution time due to increased loop iteration count in a test case.

The most important findings include:

1. The main patch modifies the `wasmedge-sys` library, specifically targeting the `function.rs` module. It removes the code that checks if the number of host functions has reached the maximum capacity, potentially leading to memory or storage issues.
2. Memory management improvements have been made by adding support for removing real functions from the `HOST_FUNCS` HashMap and introducing `HOST_FUNC_FOOTPRINTS` Mutex<HashMap>.
3. Several patches update the tests, including changes in conditions and assertions, which may impact overall test efficacy.
4. GitHub Actions workflow has been updated to include Rust versions 1.66, 1.67, and 1.68 in the version matrix, and test execution has been modified to run with `--test-threads=1`.

It is recommended to carefully review these changes, considering the implications on compatibility, performance, and test efficacy. Thorough and comprehensive testing of the modifications should be performed to verify the patches' functionality and address potential issues.

## Details

### Commit 1
This patch is for the `wasmedge-sys` library, specifically targeting the `function.rs` module. The main changes in this patch include:

1. The removal of code that checks if the number of host functions has reached the maximum capacity. This could be a potential problem if the library user inadvertently creates too many host functions, causing memory or storage issues due to unbounded scaling.
   ```
  -if map_host_func.len() >= map_host_func.capacity() {
  -   return Err(Box::new(WasmEdgeError::Func(FuncError::CreateBinding(
  -       format!(
  -           "The number of the host functions reaches the upper bound: {}",
  -           map_host_func.capacity()
  -       ),
  -   ))));
   }
   ```

2. The addition of a new test, `test_crazy_function_creation`, that creates numerous host functions in a loop. While verifying the functionality of the updated code, this test could potentially consume a significant amount of system resources due to the large number of host functions created under certain conditions.

3. A minor import change in the tests module:
   ```
  -use crate::{types::WasmValue, Executor};
  +use crate::{types::WasmValue, Executor, HOST_FUNCS};
   ```

In summary, the main change in this patch is the removal of code that checks for the maximum capacity of host functions. As a result, potential issues may arise if the library user creates an excessive number of host functions. Additionally, the added test case could consume significant system resources.

### Commit 2
This patch adds support for removing real functions from the `HOST_FUNCS` HashMap and introduces a new `HOST_FUNC_FOOTPRINTS` Mutex<HashMap>. The significant changes and potential problems are summarized below:

Key Changes:
1. Added a real function removal code block in the `impl Drop for Function` implementation. This block exchanges a footnote about running the `ffi::WasmEdge_FunctionInstanceDelete(self.inner.0)` function for the host functions in this drop implementation.
2. Modified the test function `multiple_threads()` to check the number of host functions after creating the `FuncType`. 
3. Mutex<HashMap> `HOST_FUNC_FOOTPRINTS` has been added, which creates a mapping between addresses of host functions and their keys while registering them in the `HOST_FUNCS`.

Potential Problems:
1. The introduced mechanism for cleaning up allocations might not work correctly, causing a memory leak. It's necessary to thoroughly test functionality and memory usage.
2. The drop implementation for the `Function` struct might be problematic. If several other references are still tying up the inner WasmEdge function, the early destruction of `self.inner` could lead to unexpected behavior.

Overall, these changes aim to provide better support for managing memory when using real functions, but additional testing and verification must be done to ensure the functionality and memory usage are working as expected.

### Commit 3
Summary of key changes:

1. The patch updates the test code in `bindings/rust/wasmedge-sys/src/instance/function.rs`.
2. The `assert_eq!` macro has been replaced with the `assert!` macro, modifying the assertion condition.
3. The condition for the assertion has been updated from checking if the difference between the final and initial length of `HOST_FUNCS` is equal to the length of `funcs`, to checking if the difference is greater or equal.

Potential problems:

1. The change in the assertion may produce false positives, as it will pass even if the length difference is greater than the length of `funcs`.
2. The nature of the change in the test case condition implies that there could be other issues or bugs in the code that are not being caught by the broadened assertion.

### Commit 4
This patch modifies the GitHub Actions workflow file `bindings-rust.yml`. The key changes are:

- In the jobs `build_linux`, `build_macos`, and `build_windows`, the Rust tests have been updated to run:
  - with `--test-threads=1` flag for both workspace (`cargo test --workspace`) and examples (`cargo test --examples`)
  - with the previously used features: aot, async, ffi
  - with the previously used flags: --locked

Potential problems:

1. The test performance may be slower since the tests are now forced to run with only one thread (using `--test-threads=1`). The reasoning behind this change is not provided, and we cannot assess if this is the best approach. 
2. The patch does not have any accompanying tests or documentation updates to provide context for these changes.

### Commit 5
This patch updates a test in the `bindings/rust/wasmedge-sys/src/instance/function.rs` file.

Key changes:
1. The `assert!` statement is replaced with the more specific `assert_eq!` to make the assertion more precise.
2. The expected value and the actual value for comparison are updated, reflecting the number of functions in `HOST_FUNCS`.

Potential problems:
- There do not appear to be any apparent issues in the patch itself.
- Ensure that the change in assertion does not cause any unexpected side-effects or failures in related tests.

### Commit 6
Summary of key changes:

1. The Rust version matrix in the CI workflow has been updated from Rust versions 1.64, 1.65, and 1.66 to Rust versions 1.66, 1.67, and 1.68.
2. Two cargo test commands have been updated to include an additional flag: "--test-threads=1"

Potential problems:

1. Updating the Rust version matrix might cause compatibility issues or change the behavior of the code being tested. It is essential to consider whether the codebase is compatible with the new Rust versions or if any additional changes are needed.
2. The `--test-threads=1` flag's addition may affect testing performance as it forces the tests to run sequentially which may take longer compared to running the tests in parallel.

To minimize potential problems, ensure that the codebase is compatible with the new Rust versions and consider the implications of forcing the tests to run sequentially.

### Commit 7
This patch updates the test code in the `bindings/rust/wasmedge-sys/src/instance/function.rs` file. The key change is updating the loop iteration count in the tests from 10 to 1,000,000.

Summary of changes:
1. In the file `bindings/rust/wasmedge-sys/src/instance/function.rs`, the loop iteration count has been changed from `1..=10` to `1..=1_000_000`.

Potential problem:
1. Increasing the loop count to 1,000,000 may significantly increase the test execution time, which may not be ideal for a test case. It's essential to balance between test thoroughness and the time it takes to run the test suite.

Additional note:
Consider checking the rationale behind this change with the developer, as it may impact test execution duration.

### Commit 8
Summary:
This GitHub patch modifies the Rust bindings CI workflow file `bindings-rust.yml`. The author, Xin Liu, updates the workflow by adding a flag to run the cargo test jobs sequentially.

Key changes:
1. The `cargo test` commands have been updated to include the `--test-threads=1` flag.
   - The previous command for workspace tests: `cargo test --workspace --locked --features aot,async,ffi`
   - The updated command for workspace tests: `cargo test --workspace --locked --features aot,async,ffi -- --test-threads=1`
   - The previous command for examples tests: `cargo test --examples --locked --features aot,async,ffi`
   - The updated command for examples tests: `cargo test --examples --locked --features aot,async,ffi -- --test-threads=1`

Potential problems:
1. Running tests sequentially with `--test-threads=1` might slow down the CI process due to the lack of parallelism.
2. There are no comments justifying the change, so it might be difficult for other contributors to understand why this update was made.

juntao:
flows summarize with GPT4

alabulei1:

------

Overall Summary:

This set of patches involves modifications primarily to the Rust bindings for the wasmedge-sys component, the related CI workflow, and the test files. The primary focus of these changes is to improve memory management, remove limits on the number of host functions, and update assertions in tests.

Potential issues and errors:

1. Removing the check for host function capacity and the 'MAX_HOST_FUNC_LENGTH' environment variable could lead to excessive resource consumption, affecting system stability and performance.
2. The `test_crazy_function_creation` function may consume a significant amount of system resources and take a long time to finish. Additionally, increasing the loop count from 10 to 1,000,000 in function.rs might lead to performance concerns and resource exhaustion.
3. Changing the assertions in tests to use `assert!` instead of `assert_eq!` (and vice versa) could lead to altered test outcomes and potential false test failures.
4. Mutexes and RwLocks in the code can potentially cause deadlocks if not implemented carefully.
5. Updating the Rust CI workflow to restrict the number of test threads to 1 across all platforms could slow down the CI process.

Important Findings:

1. Memory management has been improved through the addition of support for removing real functions in Rust, and the creation of a footprint for host functions.
2. Changes have been made to the Rust bindings CI workflow, including updating the operating systems matrix and Rust compiler versions matrix, as well as modifying the cargo test commands to set the test-threads to 1.
3. The various changes to the test files involve updating assertions and modifying test scenarios, affecting the conditions that are being checked and the performance of the tests.

Given these changes, it's important to ensure that potential issues such as deadlocks, resource exhaustion, and false test failures are mitigated. Also, be cautious when removing limitations on host functions, as it could impact overall system stability and performance.

## Details

### Commit 1
This patch has made changes to the following file:
- bindings/rust/wasmedge-sys/src/instance/function.rs

Key changes:
1. Removal of the code block that checks if the number of host functions has reached its capacity. It returns an error if the condition is met.
2. Addition of `test_crazy_function_creation` function to test the function creation process for 1,000,000 times.

Potential problems:
1. Removing the capacity check may lead to a situation where too many host functions are created, consuming system resources.
2. The test `test_crazy_function_creation` could use system resources excessively and it may take a long time to finish.

### Commit 2
Summary of changes in GitHub Patch:

1. Added support for removing real function in Rust.
2. Created a footprint for the host function and used it for better memory management.
3. Updated the test to check added functionality and print some debug information.
4. Removed the environment variable 'MAX_HOST_FUNC_LENGTH' and made some changes in the `HOST_FUNCS` and `HOST_FUNC_FOOTPRINTS`.

Potential Problems:
1. Removing the 'MAX_HOST_FUNC_LENGTH' environment variable might have some side effects if it was used in other parts of the code that were not updated.
2. The Mutexes and RwLocks being used can cause potential deadlocks if not implemented carefully.
3. The environment variable was meant to put limits on HashMaps, potentially the removal could lead to unbounded growth.

Overall, the changes look reasonable and improve memory management by adding support for removing real functions. However, be cautious about potential deadlocks and make sure any usage of removed environment variables is accounted for.

### Commit 3
This patch modifies a Rust test file named `function.rs` under the `wasmedge-sys` directory.

Key changes:
1. In the test module, the `assert_eq!()` macro is replaced with an `assert!()` macro.
2. The updated assertion checks that the difference between the initial number of host functions (`start_num`) and the final number of host functions is greater than or equal to the number of functions.

Potential problems:
1. If the intention was to check for an exact match in the number of new host functions, then using `assert_eq!()` would be more appropriate.
2. It might be helpful to add a comment explaining why the change from the `assert_eq!()` macro to the `assert!()` macro was made, since the context is not clear from the patch alone.

### Commit 4
The key changes in this patch are related to the Rust bindings Continuous Integration (CI) workflow. The file `.github/workflows/bindings-rust.yml` has been modified. The patch affects the testing steps of MacOS, Windows, and Fedora builds by adding the `--test-threads=1` option.

Summary of changes:
1. In the MacOS build job, `cargo test` and `cargo test --examples` have been updated with `--test-threads=1` flag.
2. Similarly, in the Windows build job, `cargo test` and `cargo test --examples` have been updated by adding `--test-threads=1` flag.
3. Lastly, in the Fedora build job, `cargo test --workspace` and `cargo test --examples` have been modified to include the `--test-threads=1` flag.

The potential problems that can arise from these changes:
- The `--test-threads=1` flag will make the tests run sequentially, which could significantly slow down the CI process. Make sure this trade-off in testing speed is acceptable.

### Commit 5
Summary of Key Changes:
1. In the `bindings/rust/wasmedge-sys/src/instance/function.rs` file, the `assert!` macro used in the test module has been replaced with the `assert_eq!` macro.
2. The old `assert!` checked if `HOST_FUNCS.read().len() - start_num` is greater than or equal to `funcs.len()`. The new `assert_eq!` checks for equality between `HOST_FUNCS.read().len() - start_num` and `funcs.len()`.

Potential Problems:
1. The change in the assertion condition might cause the test to fail in some scenarios where it previously passed. If the original logic was correct, and the `>=` condition was required, this change might lead to false test failures.
2. If there are any other dependencies or other parts of the code that rely on the previous behavior, it could lead to issues in those areas as well.

### Commit 6
Summary:

This patch updates the Rust CI workflow for building a static library.

Key changes:
1. Updated the operating systems matrix: ubuntu-20.04, ubuntu-22.04
2. Updated the Rust compiler versions matrix: 1.66, 1.67, 1.68
3. Modified the `cargo test` commands to set the test-threads to 1.

Potential problems:
- If any of the updated Rust compiler versions introduces breaking changes, it might cause issues with the project's dependencies and the codebase. Testing on multiple versions of Rust is important, but it is necessary to keep an eye on compatibility issues.
- Setting the test-threads to 1 might cause the tests to run slower, as it is not leveraging the available parallelism. It might be suitable if tests are interfering with each other, but if there is no such reason, running tests with the default test-threads might be faster.

### Commit 7
Summary of Key Changes:
- In the bindings/rust/wasmedge-sys/src/instance/function.rs file, the loop counter for creating instance of FuncType has been changed from a range of 1 to 10 to 1 to 1,000,000.

Potential Problems:
1. Performance Concern: Increasing the loop count from 10 to 1,000,000 may significantly increase the test execution time, especially if the creation of FuncType instances is expensive (memory or computation-wise).
2. Resource Exhaustion: Creating 1,000,000 FuncType instances may cause a large memory allocation, which could lead to problems if the system has limited resources. Consider checking for any memory constraints and assess whether it's necessary to create such a large number of instances.

### Commit 8
Summary of Key Changes:
1. Modified `.github/workflows/bindings-rust.yml` to restrict the number of test threads to 1 when running `cargo test` and `cargo test --examples` commands for the Rust bindings.

Potential problems:
1. The `--test-threads=1` option potentially slows down the test execution time as it forces `cargo test` to run in a single-threaded mode rather than using multiple threads.

PR #2410, opened 2023-04-15
[WASI] Enable readonly filesystem

hydai:


juntao:
Flows review

juntao:
Flows review

juntao:
Hello, I am a [serverless review bot](https://github.com/flows-network/github-pr-review/) on [flows.network](https://flows.network/). Here are my reviews of changed source code files in this PR.

------

## [lib/driver/runtimeTool.cpp](https://github.com/WasmEdge/WasmEdge/blob/b53cade5dd71ba032a4322f57507a8b880be031e/lib%2Fdriver%2FruntimeTool.cpp)

Overall, the code snippet seems to be well-structured and follows good coding practices. However, there are a few potential issues to consider:

1. There is no context provided on the purpose or the output of the "Tool" function, making it difficult to understand the overall functionality.

2. The use of "noexcept" in the function definition is not necessary since it does not throw any exceptions.

3. The use of "using namespace std::literals" is potentially problematic as it can cause naming conflicts and make the code harder to read and understand.

4. The lack of error handling for plugin loading in the "Tool" function could lead to unexpected behavior if a plugin fails to load.

5. The use of the "stdout" stream for error messages in the "Parser.parse" function call could cause confusion as the user would expect to see error messages on the standard error stream.

6. There is no error handling in case the "--dir" option is misused, which could lead to undefined behavior during execution.

7. There is no documentation on the possible values of the "ConfEnableAllStatistics" option and what they mean, making it difficult for the user to decide if they want to enable it or not.

The patch includes the addition of a new feature for the "--dir" option, which allows the user to specify a read-only mode for the guest directory. This is achieved by adding ":readonly" to the "--dir" option. The patch also adds corresponding documentation on how to use this new feature.

## [lib/host/wasi/environ.cpp](https://github.com/WasmEdge/WasmEdge/blob/b53cade5dd71ba032a4322f57507a8b880be031e/lib%2Fhost%2Fwasi%2Fenviron.cpp)

The code snippet appears to define several constants and inline functions related to the WASI (WebAssembly System Interface) environment. There is also an implementation of two functions, `init` and `fini`, both belonging to the `Environ` class, which are defined within the `WASI` and `Host` namespaces. These functions seem to initialize the WASI environment using information about the specified directories, program arguments, and environment variables.

The code snippet itself appears to be fine, without any obvious issues that could cause errors or undefined behavior. However, without the context of the larger program and system, it is difficult to assess if the implementation is indeed correct and functional. Therefore, it is recommended to perform further testing and debugging to ensure the functionality of the code before integration.

The patch adds two new constant inline functions `kPreOpenBaseRightsReadOnly` and `kPreOpenInheritingRightsReadOnly` which are read-only variants of the `kPreOpenBaseRights` and `kPreOpenInheritingRights`. It also adds a new constant `kReadOnly` representing the string "readonly".

In the `init` function, the patch handles the readonly flag by checking if `Mode` is equal to `kReadOnly`, which indicates that the host directory should be read-only. It then sets the `ReadOnly` flag accordingly.

Finally, the `const auto BaseRights` and `InheritingRights` are set according to whether the directory is read-only or not, and are then passed as arguments to `VINode::bind`.

q82419 (dismissed):

q82419 (approved):

PR #2431, opened 2023-04-24
[WASI Socket][Don't merge until the discussion is finalized] Provide multiple WASI socket API implementation

hydai:
To make our previous WASI socket functions can be compatible with the WASI spec one and the new WasmEdge WASI socket functions, we use a whitelist mechanism to filter those functions that have the same name but with different function signatures. However, this may increase the maintenance cost, so we are going to deprecate it in a future version.

Please notice that this is a workaround, and this behavior is not allowed by the Wasm spec. This is a final PR before the 0.12 release. So we should make sure that CI is grean and everything is compatible when enabling this feature.

### DIFF
--- a/.CurrentChangelog.md
+++ b/.CurrentChangelog.md
@@ -1,4 +1,4 @@
-### 0.12.0-alpha.2 (2023-02-24)
+### 0.12.0 (2023-04-24)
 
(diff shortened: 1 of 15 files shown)


alabulei1:

------

PR #2434, opened 2023-04-25
[CI] Not to test WASI when releasing.

q82419:


alabulei1:

------

### [Commit 7768f89445fbe0250876b5a7ded14093eeec6a36](https://github.com/WasmEdge/WasmEdge/pull/2434/commits/7768f89445fbe0250876b5a7ded14093eeec6a36)
The patch adds branch filtering to the on: push and on: pull_request actions. Specifically, it only runs the workflow for the master branch. This change allows the WASI test suite to run on branches other than master without being tested. Since the workflow is expensive, this provides faster feedback to the developers. 

One potential problem with this patch is that if there is a bug in the workflow, it may be missed until the next release, causing potential problems for users. Additionally, WASI functionality on non-master branches may not be adequately tested, leading to bugs when merging upstream into the master branch. It may be worth considering alternative solutions, such as decreasing the frequency or the depth of testing on non-master branches, instead of entirely skipping testing.

hydai (approved):

Issue #2335, opened 2023-03-04
Support exception handling proposal


sdeleuze:
Related to #1122 as it is usually required by GCed language like Kotlin or Java, supporting exception handling proposal is necessary to run Kotin/Wasm in WasmEdge.

An easier first step would allow to parse WebAssembly bytecode built with EH proposal enabled, in order to support positive scenarios that do not throw exceptions. That would unlock a lot of potential experimentation with Kotlin/Wasm, Dart, etc.

harry900831:
I'll do this!

hydai:
@q82419 We supported the EH proposal in the 0.14.0 release. Could you please give a summary of the current implementation status?

q82419:
Finished.

PR #2459, opened 2023-05-01
[Proposal] Add exception handling proposal on interpreter mode

harry900831:
#2335 

## Note
We may now generated the wasm with exception handling by `emcc -O1 -fwasm-exceptions a.cpp -o a.wasm` and test it on the interpreter mode.

<details> 
  <summary> Example </summary>

```cpp
#include <cstdio>
#include <stdexcept>

double divideNumbers(double num1, double num2) {
    if (num2 == 0) {
        throw std::invalid_argument("Error: Division by zero");
    }
    if (num1 < 0 || num2 < 0) {
        throw std::domain_error("Error: Negative number");
    }
    return num1 / num2;
}

void testDivideNumbers(double num1, double num2) {
    try {
        double result = divideNumbers(num1, num2);
        printf("The result is: %f\n", result);
    } catch (const std::invalid_argument& e) {
        printf("Caught an invalid_argument exception in testDivideNumbers: %s\n", e.what());
        throw; // re-throwing the exception
    } catch (const std::domain_error& e) {
        printf("Caught a domain_error exception in testDivideNumbers: %s\n", e.what());
        throw std::runtime_error("New runtime_error from testDivideNumbers");
    }
}

void anotherTest(double num1, double num2) {
    try {
        testDivideNumbers(num1, num2);
    } catch (const std::invalid_argument& e) {
        printf("Caught an invalid_argument exception in anotherTest: %s\n", e.what());
        throw std::out_of_range("New out_of_range from anotherTest");
    } catch (const std::runtime_error& e) {
        printf("Caught a runtime_error exception in anotherTest: %s\n", e.what());
    }
}

int main() {
    try {
	double num1 = 10.0;
	double num2 = 3.0;
        anotherTest(num1, num2);
    } catch (const std::out_of_range& e) {
        printf("Caught an out_of_range exception in main: %s\n", e.what());
    } catch (...) {
        printf("Caught an unknown exception in main.\n");
    }

    try {
	double num1 = -10.0;
	double num2 = 0.0;
        anotherTest(num1, num2);
    } catch (const std::out_of_range& e) {
        printf("Caught an out_of_range exception in main: %s\n", e.what());
    } catch (...) {
        printf("Caught an unknown exception in main.\n");
    }

    try {
	double num1 = -10.0;
	double num2 = 3.0;
        anotherTest(num1, num2);
    } catch (const std::out_of_range& e) {
        printf("Caught an out_of_range exception in main: %s\n", e.what());
    } catch (...) {
        printf("Caught an unknown exception in main.\n");
    }
    return 0;
}
```

Output:
```
The result is: 3.333333
Caught an invalid_argument exception in testDivideNumbers: Error: Division by zero
Caught an invalid_argument exception in anotherTest: Error: Division by zero
Caught an out_of_range exception in main: New out_of_range from anotherTest
Caught a domain_error exception in testDivideNumbers: Error: Negative number
Caught a runtime_error exception in anotherTest: New runtime_error from testDivideNumbers
```

</details>

However, if I replace printf by cout, the execution may crash. It seems that cout may contain several complicated exception handling block. I think we can merge this first and solve this issue in the future.

## References
- [Webassembly Specifications with exception handling](https://webassembly.github.io/exception-handling/)
- [Exception handling proposal](https://github.com/WebAssembly/exception-handling/blob/main/proposals/exception-handling/Exceptions.md)

### DIFF
--- a/include/ast/instruction.h
+++ b/include/ast/instruction.h
@@ -28,8 +28,10 @@ class Instruction {
 public:
   struct JumpDescriptor {
     uint32_t TargetIndex;
-    uint32_t StackEraseBegin;
-    uint32_t StackEraseEnd;
+    uint32_t ValueStackEraseBegin;
+    uint32_t ValueStackEraseEnd;
+    uint32_t HandlerStackOffset;
+    uint32_t CaughtStackOffset;
     int32_t PCOffset;
   };
 
@@ -46,12 +48,15 @@ public:
 #endif
     Flags.IsAllocLabelList = false;
     Flags.IsAllocValTypeList = false;
+    Flags.IsTryLast = false;
+    Flags.IsCatchLast = false;
+    Flags.IsDelegate = false;
   }
 
   /// Copy constructor.
   Instruction(const Instruction &Instr)
       : Data(Instr.Data), Offset(Instr.Offset), Code(Instr.Code),
-        Flags(Instr.Flags) {
+        Flags(Instr.Flags), JumpCatch(Instr.JumpCatch) {
     if (Flags.IsAllocLabelList) {
       Data.BrTable.LabelList = new JumpDescriptor[Data.BrTable.LabelListSize];
       std::copy_n(Instr.Data.BrTable.LabelList, Data.BrTable.LabelListSize,
@@ -66,7 +71,7 @@ public:
   /// Move constructor.
   Instruction(Instruction &&Instr)
       : Data(Instr.Data), Offset(Instr.Offset), Code(Instr.Code),
-        Flags(Instr.Flags) {
+        Flags(Instr.Flags), JumpCatch(Instr.JumpCatch) {
     Instr.Flags.IsAllocLabelList = false;
     Instr.Flags.IsAllocValTypeList = false;
   }
@@ -90,12 +95,21 @@ public:
   uint32_t getOffset() const noexcept { return Offset; }
 
   /// Getter and setter of block type.
-  BlockType getBlockType() const noexcept { return Data.Blocks.ResType; }
-  void setBlockType(ValType VType) noexcept {
-    Data.Blocks.ResType.setData(VType);
+  BlockType getBlockType() const noexcept {
+    return Code == OpCode::Try ? Data.TryBlock.ResType : Data.Blocks.ResType;
   }
-  void setBlockType(uint32_t Idx) noexcept { Data.Blocks.ResType.setData(Idx); }
-  void setEmptyBlockType() noexcept { Data.Blocks.ResType.setEmpty(); }
+  void setBlockType(ValType VType) noexcept {
+    Code == OpCode::Try ? Data.TryBlock.ResType.setData(VType)
+                        : Data.Blocks.ResType.setData(VType);
+  }
+  void setBlockType(uint32_t Idx) noexcept {
+    Code == OpCode::Try ? Data.TryBlock.ResType.setData(Idx)
+                        : Data.Blocks.ResType.setData(Idx);
+  }
+  void setEmptyBlockType() noexcept {
+    Code == OpCode::Try ? Data.TryBlock.ResType.setEmpty()
+                        : Data.Blocks.ResType.setEmpty();
+  }
 
   /// Getter and setter of jump count to End instruction.
   uint32_t getJumpEnd() const noexcept { return Data.Blocks.JumpEnd; }
@@ -179,6 +193,64 @@ public:
   uint8_t getMemoryLane() const noexcept { return Data.Memories.MemLane; }
   uint8_t &getMemoryLane() noexcept { return Data.Memories.MemLane; }
 
+  /// Getter and setter of jump count to each catch instruction.
+  const std::vector<uint32_t> &getTryBlockJumpCatch() const noexcept {
+    return JumpCatch;
+  }
+  std::vector<uint32_t> &getTryBlockJumpCatch() noexcept { return JumpCatch; }
+
+  /// Getter and setter of jump count to catch_all instruction.
+  uint32_t getTryBlockJumpCatchAll() const noexcept {
+    return Data.TryBlock.JumpCatchAll;
+  }
+  void setTryBlockJumpCatchAll(const uint32_t Cnt) noexcept {
+    Data.TryBlock.JumpCatchAll = Cnt;
+  }
+
+  /// Getter and setter of jump count to the end of try block.
+  uint32_t getTryBlockJumpEnd() const noexcept { return Data.TryBlock.JumpEnd; }
+  void setTryBlockJumpEnd(const uint32_t Cnt) noexcept {
+    Data.TryBlock.JumpEnd = Cnt;
+  }
+
+  /// Getter and setter of number of block type parameter.
+  uint32_t getTryBlockBlockParamNum() const noexcept {
+    return Data.TryBlock.BlockParamNum;
+  }
+  void setTryBlockBlockParamNum(const uint32_t Num) noexcept {
+    Data.TryBlock.BlockParamNum = Num;
+  }
+
+  /// Getter and setter of jump count to delegate instruction.
+  uint32_t getTryBlockDelegate() const noexcept {
+    return Data.TryBlock.DelegateIdx;
+  }
+  void setTryBlockDelegate(const uint32_t Cnt) noexcept {
+    Data.TryBlock.DelegateIdx = Cnt;
+  }
+
+  /// Getter and setter of VSize
+  uint32_t getTryBlockVSize() const noexcept { return Data.TryBlock.VSize; }
+  void setTryBlockVSize(const uint32_t Num) noexcept {
+    Data.TryBlock.VSize = Num;
+  }
+
+  /// Getter and setter of HOffset
+  uint32_t getTryBlockHOffset() const noexcept { return Data.TryBlock.HOffset; }
+  void setTryBlockHOffset(const uint32_t Num) noexcept {
+    Data.TryBlock.HOffset = Num;
+  }
+
+  /// Getter and setter of COffset
+  uint32_t getTryBlockCOffset() const noexcept { return Data.TryBlock.COffset; }
+  void setTryBlockCOffset(const uint32_t Num) noexcept {
+    Data.TryBlock.COffset = Num;
+  }
+
+  /// Getter of tag index.
+  uint32_t getTagIdx() const noexcept { return Data.Tag.TagIdx; }
+  uint32_t &getTagIdx() noexcept { return Data.Tag.TagIdx; }
+
   /// Getter and setter of the constant value.
   ValVariant getNum() const noexcept {
 #if defined(__x86_64__) || defined(__aarch64__) ||                             \
@@ -197,6 +269,18 @@ public:
     std::memcpy(&Data.Num, &N.get<uint128_t>(), sizeof(uint128_t));
 #endif
   }
+
+  /// Getter and setter of IsLast for End instruction.
+  bool isTryLast() const noexcept { return Flags.IsTryLast; }
+  void setTryLast(bool Last = true) noexcept { Flags.IsTryLast = Last; }
+
+  /// Getter and setter of IsLast for End instruction.
+  bool isCatchLast() const noexcept { return Flags.IsCatchLast; }
+  void setCatchLast(bool Last = true) noexcept { Flags.IsCatchLast = Last; }
+
+  /// Getter and setter of IsLast for End instruction.
+  bool isDelegate() const noexcept { return Flags.IsDelegate; }
+  void setDelegate(bool Last = true) noexcept { Flags.IsDelegate = Last; }
 
 private:
   /// Release allocated resources.
@@ -268,13 +352,34 @@ private:
 #endif
     // Type 9: IsLast.
     bool IsLast;
+    // Type 10: Try Block
+    struct {
+      uint32_t JumpCatchAll;
+      uint32_t JumpEnd;
+      uint32_t BlockParamNum;
+      uint32_t DelegateIdx;
+      uint32_t VSize;
+      uint32_t HOffset;
+      uint32_t COffset;
+      BlockType ResType;
+    } TryBlock;
+    // Type 11: Tag
+    struct {
+      uint32_t TagIdx;
+    } Tag;
   } Data;
   uint32_t Offset = 0;
   OpCode Code = OpCode::End;
   struct {
     bool IsAllocLabelList : 1;
     bool IsAllocValTypeList : 1;
+    bool IsTryLast : 1;
+    bool IsCatchLast : 1;
+    bool IsDelegate : 1;
   } Flags;
+
+  // TODO: Need refactor
+  std::vector<uint32_t> JumpCatch;
   /// @}
 };
 
--- a/include/ast/instruction.h
+++ b/include/ast/instruction.h
@@ -207,7 +207,7 @@ public:
   void setJumpCatchList(const std::vector<uint32_t> &CatchList) {
     reset();
     if (!CatchList.empty()) {
-      Data.TryBlock.JumpCatchListSize = CatchList.size();
+      Data.TryBlock.JumpCatchListSize = static_cast<uint32_t>(CatchList.size());
       Data.TryBlock.JumpCatchList = new uint32_t[CatchList.size()];
       Flags.IsAllocJumpCatchList = true;
       std::copy_n(CatchList.begin(), CatchList.size(),
--- a/include/ast/instruction.h
+++ b/include/ast/instruction.h
@@ -247,10 +247,6 @@ public:
     Data.TryBlock.COffset = Num;
   }
 
-  /// Getter of tag index.
-  uint32_t getTagIdx() const noexcept { return Data.Tag.TagIdx; }
-  uint32_t &getTagIdx() noexcept { return Data.Tag.TagIdx; }
-
   /// Getter and setter of the constant value.
   ValVariant getNum() const noexcept {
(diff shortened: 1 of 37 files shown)


juntao:

test/spec/spectest.cpp
```diff
@@ -652,6 +666,18 @@ void SpecTest::run(std::string_view Proposal, std::string_view UnitName) {
...
+        const auto &Action = Cmd["action"s];
+        const auto ActType = Action["type"].Get<std::string>();
+        const uint64_t LineNumber = Cmd["line"].Get<uint64_t>();
+        // TODO: Check expected exception type
+        if (ActType == "invoke"sv) {
+          ExceptionInvoke(Action, LineNumber);
+          return;
+        }
+        EXPECT_TRUE(false);
+        return;
```
harry900831, line 679: According to the spec test json file, there's an "expected" section to check the type of uncaught exception. However, it would be quite unclean to deal with it under the current code structure if we want to get the type of the uncaught exception. Therefore, I think it's enough for us to check whether there's an uncaught exception.

include/ast/instruction.h
```diff
@@ -268,13 +352,34 @@ class Instruction {
...
   struct {
     bool IsAllocLabelList : 1;
     bool IsAllocValTypeList : 1;
+    bool IsTryLast : 1;
+    bool IsCatchLast : 1;
+    bool IsDelegate : 1;
   } Flags;
+
+  // TODO: Need refactor
+  std::vector<uint32_t> JumpCatch;
```
harry900831, line 382: I would like to ask that is there any better way to deal with this vector? We have to know how many catch clause after loading the entire try-catch-end block, so it can't readily use an uint32_t* inside union to store it.

include/runtime/stackmgr.h
```diff
@@ -35,7 +37,29 @@ class StackManager {
...
   };
 
-  using Value = ValVariant;
+  struct Handler {
+    Handler(AST::InstrView::iterator E, uint32_t V, uint32_t F, uint32_t H,
+            uint32_t C)
+        : CatchCaluse(), EndIt(E), VPos(V), FPos(F), HPos(H), CPos(C) {}
+    // nullptr stands for catch all clause
+    std::vector<
+        std::pair<Runtime::Instance::TagInstance *, AST::InstrView::iterator>>
```
harry900831, line 46: I'm not sure whether it's a proper way to distinguish the tag by the address. The current implementation may compare two tag address to indicate whether they are the same tag.

harry900831:
The current implementation may pass the exception handling spec test. But, it may failed on the following test right now:
- wasmedgeAOTCoreTests
- wasmedgeExecutorCoreTests
- wasmedgeAPIVMCoreTests
- wasmedgeAPIStepsCoreTests
- wasmedgeAPIAOTCoreTests

For the AOT test, I believe it's reasonable since I hasn't implemented AOT yet.

For the rest of the tests, they all failed on the same part:

```
[2023-05-22 04:57:17.976] [error] loading failed: unexpected end, Code: 0x22
[2023-05-22 04:57:17.976] [error]     Bytecode offset: 0x0000000e
[2023-05-22 04:57:17.976] [error]     At AST node: tag section
[2023-05-22 04:57:17.976] [error]     At AST node: import section
[2023-05-22 04:57:17.976] [error]     At AST node: module
[2023-05-22 04:57:17.976] [error]     File name: "../spec/testSuites/core/binary/binary.139.wasm"
[2023-05-22 04:57:17.976] [error]    ##### expected text : malformed import kind
[2023-05-22 04:57:17.976] [error]    ######## error text : unexpected end
/Users/hchiang/WasmEdge/test/spec/spectest.cpp:530: Failure
Value of: stringContains(Text, WasmEdge::ErrCodeStr[Res.error().getEnum()])
  Actual: false
Expected: true
[2023-05-22 04:57:17.976] [error] loading failed: unexpected end, Code: 0x22
[2023-05-22 04:57:17.976] [error]     Bytecode offset: 0x0000000f
[2023-05-22 04:57:17.976] [error]     At AST node: tag section
[2023-05-22 04:57:17.977] [error]     At AST node: import section
[2023-05-22 04:57:17.977] [error]     At AST node: module
[2023-05-22 04:57:17.977] [error]     File name: "../spec/testSuites/core/binary/binary.140.wasm"
[2023-05-22 04:57:17.977] [error]    ##### expected text : malformed import kind
[2023-05-22 04:57:17.977] [error]    ######## error text : unexpected end
/Users/hchiang/WasmEdge/test/spec/spectest.cpp:530: Failure
Value of: stringContains(Text, WasmEdge::ErrCodeStr[Res.error().getEnum()])
  Actual: false
Expected: true
```

It seems that it's because the binary format has changed, the malformed part may be distinguished as another type of error. I hasn't come up with a good work around. Do yo have any idea?

harry900831:
Another question, do I have to add several tag related function to WasmEdge C API? For example: `WasmEdge_TagTypeCreate()`, `WasmEdge_TagInstanceGetTagType()`, .......

include/ast/instruction.h
hydai: What's the real types used in this vector?

hydai:
> The current implementation may pass the exception handling spec test. But, it may failed on the following test right now:
> 
> * wasmedgeAOTCoreTests
> * wasmedgeExecutorCoreTests
> * wasmedgeAPIVMCoreTests
> * wasmedgeAPIStepsCoreTests
> * wasmedgeAPIAOTCoreTests
> 
> For the AOT test, I believe it's reasonable since I hasn't implemented AOT yet.

If we are going to implement the AOT part, then we should have this PR as a standalone branch until all tests are green.


> For the rest of the tests, they all failed on the same part:
> 
> ```
> [2023-05-22 04:57:17.976] [error] loading failed: unexpected end, Code: 0x22
> [2023-05-22 04:57:17.976] [error]     Bytecode offset: 0x0000000e
> [2023-05-22 04:57:17.976] [error]     At AST node: tag section
> [2023-05-22 04:57:17.976] [error]     At AST node: import section
> [2023-05-22 04:57:17.976] [error]     At AST node: module
> [2023-05-22 04:57:17.976] [error]     File name: "../spec/testSuites/core/binary/binary.139.wasm"
> [2023-05-22 04:57:17.976] [error]    ##### expected text : malformed import kind
> [2023-05-22 04:57:17.976] [error]    ######## error text : unexpected end
> /Users/hchiang/WasmEdge/test/spec/spectest.cpp:530: Failure
> Value of: stringContains(Text, WasmEdge::ErrCodeStr[Res.error().getEnum()])
>   Actual: false
> Expected: true
> [2023-05-22 04:57:17.976] [error] loading failed: unexpected end, Code: 0x22
> [2023-05-22 04:57:17.976] [error]     Bytecode offset: 0x0000000f
> [2023-05-22 04:57:17.976] [error]     At AST node: tag section
> [2023-05-22 04:57:17.977] [error]     At AST node: import section
> [2023-05-22 04:57:17.977] [error]     At AST node: module
> [2023-05-22 04:57:17.977] [error]     File name: "../spec/testSuites/core/binary/binary.140.wasm"
> [2023-05-22 04:57:17.977] [error]    ##### expected text : malformed import kind
> [2023-05-22 04:57:17.977] [error]    ######## error text : unexpected end
> /Users/hchiang/WasmEdge/test/spec/spectest.cpp:530: Failure
> Value of: stringContains(Text, WasmEdge::ErrCodeStr[Res.error().getEnum()])
>   Actual: false
> Expected: true
> ```
> 
> It seems that it's because the binary format has changed, the malformed part may be distinguished as another type of error. I hasn't come up with a good work around. Do yo have any idea?

Does the binary format change during this proposal?

hydai:
> Another question, do I have to add several tag related function to WasmEdge C API? For example: `WasmEdge_TagTypeCreate()`, `WasmEdge_TagInstanceGetTagType()`, .......

Yes.

test/spec/spectest.cpp
harry900831: I'm not sure whether this is require by the spec. 

Below is the example json config for spec test. If I understand correctly, "type" inside "expected" is the type of the value associated with the exception. Our implementation to handle "expected" actually suppose there should be a "value" after "type" just like the below "assert_return". Furthermore, "assert_trap" also have the "expected" section, but it seems that our spectest.cpp doesn't deal with it either. So, I believe it's acceptable to not deal with the "expected" section with "assert_exception".
```
  {"type": "assert_trap", "line": 160, "action": {"type": "invoke", "field": "trap-in-callee", "args": [{"type": "i32", "value": "1"}, {"type": "i32", "value": "0"}]}, "text": "integer divide by zero", "expected": [{"type": "i32"}]}, 
  {"type": "assert_return", "line": 162, "action": {"type": "invoke", "field": "catch-complex-1", "args": [{"type": "i32", "value": "0"}]}, "expected": [{"type": "i32", "value": "3"}]}, 
  {"type": "assert_return", "line": 163, "action": {"type": "invoke", "field": "catch-complex-1", "args": [{"type": "i32", "value": "1"}]}, "expected": [{"type": "i32", "value": "4"}]}, 
  {"type": "assert_exception", "line": 164, "action": {"type": "invoke", "field": "catch-complex-1", "args": [{"type": "i32", "value": "2"}]}, "expected": [{"type": "i32"}]}, 
  {"type": "assert_return", "line": 166, "action": {"type": "invoke", "field": "catch-complex-2", "args": [{"type": "i32", "value": "0"}]}, "expected": [{"type": "i32", "value": "3"}]}, 
  {"type": "assert_return", "line": 167, "action": {"type": "invoke", "field": "catch-complex-2", "args": [{"type": "i32", "value": "1"}]}, "expected": [{"type": "i32", "value": "4"}]}, 
  {"type": "assert_exception", "line": 168, "action": {"type": "invoke", "field": "catch-complex-2", "args": [{"type": "i32", "value": "2"}]}, "expected": [{"type": "i32"}]}, 
```

include/ast/instruction.h
harry900831: Do you mean what value does this vector store? It is used only on try instruction, it stores the index offset on the instruction vector for each catch clause associated with it.

harry900831: One possible way to refactor is to imitate the way to store the lablelist for brtable instruction.

https://github.com/WasmEdge/WasmEdge/blob/9c6d6eef435907e87a64b90e9ae0f8461cb253f3/include/ast/instruction.h#L113-L130

But, we have to go through to the end of the try block to know the size we need for the array to store the index offset (Handling raw pointer is dangerous too). I'm wondering whether there's any better solution.

harry900831:
> If we are going to implement the AOT part, then we should have this PR as a standalone branch until all tests are green.

Yes, so this would be merged into [WasmEdge:proposal/exception-handling](https://github.com/WasmEdge/WasmEdge/tree/proposal/exception-handling)

> Does the binary format change during this proposal?

To be more specific, some wasm may be parsed successfully after enabling exception handling while they are malformed before.

The current problem is that our implementation may generate a different error message compare to the spec test. I believe it's because part of the malformed part become valid after enabling exception handling proposal, but it's still malformed. And I believe the error message generated by the current implementation is reasonable. We can check the failed binary.139.wasm and binary.140.wasm by wasm2wat. It has the different error message too.

image

q82419 (changes requested):
Suggestion of the tag section:

* TagSection: vector of TagType
* TagType: 0x00 (uint8_t in current), TypeIndex (uint32_t), FunctionType*
   - At the end of the module loading, find the pointer of function types and set into the content of tag section.
* TagInstanance: TagType

include/ast/type.h
```diff
@@ -206,5 +206,38 @@ class GlobalType {
...
+  // Getter of the size of value that is associated with the tag
+  size_t getAssocValSize() const noexcept {
+    return Type.getParamTypes().size();
+  }
+
+private:
+  FunctionType Type;
+};
+
+class Tag {
```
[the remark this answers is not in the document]
harry900831: > TagSection: vector of TagType
> TagType: 0x00 (uint8_t in current), TypeIndex (uint32_t), FunctionType*
> At the end of the module loading, find the pointer of function types and set into the content of tag section.
> TagInstanance: TagType

Good idea!
One last concern, would it be better to set the pointer of function types to the tag type on validation phase? Since we have to validate the type idx of a tag to set the pointer of function types.

include/ast/instruction.h
```diff
@@ -28,8 +28,10 @@ class Instruction {
 public:
   struct JumpDescriptor {
     uint32_t TargetIndex;
-    uint32_t StackEraseBegin;
-    uint32_t StackEraseEnd;
+    uint32_t ValueStackEraseBegin;
```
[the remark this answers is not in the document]
harry900831: size of Instruction would be from 32 to 48 if we ignore the size of the below TryBlock inside union.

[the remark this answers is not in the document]
harry900831: According to the offline discussion, we may ignore the size problem over here.

include/ast/instruction.h
```diff
@@ -268,13 +352,34 @@ class Instruction {
 #endif
     // Type 9: IsLast.
     bool IsLast;
+    // Type 10: Try Block
+    struct {
+      uint32_t JumpCatchAll;
```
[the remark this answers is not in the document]
harry900831: Ditto

hydai:
Hi @harry900831 
Since we are going to put the exception handling proposal into 0.14.x release, we will not merge it in the master branch. Instead, we will merge it into the `proposal/exception-handling` branch. So you can keep going on the AOT part.

lib/driver/runtimeTool.cpp
```diff
@@ -80,6 +80,9 @@ int Tool(int Argc, const char *Argv[]) noexcept {
       PO::Description("Enable Extended-const proposal"sv));
   PO::Option<PO::Toggle> PropThreads(
       PO::Description("Enable Threads proposal"sv));
+  PO::Option<PO::Toggle> PropExceptionHandling(
+      PO::Description("Enable Exception Handling"sv));
```
harry900831, line 84: Enable Exception handling proposal

hydai:
@harry900831 
We will merge this PR first. The typed function reference proposal is for 0.14.0. Due to several breaking changes, we plan to have it at the end of Sep.

q82419:
> @hydai @q82419 The AOT mode for exception handling proposal is unable to be completed in a few days, will we merge this first?
> 
> Furthermore, is it going to be merge into master? I'm afraid that there may be several conflicts with the latest master right now or even after #2691 be merged.

I'll rebase the `proposal/exception-handling` onto the latest master first.
You can only considering the latest master branch only.
Please help to resolve the conflicts. Thanks!

harry900831:
I accidentally found a small bug during rebase, I'm still checking it.

harry900831:
> However, if I replace printf by cout, the execution may crash. It seems that cout may contain several complicated exception handling block. I think we can merge this first and solve this issue in the future.

An update for this issue.

```cpp
// a.cpp
#include <iostream>

int main() {
  double result = 10.0 / 3.0;
  // int result = 10.0 / 3.0;
  // printf("%lf\n", result);
  std::cout << result << std::endl;
  return 0;
}
```
```bash
emcc -O1 -fwasm-exceptions a.cpp -o a.wasm
wasmedge --enable-exception-handling a.wasm
```

If we run the above command, then it will have the following error:

```
[2023-08-30 00:47:00.375] [error] execution failed: unreachable, Code: 0x89
[2023-08-30 00:47:00.376] [error]     In instruction: unreachable (0x00) , Bytecode offset: 0x0001760a
[2023-08-30 00:47:00.376] [error]     When executing function name: "_start"
```

However, if we replace `double result = 10.0 / 3.0;` to `int result = 10.0 / 3.0` **or** replace `std::cout` by `printf`, it may execute correctly. So it seems that the problem is because of the special exception handling from cout or division.

`wasm2wat --enable-exception a.wasm`

After further investigate the wat from the generated wasm file of  `double result = 10.0 / 3.0;` and `int result = 10.0 / 3.0`, the two files are slightly different. [Link](https://gist.github.com/harry900831/6a8f9fef60858372cab1c11308543404) is the diff of two files.

I couldn't find the cause for this situation. And I'm not sure whether it's because of the incorrect implementation for exception handling proposal.

harry900831:
@q82419 I've rebased on the latest commit, and the error on CI are all because of aot failed now.

hydai (changes requested):

test/spec/spectest.cpp
```diff
@@ -229,6 +230,7 @@ static const TestsuiteProposal TestsuiteProposals[] = {
     {"tail-call"sv, {Proposal::TailCall}},
     {"extended-const"sv, {Proposal::ExtendedConst}},
     {"threads"sv, {Proposal::Threads}},
+    {"exception-handling"sv, {Proposal::ExceptionHandling, Proposal::TailCall}},
```
q82419, line 233: @harry900831 
After the PR #2801 being merged, I'll rebase the branch.You can add the third param `WasmEdge::SpecTest::TestMode::Interpreter` here. Then the CI will pass.

q82419 (changes requested):

test/spec/spectest.cpp
harry900831: Done

harry900831:
@q82419 
It seems that there’s a lot of windows compile warning when it comes to the conversion from size_t to  uint32_t.
I’m wondering whether it’s fine to change the VPos, HPos, FPos, … from stackmgr.h to size_t

q82419:
> @q82419 It seems that there’s a lot of windows compile warning when it comes to the conversion from size_t to uint32_t. I’m wondering whether it’s fine to change the VPos, HPos, FPos, … from stackmgr.h to size_t

You can use `static_cast` the type into `uint32_t` instead on line 165.

q82419 (approved):

PR #2513, opened 2023-05-18
[WASI-NN] Refactor and reorganize codes

ibmibmibm:
* Separate different backand codes to different file
* Check all parameters before process


hydai (approved):
LGTM

Issue #2502, opened 2023-05-16
Standalone Repository for WasmEdge Rust Binding from Version 0.13.0 Onwards


hydai:
Hello, WasmEdge Community,

I would like to propose a new feature request regarding the organization of our repositories. This request involves the creation of a standalone repository specifically for the Rust binding, starting from the 0.13.0 release.

**Motivation**

The Rust binding has been an integral part of WasmEdge, primarily relying on the C API provided by WasmEdge. However, as the project evolves, it has become apparent that the Rust binding has a unique development roadmap. Hence, it is sensible to consider a structural change in how we organize these components to enable more focused development.

**Proposal**

I propose to split the Rust binding into a separate, standalone repository. This move would lead to several advantages:

1. **Independent Versioning**: This would allow for independent versioning of the Rust binding, which can evolve based on its own development and release cycles. This independence would ensure that enhancements, fixes, or changes in the Rust binding do not necessitate a new release of the WasmEdge main project.

2. **Focused Issue Tracking**: By separating the repositories, we can better track issues and improvements related to the Rust binding, making it easier for contributors to participate and for maintainers to manage the project.

3. **Clearer Roadmap**: A standalone repository would allow for the development of a distinct roadmap for the Rust binding, providing clarity for developers and users alike about the future direction of this component.

4. **Reduced Complexity**: It will simplify the WasmEdge repository by reducing the number of components it directly manages, resulting in a cleaner and more manageable codebase.

**Implementation Details**

The Rust binding will be split from the WasmEdge repository starting from version 0.13.0. The new repository will maintain the history of the Rust binding's development and will be managed by the same maintainers.

The Rust binding will continue to rely on the C API provided by WasmEdge, and efforts will be made to ensure compatibility and synchronization between the two projects.

**Conclusion**

This proposal aims to enhance the efficiency of our development processes and provide a clearer roadmap for both the WasmEdge main project and the Rust binding. We believe this change will benefit the WasmEdge community in the long run.

I kindly request feedback on this proposal, and I am looking forward to hearing your thoughts and suggestions.

Thank you for your attention and your continued contributions to the WasmEdge project.

/cc WasmEdge Rust binding maintainers @apepkuss @yanganto

apepkuss:
@hydai The issue has been fixed by https://github.com/WasmEdge/wasmedge-rust-sdk/pull/1 and https://github.com/WasmEdge/wasmedge-rust-sdk/pull/2

PR #2532, opened 2023-05-24
[Misc] Remove precompiled wasm assets in the source tree

### LINKED ISSUE #1869: Shipping binary precompiled WASM files is problematic for downstreams
## Motivation

I'm a Debian Developer and interested in packaging WasmEdge and including it in Debian. (Also [see Debian bug #1003128](https://bugs.debian.org/1003128)). I already have (mostly) working packages.

Debian has strict rules for what can be included in its "main" archive. Among those, is, to put it simply that that every source package must always include the source for every file it ships (as in, the preferred source of modification), and every binary needs to be built from its source, and using only software from Debian.

The WasmEdge tarball right now ships WASM binaries under `examples`, `tests` (and examples/tests of `bindings/rust`). While I can "just" repackage the tarball and remove these, I'd like for this to be the last resort, and to work with you in case we can find some other solution which lets Debian use the upstream tarball as-is.

The good news are that **none of the shipped WASM files seem to be necessary for a default release build** (`cmake -DCMAKE_BUILD_TYPE=Release`).

Finally, note that other distributions may have similar rules, and thus any upstream fix here may be beneficial more broadly than just Debian.

## Details

The shipped WASM files can be split in the following categories:

### Handcrafted WAT, shipped alongside WASM 

These are actually the majority. Debian and derivatives (incl. Ubuntu) have been shipping wabt for a few years now. I'm hoping that these can be dropped and regenerated by the build system when necessary (e.g. when building tests or examples). If not, there is a possibility that the Debian archive will accept those as binaries shipped alongside their sources.

- bindings/rust/wasmedge-sys/tests/data/fibonacci.wasm
- bindings/rust/wasmedge-sys/tests/data/import.wasm
- bindings/rust/wasmedge-sys/tests/data/test.wasm
- bindings/rust/wasmedge-sys/examples/data/funcs.wasm
- bindings/rust/wasmedge-sys/examples/data/fibonacci.wasm
- bindings/rust/wasmedge-sys/examples/data/module2.wasm
- bindings/rust/wasmedge-sys/examples/data/wasi_print_env.wasm
- bindings/rust/wasmedge-sys/examples/data/module1.wasm
- test/mixcall/mixcallTestData/module2.wasm
- test/mixcall/mixcallTestData/module1.wasm
- test/externref/externrefTestData/funcs.wasm
- test/externref/externrefTestData/stl.wasm
- test/api/apiTestData/import.wasm
- test/api/apiTestData/test.wasm
- examples/plugin/get-string/test.wasm
- examples/capi/host_functions/parse_json/parse-json.wasm
- examples/wasm/fibonacci.wasm
- examples/wasm/factorial.wasm
- examples/wasm/factorial2.wasm

### Rust sources shipped in tree 

- examples/capi/wasi-env/wasi_get_env.wasm

This can be rebuilt in Debian assuming the Rust dependencies are installed (`apt install rustc cargo lld-14 libstd-rust-dev-wasm32; cargo build --offline --release --target=wasm32-wasi`). 

Given the build system does not really build the corresponding C test program, or execute this example, removing this precompiled .wasm the tarball would be the safest option for Debian inclusion.

### C sources shipped in tree 

The two copies are byte-for-byte identical. The former includes a Makefile, which regenerates the .wasm assuming the build dependencies are installed (`apt install --no-install-recommends clang lld make`).

Same as above: given these examples are really there as references and nothing relies on them at build time, it seems possible to remove this precompiled .wasm from the tarball.

- examples/capi/mandelbrot-set-in-threads/mandelbrot.wasm
- bindings/rust/wasmedge-sdk/examples/data/mandelbrot.wasm

### Rust sources published online

The source for these is not shipped in the tarball, which is definitely a blocker for Debian. I had a look, and these look possible to rebuild them using the Rust toolchain outlined above. Assuming the sources were to be included, and given the possibily for reproducing the binaries, a (small) possibility exists that the Debian archive will accept those as binaries shipped alongside their sources.

- examples/wasm/add.wasm
- examples/wasm/hello.wasm

### C sources published online

The source for these is not shipped in the tarball, which is definitely a blocker for Debian. Even if it was, I took a look and it looks like it would be very complicated for this to be rebuilt in a clean Debian system (and using components only from Debian). I'd recommend removing these files from the tarball entirely, and perhaps leaving URLs or CMake files to fetch those from their respective repository for anyone installing from source.

- examples/js/qjs_tf.wasm
- examples/js/qjs.wasm

hydai:
Fix #1869

The Rust part will be removed after #2502. No need to care about them now.

## Handcrafted WAT, shipped alongside WASM

- [x] bindings/rust/wasmedge-sys/tests/data/fibonacci.wasm
- [x] bindings/rust/wasmedge-sys/tests/data/import.wasm
- [x] bindings/rust/wasmedge-sys/tests/data/test.wasm
- [x] bindings/rust/wasmedge-sys/examples/data/funcs.wasm
- [x] bindings/rust/wasmedge-sys/examples/data/fibonacci.wasm
- [x] bindings/rust/wasmedge-sys/examples/data/module2.wasm
- [x] bindings/rust/wasmedge-sys/examples/data/wasi_print_env.wasm
- [x] bindings/rust/wasmedge-sys/examples/data/module1.wasm
- [x] test/mixcall/mixcallTestData/module2.wasm
- [x] test/mixcall/mixcallTestData/module1.wasm
- [x] test/externref/externrefTestData/funcs.wasm
- [x] test/externref/externrefTestData/stl.wasm
- [x] test/api/apiTestData/import.wasm
- [x] test/api/apiTestData/test.wasm
- [x] examples/plugin/get-string/test.wasm
- [x] examples/capi/host_functions/parse_json/parse-json.wasm
- [x] examples/wasm/fibonacci.wasm
- [x] examples/wasm/factorial.wasm
- [x] examples/wasm/factorial2.wasm

## Rust sources shipped in tree
- [x] examples/capi/wasi-env/wasi_get_env.wasm

## C sources shipped in tree

- [x] examples/capi/mandelbrot-set-in-threads/mandelbrot.wasm
- [x] bindings/rust/wasmedge-sdk/examples/data/mandelbrot.wasm

## Rust sources published online

- [x] examples/wasm/add.wasm
- [x] examples/wasm/hello.wasm

## C sources published online

- [x] examples/js/qjs_tf.wasm
- [x] examples/js/qjs.wasm


hydai:
After discussing with @ibmibmibm, we will add a helper function in CMake which will try to compile the WAT files for the test suites via wat2wasm. If the wat2wasm is not found, we will try to fetch and build the wabt project.

hydai:
For the remaining test wasm/wat files, I think we can just hardcoding them in the test directly.

.github/workflows/reusable-build-on-macos.yml
```diff
@@ -29,7 +29,7 @@ jobs:
           fetch-depth: 0
       - name: Setup build environment
         run: |
-          brew install llvm ninja cmake
+          brew install llvm ninja boost cmake wabt
```
q82419, line 32: Why boost?

hydai: It's gone.

q82419 (approved):

PR #2560, opened 2023-06-04
[TEST] Enable test when boost not found.

MediosZ:
When boost is not installed, it will be built from the source. But setting `set(BUILD_TESTING OFF)` will disable all subsequent tests.  This can be observed in Fedora CI and WIndows CI.
So we need a better way to disable the boost tests.


juntao:

MediosZ:
Actually, I don't get why adding `set(BUILD_TESTING OFF)` there. Since the unit tests inside boost are not executed by default during `FetchContent_MakeAvailable`.
Maybe it would be better just to delete [that line](https://github.com/WasmEdge/WasmEdge/blob/e0e6dcc6847595ddf3cb152836c7b89578d1da1c/CMakeLists.txt#L56).

CMakeLists.txt
```diff
@@ -53,7 +53,7 @@ else()
   )
   set(BOOST_ENABLE_CMAKE ON)
   set(BOOST_RUNTIME_LINK static)
-  set(BUILD_TESTING OFF)
+  set(BUILD_TESTING ON)
```
hydai, line 56: How about disabling this here and enabling it again after the boost part?

hydai (changes requested):

hydai:
> Actually, I don't get why adding `set(BUILD_TESTING OFF)` there. Since the unit tests inside boost are not executed by default during `FetchContent_MakeAvailable`. Maybe it would be better just to delete [that line](https://github.com/WasmEdge/WasmEdge/blob/e0e6dcc6847595ddf3cb152836c7b89578d1da1c/CMakeLists.txt#L56).

Oh, so the Fetch and build from source will not run the test part inside boost. If so, then we can remove it.

hydai (approved):

PR #2589, opened 2023-06-14
[WASI-NN] Fix TfLite model data ownership

yanghaku:
In TfLite C API:
```c
// Returns a model from the provided buffer, or null on failure.
//
// NOTE: The caller retains ownership of the `model_data` and should ensure that
// the lifetime of the `model_data` must be at least as long as the lifetime
// of the `TfLiteModel`.
TFL_CAPI_EXPORT extern TfLiteModel* TfLiteModelCreate(const void* model_data,
                                                      size_t model_size);
```

The model data cannot be modified or freed during the inference, so the host must have ownership of the model data.

Now the WasmEdge cannot ensure the model data lifetime, for example in the wasm user code:
```rust
let data = std::fs::read("./test.tflite").unwrap();
// Load graph.
let graph = unsafe {
        wasi_nn::load(
            &[&data],
            wasi_nn::GRAPH_ENCODING_TENSORFLOWLITE,
            wasi_nn::EXECUTION_TARGET_CPU,
        )
        .unwrap()
    };
// Free the model data
drop(data);
// The following code may cause segment fault.
let context = unsafe { wasi_nn::init_execution_context(graph).unwrap() };
```

In this PR, simply copy data to the host functions to avoid the problem above.

References
* The discussion for this: https://github.com/WasmEdge/WasmEdge/pull/2360.
* TfLite model format with metadata: https://www.tensorflow.org/lite/models/convert/metadata#model_with_metadata_format


hydai (changes requested):
If users try to drop the model data before the inference, they should take care of it themselves. But I would like to know your feedback. Do you think we should take care of it on the plugin side?
Since there is no way to unload the model, it will make the memory usage grow, but it doesn't get any pros in such a situation.

yanghaku:
1. In my opinion, the [WASI-NN spec](https://github.com/WebAssembly/wasi-nn/) allows users free `graph-builder-array` after `load` graph. I think WASI-NN spec may add the graph unload function in the future.
2. I tested the `OpenVINO` and `Pytorch` (modify/free model data when inference), and they can run correctly.  The Implementation of the plugin OpenVINO also maintains the model data copy `OpenVINOWeightBlob`, `OpenVINONetwork`, and `OpenVINOExecNetwork`.
3. For security, We can't let security depend on the wasm user code. The user code especially malicious code, will cause segment fault or other potential errors.

Although it's not the best solution, I think we need this patch now.

plugins/wasi_nn/tfl.h
```diff
@@ -24,6 +24,7 @@ struct Graph {
       TfLiteModelDelete(TFLiteMod);
```
hydai, line 24: The `TfLiteModData` should be clean after the `TFLiteMod` is deleted.

hydai (changes requested):

hydai:
> Although it's not the best solution, I think we need this patch now.

Sounds great. Let's get this PR merged.

plugins/wasi_nn/tfl.h
yanghaku: In C++, it will run `~Graph()` first, then free the class member data in Graph, so the ```TfLiteModelDelete``` will run before the `TfLiteModData` cleaned.

hydai: Ah, yes. The destructor will do this for us. No need to call it on our side.

hydai (approved):

PR #2593, opened 2023-06-14
[Docs] Remove Unnecessary Docs and Reorganised Others

adithyaakrishna:
**Description:**

- This PR removes unnecessary/unwanted docs from the books site.
- Reorganised other docs


juntao:

hydai:
@alabulei1 PTAL

adithyaakrishna:
@hydai TIL - PTAL stands for Please take another look 🤯

alabulei1:
Hi @adithyaakrishna 

Thank you for this PR. 

For this PR, we don't have special docs for Swift, AssemblyScript, and Kotlin, which doesn't mean that WasmEdge can't run programs in these languages. 

Again, the first priority is to migrate the content from the old doc and delete adding notes for the old docs. While the format of the old docs was nice, our main objective is to ensure that the WasmEdge/docs are synchronized with the old book. This is an essential aspect that we must focus on.  I have a document to keep track of what is updated or missing. Let me know if you have any questions. https://docs.google.com/document/d/1wYOLvhuWt50M0sMPGN2TlA5JMICzv5BSczPmV4GsilQ/edit?usp=sharing

alabulei1:
@hydai  Are you ok with Moving `CODE_OF_CONDUCT.md`, `Ecosystem.md`, `GOVERNANCE.md`, and `OWNER.md`  from the docs directory to the project root?

hydai:
Hi @adithyaakrishna 
I would like to know the intention of moving these files from the docs folder to the root folder. 
Also, replacing `*` with `-` doesn't make sense to me. It should be the same in the markdown. Is there any rule forbidding `*`?

adithyaakrishna:
@alabulei1 Gotcha, could you pls lmk if I should revert any changes with respect to WasmEdge being able to  run programs in those languages?

@hydai I don't have any specific reason, if we decide to sunset `docs/` directory completely, then it would make much sense to move those files to the root directory. Also, regarding `-` and `*`, I thought to make them consistent across the whole project as we use `-` in some MD files, I updated them here as well 👀

alabulei1:
Since the linting and moving `CODE_OF_CONDUCT.md`, `Ecosystem.md`, `GOVERNANCE.md`, and `OWNER.md` from the docs directory to the project root is not in a hurry, let's hold this PR before the migration is done.  Plus, the migration way will affect what the old book looks like.

cc @adithyaakrishna @hydai

hydai:
1. IMO, I would like to keep these files inside the docs folder. Making it flatten into the root folder is unnecessary.
2. `*` and `-` are exactly the same. Humans should not modify it, or it should be on our checklist to make sure nobody will use another one. If we want to keep it in the same style, we should figure out how to use tools to fix it.

adithyaakrishna:
@alabulei1 Okay sure, that makes sense :)

@hydai Actually prettier formats it to use `-` instead of `*`

PR #2603, opened 2023-06-16
Add TCP_NODELAY socket option

Pryz:
While working on using some network libraries with Python, we've noticed that TCP_NODELAY is set by default on sockets since Python 3.6. This PR adds the IPPROTO_TCP option level as well as the TCP_NODELAY socket option.

I've added the TCP_NODELAY option in the `__wasi_sock_opt_so_t` enum since it is used in `sockSetOpt` and `sockGetOpt` but maybe these should be separated.

Also, note that I've used the value 1 for `__WASI_SOCK_OPT_LEVEL_IPPROTO_TCP` even though the level value on Linux side is 6. Should we match the values there?

Please let me know if there is interest in such a change and if there is any feedback about the implementation.


juntao:

thirdparty/wasi/api.hpp
```diff
@@ -1715,6 +1715,7 @@ static_assert(offsetof(__wasi_address_t, buf_len) == 4,
 
 enum __wasi_sock_opt_level_t : uint32_t {
   __WASI_SOCK_OPT_LEVEL_SOL_SOCKET = 0,
+  __WASI_SOCK_OPT_LEVEL_IPPROTO_TCP = 1,
```
hydai, line 1718: I think we can use `6` instead of `1`. Follow the https://github.com/leostratus/netinet/blob/master/in.h#L43

thirdparty/wasi/api.hpp
```diff
@@ -1751,6 +1752,7 @@ enum __wasi_sock_opt_so_t : uint32_t {
 
   __WASI_SOCK_OPT_SO_BINDTODEVICE = 14,
 
+  __WASI_SOCK_OPT_TCP_NODELAY = 15,
```
hydai, line 1755: The `__wasi_sock_opt_so_t` is for the socket options. Should we create a new struct `__wasi_sock_opt_tcp_t` for the TCP options?

hydai (changes requested):

thirdparty/wasi/api.hpp
Pryz: yes, pretty much what I was wondering. Was not sure WasmEdge was following the same number assignment. Will change to 6! thanks

thirdparty/wasi/api.hpp
Pryz: I guess so? Not sure how to adapt the set and get sock opt tho. Like this one for example https://github.com/WasmEdge/WasmEdge/blob/8c48ad0fb3feb1948cb0996f4afefae0957a459e/lib/host/wasi/inode-linux.cpp#L1214

hydai: The function you said is internal and can be modified without breaking the external ABI.

This external function is used as ABI and interacts with other languages, such as Rust.
https://github.com/WasmEdge/WasmEdge/blob/master/lib/host/wasi/wasifunc.cpp#L2093-L2096

hydai:
Some issues need to fix:
1. Please sign the DCO on your commits. Ref: https://www.secondstate.io/articles/dco/
2. Fedora/Windows/MacOS CI failed.

PR #2644, opened 2023-07-03
[Bug] - Fixed Broken Link

adithyaakrishna:
**Description:**

- Fixes Broken Links


juntao:
I would like to see contents of those pages moved to the new Docsauraus site and then removed here. Thanks

juntao:
We should not making any changes in this docbook site other than removing content and setting up redirect links.

adithyaakrishna:
@juntao Okay sure, that makes sense 👀

PR #2659, opened 2023-07-12
[Test] Migrating spec test from RapidJSON to SIMDJSON

### LINKED ISSUE #2598: Migrate from RapidJSON to SIMDJSON for the better performance in the spec test
## Motivation

[SIMDJSON](https://github.com/simdjson/simdjson) is a faster JSON library than RapidJSON. 

We can replace the RapidJSON to make the spec test faster.

## Details

Replace everywhere that contains RapidJSON with SIMDJSON.

vrnimje:
Resolves #2598


juntao:

test/spec/CMakeLists.txt
```diff
@@ -12,45 +12,15 @@ FetchContent_Declare(
...
-endif()
-
-add_library(RapidJSON INTERFACE)
-if(RAPIDJSON_INCLUDE_DIRS)
-  target_include_directories(RapidJSON INTERFACE ${RAPIDJSON_INCLUDE_DIRS})
-else()
-  target_include_directories(RapidJSON INTERFACE ${RapidJSON_SOURCE_DIR}/include)
-endif()
+message(STATUS "Downloading SIMDJSON")
+include(FetchContent)
```
hydai, line 16: Since some distro will provide simdjson package, we should keep the `find_package()` to get simdjson from the system.

test/spec/spectest.cpp
```diff
@@ -14,18 +14,10 @@
 ///
 //===----------------------------------------------------------------------===//
 
-#define RAPIDJSON_HAS_STDSTRING 1
-#if defined(__SSE4_2__)
-#define RAPIDJSON_SSE42 1
-#elif defined(__SSE2__)
-#define RAPIDJSON_SSE2 1
-#elif defined(__ARM_NEON__) || defined(__ARM_NEON) || defined(__ARM_NEON_FP)
-#define RAPIDJSON_NEON 1
```
hydai, line 23: Does simdjson provide such options?

test/spec/spectest.cpp
```diff
@@ -46,34 +36,37 @@ using namespace WasmEdge;
...
-  rapidjson::Value::ValueIterator ItMod = CmdArray.Begin();
-  for (rapidjson::Value::ValueIterator It = CmdArray.Begin();
-       It != CmdArray.End(); ++It) {
-    const auto CmdType = It->GetObject()["type"].Get<std::string>();
+                     simdjson::dom::array CmdArray) {
+  std::string_view OrgName;
+  uint64_t Mod;
+  for (simdjson::dom::object Obj : CmdArray) {
+    std::string_view CmdType = Obj["type"];
+    int flag = 0;
```
hydai, line 44: Please follow the coding style of the variable naming.
Also, the `flag` doesn't provide any information about what the flag is recorded. Please use a meaningful naming.

test/spec/spectest.cpp
```diff
@@ -102,23 +95,62 @@ SpecTest::CommandID resolveCommand(std::string_view Name) {
 
 // Helper function to parse parameters from json to vector of value.
 std::pair<std::vector<WasmEdge::ValVariant>, std::vector<WasmEdge::ValType>>
-parseValueList(const rapidjson::Value &Args) {
+parseValueList(simdjson::dom::array &Args) {
```
hydai, line 98: Why does the `const` get removed? Do you need to change the elements inside Args?

hydai (changes requested):

test/spec/spectest.cpp
vrnimje: Ohh, my bad. The const keyword was causing some issues with the Ondemad parser, so had to remove it.
But I have checked now, it works well with the DOM parser. Will add it back

test/spec/spectest.cpp
vrnimje: Simdjson actually automatically detects and sets these options while compiling, for the specific architecture, as per this [doc](https://github.com/simdjson/simdjson/blob/master/doc/implementation-selection.md)

test/spec/spectest.cpp
vrnimje: Ok, will do

test/spec/CMakeLists.txt
vrnimje: Ok, will do

vrnimje: I have made the changes in [f07568f](https://github.com/WasmEdge/WasmEdge/pull/2659/commits/f07568ff4216c571014bd3dae3e3d4f41982074a), and tested it on my system successfully.

test/spec/spectest.cpp
vrnimje: I have added the const arguments back here : [4020806](https://github.com/WasmEdge/WasmEdge/pull/2659/commits/4020806f88ab6fa990867cae568026c0782518df)

test/spec/spectest.cpp
vrnimje: I have changed the variable names [here](https://github.com/WasmEdge/WasmEdge/pull/2659/commits/4020806f88ab6fa990867cae568026c0782518df#:~:text=uint64_t%20LastModLine%3B,%3D%20false%3B), and changed `flag` to bool `Replaced`. 
Please let me know if the names are acceptable.

hydai:
Hi @vrnimje 
I found the SIMDJSON failed the building workflow on the Windows platform. Could you please check if this library can work on such a platform? We use `clang(clang-cl)` to build WasmEdge on Windows, so you don't need to worry about the MSVC-compatible issue.

vrnimje:
Hi @hydai 
I am aware of the issue and currently working on it. 
But I am not sure why the build is failing, SIMDJSON does support the clang-cl compiler under Visual Studio, as per the [doc](https://simdjson.github.io/simdjson/md_doc_basics.html#:~:text=Visual%20Studio%202017%20or%20better%20under%2064%2Dbit%20Windows.%20Users%20should%20target%20a%2064%2Dbit%20build%20(x64%20or%20ARM64)%20instead%20of%20a%2032%2Dbit%20build%20(x86).%20We%20support%20the%20LLVM%20clang%20compiler%20under%20Visual%20Studio%20(clangcl)%20as%20well%20as%20as%20the%20regular%20Visual%20Studio%20compiler.%20We%20also%20support%20MinGW%2064%2Dbit%20under%20Windows.)

test/spec/CMakeLists.txt
```diff
@@ -71,9 +80,8 @@ wasmedge_add_library(wasmedgeTestSpec
 )
 
 target_link_libraries(wasmedgeTestSpec
-  PRIVATE
-  RapidJSON
   PUBLIC
+  simdjson
```
hydai, line 84: Do you have any intention of making `simdjson` a PUBLIC scope?

test/spec/spectest.cpp
```diff
@@ -562,116 +571,118 @@ void SpecTest::run(std::string_view Proposal, std::string_view UnitName) {
...
     // Unknown command.
     EXPECT_TRUE(false);
   };
 
   // Get command list.
-  if (auto Commands = Doc.FindMember("commands"s);
-      Commands != Doc.MemberEnd()) {
-    rapidjson::Value CmdArray;
-    CmdArray.CopyFrom(Commands->value, Allocator);
+  simdjson::dom::array CmdArray = Doc["commands"];
```
hydai, line 679: What if the `commands` is not in `Doc`? We should have a check for this.

test/spec/CMakeLists.txt
vrnimje: No, apologies. Will use SIMDJSON under PRIVATE scope

test/spec/spectest.cpp
vrnimje: Ohh, didn't consider this. Will add a check for the same.

vrnimje: I have added the check for commands list with [e7162ac](https://github.com/WasmEdge/WasmEdge/pull/2659/commits/e7162acd4ff9812fa00a628c6d35802cb09d6ea6)

test/spec/CMakeLists.txt
vrnimje: I have added SIMDJSON under PRIVATE scope in [e7162ac](https://github.com/WasmEdge/WasmEdge/pull/2659/commits/e7162acd4ff9812fa00a628c6d35802cb09d6ea6)

test/spec/spectest.cpp
```diff
@@ -46,34 +36,37 @@ using namespace WasmEdge;
...
-      // Record last module in order.
-      ItMod = It;
+      // Record last module in order
+      if (Cmd["name"].get(OrgName)) {
+        OrgName = {};
+      }
+      LastModLine = Cmd["line"];
     } else if (CmdType == "register"sv) {
-      const auto NewNameStr = It->GetObject()["as"].Get<std::string>();
-      auto OrgName = ItMod->FindMember("name");
```
hydai, line 60: The original code retrieves the `OrgName`. However, your new code doesn't do the same thing.

test/spec/spectest.cpp
```diff
@@ -46,34 +36,37 @@ using namespace WasmEdge;
...
+  std::string_view OrgName;
+  uint64_t LastModLine;
+  for (simdjson::dom::object Cmd : CmdArray) {
+    std::string_view CmdType = Cmd["type"];
+    bool Replaced = false;
     if (CmdType == "module"sv) {
-      // Record last module in order.
-      ItMod = It;
+      // Record last module in order
+      if (Cmd["name"].get(OrgName)) {
```
hydai, line 47: Why do we need this?

test/spec/spectest.cpp
```diff
@@ -46,34 +36,37 @@ using namespace WasmEdge;
...
+        Alias.emplace(std::string(Value), std::string(NewNameStr));
+      } else if (!OrgName.empty()) {
+        // Register command does not record the original name. Get name from the
         // module.
-        Alias.emplace(OrgName->value.Get<std::string>(), NewNameStr);
+        Replaced = true;
+        Alias.emplace(std::string(OrgName), std::string(NewNameStr));
       }
-      if (OrgName != ItMod->MemberEnd()) {
+      if (!OrgName.empty() && !Replaced) {
```
hydai, line 63: The `Replaced` variable is introduced in this PR. What's the purpose of this?

test/spec/spectest.cpp
```diff
@@ -46,34 +36,37 @@ using namespace WasmEdge;
...
         // Module has origin name. Replace to aliased one.
-        OrgName->value.SetString(NewNameStr, Allocator);
+        Alias.emplace(std::string(OrgName), NewNameStr);
```
hydai, line 65: This behavior is not the same as the original one. The old one is to update the `OrgName`. However, your new code is replaced the `OrgName`, which is in the `Alias`.

test/spec/spectest.cpp
```diff
@@ -102,23 +95,62 @@ SpecTest::CommandID resolveCommand(std::string_view Name) {
...
-  Result.reserve(Args.Size());
-  ResultTypes.reserve(Args.Size());
-  for (const auto &Element : Args.GetArray()) {
-    const auto &Type = Element["type"].Get<std::string>();
-    const auto &ValueNode = Element["value"];
-    if (ValueNode.IsString()) {
-      const auto &Value = ValueNode.Get<std::string>();
+  Result.reserve(Args.size());
+  ResultTypes.reserve(Args.size());
+  for (simdjson::dom::object Element : Args) {
```
hydai, line 103: The `Element` is read-only, use `const simdjson::dom::object &Element` instead.

test/spec/spectest.cpp
```diff
@@ -128,85 +160,53 @@ parseValueList(const rapidjson::Value &Args) {
...
-      Result.emplace_back(Type, ValueNode.Get<std::string>());
-    } else if (ValueNode.IsArray()) {
-      const auto LaneType = Element["lane_type"].Get<std::string>();
-      std::string Value;
-      for (auto Iter = ValueNode.Begin(); Iter != ValueNode.End(); ++Iter) {
-        Value += Iter->Get<std::string>();
-        Value += ' ';
+  Result.reserve(Args.size());
+  simdjson::dom::array ValueNodeArray;
+  for (simdjson::dom::object Element : Args) {
```
hydai, line 196: Ditto.

test/spec/spectest.cpp
```diff
@@ -128,85 +160,53 @@ parseValueList(const rapidjson::Value &Args) {
...
+      std::string StrValue;
+      std::string_view LaneType = Element["lane_type"];
+      for (std::string_view X : ValueNodeArray) {
+        StrValue += std::string(X);
+        StrValue += ' ';
       }
-      Value.pop_back();
-      Result.emplace_back(Type + LaneType, std::move(Value));
+      StrValue.pop_back();
+      Result.emplace_back(std::string(Type) + std::string(LaneType), StrValue);
```
hydai, line 206: `std::move(StrValue)`.

test/spec/spectest.cpp
```diff
@@ -128,85 +160,53 @@ parseValueList(const rapidjson::Value &Args) {
...
     } else {
-      assumingUnreachable();
+      std::string_view Value = Element["value"];
```
hydai, line 208: The condition could be `array`, `string`, or `unexpected error`. This modification only maintains the first two conditions.

test/spec/spectest.cpp
```diff
@@ -465,55 +466,61 @@ bool SpecTest::stringContains(std::string_view Expected,
...
+  auto FName =
+      (TestsuiteRoot / Proposal / UnitName / (std::string(UnitName) + ".json"s))
+          .string();
+
+  simdjson::dom::parser parser;
+  simdjson::dom::element Doc;
+  try {
+    Doc = parser.load(FName);
+  } catch (simdjson::error_code &err) {
+    ;
```
hydai, line 478: Should it fail? If not, we should print some error messages.

test/spec/spectest.cpp
```diff
@@ -465,55 +466,61 @@ bool SpecTest::stringContains(std::string_view Expected,
...
 
-  auto Invoke = [&](const rapidjson::Value &Action,
-                    const rapidjson::Value &Expected, uint64_t LineNumber) {
+  auto Invoke = [&](const simdjson::dom::object &Action,
+                    const simdjson::dom::array &Expected, uint64_t LineNumber) {
     const auto ModName = GetModuleName(Action);
-    const auto Field = Action["field"s].Get<std::string>();
-    const auto Params = parseValueList(Action["args"s]);
+    const std::string_view Field = Action["field"];
+    simdjson::dom::array Args = Action["args"];
```
hydai, line 501: Do we need this temp variable? Is it possible to use `parseValueList(Action["args"])`.

test/spec/spectest.cpp
```diff
@@ -528,13 +535,15 @@ void SpecTest::run(std::string_view Proposal, std::string_view UnitName) {
...
   };
-  auto TrapInvoke = [&](const rapidjson::Value &Action, const std::string &Text,
-                        uint64_t LineNumber) {
+  auto TrapInvoke = [&](const simdjson::dom::object &Action,
+                        const std::string &Text, uint64_t LineNumber) {
     const auto ModName = GetModuleName(Action);
-    const auto Field = Action["field"s].Get<std::string>();
-    const auto Params = parseValueList(Action["args"s]);
+    const std::string_view Field = Action["field"];
+    simdjson::dom::array Args = Action["args"];
```
hydai, line 542: Ditto.

test/spec/spectest.cpp
```diff
@@ -562,117 +571,121 @@ void SpecTest::run(std::string_view Proposal, std::string_view UnitName) {
...
-          // Module has name. Register module with module name.
-          LastModName = Name->value.Get<std::string>();
-        } else {
-          // Instantiate the anonymous module.
-          LastModName.clear();
-        }
-        if (onModule(LastModName, FileName)) {
-          EXPECT_TRUE(true);
+  auto RunCommand = [&](const simdjson::dom::object &Cmd) {
+    std::string_view TypeField = Cmd["type"];
```
hydai, line 575: In the old code, it will check `Type != Cmd.MemberEnd()`. However, the new one doesn't.

hydai (changes requested):

test/spec/spectest.cpp
```diff
@@ -46,34 +36,37 @@ using namespace WasmEdge;
...
-                     rapidjson::Value &CmdArray,
-                     rapidjson::Document::AllocatorType &Allocator) {
-  rapidjson::Value::ValueIterator ItMod = CmdArray.Begin();
-  for (rapidjson::Value::ValueIterator It = CmdArray.Begin();
-       It != CmdArray.End(); ++It) {
-    const auto CmdType = It->GetObject()["type"].Get<std::string>();
+                     simdjson::dom::array CmdArray) {
+  std::string_view OrgName;
+  uint64_t LastModLine;
+  for (simdjson::dom::object Cmd : CmdArray) {
```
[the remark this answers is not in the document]
vrnimje: Ok, will make this change

test/spec/spectest.cpp
vrnimje: This is part of a major logic change, caused due to the fact that SIMDJSON doesn't allow us to modify or add fields to JSON objects. I had to retrieve the `OrgName` here, instead of the block mentioned above, as I faced some errors using SIMDJSON object variable to store ItMod.

test/spec/spectest.cpp
vrnimje: If it's not already added to the `Alias` [here](https://github.com/vrnimje/WasmEdge_fork/blob/172c416c068e120b02b3ed15390070a920337277/test/spec/spectest.cpp#L57-L62), then Replaced is false, so it is added here.

test/spec/spectest.cpp
vrnimje: As I mentioned above, SIMDJSON doesn't allow us to modify JSON objects. So, I used the Alias to store the `OrgName`s and their corresponding `NewNameStr`s, which is later used [here](https://github.com/vrnimje/WasmEdge_fork/blob/172c416c068e120b02b3ed15390070a920337277/test/spec/spectest.cpp#L577-L603). I just used the Alias, which was already used to store aliases, to also store the names that needed to be modified or added.

test/spec/spectest.cpp
vrnimje: Ok, will make this change as well.

test/spec/spectest.cpp
vrnimje: Ohh, missed it. Will add another condition for the same.

test/spec/spectest.cpp
vrnimje: It shouldn't fail. I was using this for testing purposes. Will remove this block.

test/spec/spectest.cpp
vrnimje: The problem is by default, when indexing the parsed objects, SIMDJSON returns `simdjson::result<simdjson::dom::value>`, which doesn't convert to `simdjson::dom::array` implicitly
 Hence, I had to use this temp variable, to type cast it explicitly.

test/spec/spectest.cpp
vrnimje: Same as above.

test/spec/spectest.cpp
vrnimje: Will add a check for the same, if whether or not the field "type" exists or not.

test/spec/spectest.cpp
vrnimje: Explained below.

test/spec/spectest.cpp
```diff
@@ -46,34 +36,37 @@ using namespace WasmEdge;
 
 // Preprocessing for set up aliasing.
 void resolveRegister(std::map<std::string, std::string> &Alias,
-                     rapidjson::Value &CmdArray,
-                     rapidjson::Document::AllocatorType &Allocator) {
-  rapidjson::Value::ValueIterator ItMod = CmdArray.Begin();
-  for (rapidjson::Value::ValueIterator It = CmdArray.Begin();
-       It != CmdArray.End(); ++It) {
-    const auto CmdType = It->GetObject()["type"].Get<std::string>();
+                     simdjson::dom::array CmdArray) {
```
hydai, line 39: Why do we need to remove the reference of `CmdArray`?

test/spec/spectest.cpp
```diff
@@ -612,44 +628,45 @@ void SpecTest::run(std::string_view Proposal, std::string_view UnitName) {
...
       case CommandID::AssertUninstantiable: {
-        const auto Filename = (TestsuiteRoot / Proposal / UnitName /
-                               Cmd["filename"s].Get<std::string>())
-                                  .u8string();
-        const auto &Text = Cmd["text"s].Get<std::string>();
-        TrapInstantiate(Filename, Text);
+        std::string_view Name = Cmd["filename"];
+        const auto Filename =
+            (TestsuiteRoot / Proposal / UnitName / Name).u8string();
+        Name = Cmd["text"];
```
hydai, line 668: Don't use the `Name` for different purposes.

test/spec/spectest.cpp
```diff
@@ -612,44 +628,45 @@ void SpecTest::run(std::string_view Proposal, std::string_view UnitName) {
...
       case CommandID::AssertInvalid: {
-        const auto Filename = (TestsuiteRoot / Proposal / UnitName /
-                               Cmd["filename"s].Get<std::string>())
-                                  .u8string();
-        const auto &Text = Cmd["text"s].Get<std::string>();
-        TrapValidate(Filename, Text);
+        std::string_view Name = Cmd["filename"];
+        const auto Filename =
+            (TestsuiteRoot / Proposal / UnitName / Name).u8string();
+        Name = Cmd["text"];
```
hydai, line 659: Reuse variables with different purposes.

test/spec/spectest.cpp
```diff
@@ -612,44 +628,45 @@ void SpecTest::run(std::string_view Proposal, std::string_view UnitName) {
...
       }
       case CommandID::AssertExhaustion: {
         // TODO: Add stack overflow mechanism.
         return;
       }
       case CommandID::AssertMalformed: {
-        const auto &ModType = Cmd["module_type"s].Get<std::string>();
-        if (ModType != "binary") {
+        std::string_view ModType = Cmd["module_type"];
+        std::string Binary = "binary";
```
hydai, line 643: Redundant variables. Use string literals instead.

test/spec/spectest.cpp
```diff
@@ -612,44 +628,45 @@ void SpecTest::run(std::string_view Proposal, std::string_view UnitName) {
...
         }
-        const auto Filename = (TestsuiteRoot / Proposal / UnitName /
-                               Cmd["filename"s].Get<std::string>())
-                                  .u8string();
-        const auto &Text = Cmd["text"s].Get<std::string>();
-        TrapLoad(Filename, Text);
+        std::string_view Name = Cmd["filename"];
+        const auto Filename =
+            (TestsuiteRoot / Proposal / UnitName / Name).u8string();
+        Name = Cmd["text"];
```
hydai, line 651: Reuse variables with different purposes.

test/spec/spectest.cpp
```diff
@@ -612,44 +628,45 @@ void SpecTest::run(std::string_view Proposal, std::string_view UnitName) {
         return;
       }
       case CommandID::AssertTrap: {
-        const auto &Action = Cmd["action"s];
-        const auto &Text = Cmd["text"s].Get<std::string>();
-        const uint64_t LineNumber = Cmd["line"].Get<uint64_t>();
-        TrapInvoke(Action, Text, LineNumber);
+        const simdjson::dom::object Action = Cmd["action"].get_object();
```
hydai, line 631: Use reference instead of copying a new data structure.

test/spec/spectest.cpp
```diff
@@ -597,10 +613,10 @@ void SpecTest::run(std::string_view Proposal, std::string_view UnitName) {
         return;
       }
       case CommandID::AssertReturn: {
-        const auto &Action = Cmd["action"s];
-        const auto &Expected = Cmd["expected"s];
-        const auto ActType = Action["type"].Get<std::string>();
-        const uint64_t LineNumber = Cmd["line"].Get<uint64_t>();
+        const uint64_t LineNumber = Cmd["line"];
+        const simdjson::dom::object Action = Cmd["action"].get_object();
```
hydai, line 617: Use reference instead of copying a new data structure.

test/spec/spectest.cpp
```diff
@@ -597,10 +613,10 @@ void SpecTest::run(std::string_view Proposal, std::string_view UnitName) {
...
+        const simdjson::dom::array Expected = Cmd["expected"];
```
hydai, line 618: Ditto.

test/spec/spectest.cpp
```diff
@@ -586,9 +602,9 @@ void SpecTest::run(std::string_view Proposal, std::string_view UnitName) {
         return;
       }
       case CommandID::Action: {
-        const auto &Action = Cmd["action"s];
-        const auto &Expected = Cmd["expected"s];
-        const uint64_t LineNumber = Cmd["line"].Get<uint64_t>();
+        const simdjson::dom::object Action = Cmd["action"].get_object();
```
hydai, line 605: Use reference instead of copying a new data structure.

test/spec/spectest.cpp
```diff
@@ -586,9 +602,9 @@ void SpecTest::run(std::string_view Proposal, std::string_view UnitName) {
...
+        const simdjson::dom::array Expected = Cmd["expected"];
```
hydai, line 606: Ditto.

test/spec/spectest.cpp
```diff
@@ -562,18 +569,27 @@ void SpecTest::run(std::string_view Proposal, std::string_view UnitName) {
...
+      switch (resolveCommand(TypeField)) {
       case SpecTest::CommandID::Module: {
-        const auto FileName = (TestsuiteRoot / Proposal / UnitName /
-                               Cmd["filename"].Get<std::string>())
-                                  .u8string();
-        const uint64_t LineNumber = Cmd["line"].Get<uint64_t>();
-        if (const auto Name = Cmd.FindMember("name"); Name != Cmd.MemberEnd()) {
+        std::string_view Name = Cmd["filename"];
+        const auto FileName =
+            (TestsuiteRoot / Proposal / UnitName / Name).u8string();
```
hydai, line 580: Why not use `Cmd["filename"]` instead?

test/spec/spectest.cpp
```diff
@@ -562,18 +569,27 @@ void SpecTest::run(std::string_view Proposal, std::string_view UnitName) {
...
+        const uint64_t LineNumber = Cmd["line"];
+        std::string LineStr = std::to_string(LineNumber);
+        std::string_view TempName;
+        if (!Cmd["name"].get(TempName)) {
           // Module has name. Register module with module name.
-          LastModName = Name->value.Get<std::string>();
+          if (auto It = Alias.find(std::string(TempName)); It != Alias.end()) {
```
hydai, line 586: Why do we have the logic of `Alias` here? It seems like the original codes don't have this logic. Is the original code wrong?

test/spec/spectest.cpp
hydai: Does SIMDJSON provide `get_array` or something? This sounds so weird.

test/spec/spectest.cpp
```diff
@@ -465,55 +469,56 @@ bool SpecTest::stringContains(std::string_view Expected,
...
-                       (std::string(UnitName) + ".json"s));
-  rapidjson::IStreamWrapper JSONISWrapper(JSONIS);
-  rapidjson::Document Doc;
-  auto &Allocator = Doc.GetAllocator();
-  Doc.ParseStream(JSONISWrapper);
+  auto FName =
+      (TestsuiteRoot / Proposal / UnitName / (std::string(UnitName) + ".json"s))
+          .string();
+
+  simdjson::dom::parser parser;
```
hydai, line 476: Wrong naming style.

test/spec/spectest.cpp
```diff
@@ -128,83 +160,54 @@ parseValueList(const rapidjson::Value &Args) {
...
-      std::string Value;
-      for (auto Iter = ValueNode.Begin(); Iter != ValueNode.End(); ++Iter) {
-        Value += Iter->Get<std::string>();
-        Value += ' ';
+  Result.reserve(Args.size());
+  simdjson::dom::array ValueNodeArray;
+  std::string_view Value;
+  for (const simdjson::dom::object &Element : Args) {
+    std::string_view Type = Element["type"];
+    if (!Element["value"].get(ValueNodeArray)) {
```
hydai, line 199: Does SIMDJSON doesn't provide any function like `is_array`? Why do we always have to create a temp variable for testing the result?

hydai (changes requested):

test/spec/spectest.cpp
vrnimje: Didn't mean to remove it, will add it back

test/spec/spectest.cpp
vrnimje: It does, but then it's still gives `simdjson::result<simdjson::dom::array>`, which still needs to be converted

test/spec/spectest.cpp
vrnimje: Then it returns `simdjson::result<std::string_view>`, which presents the same problem I mentioned [above](https://github.com/WasmEdge/WasmEdge/pull/2659/files#r1270234563)

test/spec/spectest.cpp
vrnimje: As I mentioned above, SIMDJSON doesn't allow us to modify JSON objects. In the previous code, the module names were added or modified in the JSON objects itself. Since I couldn't do it here, I used `Alias` to replace the Module names.

test/spec/spectest.cpp
vrnimje: Will look into whether or not is there a better implementation for the same.

vrnimje: I am now using the `type()` function in the latest commit

test/spec/spectest.cpp
vrnimje: Hey @hydai , I tried using reference here, but unfortunately it was causing SegFault in the spectests

test/spec/spectest.cpp
vrnimje: Same as above

test/spec/spectest.cpp
vrnimje: Same as above

test/spec/spectest.cpp
vrnimje: Same as above

test/spec/spectest.cpp
vrnimje: I have added the references in https://github.com/WasmEdge/WasmEdge/pull/2659/commits/4e71119fc68455dceb5b27fc933a93c451f63ecc, it was failing due to `get_object()` function calls.

test/spec/spectest.cpp
vrnimje: I have added it back in https://github.com/WasmEdge/WasmEdge/pull/2659/commits/b903dff5d269911bd3a841257eba8f95fe440630

test/spec/spectest.cpp
vrnimje: Added variable `Text` in https://github.com/WasmEdge/WasmEdge/pull/2659/commits/b903dff5d269911bd3a841257eba8f95fe440630

test/spec/spectest.cpp
vrnimje: Added variable `Text` in https://github.com/WasmEdge/WasmEdge/pull/2659/commits/b903dff5d269911bd3a841257eba8f95fe440630

test/spec/spectest.cpp
vrnimje: Using string variables in https://github.com/WasmEdge/WasmEdge/pull/2659/commits/b903dff5d269911bd3a841257eba8f95fe440630

test/spec/spectest.cpp
```diff
@@ -188,25 +192,26 @@ parseValueList(const rapidjson::Value &Args) {
...
+      std::string StrValue;
+      std::string_view LaneType = Element["lane_type"];
+      for (std::string_view X : ValueNodeArray) {
+        StrValue += std::string(X);
+        StrValue += ' ';
       }
-      Value.pop_back();
-      Result.emplace_back(Type + LaneType, std::move(Value));
-    } else {
-      assumingUnreachable();
```
hydai, line 209: You should keep `assumingUnreachable()`.

hydai (changes requested):

test/spec/spectest.cpp
```diff
@@ -46,34 +36,37 @@ using namespace WasmEdge;
...
         // Module has origin name. Replace to aliased one.
-        OrgName->value.SetString(NewNameStr, Allocator);
+        Alias.emplace(std::string(OrgName), NewNameStr);
       } else {
         // Module has no origin name. Add the aliased one.
-        rapidjson::Value Text;
-        Text.SetString(NewNameStr, Allocator);
-        ItMod->AddMember("name", Text, Allocator);
+        std::string X = std::to_string(LastModLine);
+        Alias.emplace(X, NewNameStr);
```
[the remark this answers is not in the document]
vrnimje: Yes, will make that change.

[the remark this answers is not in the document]
vrnimje: Replaced it in [2def666](https://github.com/WasmEdge/WasmEdge/pull/2659/commits/2def666c877c49e25752cbebb63f04a8ad8b45c8)

test/spec/spectest.cpp
vrnimje: Added missing `assumingReachable()` in [2def666](https://github.com/WasmEdge/WasmEdge/pull/2659/commits/2def666c877c49e25752cbebb63f04a8ad8b45c8)

hydai (approved):
LGTM. @q82419 PTAL.

q82419 (approved):

PR #2662, opened 2023-07-16
[WASI] Fix blocking when multiple requests have the same fds.

MediosZ:
Fix this CI error. https://github.com/WasmEdge/wasmedge_hyper_demo/actions/runs/5511875753/jobs/10047996074

This PR remove the unused file descriptor from FdDatas when the socket is closed.

PR #2675, opened 2023-07-21
[WASI] Remove unused fds before closing them.

MediosZ:
Fix PR #2662 so that the unused fds are closed successfully.

### DIFF
--- a/lib/host/wasi/wasifunc.cpp
+++ b/lib/host/wasi/wasifunc.cpp
@@ -560,8 +560,6 @@ Expect<uint32_t> WasiFdAllocate::body(const Runtime::CallingFrame &, int32_t Fd,
 Expect<uint32_t> WasiFdClose::body(const Runtime::CallingFrame &, int32_t Fd) {
   const __wasi_fd_t WasiFd = Fd;
 
-  Env.close(WasiFd);
-
   if (auto Res = Env.fdClose(WasiFd); unlikely(!Res)) {
     return Res.error();
   }
--- a/lib/host/wasi/wasifunc.cpp
+++ b/lib/host/wasi/wasifunc.cpp
@@ -560,11 +560,11 @@ Expect<uint32_t> WasiFdAllocate::body(const Runtime::CallingFrame &, int32_t Fd,
 Expect<uint32_t> WasiFdClose::body(const Runtime::CallingFrame &, int32_t Fd) {
   const __wasi_fd_t WasiFd = Fd;
 
+  Env.close(WasiFd);
+
   if (auto Res = Env.fdClose(WasiFd); unlikely(!Res)) {
     return Res.error();
   }
-
-  Env.close(WasiFd);
   return __WASI_ERRNO_SUCCESS;
 }
 
--- a/include/host/wasi/environ.h
+++ b/include/host/wasi/environ.h
@@ -255,6 +255,7 @@ public:
     if (auto It = FdMap.find(Fd); It == FdMap.end()) {
       return WasiUnexpect(__WASI_ERRNO_BADF);
     } else {
+      close(It->second);
       FdMap.erase(It);
       return {};
     }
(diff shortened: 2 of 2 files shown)


lib/host/wasi/wasifunc.cpp
```diff
@@ -560,11 +560,11 @@ Expect<uint32_t> WasiFdAllocate::body(const Runtime::CallingFrame &, int32_t Fd,
 Expect<uint32_t> WasiFdClose::body(const Runtime::CallingFrame &, int32_t Fd) {
   const __wasi_fd_t WasiFd = Fd;
 
+  Env.close(WasiFd);
```
hydai, line 563: Please add comments to describe why we have to `close` before `fdClose`.

hydai (changes requested):

MediosZ:
@hydai  
I agree that the `close` should be moved after `fdClose`. The problem is that after `fdClose`, the Node is removed from the FdMap, so that `close` cannot get the Fd to close.
Maybe we should add a function to get the Fd of the node, and then we will have something like this:
```c++
Expect<uint32_t> WasiFdClose::body(const Runtime::CallingFrame &, int32_t Fd) {
  const __wasi_fd_t WasiFd = Fd;

  // get the fd to be closed.
  __wasi_fd_t NodeFd;
  if (auto Res = Env.getNodeFd(WasiFd); unlikely(!Res)) {
    return Res.error();
  } else {
    NodeFd = *Res;
  }

  if (auto Res = Env.fdClose(WasiFd); unlikely(!Res)) {
    return Res.error();
  }
  // close the fd
  Env.close(NodeFd);

  return __WASI_ERRNO_SUCCESS;
}
``` 

I can confirm that this works, but I don't know if it's a good idea to expose the fd of the Node publicly.

hydai:
@MediosZ 
How about moving the `Env.close` inside the `Env.fdClose` function if `Env.close` will always execute after `Env.fdClose`?

MediosZ:
Agreed.
Seems that a separate function is unnecessary.

hydai:
Hi @MediosZ 
Could you confirm again if the latest modification is still working in the examples?

MediosZ:
@hydai I confirm the modifcation works.

MediosZ:
This is what I get after running the `wasmedge_hyper_demo`
```
GET as byte stream: http://eu.httpbin.org/get?msg=Hello
Response: 200 OK
Headers: {
    "date": "Mon, 24 Jul 2023 02:05:51 GMT",
    "content-type": "application/json",
    "content-length": "237",
    "connection": "keep-alive",
    "server": "gunicorn/19.9.0",
    "access-control-allow-origin": "*",
    "access-control-allow-credentials": "true",
}

b"{\n  \"args\": {\n    \"msg\": \"Hello\"\n  }, \n  \"headers\": {\n    \"Host\": \"eu.httpbin.org\", \n    \"X-Amzn-Trace-Id\": \"Root=1-64bddc7f-19e3a62e0007c9342b8a2512\"\n  }, \n  \"origin\": \"121.249.15.204\", \n  \"url\": \"http://eu.httpbin.org/get?msg=Hello\"\n}\n"

GET and get result as string: http://eu.httpbin.org/get?msg=WasmEdge
{
  "args": {
    "msg": "WasmEdge"
  }, 
  "headers": {
    "Host": "eu.httpbin.org", 
    "X-Amzn-Trace-Id": "Root=1-64bddc80-51f9518828c785f15fef0085"
  }, 
  "origin": "121.249.15.204", 
  "url": "http://eu.httpbin.org/get?msg=WasmEdge"
}


POST and get result as string: http://eu.httpbin.org/post
with a POST body: hello wasmedge
{
  "args": {}, 
  "data": "hello wasmedge", 
  "files": {}, 
  "form": {}, 
  "headers": {
    "Content-Length": "14", 
    "Host": "eu.httpbin.org", 
    "X-Amzn-Trace-Id": "Root=1-64bddc81-6a44ff397d5fe7cf2a00fdc9"
  }, 
  "json": null, 
  "origin": "121.249.15.204", 
  "url": "http://eu.httpbin.org/post"
}

```

hydai (approved):

PR #2715, opened 2023-08-03
[Do not merge][Preview] This is a preview version of WasmEdge coroutine.

hydai:
The API may change if there is any design decision update

### DIFF
--- a/include/api/wasmedge/wasmedge.h
+++ b/include/api/wasmedge/wasmedge.h
@@ -141,6 +141,9 @@ typedef struct WasmEdge_CallingFrameContext WasmEdge_CallingFrameContext;
 /// Opaque struct of WasmEdge asynchronous result.
 typedef struct WasmEdge_Async WasmEdge_Async;
 
+/// Opaque struct of WasmEdge coroutine result.
+typedef struct WasmEdge_Coroutine WasmEdge_Coroutine;
+
 /// Opaque struct of WasmEdge VM.
 typedef struct WasmEdge_VMContext WasmEdge_VMContext;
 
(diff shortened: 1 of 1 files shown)


juntao:

include/api/wasmedge/wasmedge.h
```diff
@@ -2636,6 +2658,13 @@ WASMEDGE_CAPI_EXPORT extern WasmEdge_MemoryInstanceContext *
 WasmEdge_CallingFrameGetMemoryInstance(const WasmEdge_CallingFrameContext *Cxt,
                                        const uint32_t Idx);
 
+/// Suspend the execution from the current calling frame.
+/// This is only available when using WasmEdge_Coroutine.
+///
+/// \param Cxt the WasmEdge_CallingFrameContext.
+WASMEDGE_CAPI_EXPORT extern void
+WasmEdge_CallingFrameCoroutineSuspend(const WasmEdge_CallingFrameContext *Cxt);
```
rockwotj, line 2666: How is this expected to be used? Is another thread expected to call this?

Ideally for our use case we would suspend execution periodically using gas metering

hydai: Hi @rockwotj 
This feature is for those who want to suspend the execution from the host functions. We have some use cases that will spend much time waiting for IO. They would like to suspend the execution until the IO is ready and get it back.

rockwotj: I see. We do this today using `std::longjmp` to jump out from the host function then back in on a separate stack.

An important use case for our thread-per-core architecture is that one task doesn't hog the CPU for too long. Otherwise that means other requests that thread is responsible for cannot be responded, increasing tail latencies in these situations. See https://github.com/scylladb/seastar/blob/master/doc/tutorial.md#breaking-up-long-running-computations

Anyways, it seems this mechanism and set of APIs won't solve this use case and instead we need some callback function to be periodically invoked from the runtime while executing.

include/api/wasmedge/wasmedge.h
```diff
@@ -2701,6 +2730,71 @@ WASMEDGE_CAPI_EXPORT void WasmEdge_AsyncDelete(WasmEdge_Async *Cxt);
 
 // <<<<<<<< WasmEdge Async functions <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
 
+// >>>>>>>> WasmEdge Coroutine functions >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
+
+/// Start to execute a WasmEdge_Coroutine execution.
+///
+/// \param Cxt the WasmEdge_Coroutine.
+WASMEDGE_CAPI_EXPORT void WasmEdge_CoroutineSpawn(WasmEdge_Coroutine *Cxt);
```
rockwotj, line 2738: Are you going to be creating a separate stack here (or is this a stackful or stackless coroutine)? If so it's useful to add the amount of memory used here.

hydai: I am going to have a stackful coroutine. Would you like any other information (memory usage sounds useful)?

include/api/wasmedge/wasmedge.h
```diff
@@ -2636,6 +2658,13 @@ WASMEDGE_CAPI_EXPORT extern WasmEdge_MemoryInstanceContext *
 WasmEdge_CallingFrameGetMemoryInstance(const WasmEdge_CallingFrameContext *Cxt,
                                        const uint32_t Idx);
 
+/// Suspend the execution from the current calling frame.
+/// This is only available when using WasmEdge_Coroutine.
+///
+/// \param Cxt the WasmEdge_CallingFrameContext.
+WASMEDGE_CAPI_EXPORT extern void
```
L-jasmine, line 2665: Perhaps developers working on the host section need to know which module's host function triggered the suspend. Therefore, I hope that it can return a pointer to the module instance to which the triggering host function belongs.

hydai:
Closing, as this will no longer be a milestone on our roadmap at the moment, and it requires further discussions regarding the details before we proceed with implementing this feature.

PR #2736, opened 2023-08-11
[AOT] Fix fallback impl of `compileVectorExtAddPairwise`.

am009:
LLVM vector shift operator (e.g., [ashr](https://llvm.org/docs/LangRef.html#ashr-instruction)) requires two operands of the same type. Use a `splat` instruction to convert the shift count to vector type.

The tests that cover this is passed because of the alternate implementation using SSE on x64_64. These implementations come before the fallback impl but guarded by feature checking (check for SSE on x64_64, NEON on aarch64). The problem is confirmed by removing other implementations (two `ifdef` parts in the same function) and rerun the tests.
image

After applying this fix, these tests are passed (still keep other impl removed). (Under Linux in docker image wasmedge/wasmedge:latest)

### DIFF
--- a/lib/aot/compiler.cpp
+++ b/lib/aot/compiler.cpp
@@ -4376,8 +4376,10 @@ private:
           // If the XOP, SSSE3, or SSE2 is not supported on the x86_64 platform
           // or the NEON is not supported on the aarch64 platform,
           // then fallback to this.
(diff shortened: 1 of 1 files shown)


juntao:

hydai (approved):
Nice catch. LGTM.

Issue #2802, opened 2023-09-12
llama.cpp fails to build with g++ 13


paravoid:
Building WasmEdge 0.13.4 on Debian unstable with gcc/g++ 13, results into this build failure:

<details>
  <summary>Build log</summary>
```
cd <<BUILDDIR>>/wasmedge/obj-x86_64-linux-gnu/lib/validator && /usr/lib/ccache/c++ -DFMT_SHARED -DSPDLOG_COMPILED_LIB -DSPDLOG_FMT_EXTERNAL -DSPDLOG_SHARED_LIB -I<<BUILDDIR>>/wasmedge/obj-x86_64-linux-gnu/include -I<<BUILDDIR>>/wasmedge/include -g -O2 -ffile-prefix-map=<<BUILDDIR>>/wasmedge=. -fstack-protector-strong -fstack-clash-protection -Wformat -Werror=format-security -fcf-protection -Wdate-time -D_FORTIFY_SOURCE=2 -std=c++17 -fPIC -fvisibility=hidden -fvisibility-inlines-hidden -Wall -Wextra -Werror -Wno-error=pedantic -Wno-psabi -MD -MT lib/validator/CMakeFiles/wasmedgeValidator.dir/validator.cpp.o -MF CMakeFiles/wasmedgeValidator.dir/validator.cpp.o.d -o CMakeFiles/wasmedgeValidator.dir/validator.cpp.o -c <<BUILDDIR>>/wasmedge/lib/validator/validator.cpp
In file included from /usr/include/c++/13/string:51,
                 from <<BUILDDIR>>/wasmedge/thirdparty/ggml/llama-util.h:15,
                 from <<BUILDDIR>>/wasmedge/thirdparty/ggml/llama.cpp:9:
In static member function ‘static void std::__copy_move<false, false, std::random_access_iterator_tag>::__assign_one(_Tp*, _Up*) [with _Tp = const llama_grammar_element*; _Up = const llama_grammar_element* const]’,
    inlined from ‘static _Up* std::__copy_move<_IsMove, true, std::random_access_iterator_tag>::__copy_m(_Tp*, _Tp*, _Up*) [with _Tp = const llama_grammar_element* const; _Up = const llama_grammar_element*; bool _IsMove = false]’ at /usr/include/c++/13/bits/stl_algobase.h:440:20,
[...]

PR #2805, opened 2023-09-13
[WASI-NN] Update ggml

dm4:
- Update ggml library (#2802)
- Use gguf model for tests
- Enable `LLAMA_BLAS` by default


hydai (changes requested):

LICENSE.spdx
```diff
@@ -39,7 +39,7 @@ PackageFileName: ./thirdparty/ggml
 PackageHomePage: https://github.com/ggerganov/llama.cpp
 PackageOriginator: Georgi Gerganov
 PackageLicenseDeclared: MIT
-PackageDownloadLocation: git://github.com/ggerganov/llama.cpp.git@da39a3ee5e6b4b0d3255bfef95601890afd80709
```
hydai, line 42: The `PackageFileName` should be updated at the same time.

hydai (changes requested):

hydai:
Is it possible to know if the OpenBLAS is set correctly?

dm4:
> Is it possible to know if the OpenBLAS is set correctly?

In cmake configure log:
```
-- BLAS found, Libraries: /usr/lib/x86_64-linux-gnu/libopenblas.so
```

or check the plugin binary dependency:
```
$ ldd ~/.wasmedge/plugin/libwasmedgePluginWasiNN.so | grep openblas
        libopenblas.so.0 => /lib/x86_64-linux-gnu/libopenblas.so.0 (0x00007f749cf88000)
```

dm4:
Continue at the new PR #2831

PR #2995, opened 2023-11-01
[AOT] Add invariant attribute for memory and global pointer

### LINKED ISSUE #2963: Bad performance in AOT mode
### Summary

[case.zip](https://github.com/WasmEdge/WasmEdge/files/13071445/report12.zip)

I run `case.wasm` in WasmEdge(AOT mode) and find that it use more time than expected. 
It takes almost twice as long as `wasmtime`, and is slower than wamr(both `llvm-jit` mode and `fast-jit` mode).

### Current State

On my machine, `WasmEdge` takes 3.79s while `Wasmtime` takes 1.97s to finish running `case.wasm`.
image

### Expected State

According to the performance of other cases. I think `WasmEdge` should take less time than other runtimes. So maybe the abnormal long execution time is caused by some performance problems in `WasmEdge`.

### Reproduction steps

```bash
1. Build with default options.
2. Compile with `wasmedgec case.wasm case-aot.wasm`
3. Execute with `time wasmedge case-aot.wasm > case.out`
```

### Screenshots

_No response_

_No response_

### Components

CLI

0.13.4-92-g278904bc

Ubuntu 20.04

### Hardware Architecture

x86_64

GCC version: 9.4.0
CMake version: 3.16.3

ibmibmibm:
* Add invariant.group attribute for global pointers and memory pointers
* fix #2963

### DIFF
--- a/lib/aot/compiler.cpp
+++ b/lib/aot/compiler.cpp
@@ -453,6 +453,8 @@ struct WasmEdge::AOT::Compiler::CompileContext {
     auto VPtr = Builder.createLoad(
         Int8PtrTy, Builder.createInBoundsGEP1(Int8PtrTy, Array,
                                               LLContext.getInt64(Index)));
+    VPtr.setMetadata(LLContext, LLVM::Core::InvariantGroup,
+                     LLVM::Metadata(LLContext, {}));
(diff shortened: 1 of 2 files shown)


hydai (approved):
LGTM

PR #3008, opened 2023-11-03
[Installer] Enable WASI_NN_GGML_CUDA

hydai:


juntao:

utils/install.py
```diff
@@ -610,6 +615,27 @@ def fix_gnu_sparse(args):
...
+                e.output,
+                e.cmd,
+            )
+            return False
+        logging.debug("%s: %s", cmd, output)
+        if "nvcc: NVIDIA (R) Cuda compiler driver" in output:
+            return True
+    else:
+        logging.warning("CUDA should be only available on Linux")
+    return False
```
SAtacker, line 636: Can you try to use https://github.com/WasmEdge/WasmEdge/blob/5cfcabb6ab3a1009e15f92244ea30f506d1c080f/utils/install.py#L1256 ?

SAtacker (changes requested):
Maybe use https://github.com/WasmEdge/WasmEdge/pull/3008/files#r1381380504

utils/install.py
hydai: Because `run_shell_command` will print an error log like the following section:

```
ERROR:root:Exception on process - rc= 127 output=  command= [u'/usr/local/cuda/bin/nvcc --version']
False
```

However, the only thing I want to do is to detect if the environment has cuda or not. Hence, using this function will make users confused.

SAtacker: `/usr/local/cuda/bin/nvcc --version 2>/dev/null` if empty then it does not exist.

Also should we check for /usr/local/cuda/bin path only or is it possible for it to be in other paths?

SAtacker: Alrighty, thanks!

hydai: > `/usr/local/cuda/bin/nvcc --version 2>/dev/null` if empty then it does not exist.
> 

Nice catch! Will process this.

hydai: No luck. The issue persists.

```
ERROR:root:Exception on process - rc= 127 output=  command= [u'/usr/local/cuda/bin/nvcc --version 2>/dev/null']
False
```

hydai: Since this is a minor issue, I will use `run_shell_command` instead from now.

juntao:
Let's try to get this merged today. Thanks!

SAtacker (approved):
Thanks!

utils/install.py
```diff
@@ -1317,6 +1338,7 @@ def __init__(
         self.ld_library_path = None
         self.dist = dist_
         self.release_package_wasmedge = None
+        self.cuda = check_nvcc(self.platform)
```
SAtacker, line 1341: You might want to move this later in the code

hydai: Sure. Let's do this in another PR.

Issue #2975, opened 2023-10-26
Add the software bill of materials


alabulei1:
Check out docs: https://clomonitor.io/docs/topics/checks/#software-bill-of-materials-sbom

jaydee029:
Hi, I would like to work on this.

jaydee029:
Looking at other CNCF projects , maybe we can move forward with SPDX format , I found this [https://github.com/kubernetes-sigs/bom](url) to satisfy our requirements, some projects add the sbom along with every release. @hydai @dannypsnl

dannypsnl:
Sounds good!

jaydee029:
So ,how do we go about this? Does the maintainer generate a file and make it a part of the release ? or do we work on this and add it to the repository in a specific place?. i'm a bit unaware of how the release process works. @dannypsnl

alabulei1:
Do you mean this changelog file? https://github.com/WasmEdge/WasmEdge/blob/master/Changelog.md

jaydee029:
So, not exactly I think i'll have to edit the .github/workflows/release.yml and add a job to generate sbom

hydai:
Fixed by #3090

PR #3090, opened 2023-12-12
Added Sbom generator action

jaydee029:
Tries to fix issue #2975.
Uses third party Github action [https://github.com/marketplace/actions/sbom-generator-action.](url)
I was not able to test it locally and it is first time editing such a huge workflow file, hence it may result in some error.
I would like to request a review , and also request a feedback about how it can be done in a better way.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -785,3 +785,4 @@ jobs:
         run: |
           tar -zcvf sbom.tar.gz ${{steps.sbom.outputs.fileName}}
           gh release upload ${{needs.create_release.outputs.version}} sbom.tar.gz --clobber
+      
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -754,3 +754,34 @@ jobs:
       - name: Push to DockerHub
         run: |
           docker push ${{ steps.docker_env.outputs.docker_image }}
+
+  generate_sbom:
+    runs-on: ubuntu-latest
+    needs: create_release
+    steps:
+      - name: checkout code
+        uses: actions/checkout@v3 
+
+      - name: Grant the safe directory for git
+        run: |
+          git config --global --add safe.directory $(pwd)
+      - name: generate sbom
+        uses: advanced-security/sbom-generator-action@v0.0.1
+        id: sbom
+        env:
+          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
+
+      - name: install gh
+        run: |
+          type -p curl >/dev/null || (apt update && apt install curl -y)
+          curl -fsSL https://cli.github.com/packages/githubcli-archive-keyring.gpg | dd of=/usr/share/keyrings/githubcli-archive-keyring.gpg \
+          && chmod go+r /usr/share/keyrings/githubcli-archive-keyring.gpg \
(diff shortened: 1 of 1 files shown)


juntao:

jaydee029:
Hey, @dannypsnl please review, can you recommend what changes are to be made.

dannypsnl:
You will need to remove EOF, but the rest I think hydai is more proper reviewer.

jaydee029:
EOF removed , please review @hydai

.github/workflows/release.yml
```diff
@@ -754,3 +754,35 @@ jobs:
...
+          && chmod go+r /usr/share/keyrings/githubcli-archive-keyring.gpg \
+          && echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/githubcli-archive-keyring.gpg] https://cli.github.com/packages stable main" | tee /etc/apt/sources.list.d/github-cli.list > /dev/null \
+          && apt update \
+          && apt install gh -y
+        
+      - name: Generate tar.gz file and upload release
+        shell: bash
+        run: |
+          tar -zcvf sbom.tar.gz ${{steps.sbom.outputs.fileName}}
+          gh release upload ${{needs.create_release.outputs.version}} sbom.tar.gz --clobber
```
hydai, line 787: Why do we have to create a tarball for this file? The SBOM file should be pretty small.

.github/workflows/release.yml
```diff
@@ -754,3 +754,35 @@ jobs:
...
+      - name: Grant the safe directory for git
+        run: |
+          git config --global --add safe.directory $(pwd)
+      - name: generate sbom
+        uses: advanced-security/sbom-generator-action@v0.0.1
+        id: sbom
+        env:
+          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
+
+      - name: install gh
```
hydai, line 774: If we will install `gh`, why don't we use [gh-sbom](https://github.com/advanced-security/gh-sbom) instead? This enables an extension of the `gh` tool and generates the SBOM with one command.

hydai (changes requested):

.github/workflows/release.yml
jaydee029: Right , So I looked at other cncf projects who have a good ossf score , and they had created a tar file for this , while my initial intuition was just uploading it as an artifact.

.github/workflows/release.yml
jaydee029: For this , i thought of being consistent with other jobs in the release.yml file , which had gh installed this way, the only purpose of installing gh was to upload the release tar.gz package. if we choose to move forward with uploading it as an artifact we don't need gh.

.github/workflows/release.yml
hydai: Tarball is also fine. May I ask which repo you are referring to?

.github/workflows/release.yml
hydai: You still need gh to upload the assets to the release page. Just want to make sure that the reason you choose the action workflow instead of the gh extension.

.github/workflows/release.yml
jaydee029: ArgoCD did

.github/workflows/release.yml
jaydee029: ohkay, so someone in cncf slack recommended the above sbom workflow action, hence i went forward with this.

.github/workflows/release.yml
jaydee029: should i close these conversations then?

hydai (approved):

PR #3094, opened 2023-12-15
Extend `RefVariant` to 128-bit to contain the value type information.

q82419:


juntao:

q82419:
Hi @am009 ,

As the PR #3129 , I think we should update due to the `RefVariant` extended into 128-bit size.
Can you help this for MSVC? Thanks!

am009:
> Hi @am009 ,
> 
> As the PR #3129 , I think we should update due to the `RefVariant` extended into 128-bit size. Can you help this for MSVC? Thanks!

For that part of code(PR #3129), I think the current implementation is correct. Just return the internal intrinsic type(int64x2) instead of the class, which is consistent with the LLVM IR function signature of `tableGet`.

https://github.com/WasmEdge/WasmEdge/blob/e7f15a93fb37ba3d2433d4c9763aaa351a4b644c/lib/aot/compiler.cpp#L1035-L1038

```diff
    if constexpr (std::is_same_v<RetT, RefVariant>) {
---      return (*Res).Ptr;
+++      return (*Res).getRawData();
    } else if constexpr (!std::is_void_v<RetT>) {
      return *Res;
    }
```

q82419:
> > Hi @am009 ,
> > As the PR #3129 , I think we should update due to the `RefVariant` extended into 128-bit size. Can you help this for MSVC? Thanks!
> 
> For that part of code(PR #3129), I think the current implementation is correct. Just return the internal intrinsic type(int64x2) instead of the class, which is consistent with the LLVM IR function signature of `tableGet`.
> 
> https://github.com/WasmEdge/WasmEdge/blob/e7f15a93fb37ba3d2433d4c9763aaa351a4b644c/lib/aot/compiler.cpp#L1035-L1038
> 
> ```diff
>     if constexpr (std::is_same_v<RetT, RefVariant>) {
> ---      return (*Res).Ptr;
> +++      return (*Res).getRawData();
>     } else if constexpr (!std::is_void_v<RetT>) {
>       return *Res;
>     }
> ```

Hi @am009 ,

Thanks for the comment. But it still not works with returning the internal type instead of the class.
Can you help this? Thanks.

q82419:
Hi @am009 ,

A friendly ping, would you help for this? Thanks!

am009:
> Hi @am009 ,
> 
> A friendly ping, would you help for this? Thanks!

This is somewhat troublesome. Can we pass the return value by pointer? (when the return type is refvarient)

q82419:
> > Hi @am009 ,
> > A friendly ping, would you help for this? Thanks!
> 
> This is somewhat troublesome. Can we pass the return value by pointer? (when the return type is refvarient)

Well, actually yes but -- the AOT part should fix for that. @ibmibmibm

am009:
I found another solution. Revert the last commit that changed the return type of `Executor::tableGet` and `Executor::refFunc`. Then do this:

```diff
diff --git a/lib/executor/engine/proxy.cpp b/lib/executor/engine/proxy.cpp
index f0921709..cf82b698 100644
--- a/lib/executor/engine/proxy.cpp
+++ b/lib/executor/engine/proxy.cpp
@@ -27,7 +27,11 @@ struct Executor::ProxyHelper<Expect<RetT> (Executor::*)(Runtime::StackManager &,
     if (unlikely(!Res)) {
       Fault::emitFault(Res.error());
     }
-    if constexpr (!std::is_void_v<RetT>) {
+    if constexpr (std::is_same_v<RetT, RefVariant>) {
+      uint64x2_t Ref = (*Res).getRawData();
+      return *reinterpret_cast<__m128*>(Ref.data());
+    } else if constexpr (!std::is_void_v<RetT>) {
       return *Res;
     }
   }
@@ -290,7 +294,7 @@ Expect<void> Executor::dataDrop(Runtime::StackManager &StackMgr,
   return {};
 }
```

This probably needs to be guarded by `#if WASMEDGE_OS_WINDOWS`.

am009:
Currently, MSVC support uses `std::array<uint64_t,2>` as `uint64x2_t`. This is also a cpp class, if it is used as the return type, the RVO will change the signature again to add an additional parameter.

For clang, the uint64x2_t is of vector type. and the return value is passed in the xmm0 register. Here we use __m128 (the equivalent vector type) to match it.

q82419:
> Currently, MSVC support uses `std::array<uint64_t,2>` as `uint64x2_t`. This is also a cpp class, if it is used as the return type, the RVO will change the signature again to add an additional parameter.
> 
> For clang, the uint64x2_t is of vector type. and the return value is passed in the xmm0 register. Here we use __m128 (the equivalent vector type) to match it.

@am009 ,
Thanks for help. It works!

q82419:
@ibmibmibm  please take a look.

ibmibmibm (approved):

Issue #3062, opened 2023-11-28
bug: `fd_pwrite` after setting `fdflags::append` writes from the wrong offset


yagehu:
### Summary

This is a subtle platform-specific WASI bug. The code sample I've included creates a file `tmp/a` and performs this sequence of actions:

1. Write 67 bytes to `tmp/a`.
2. Set the `fdflags::append` flag via `fcntl()`.
3. Write 102 bytes at offset 0 to `tmp/a` with `fd_pwrite`.

The resulting file should be 102 bytes because `fd_pwrite` should write from offset 0 regardless of the append flag per specification (WASI specifies `fd_pwrite` to be similar to [POSIX `pwritev`](https://pubs.opengroup.org/onlinepubs/9699919799/functions/write.html) which specifies this behavior).

Wasmtime behaves correctly, producing a 102-byte file. WasmEdge produces a 169-byte file.

### Reproduction steps

```c
#include <fcntl.h>
#include <unistd.h>
#include <stdio.h>
#include <stdlib.h>

int main() {
    int f = open("tmp/a", O_CREAT | O_WRONLY);
    if (f == -1) {
        perror("open");
	return 1;
    }

    char buf[67];

    int written = write(f, buf, 67);
    printf("written %d\n", written);

    int ret = fcntl(f, F_SETFL, O_APPEND);
    if (ret == -1) {
	perror("fcntl");
	return 1;
    }


    char buf2[102];

    int written2 = pwrite(f, buf2, 102, 0);
    printf("written %d\n", written2);
}
```

1. Build with wasi-sdk producing `a.out`
2. Execute `mkdir tmp && wasmedge run --dir=tmp a.out`
3. Check the newly created file size: `stat tmp/a`

### Components

CLI, C SDK, Rust SDK

### WasmEdge Version or Commit you used

0.13.5

### Operating system information

Fedora 39

### Hardware Architecture

amd64

yagehu:
Definitely a Linux bug as you linked. It was [first raised by Micharl Kerrisk](https://bugzilla.kernel.org/show_bug.cgi?id=43178). You can argue WASI aims to be platform independent, so a specification compliant implementation should fix this in userland. The Wasmtime team has probably fixed it by using cap-std. I leave it up to you WasmEdge folks to decide if you wontfix it or not. I'm just reporting this deviation from the specification :)

hydai:
[...]

PR #3106, opened 2023-12-21
Fix large timestamp causing overflow

### LINKED ISSUE #3105: `fd_filestat_set_times` with a large timestamp leads to errno `inval`
### Summary

Calling `fd_filestat_set_times` with a large timestamp causes it to return errno 28 (inval).  In other runtimes, it succeeds.  Specifically, a large timestamp is something larger than int64 max like `13136941795642118959`.

Note that you cannot test this case using wasi-sdk's `futimens` because `struct timespec` only contains int64 `tv_sec`.

### Current State

_No response_

### Expected State

_No response_

### Reproduction steps

Run this WAT with wasmedge, mounting any directory.  It should exit without error. But with WasmEdge it exits with 28 (inval).

```wat
(module
  (type (func))
  (type (func (param i32 i64 i64 i32) (result i32)))
  (type (func (param i32)))
  (import "wasi_snapshot_preview1" "fd_filestat_set_times" (func $fd_filestat_set_times (type 1)))
  (import "wasi_snapshot_preview1" "proc_exit" (func $proc_exit (type 2)))
  (func $_start
    (local $err i32)

    (call $fd_filestat_set_times
      (i32.const 3) ;; preopened dir
      (i64.const 13136941795642118959) ;; very large timestamp
      (i64.const 0)
      (i32.const 1) ;; atim
    )
    (local.set $err)
    (if
      (i32.ne (i32.const 0) (local.get $err))
      (then (call $proc_exit (local.get $err)))
    )

    (call $proc_exit (i32.load (i32.const 2048)))
  )
  (memory 2)
  (export "memory" (memory 0))
  (export "_start" (func $_start))
)
```

### Screenshots

_No response_

_No response_

### Components

CLI

d1a4df1827f789cc109fdbc66e2ae8dc35fa7008

Fedora 39

### Hardware Architecture

amd64

_No response_

yagehu:
This commit fixes a bug in `toTimespec`.  The duration overflows with a large timestamp because `std::chrono::nanoseconds` is signed 64-bit integer.  Instead, we should explicitly use `uint64_t` because `__wasi_timestamp_t` is unsigned 64-bit.

fixes #3105

### DIFF
--- a/lib/host/wasi/linux.h
+++ b/lib/host/wasi/linux.h
@@ -264,7 +264,7 @@ inline constexpr clockid_t toClockId(__wasi_clockid_t Clock) noexcept {
 
 inline constexpr timespec toTimespec(__wasi_timestamp_t Timestamp) noexcept {
   using namespace std::chrono;
-  const auto Total = nanoseconds(Timestamp);
+  const auto Total = duration<uint64_t, std::nano>(Timestamp);
   const auto Second = duration_cast<seconds>(Total);
   const auto Nano = Total - Second;
   timespec Result{};


juntao:

alabulei1:
@hydai  Can you please take a look? Thanks.

hydai:
Hi @yagehu 
I found some WASI tests failed. Could you check if this is related to your commit or not?

yagehu:
@hydai The failing wasi-testsuite case is `pwrite-with-append`, which is a newly added case related to https://github.com/WasmEdge/WasmEdge/issues/3062. So it is not related. Maybe I should remove this test case, it seems most runtimes are failing this and it is not trivial to fix and there's no consensus.

hydai (approved):

Issue #3014, opened 2023-11-05
Potential Uninitialized Memory Usage in WasmEdge


khagankhan:
### Summary

A MemorySanitizer warning is triggered when running a WasmEdge tool with a specific .wasm file. The warning indicates a `use-of-uninitialized-value within the std::filesystem::path` destructor, which is invoked in the `WasmEdge::Plugin::Plugin::getDefaultPluginPaths[abi:cxx11]()` function.

I believe the issue may be related to in `WasmEdge/lib/plugin/plugin.cpp`

```
// ...
auto LibPath = std::filesystem::u8path(DLInfo.dli_fname)
                       .parent_path()
                       .lexically_normal();
// ...
```

### Current State

MemorySanitizer warning is triggered, indicating a use-of-uninitialized-value in the `std::filesystem::path` destructor, specifically in the `std::unique_ptr<std::filesystem::__cxx11::path::_List::_Impl, std::filesystem::__cxx11::path::_List::_Impl_deleter>::~unique_ptr()` function.

The full MSAN log:

```
==2902116==WARNING: MemorySanitizer: use-of-uninitialized-value
    #0 0x7fa331ba47cf in std::unique_ptr<std::filesystem::__cxx11::path::_List::_Impl, std::filesystem::__cxx11::path::_List::_Impl_deleter>::~unique_ptr() /usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/unique_ptr.h:360:6
    #1 0x7fa331d2b49c in std::filesystem::__cxx11::path::_List::~_List() /usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/fs_path.h:647:24
    #2 0x7fa331d2b49c in std::filesystem::__cxx11::path::~path() /usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/bits/fs_path.h:313:21
    #3 0x7fa331d2b49c in WasmEdge::Plugin::Plugin::getDefaultPluginPaths[abi:cxx11]() /root/wasmoi/targets/WasmEdge/lib/plugin/plugin.cpp:254:20
    #4 0x7fa331c1594f in WasmEdge::Driver::DriverToolOptions::add_option(WasmEdge::PO::ArgumentParser&) /root/wasmoi/targets/WasmEdge/include/driver/tool.h:147:29
[...]

PR #3135, opened 2024-01-03
Proper initialization of DLInfo

khagankhan:
Fixed https://github.com/WasmEdge/WasmEdge/issues/3014

### DIFF
--- a/lib/plugin/plugin.cpp
+++ b/lib/plugin/plugin.cpp
@@ -252,8 +252,8 @@ std::vector<std::filesystem::path> Plugin::getDefaultPluginPaths() noexcept {
       dladdr(reinterpret_cast<void *>(Plugin::getDefaultPluginPaths), &DLInfo);
   if (Status != 0) {
     if (DLInfo.dli_fname == nullptr) {
-      spdlog::error("Address matched to a shared object but not to any symbol "
-                    "within the object. dli_fname is null.");
+      spdlog::error("Address matched to a shared object but not to any symbol "sv
+              "within the object. dli_fname is null."sv);
       return std::vector<std::filesystem::path>();
     }
     auto LibPath = std::filesystem::u8path(DLInfo.dli_fname)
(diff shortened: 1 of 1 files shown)


hydai (changes requested):
According to the man page:
```
RETURN VALUE         [top](https://man7.org/linux/man-pages/man3/dladdr.3.html#top_of_page)
       On success, these functions return a nonzero value.  If the
       address specified in addr could be matched to a shared object,
       but not to a symbol in the shared object, then the
       info->dli_sname and info->dli_saddr fields are set to NULL.

       If the address specified in addr could not be matched to a shared
       object, then these functions return 0.  In this case, an error
       message is not available via [dlerror(3)](https://man7.org/linux/man-pages/man3/dlerror.3.html).
```

The error message should be more accurate.

dannypsnl:
Please read https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

khagankhan:
Oh, I see! I will sign it off now when adding detailed error message.

khagankhan:
It is almost done. If anything needed please let me know - "we've come too far to give up".

Thank you for your guidance and the chance to work on this issue! o7

lib/plugin/plugin.cpp
```diff
@@ -272,6 +277,11 @@ std::vector<std::filesystem::path> Plugin::getDefaultPluginPaths() noexcept {
       Result.push_back(LibPath / std::filesystem::u8path(".."sv) /
                        std::filesystem::u8path("plugin"sv));
     }
+  } else {
+    spdlog::error(ErrCode::Value::NonNullRequired);
+    spdlog::error("Address could not be matched to any shared object. "
+                  "Detailed error information is not available.");
```
hydai, line 283: Use `"..."sv` instead of a normal string, like line 278 `"plugin"sv`.

hydai (changes requested):

hydai (approved):

PR #3149, opened 2024-01-10
the unique_ptr at the place is not required

### LINKED ISSUE #3148: bug: Unnecessary copy of full module when loading
### Summary

After applying the loader of compoent model (#3036), the loader load create a unique ptr and load from file / bytes.
https://github.com/WasmEdge/WasmEdge/blob/master/lib/loader/loader.cpp#L108

But this function return the raw pointer from the unique ptr, which causes the copy of full module.
https://github.com/WasmEdge/WasmEdge/blob/master/lib/loader/loader.cpp#L126

Even though the function is safe (the module is copied into the returned variant container and then the unique ptr destroyed), but this will cause the overhead.

### Current State

_No response_

### Expected State

Return the unique_ptr directly.
This should be fixed before the 0.14.0 release.

### Reproduction steps

None.

### Screenshots

_No response_

_No response_

### Components

C SDK

master

Ubuntu 20.04

### Hardware Architecture

x86_64

_No response_

dannypsnl:
resolve #3148

### DIFF
--- a/lib/loader/loader.cpp
+++ b/lib/loader/loader.cpp
@@ -95,7 +95,7 @@ Loader::parseWasmUnit(const std::filesystem::path &FilePath) {
       return Unexpect(Res);
     }
 
-    std::unique_ptr<AST::Module> Mod;
+    AST::Module Mod;
     if (auto Code = LMgr.getWasm()) {
       // Set the binary and load module.
       // Not to use parseModule() here to keep the `WASMType` value.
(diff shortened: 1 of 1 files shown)


juntao:

hydai:
@q82419 
PTAL

dannypsnl:
@hydai The problem here is we are trying to avoid copy constructor get called, it seems need to redesign these API surface.

dannypsnl:
@q82419 The copy constructors won't get called in loader now

q82419 (approved):

PR #3152, opened 2024-01-10
[component model] wrong loading

dannypsnl:
1. fix the concept of various type index
2. fix import/export name rule
3. some section of component model can be repeated
4. fix core-module and nested component loading
5. fix core-module's bound loading


dannypsnl:
There is a bigger trouble in, the section loader says it's expecting a vector with size 1, though the wasm-tools also tells there have 4 types. The implementation correctly load the four corresponding types, but the binary contains wrong vector size.

dannypsnl:
Follows https://github.com/second-state/component-model-tests?tab=readme-ov-file#prepare-wasmedge-and-test to test the branch.

dannypsnl:
@q82419 Please take a look.

q82419 (approved):

PR #3206, opened 2024-02-08
Fix the linking error with simdjson

### LINKED ISSUE #2791: Cannot find simdjson header when library is found
## Description

In `cmake` log can check the simdjson library is found

```
-- SIMDJSON found
```

, but later `ninja` command failed with

```
WasmEdge/test/spec/spectest.cpp:20:10: fatal error: 'simdjson.h' file not found
```

### Current State

simdjson.h file not found, build failed

### Expected

If library is found, build should pass

## Environment

- Hardware Architecture: aarch64
- Operating system: MacOS 13.4.1

- C++ Compiler version: AppleClang 14.0.3.14030022
- CMake version: 3.26.1
- CMake flags: `cmake -DCMAKE_BUILD_TYPE=Debug -DWASMEDGE_BUILD_TESTS=ON -GNinja ..`
- Path of header: `/usr/local/include/simdjson.h`

## Steps to Reproduce

```shell
mkdir build && cd build
cmake -DCMAKE_BUILD_TYPE=Debug -DWASMEDGE_BUILD_TESTS=ON -GNinja ..
ninja
```

hankluo6:
Fix #2791 

The issue was due to simdjson being exported with a namespace, which led to `find_package` successfully locating simdjson but failing to link the library without specifying the namespace. Prefixing simdjson with its namespace in `target_link_libraries` resolves this linking issue.

### DIFF
--- a/plugins/wasi_nn/CMakeLists.txt
+++ b/plugins/wasi_nn/CMakeLists.txt
@@ -143,7 +143,7 @@ target_include_directories(wasmedgePluginWasiNN
 
 if(BACKEND STREQUAL "ggml")
   target_include_directories(wasmedgePluginWasiNN PUBLIC ${CMAKE_BINARY_DIR}/_deps/llama-src)
-  target_link_libraries(wasmedgePluginWasiNN PRIVATE common simdjson)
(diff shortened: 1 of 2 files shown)


hydai (approved):
Nice catch. Thanks.

PR #3246, opened 2024-02-27
[WASI] fix special case in vinode path resolving

### LINKED ISSUE #3231: Bug: (WASI) Failed to create directory with trailing backslash
### Summary

```echo "`wasmedge -v` | `rustc -V` | `uname -m`"```
```wasmedge version 0.13.5 | rustc 1.76.0 (07dca489a 2024-02-04) | x86_64```

Compiling a Rust test case to `wasm32-wasi` and running it with the WasmEdge CLI raises an exception, when the expected behavior is to succeed. The test case (shown below) is relatively simple. It attempts to create a new directory.
```
use std::fs;

fn main() {
    fs::create_dir("foo/").unwrap();
}
```
The test case runs successfully if the trailing slash in the directory name is omitted (`foo/` is replaced with `foo`). The exception raised by the test case is `ENOENT`.

The [2018 edition of POSIX](https://pubs.opengroup.org/onlinepubs/9699919799.2018edition/) states that `mkdir` raises an `ENOENT` exception when "A component of the path prefix specified by path does not name an existing directory or path is an empty string", where a path prefix prefix is defined as "The part of a pathname up to, but not including, the last component and any trailing <slash> characters"

The path prefix of `foo/` is the empty string, which does not contain components naming a non-existent directory. Therefore, a `ENOENT` exception should not be raised.

The test case runs successfully when given to other Wasm runtimes (Wasmtime, Wasmer, WAMR, and Wasmi) as well as when compiled to `x86_64_unknown_linux_gnu` and run natively.

This test case was derived from a program generated by Wasimilar, an [Xsmith](https://www.flux.utah.edu/project/xsmith)-based random program generator.

### Current State

The test case fails to create a new directory and rasies an `ENOENT` exception.

Here is my observed output:
```
thread 'main' panicked at src/main.rs:4:28:
called `Result::unwrap()` on an `Err` value: Os { code: 44, kind: NotFound, message: "No such file or directory" }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
[2024-02-21 23:47:10.078] [error] execution failed: unreachable, Code: 0x89
[2024-02-21 23:47:10.078] [error]     In instruction: unreachable (0x00) , Bytecode offset: 0x00006b2a
[2024-02-21 23:47:10.078] [error]     When executing function name: "_start"
```

### Expected State

The test case should create a directory, `foo`, without raising an exception.

### Reproduction steps

Here is a zip file of a rust project that reproduces the behavior: 
[wasmedge-mkdir-bug.zip](https://github.com/WasmEdge/WasmEdge/files/14370043/wasmedge-mkdir-bug.zip)

1. Unzip wasmedge-mkdir-bug.zip
2. Go to `wasmedge-mkdir-bug`
3. Compile with `cargo build --target wasm32-wasi`
4. Run with `wasmedge --dir=. target/wasm32-wasi/debug/wasmedge-mkdir-bug.wasm`
5. Observe that an exception is thrown with the message "No such file or directory"

### Screenshots

_No response_

_No response_

### Components

CLI

0.13.5

Ubuntu 22.04

### Hardware Architecture

x86_64

_No response_

### PR DESCRIPTION
**Pull request by tannal**
This patch attempt to fix #3231.

The resolvePath function uses fstatat for path resolution, which will return error when the relative path (foo) is not exist.

### DIFF
--- a/lib/host/wasi/vinode.cpp
+++ b/lib/host/wasi/vinode.cpp
@@ -106,7 +106,7 @@ WasiExpect<std::shared_ptr<VINode>> VINode::bind(__wasi_rights_t FRB,
 WasiExpect<void> VINode::pathCreateDirectory(std::shared_ptr<VINode> Fd,
                                              std::string_view Path) {
   std::vector<char> Buffer;
-  if (auto Res = resolvePath(Fd, Path); unlikely(!Res)) {
+  if (auto Res = resolvePath(Fd, Path, false); unlikely(!Res)) {
     return WasiUnexpect(Res);
   } else if (!Fd->can(__WASI_RIGHTS_PATH_CREATE_DIRECTORY)) {
     return WasiUnexpect(__WASI_ERRNO_NOTCAPABLE);

### CODE REVIEW

**Comment by juntao** [comment]

File: lib/host/wasi/vinode.cpp [hunk 1]
```diff
@@ -105,15 +105,9 @@ WasiExpect<std::shared_ptr<VINode>> VINode::bind(__wasi_rights_t FRB,
 
 WasiExpect<void> VINode::pathCreateDirectory(std::shared_ptr<VINode> Fd,
                                              std::string_view Path) {
-  std::vector<char> Buffer;
-  if (auto Res = resolvePath(Fd, Path); unlikely(!Res)) {
```
hydai, line 109: We should handle the special cases inside the `resolvePath` function. Removing the path checking is not the correct way to fix this issue.

hydai (changes requested):

dannypsnl:
Please check the failed tests and sign your commits by adding email line.

tannal:
For ubuntu machine, we might need to install liblld-14-dev(depends on the llvm version) here.

https://github.com/WasmEdge/WasmEdge/actions/runs/8111354854/workflow?pr=3246#L46

ibmibmibm (approved):

ibmibmibm (approved):

hydai (approved):

PR #3265, opened 2024-03-08
[Installer] Add rustls plugin's name into env file

hankluo6:
Previously, the `rustls` plugin could not be removed by `uninstall.sh` because it was not correctly placed in the `env` due to its name not containing `Plugin`. 

Check the end of `to_path` can prevent the need for hard coding `ggml-metal.metal` and ensure the plugin's name is properly added.


juntao:

utils/install.py
```diff
@@ -178,8 +178,8 @@ def extract_archive(
                     fname = fname[:-5] + "lib"
                 if fname.startswith("/usr") and "lib64" in fname:
                     fname = fname.replace("lib64", "lib", 1)
-                # ggml-metal.metal is downloaded if we download ggml plugin on macOS
-                if "Plugin" in fname or fname == "ggml-metal.metal":
+
+                if to_path.endswith("Plugins"):
```
hydai, line 182: This seems incorrect.
The default installed path is `~/.wasmedge/plugin`.

hydai (changes requested):

utils/install.py
hankluo6: `to_path` is the path in the `TEMP_PATH`, where the archive file gets extracted, so I think it is not related to `~/.wasmedge`.

When `extract_archive` is [called](https://github.com/WasmEdge/WasmEdge/blob/b74598db9aa0a07f98526d36e6a2e6e3e17e9d4e/utils/install.py#L1247) for installing the plugin, "Plugins" will be the final segment in the `to_path`.

hydai:
Hi @SAtacker 
Could you help review this PR?
This PR aims to fix an issue: the rustls plugin is incorrectly handled in the env file.

utils/install.py
SAtacker: Then how about `if "Plugin" in fname or "Plugins" in fname:` ?
Reason is that there could be something that does not end with `Plugins`

SAtacker: Besides If there is not `Plugin` in fname, then for sure there is not `Plugins` in fname

SAtacker: Any way we can get the complete test logs @hydai ?

SAtacker: So it's trying to download github.com/second-state/WasmEdge-tensorflow/releases/download/0.12.0/WasmEdge-tensorflow-0.12.0-darwin_arm64.tar.gz which does not exist

SAtacker: This CI fail does not seem to be related to this PR.

hydai: @SAtacker We should ignore the macOS latest CI, it's not related to this PR. And the Intel model macOS doesn't have the metal files.

hydai:
> > Hi @SAtacker Could you help review this PR? This PR aims to fix an issue: the rustls plugin is incorrectly handled in the env file.
> 
> This looks a better fix than the one already there.

Would you approve of this? It seems like this PR can fix the above issue.

SAtacker (approved):

hydai (approved):

hydai:
Thanks

PR #3321, opened 2024-04-03
[Installer] If CUDA is detected, use ubuntu20.04 instead of manylinux2014

hydai:
Fixes https://github.com/LlamaEdge/LlamaEdge/issues/142

### DIFF
--- a/utils/install.py
+++ b/utils/install.py
@@ -1188,6 +1188,8 @@ def install_plugins(args, compat):
                 # Re-write the plugin name if CUDA is available
                 if plugin_name == WASI_NN_GGML and compat.cuda:
                     plugin_name = WASI_NN_GGML_CUDA
+                    # If CUDA is detected, we will use the ubuntu20.04 assets instead of any other linux distro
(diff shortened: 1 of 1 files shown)


utils/install.py
```diff
@@ -1188,6 +1188,8 @@ def install_plugins(args, compat):
                 # Re-write the plugin name if CUDA is available
                 if plugin_name == WASI_NN_GGML and compat.cuda:
                     plugin_name = WASI_NN_GGML_CUDA
+                    # If CUDA is detected, we will use the ubuntu20.04 assets instead of any other linux distro
+                    compat.dist = "ubuntu20.04"
```
SAtacker, line 1192: If we are changing this value just for the plugins we should restore it's old value at the end of the function, just to avoid future accidental changes that would break things.

SAtacker (changes requested):

hydai:
The manylinux2014 will be deprecated soon, closing this. We prefer to make the default installation distribution from manylinux2014 to ubuntu20.04 or manylinux2_28.

PR #3334, opened 2024-04-12
[WASI-NN] ggml: fix embedding with batch decode

dm4:
This PR updates the `getEmbedding` function to generate the same results as the [embedding example in llama.cpp](https://github.com/ggerganov/llama.cpp/tree/master/examples/embedding).

I have tested it locally to ensure that it generates the same results for the models `llama-2-7b-chat.Q5_K_M.gguf` and `gritlm-7b_q4_1.gguf` as those in `llama.cpp`. I think we should include one of those models in our example CI later to ensure that it executes successfully.


juntao:

hydai (approved):

PR #3380, opened 2024-05-02
[Component] interface types

dannypsnl:
1. `InterfaceType` inherit `public ValType`
2. `using ValInterface = std::variant<ValVariant, std::string>`
3. extract component host function
4. plugin: wasi-poll
5. misc: related updates


juntao:

dannypsnl:
NOTE: If I update existed execute signatures, will fail many tests and break API, so I will keep both versions for a while.

dannypsnl:
```c++
include/runtime/component/hostfunc.h --- C++
107 107 
108 108   template <typename Tuple, typename SpanT, size_t... Indices>
109 109   static Tuple toTuple(SpanT &&Args, std::index_sequence<Indices...>) {
110 110     return Tuple(std::holds_alternative<ValVariant>((Args)[Indices])
... 111                      ? std::get<ValVariant>(std::forward<SpanT>(Args)[Indices])
111 112                            .template get<std::tuple_element_t<Indices, Tuple>>()
... 113                      : std::get<std::tuple_element_t<Indices, Tuple>>(
... 114                            std::forward<SpanT>(Args)[Indices])...);
112 115   }
113 116 
114 117   template <typename Tuple, typename SpanT, size_t... Indices>
```

This patch should be valid typed, but the `std::get<std::tuple_element_t<Indices, Tuple>>` argue that cannot find `std::tuple_element_t<Indices, Tuple>` in type list.

include/common/types.h
```diff
@@ -599,10 +576,6 @@ template <> inline ValType ValTypeFromType<float>() noexcept {
 template <> inline ValType ValTypeFromType<double>() noexcept {
   return ValType(TypeCode::F64);
 }
-// wasm interface types
-template <> inline ValType ValTypeFromType<StrVariant>() noexcept {
-  return ValType(TypeCode::String);
-}
```
dannypsnl, line 605: TODO: should be
```c++
template <> inline ValType ValTypeFromType<std::string>() noexcept {
  return InterfaceType(...);
}
```

dannypsnl: besides that, we will need a concrete way to handle the case like
```
template<> inline ValType ValTypeFromType<std::vector<T>>() noexcept {
  return InterfaceType(
    TypeCode::List,
    ValTypeFromType<T>
  );
}
```

@q82419 Do you have any idea about such construction?

dannypsnl:
```c++
template <typename T>
inline ValType ValTypeFromType<std::vector<T>>() noexcept {
  return InterfaceType(TypeCode::String, ValTypeFromType<T>());
}
```
is invalid, but above also invalid due to `T` is missing.

include/common/types.h
```diff
@@ -599,9 +580,8 @@ template <> inline ValType ValTypeFromType<float>() noexcept {
 template <> inline ValType ValTypeFromType<double>() noexcept {
   return ValType(TypeCode::F64);
 }
-// wasm interface types
-template <> inline ValType ValTypeFromType<StrVariant>() noexcept {
-  return ValType(TypeCode::String);
+template <> inline ValType ValTypeFromType<std::string>() noexcept {
+  return InterfaceType(TypeCode::String);
```
dannypsnl, line 584:
```c++
template<> inline ValType ValTypeFromType<std::vector<T>>() noexcept {
  return InterfaceType(
    TypeCode::List,
    ValTypeFromType<T>()
  );
}
```
vs
```c++
template <typename T>
inline ValType ValTypeFromType<std::vector<T>>() noexcept {
  return InterfaceType(TypeCode::List, ValTypeFromType<T>());
}
```

dannypsnl:
Take note and mark as resolved.

### Problem 1

```c++
template <typename Tuple, typename SpanT, size_t... Indices>
static Tuple toTuple(SpanT &&Args, std::index_sequence<Indices...>) {
  return Tuple(std::forward<SpanT>(Args)[Indices]
                   .template get<std::tuple_element_t<Indices, Tuple>>()...);
}
```
cannot get proper handled after modification, even the type is correct, since `std::variant` strictly need a concrete type, we need to ensure `std::tuple_element_t<Indices, Tuple>`.

dannypsnl:
### Problem 2

Due to limitation of C++ function specialization, I cannot do below, I will try class template instead.
```c++
template<> inline ValType ValTypeFromType<std::vector<T>>() noexcept {
  return InterfaceType(
    TypeCode::List,
    ValTypeFromType<T>()
  );
}
```

```c++
template <typename T>
inline ValType ValTypeFromType<std::vector<T>>() noexcept {
  return InterfaceType(TypeCode::List, ValTypeFromType<T>());
}
```

dannypsnl:
### Problem 3

If we consider the definition:
```c++
using ValInterface = std::variant<std::string, ValVariant>;
```

There is a trouble: we would like to introduce any `vector<T>` as a variant of `ValInterface`, but we can't do it without template. cc @q82419 if any solution comes up.

dannypsnl (commented):
Seems like I accidentally remove the `return Comp...` part......

dannypsnl (commented):
NOTE: This should fix MacOS x64

dannypsnl:
### Ubuntu/Debian
```
/usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/variant:1409:2: note: in instantiation of default argument for 'variant<std::basic_string<char>, void, void>' required here
        variant(_Tp&& __t)
        ^~~~~~~~~~~~~~~~~~
/__w/WasmEdge/WasmEdge/lib/executor/instantiate/component/canon.cpp:230:30: note: while substituting deduced template arguments into function template 'variant' [with _Tp = typename std::remove_reference<basic_string<char> &>::type, $1 = (no value), $2 = (no value), _Tj = (no value), $4 = (no value)]
        HigherArgs.push_back(std::move(S));
     
```

### Alpine
```
#30 34.52             _Alloc_traits::construct(this->_M_impl, this->_M_impl._M_finish,
#30 34.52                            ^
#30 34.52 /src/lib/driver/runtimeTool.cpp:158:16: note: in instantiation of function template specialization 'std::vector<std::variant<bool, std::basic_string<char>, std::shared_ptr<WasmEdge::ValComp>, WasmEdge::Variant<unsigned int, int, unsigned long, long, float, double, unsigned __int128, __int128, __attribute__((__vector_size__(2 * sizeof(unsigned long)))) unsigned long, __attribute__((__vector_size__(2 * sizeof(long)))) long, __attribute__((__vector_size__(4 * sizeof(unsigned int)))) unsigned int, __attribute__((__vector_size__(4 * sizeof(int)))) int, __attribute__((__vector_size__(8 * sizeof(unsigned short)))) unsigned short, __attribute__((__vector_size__(8 * sizeof(short)))) short, __attribute__((__vector_size__(16 * sizeof(unsigned char)))) unsigned char, __attribute__((__vector_size__(16 * sizeof(signed char)))) signed char, __attribute__((__vector_size__(4 * sizeof(float)))) float, __attribute__((__vector_size__(2 * sizeof(double)))) double, WasmEdge::RefVariant>>>::emplace_back<std::basic_string<char>>' requested here
#30 34.52       FuncArgs.emplace_back(std::move(Value));
#30 34.52 
```

dannypsnl:
Even more crashed......

dannypsnl:
Next idea: still push_back, but use current `std::string` immediate construction.

dannypsnl:
@hydai Do we still want to fix manylinux2014?

dannypsnl:
By introducing `ValInterface` explicit, we fix build failed on each platform, which caused by ambiguous semantic when we do construction.

1. Alpine clang cannot decide `push_back` should use `T` of `vector<T>`
2. Ubuntu/Debian clang cannot decide `emplace_back` should use `T` of `vector<T>`

dannypsnl:
@q82419 need a review here.

plugins/wasi_http/func.cpp
```diff
@@ -13,21 +13,18 @@
 namespace WasmEdge {
 namespace Host {
 
-Expect<void> WasiHttpPrint::body(const Runtime::CallingFrame &,
```
hydai, line 16: Do the components not need the calling frame information?

lib/vm/vm.cpp
```diff
@@ -523,8 +525,20 @@ VM::unsafeExecute(std::string_view Func, Span<const ValVariant> Params,
...
     // Execute function and return values with the module instance.
-    return unsafeExecute(ActiveModInst.get(), Func, Params, ParamTypes);
+    auto Res = unsafeExecute(ActiveModInst.get(), Func, Params, ParamTypes);
+    if (!Res) {
+      return Unexpect(Res);
+    }
+    return *Res;
   }
+  spdlog::error(ErrCode::Value::WrongInstanceAddress);
+  spdlog::error(ErrInfo::InfoExecuting("", Func));
```
hydai, line 535: Why do we put an empty string here?

lib/executor/instantiate/component/canon.cpp
```diff
@@ -79,30 +80,28 @@ AST::FunctionType convert(Runtime::Instance::ComponentInstance &Comp,
...
 public:
-  LiftTrans(Executor *Exec, const FuncType &DefinedType,
+  LiftTrans(Executor *Exec, const AST::Component::FuncType &DefinedType,
             Instance::FunctionInstance *Func, Instance::MemoryInstance *M,
             Instance::FunctionInstance *R,
             Runtime::Instance::ComponentInstance &Comp)
-      : HostFunctionBase(0), Exec(Exec), LowerFunc(Func), Memory(M),
-        Realloc(R) {
-    auto &FT = DefType.getCompositeType().getFuncType();
+      : HostFunctionBase(), Exec(Exec), LowerFunc(Func), Memory(M), Realloc(R) {
```
hydai, line 89: The initialized value, `0`, is removed. Is this expected?

lib/executor/instantiate/component/canon.cpp
```diff
@@ -218,21 +217,22 @@ class LowerTrans : public HostFunctionBase {
...
     for (auto &ParamTy : HigherFuncType.getParamTypes()) {
       switch (ParamTy.getCode()) {
       case TypeCode::String: {
         auto Idx = Args[PI++];
         auto Len = Args[PI++];
         std::string_view V =
             Memory->getStringView(Idx.get<uint32_t>(), Len.get<uint32_t>());
-        std::string S{V.begin(), V.end()};
-        HigherArgs.push_back(StrVariant(std::move(S)));
+        // FIXME: This crashed on Ubuntu/Debian clang
```
hydai, line 228: Does this crash still occur?

hydai (changes requested):

plugins/wasi_http/func.cpp
dannypsnl: Yes, because component cannot directly touch any CallingFrame, but indirectly via original module.

lib/vm/vm.cpp
dannypsnl: Missing message

lib/executor/instantiate/component/canon.cpp
dannypsnl: That's fixed, remove it.

include/runtime/instance/component.h
```diff
@@ -39,11 +41,26 @@ namespace Instance {
 using namespace AST::Component;
 
 namespace {
-void typeConvert(ValueType &VT, const ValType &ValTy) noexcept {
-  switch (ValTy.getCode()) {
+void typeConvert(ValueType &VT, const ValType &Ty) noexcept {
+  switch (Ty.getCode()) {
+  case TypeCode::I8:
+    VT.emplace<PrimValType>(PrimValType::S8);
```
hydai, line 47: `I8` is an unsigned type. Does this mean all of the unsigned integers will be interpreted as signed integers?

hydai (changes requested):
And I found the coverage downgrades. Do you have a plan to add more tests in this PR?

dannypsnl:
> And I found the coverage downgrades. Do you have a plan to add more tests in this PR?

Yes, #3590 is preparing some handcraft test data.

include/runtime/instance/component.h
dannypsnl: Oops, I think `I8` is signed one, the final target should be we have both at the surface.

dannypsnl: I though `Ix` is signed and `Ux` is unsigned.

hydai: I see. Misunderstood.

hydai (commented):
LGTM now. @q82419 PTAL

q82419 (changes requested):
Comment:

With using the `std::string` directly for the component type, maybe we should concern about the call-by-reference case in the future because the parameter is a local variable in the host function body.

lib/vm/vm.cpp
```diff
@@ -541,12 +574,15 @@ VM::unsafeExecute(std::string_view ModName, std::string_view Func,
   const auto *FindModInst = StoreRef.findModule(ModName);
   if (FindModInst != nullptr) {
     // Execute function and return values with the module instance.
-    return unsafeExecute(FindModInst, Func, Params, ParamTypes);
-  } else {
-    spdlog::error(ErrCode::Value::WrongInstanceAddress);
-    spdlog::error(ErrInfo::InfoExecuting(ModName, Func));
-    return Unexpect(ErrCode::Value::WrongInstanceAddress);
+    auto Res = unsafeExecute(FindModInst, Func, Params, ParamTypes);
```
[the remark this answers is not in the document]
dannypsnl: some legacy I thought, removed.

dannypsnl:
Debian:
```
/__w/WasmEdge/WasmEdge/build/_deps/fmt-src/include/fmt/ranges.h:211:49: error: self-comparison always evaluates to true [-Werror=tautological-compare]
  211 |                      integer_sequence<bool, (Is == Is)...>) -> std::true_type;
      |                                              ~~~^~~~~
cc1plus: all warnings being treated as errors
```

q82419 (approved):

dannypsnl:
@hydai This is ready to merge.

hydai (changes requested):
We are not allowed to have wasm file here. Please concider using byte array inside the unit test files.
* test/component/data/core.wasm
* test/component/data/http.wasm

hydai:
If we ship these files, then we are not able to ship WasmEdge on Debian due to their policy.

dannypsnl:
@hydai I put files into vectors, also I'm not going to mark semantic of bytes, content can change in the near future due to unstable tool.

hydai (approved):

Issue #3319, opened 2024-04-02
feat: support WASI-NN streaming extension for NNRPC


hydai:
### Summary

WasmEdge forked the WASI-NN spec with the following extension for a better LLM experience.
Three new functions need to be supported:
* compute_single
* get_output_single
* fini_single

### Details

Make the NNRPC support streaming LLM mode.
1. Update the wasi-nn proto file: https://github.com/WasmEdge/WasmEdge/blob/master/lib/wasi_nn_rpc/wasi_ephemeral_nn.proto
2. Update the NNRPC implementation

### Appendix

* https://github.com/second-state/wasmedge-wasi-nn/blob/ggml/rust/src/graph.rs#L270-L280
* https://github.com/second-state/wasmedge-wasi-nn/blob/ggml/rust/src/graph.rs#L294-L309

arvganesh:
Hi @hydai, I'm a student at the University of Texas at Austin! I was interested in working on this issue as a part of my virtualization class (CS 360V). I've got the necessary background on LLMs but am new to WasmEdge – do you think it would make sense for me to work on this? Thanks!

juntao:
@arvganesh Please have a look at Akihiro Suda's talk on NNRPC here (from 2:08)

https://youtu.be/D0D8ufWtILI?si=nTF5nklApkmO7cCu&t=128

If you can build and run NNRPC from source, then you are ready for this. :)

arvganesh:
Would you be able to assign this to me? Thanks!

juntao:
Done. Look forward to your contributions!

arvganesh:
Hi @juntao, @hydai! I wanted to outline my understanding of this issue / WasmEdge, let me know if this makes sense before I begin working on it:

Background:
- WasmEdge is a WebAssembly runtime
- Wasi-NN defines a set of APIs and abstractions (Tensors, Backends, Graphs) enabling neural network inference.
    - WASI = WebAssembly System Interface, which provides low-level system interfaces for WebAssembly applications.
    - Wasi-NN provides an interface to use existing NN-backends (TensorFlow, OpenVINO, etc) for inference.
- https://github.com/second-state/wasmedge-wasi-nn/tree/ggml – this project contains Rust bindings implementing the Wasi-NN spec. 
[...]

PR #3506, opened 2024-06-26
[WASI-NN] ggml: support compute single in RPC mode

dm4:
As #3319, this PR adds support for `ComputeSingle`, `GetOutputSingle`, and `FiniSingle` in RPC mode for the ggml backend.

- Use grpc metadata to handle the `BackendError`.
- Add the `GGMLBackendComputeSingleWithRPC` test.
- Since the CI test of NNRPC is no longer running after we disabled NNRPC by default, I tried to re-enable it in this PR.


juntao:
Hello, I am a [code review bot](https://github.com/flows-network/github-pr-summary/) on [flows.network](https://flows.network/). Here are my reviews of code commits in this PR.

------

dm4:
Hi @AkihiroSuda, could you please review this pull request when you have a moment? Thank you!

.github/workflows/build-extensions.yml
```diff
@@ -143,17 +143,35 @@ jobs:
...
+            --nn-preload default:GGML:AUTO:build_rpc/test/plugins/wasi_nn/wasinn_ggml_fixtures/orca_mini.gguf &
+          RPC_SERVER_PID=$!
           sleep 3
-          cd ${test_dir}
           # The test binary consumes $WASI_NN_RPC_TEST_URI
-          ./${test_bin} --gtest_filter=WasiNNTest.GGMLBackendWithRPC
+          (cd ${nnrpc_test_dir} && ./${test_bin} --gtest_filter=WasiNNTest.GGMLBackendWithRPC)
+          kill -9 "$RPC_SERVER_PID"
+
+          # Restart the server for the compute single test
```
AkihiroSuda, line 167: Why restart?

lib/wasi_nn_rpc/wasi_ephemeral_nn.proto
```diff
@@ -18,6 +18,22 @@ enum TensorType {
   I64 = 6;
 }
 
+enum BackendError {
```
AkihiroSuda, line 21: Wondering if this should be grpc metadata
https://grpc.io/docs/guides/metadata/

dm4: Using grpc metadata is indeed a good idea, but we should also return the backend error type to the client, because the client may behave differently depending on the error. For example, in the [llama-stream example](https://github.com/second-state/WasmEdge-WASINN-examples/blob/master/wasmedge-ggml/llama-stream/src/main.rs#L240-L252), the backend error type is used to determine whether to continue inference. If we use grpc metadata, we might also need to ensure that the client still has the opportunity to continue execution in the event of a backend error.

@hydai, do you have any thoughts?

.github/workflows/build-extensions.yml
dm4: The RPC server is restarted here because the official WASI-NN does not support unloading graphs. If the RPC server is not restarted, the graph IDs obtained in the tests will continually increase, resulting in behavior that is inconsistent with standalone execution.

lib/wasi_nn_rpc/wasi_ephemeral_nn.proto
hydai: Please correct me if I am wrong.
The gprc is used between nnrpc-server and nnrpc-client. Once the client receives the grpc metadata, it will convert the metadata into a backend error and send it to the application.

In this case, we will need to:
1. Error occurs on the server side; wrap the backend error into grpc metadata
2. Server sends the grpc metadata to the client
3. The client decodes the grpc metadata and converts it back to the backend error
4. The client returns the backend error code to the WASM application

If we keep the backend error:
1. Error occurs on the server side; set the backend error as the error code.
2. The client receives the error code and returns it directly

I am not familiar with using grpc. Is it a best practice to use metadata to handle the errors? If so, we would love to adopt it.

AkihiroSuda: I think using metadata is fine, as the error should be supported for all RPC calls, not just `ComputeSingleResult`.

An alternative idea would be to just add `BackendError error = 1` to all the RPC calls without using metadata.

dm4:
Hi @hydai and @AkihiroSuda, I just replace the `BackendError` part with grpc metadata. Could you please review this pull request when you have a moment? Thank you!

hydai (approved):
LGTM

PR #3510, opened 2024-06-28
[WASI-NN] ggml: bump llama.cpp b3259

dm4:


hydai (approved):
Note that the Windows CI failed, we plan to fix it later.

PR #3521, opened 2024-07-01
Implementing Coredump for Post-Mortem Debugging in WebAssembly

sid-008:
### Implementation status
- [x]  Define new CLI flags to enable coredump file generation, i.e being able to run a WASM module in debug mode. Read more about the coredump spec and about the various custom sections, memory sections, data sections in WASM .
- [x] Modify the section of WasmEdge code that traps execution in case of abnormal behaviour/errors to support coredump file generation(which is just a dummy file for now) and understanding how coredump runtime code was injected into the WASM module(referencing this: [wasm-edit](https://github.com/xtuc/wasm-coredump/tree/main/bin/rewriter)). This is an intermediate goal to better understand how the information would be logged and how the coredump file would be generated.
- [x] Write the module for coredump file generation. The coredump file is to be encoded using the [Wasm binary format](https://webassembly.github.io/spec/core/binary/index.html). Be able to generate a dummy coredump.
- [x] **Process and Instance Information Custom Section:** Implementing support for collecting information about processes and module information as defined in: [Process information](https://github.com/WebAssembly/tool-conventions/blob/main/Coredump.md#process-information) and in [instance information](https://github.com/WebAssembly/tool-conventions/blob/main/Coredump.md#module-information), we will collect information from their relevant custom sections
- [ ] **Threads and Stack frames Custom Section:** Implementing support for collecting debug info on threads and stack frames, as defined [here](https://github.com/WebAssembly/tool-conventions/blob/main/Coredump.md#threads-and-stack-frames). Here as well, we will collect information from their relevant custom sections.
- [ ] **Linear Memory Custom Section:** Adding support for linear memory screenshotting. As defined [here](https://github.com/WebAssembly/tool-conventions/blob/main/Coredump.md#memory). We shall collect information from the memory section and data section here
- [x] **Globals Custom Section:** Adding support for capturing Globals from the global section as defined in the spec [here](https://github.com/WebAssembly/tool-conventions/blob/main/Coredump.md#globals).
- [x] Making any changes if required to ensure that the coredump file is compatible with the current toolchain, including wasmgdb and wasm-edit.
- [ ] Writing tutorials, examples, and demonstrations of post-mortem debugging using the implemented coredump feature.


include/executor/executor.h
```diff
@@ -137,6 +138,9 @@ class Executor {
     if (Stat) {
       Stat->setCostLimit(Conf.getStatisticsConfigure().getCostLimit());
     }
+    if (Conf.getRuntimeConfigure().isEnableCoredump()) {
+      std::cout<<"Coredump has been enabled in executor";
```
dannypsnl, line 142: please config your editor so these sort of format problem can be automatically eliminated.

include/common/configure.h
```diff
@@ -25,6 +25,7 @@
 #include <mutex>
 #include <optional>
 #include <shared_mutex>
+#include <sys/wait.h>
```
hydai, line 28: Why do we need this header?

include/common/configure.h
```diff
@@ -115,8 +116,9 @@ class RuntimeConfigure {
   RuntimeConfigure(const RuntimeConfigure &RHS) noexcept
       : MaxMemPage(RHS.MaxMemPage.load(std::memory_order_relaxed)),
         EnableJIT(RHS.EnableJIT.load(std::memory_order_relaxed)),
+        EnableCoredump(RHS.EnableCoredump.load(std::memory_order_relaxed)),
         ForceInterpreter(RHS.ForceInterpreter.load(std::memory_order_relaxed)),
-        AllowAFUNIX(RHS.AllowAFUNIX.load(std::memory_order_relaxed)) {}
```
hydai, line 119: Please apply clang-format and clang-tidy

include/common/coredump.h
```diff
@@ -0,0 +1,30 @@
...
+#pragma once
+#include <iostream>
+
+namespace WasmEdge{
+namespace Coredump{
+
+class Coredump {
+public:
+  Coredump() {
+    std::cout<<("Coredump class Constructor called\n");
```
hydai, line 24: Use `spdlog` from the log.h instead. Don't use stdout/stderr.

include/executor/executor.h
```diff
@@ -124,7 +125,7 @@ class HostFuncHandler {
 
 /// Executor flow control class.
 class Executor {
-public:
+public: /// add another argument to this for coredump to initialise it.
```
hydai, line 128: Why do we need this comment here? Is it a TODO item?

lib/api/wasmedge.cpp
```diff
@@ -1050,6 +1050,11 @@ WasmEdge_ConfigureDelete(WasmEdge_ConfigureContext *Cxt) {
 
 // <<<<<<<< WasmEdge configure functions <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
 
+// WASMEDGE_CAPI_EXPORT WasmEdge_CoredumpContext *
```
hydai, line 1053: If you want to add any commented out section, leave `TODO` with comments.

lib/driver/runtimeTool.cpp
```diff
@@ -128,6 +128,10 @@ int Tool(struct DriverToolOptions &Opt) noexcept {
     Conf.getCompilerConfigure().setOptimizationLevel(
         WasmEdge::CompilerConfigure::OptimizationLevel::O1);
   }
+  if (Opt.ConfEnableCoredump.value()) {
+    std::cout<<"coredump enabled\n";
```
hydai, line 132: Use spdlog instead.

hydai (changes requested):

include/executor/executor.h
sid-008: Sure. Seems like clang-format wasn't running correctly on CLion. Has been fixed on my end now. Since this is a draft PR, I'll remove the print/log statements later on. This draft PR is mostly intended to inform GSoC mentors of my progress.

include/common/configure.h
sid-008: Sure thing. Seems like clang-format wasn't running correctly in my IDE. Apologies

include/common/configure.h
sid-008: I think the LSP added it for some reason while I was making some changes. We do not require it and I have removed it.

sid-008: I think the LSP added it for some reason while I was making some changes. We do not require it and I have removed it.

include/executor/executor.h
sid-008: Yes, it is a TODO item. Will label it accordingly.

sid-008:
Currently, there is an easy way to get all required information. Once we are able to push in the Core section of the coredump which contains just the executable name, we should be able to just pick up all the information like execution frames from the StackMgr, which is why it is passed to the coredump class.

A roadblock currently exists in the final stage where the custom sections should be pushed into the module class for serialisation. I am unable to get the correct reference to the vector that contains the module's custom sections. It gives me a span which is a read only view. I require the vector itself.

Do let me know if there need to be any changes in the code or if there is a better way of doing it.

Current Status:
I have not included code for the memory or data sections in this commit since it was filled with too much code that wasn't really working and needs reconsidering. 
I also require a way to get all the frames from the StackManager. Would using `getTopN` work for this but I think that returns only the valuestack? Or is there a better way to get that information. The corestack section will contain this and the thread information for that. Do let me know how I could find the thread information/name.

hydai:
> * I also require a way to get all the frames from the StackManager. Would using `getTopN` work for this but I think that returns only the valuestack? Or is there a better way to get that information. The corestack section will contain this and the thread information for that. Do let me know how I could find the thread information/name.

Currently, you can get the active frame because it will only use the first top frame. However, adding a new method to get a `Span` of all frames is fine. The `getTopN` is used to receive the values from the stack.
Let's ignore the thread information here since the Wasm execution is a single-thread model. Even WasmEdge itself can be used in multi-threading use cases, but we should focus on the single thread during this GSoC period.

> * Any help on figuring out how to access the storemanager class or the VM class itself would be helpful. For reference, the coredump builder is initialized in the executor.

When you have a frame, it should contain the information of a module, then you should find the Linked Store(`storemanager class`). Ref: https://github.com/WasmEdge/WasmEdge/blob/master/include/runtime/instance/module.h#L521-L532
However, the current implementation only allows you to do `link` and `unlink`, but I think adding a new function called `getLinkedStore` for this use case is okay.

> * Is there a way to pass through the execution context struct itself to the coredump builder or a reference to the current active executor(idk if this is possible however)? Maybe I could do some setup through the Executor constructor? This would be helpful in accessing the memories and global sections.

If you have the frame, then you have the executor.

> * Would changing the namespace of the coredump file help with accessing the VM class?

This may not help.

> * A last question. Why are there two statistics classes declared in VM and Executor? Is one of them unused?

Which two kinds of classes do you talk about?

sid-008:
Thank you for the inputs

## Current Thoughts/Implementation:
- The core custom section is now implemented.
- The coremodule section is done but is only applicable for multi-instance support. The same is true for coreinstances section. This was added much later on and does not seem to be supported by other runtimes implementing the coredump spec. I am unfamiliar with multiple module execution but wasmedge does support it. [https://wasmedge.org/docs/embed/c/multiple_modules/](https://wasmedge.org/docs/embed/c/multiple_modules/). Currently the other runtimes do not support it as is evident by [https://github.com/xtuc/wasm-coredump/issues/5](https://github.com/xtuc/wasm-coredump/issues/5). I will be implementing this however.
- Question now is if a wasm program has only a single instance and a single module would it be wiser to omit these custom sections for simplicity? The spec does not define anything about this.


## Appendix:
- [https://github.com/WebAssembly/tool-conventions/blob/main/Coredump.md#module-information](Module Information)
- [https://github.com/WebAssembly/tool-conventions/blob/main/Coredump.md#instance-information](Instance Information)

sid-008:
The most latest commit is a WIP draft. There may be a slight mess. Apologies

sid-008:
While trying to create a getter for the Store, I have run into an issue.

Context:
The linked store is defined by   std::map<StoreManager *, std::function<BeforeModuleDestroyCallback>>
      LinkedStore;

i.e a StoreManager* and an associated destroy callback.

What I am trying to do:
Get the associated LinkedStore's StoreManager entry for a particular module. 

The issue:
 
public:
void getStores() const {
    std::unique_lock Lock(Mutex);
    std::cout<<LinkedStore.size();
    for (auto &item : LinkedStore) {
      std::cout << "Store: " << item.first << "\n";
    }
  }


How I am using this:
For any given frame get the ModuleInstance. The ModuleInstance should have an associated LinkedStore to it.

For this, the size is 0 for the linked store and no items are being printed out. I shall also push this code to the draft PR for your reference. https://github.com/WasmEdge/WasmEdge/pull/3521/commits/044a066f84ed03fa7e4b25ee2c2cde0fb4f01d27

In contrast creating a getter for the OwnedMemInsts similarly works fine. Any help would be much appreciated. 
tldr; I am unable to get the linkedStore for the ModuleInstance

sid-008:
### Update
- Working on the data section and memory sections. These can be pulled from the `ModuleInstance` through a getter function for `OwnedDataInsts` and `OwnedMemInsts`.
- The coredump is valid WASM and can be validated as well run in wasmgdb(incompletely)

sid-008:
### Update
- it seems like a few changes are breaking changes with regards to Wasmgdb tooling compatibility for multimodule support as has been noted by the spec's authors. LinkRel: [https://github.com/WebAssembly/tool-conventions/pull/207#pullrequestreview-1417478535](https://github.com/WebAssembly/tool-conventions/pull/207#pullrequestreview-1417478535)
- I think multi-module is tricky considering the maturity of the spec and it seems to have not been updated for a year or so now, so getting changes to it would take a while. 
- As far as I can tell, Wasmgdb does not support multimodule coredumps either. I've looked through the commits on both wasmgdb as well as the Coredump spec. Since the spec was updated with the multimodule support, wasmdgb has not had any commits related to supporting multimodule + I've tried it as well, there's no way to interact with that information and adding the `instanceidx` field in `corestack` section(as mentioned above) breaks the debugging flow in Wasmgdb. 
- There is also an open issue [https://github.com/xtuc/wasm-coredump/issues/5](https://github.com/xtuc/wasm-coredump/issues/5) on the coredump-rewriter and wasmgdb monorepo regarding this.
- In my opinion it would be wiser to move multimodule coredump support out of the GSoC work period, since it would be tricky to implement or would take far too long considering various challenges related to tooling and the spec.
- I'd like to try and implement multimodule support post the GSoC period if I can, the code for a few sections to support it has been written pretty much, I've just commented it out for now. It just seems to ignore the `coreinstances` and `coremodules` sections. But for the change in the `corestack` section it won't work.

cc: @hydai

sid-008:
### Current blockers
- I am having a hard time figuring out the corestack section. I need in particular: `locals:vec(value)` and `stack:vec(value)`
- Currently I am iterating over all frames and trying to get the locals and stack values through that. The issue is the Frame itself only contains `vpos` and `locals` indicating the number of locals for the frame. 
- I also need a way to be able to get the corresponding binary type encoding of these values. I don't know how this would be possible since the `ValueStack` uses `ValVariant` as its data type and there seems to be no way to actually get the binary encoding(i.e 0x7f for i32 for example) for a `ValVariant`
- Until then I shall finish up the data and global sections, since those are non blocking with regards to progress.

sid-008:
### Update
- The most recent commit contains `corestack`, partially implemented `data` and `global` sections.
- There may be no need for a globals section. I think it comes with the DWARF information in the original WASM [https://yurydelendik.github.io/webassembly-dwarf/#location-descriptions-globals](https://yurydelendik.github.io/webassembly-dwarf/#location-descriptions-globals). I was able to list globals even without the globals custom section in the coredump.
- Please do have a look at the commit and leave reviews. I am actively working on it so its a Work In Progress commit.

sid-008:
### Current Concerns/doubts/reviews required:
- In `CoreStack` I require `thread-info:thread name` and need to verify if the way I am getting Locals and Stack values is valid or not. Need validation also on whether the `funcidx` is correct or not.
- `MemorySection` The initial `limit` is being set to max `limit` for some reason. Need to fix that. Apart from that it looks fine.
- `DataSection` Section is malformed, needs further debugging. Any advice on capturing all active data segments and inserting them into one contiguous active segment would be appreciated(wasmgdb does not support multiple data segments)
- A limited set of commands works. 

cc: @hydai

hydai:
We are working on the 0.14.1 release, so your PR will be reviewed on Tuesday.

You posted several comments over the weekend. Are all of them up to date, or could you summarize your blockers and questions in a single comment?

sid-008:
> You posted several comments over the weekend. Are all of them up to date, or could you summarize your blockers and questions in a single comment?

Yes, I had a few questions/some code for you to take a look at:
- I require a way to obtain the Thread Info i.e thread name for `corestack`
- Required some code validation if the information is correctly collected in `corestack` i.e the `funcidx` and locals and stack values vectors.
- Any help on capturing active datasegments into the data section for coredump would be appreciated. I suffered a setback here due to wasmgdb not supporting more than one active contiguous data segment within the data section.
- The memory section's `Limit` is being captured with initial as the maximum and not the other way around for some reason, any help with that would be appreciated if you could have a look



> Also, it seems that you only add the implementation without unit tests. Would you plan to have tests for your component?

Yes, I think I've not pushed the commits for the tests because I did `git add` only in a certain file. I'll get back to you on that shortly.

sid-008:
> You posted several comments over the weekend. Are all of them up to date, or could you summarize your blockers and questions in a single comment?

Yes, I had a few questions/some code for you to take a look at:
- I require a way to obtain the Thread Info i.e thread name for `corestack`
- Required some code validation if the information is correctly collected in `corestack` i.e the `funcidx` and locals and stack values vectors.
- Any help on capturing active datasegments into the data section for coredump would be appreciated. I suffered a setback here due to wasmgdb not supporting more than one active contiguous data segment within the data section.
- The memory section's `Limit` is being captured with initial as the maximum and not the other way around for some reason, any help with that would be appreciated if you could have a look



> Also, it seems that you only add the implementation without unit tests. Would you plan to have tests for your component?

Yes, I think I've not pushed the commits for the tests because I did `git add` only in a certain file. I'll get back to you on that shortly.

sid-008:
> You posted several comments over the weekend. Are all of them up to date, or could you summarize your blockers and questions in a single comment?

Yes, I had a few questions/some code for you to take a look at:
- I require a way to obtain the Thread Info i.e thread name for `corestack`
- Required some code validation if the information is correctly collected in `corestack` i.e the `funcidx` and locals and stack values vectors.
- Any help on capturing active datasegments into the data section for coredump would be appreciated. I suffered a setback here due to wasmgdb not supporting more than one active contiguous data segment within the data section.
- The memory section's `Limit` is being captured with initial as the maximum and not the other way around for some reason, any help with that would be appreciated if you could have a look



> Also, it seems that you only add the implementation without unit tests. Would you plan to have tests for your component?

Yes, I think I've not pushed the commits for the tests because I did `git add` only in a certain file. I'll get back to you on that shortly.

sid-008:
> You posted several comments over the weekend. Are all of them up to date, or could you summarize your blockers and questions in a single comment?

Yes, I had a few questions/some code for you to take a look at:
- I require a way to obtain the Thread Info i.e thread name for `corestack`
- Required some code validation if the information is correctly collected in `corestack` i.e the `funcidx` and locals and stack values vectors.
- Any help on capturing active datasegments into the data section for coredump would be appreciated. I suffered a setback here due to wasmgdb not supporting more than one active contiguous data segment within the data section.
- The memory section's `Limit` is being captured with initial as the maximum and not the other way around for some reason, any help with that would be appreciated if you could have a look



> Also, it seems that you only add the implementation without unit tests. Would you plan to have tests for your component?

Yes, I think I've not pushed the commits for the tests because I did `git add` only in a certain file. I'll get back to you on that shortly.

hydai:
Here are some issues you can do before we review your PR:

1. Some Linux-related CIs failed in the testing process with a segfault. You can reproduce it with docker and figure out these problems
2. Windows requires strict type conversion. If you have some codes that will convert type, please use explicit type conversion.
3. The macOS-related CIs also failed; if you don't have a macOS machine, you can create a PR to your own repo, and then GitHub will have free runners for running these CIs for you.

hydai:
> * I require a way to obtain the Thread Info i.e thread name for `corestack`

The current implementation is a single thread, so I think it's fine to use "main" or "" as the thread name.

> * Required some code validation if the information is correctly collected in `corestack` i.e the `funcidx` and locals and stack values vectors.

Are you asking us to generate the test data for you?

> * Any help on capturing active datasegments into the data section for coredump would be appreciated. I suffered a setback here due to wasmgdb not supporting more than one active contiguous data segment within the data section.

Since this is a toolchain issue, let's reduce the situation. For your GSoC period, you only need to focus on one active data segment.

> * The memory section's `Limit` is being captured with initial as the maximum and not the other way around for some reason, any help with that would be appreciated if you could have a look

According to the wasm spec, users can provide a maximum only for the `Limit`, and the `Limit` will use this maximum value as the minimum and maximum. Of course, users can also provide both maximum and minimum value for the `Limit`.

sid-008:
> Are you asking us to generate the test data for you?

No, I'd just like you to have a look at the code and give your opinions on if you feel it could be implemented better

include/common/coredump.h
```diff
@@ -0,0 +1,279 @@
...
+    std::vector<Byte> &Version = Mod.getVersion();
+    Version.insert(Version.begin(), {0x01, 0x00, 0x00, 0x00});
+
+    Mod.getCustomSections().emplace_back(Core);
+    // Mod.getCustomSections().emplace_back(CoreModules);
+    // Mod.getCustomSections().emplace_back(CoreInstances);
+    Mod.getCustomSections().emplace_back(CoreStack);
+    // Mod.getDataSection() = std::move(DataSection);
+    Mod.getMemorySection() = std::move(MemSec);
+    // TODO globals works with DWARF data, is this section needed?
```
hydai, line 64: `Globals` should be accessible from a module. I didn't get the point about the relationship between `Globals` and `DWARF data`.

include/common/coredump.h
```diff
@@ -0,0 +1,279 @@
...
+    //  Mod.getGlobalSection() = std::move(Globals);
+    auto Res = Ser.serializeModule(Mod);
+    std::ofstream File("coredump.wasm", std::ios::out | std::ios::binary);
+    if (File.is_open()) {
+      File.write(reinterpret_cast<const char *>(Res->data()), Res->size());
+      File.close();
+    } else {
+      // Handle error opening file
+      throw std::ios_base::failure("Failed");
```
hydai, line 73: You should not use `throw` without any try-catch block. Instead, use `excepted<T>` for the return type, allowing you to handle the error.

include/common/coredump.h
```diff
@@ -0,0 +1,279 @@
...
+      const Runtime::Instance::ModuleInstance *CurrentInstance) {
+    spdlog::info("Constructing Core");
+    // AST node
+    AST::CustomSection Core;
+    Core.setName("core");
+
+    // Insert in data and set Name of section
+    auto Name = CurrentInstance->getModuleName();
+    std::vector<Byte> Data(reinterpret_cast<const Byte *>(Name.data()),
+                           reinterpret_cast<const Byte *>(Name.data()) +
```
hydai, line 87: A minor issue here. The class of `Name` should have `begin()` and `end()`. You can use it to construct the `Data` vector.

include/common/coredump.h
```diff
@@ -0,0 +1,279 @@
...
+
+    AST::MemorySection MemSec;
+    const auto *CurrentInstance = StackMgr.getModule();
+    auto &MemInstances = CurrentInstance->getOwnedMemoryInstances();
+    // In the current version of WebAssembly, at most one memory may be defined
+    // or imported in a single module, and all constructs implicitly reference
+    // this memory.
+    std::cout << MemInstances[0]->getMemoryType().getLimit().getMin();
+    // auto &Content = MemSec.getContent();
+    // TODO why is this setting max and min to the same value?
```
hydai, line 111: Please read the spec of `Limit` about the meaning of the maximum and minimum values.

include/runtime/stackmgr.h
```diff
@@ -74,6 +74,17 @@ class StackManager {
...
   }
 
+  Span<Value> getRangeSpan(uint32_t Start, uint32_t End) {
+    if (Start >= End) {
+      // Handle invalid range
+      spdlog::info(Start);
+      spdlog::info(End);
+      spdlog::info(ValueStack.size());
+      throw std::out_of_range("Invalid range for getRangeSpan");
+    }
```
hydai, line 84: Use `assuming` instead of throwing an error.

lib/driver/runtimeTool.cpp
```diff
@@ -128,6 +128,10 @@ int Tool(struct DriverToolOptions &Opt) noexcept {
     Conf.getCompilerConfigure().setOptimizationLevel(
         WasmEdge::CompilerConfigure::OptimizationLevel::O1);
   }
+  if (Opt.ConfEnableCoredump.value()) {
+    spdlog::info("coredump enabled in options");
```
hydai, line 132: Remove the useless log.

hydai (changes requested):

hydai:
> Most CIs should now be passing. Working on fixing the windows one now. Some spots require strict type conversion. Will check on it. Will push related tests as well by tomorrow. Bit late here now. Thanks

The CIs which are passing now because you removed the initialization of the `Coredump` class. It seems incorrect. Anyway, I will wait for your related tests tomorrow.

sid-008:
> > Most CIs should now be passing. Working on fixing the windows one now. Some spots require strict type conversion. Will check on it. Will push related tests as well by tomorrow. Bit late here now. Thanks
> 
> The CIs which are passing now because you removed the initialization of the `Coredump` class. It seems incorrect. Anyway, I will wait for your related tests tomorrow.

So the issues namely were:
- Removal of initialisation of coredump class in the VM did not make a difference. For both debug and release the same issues persisted. 
- Issue on release builds: there was a TODO item to check if coredump was enabled or not in the options for the unreachable instruction. Adding the check made sure it did not interfere with the respective tests.
- Issue on debug builds: there was a need to add the `wasmedgeSerialiser`(iirc) in the cmake list for wasi socket tests. Or else it would recognise it as an undefined symbol.
- Removing the initialisation of coredump and executor made no difference to the tests passing/build suceeding. Coredump right noe seems to be behaving correctly.

include/common/coredump.h
```diff
@@ -0,0 +1,287 @@
...
+      auto Funcidx = Item.From->getSourceIndex(); // function index
+      // TODO which to use?
+      //  auto Funcidx = Item.From->getTargetIndex(); // function index
+      auto Codeoffset = Item.From->getOffset();
+      // auto Codeoffset = 0; // only if codeoffset is unknown?
+      auto LocalsN = Item.Locals;
+
+      auto Locals = StackMgr.getRangeSpan(Item.VPos, LocalsN + Item.VPos);
+      auto Stack = StackMgr.getTopSpan(StackMgr.size() -
+                                       LocalsN); // TODO is this correct?
```
hydai, line 189: You will need to use the relationship between N-1, N, and N+1 to retrieve the locals and stacks.

```
Frames |...|        N-1 frame          |        N frame        |         N+1 frame         | ... |
Stack  |...| Locals(N-1) | Stacks(N-1) | Locals(N) | Stacks(N) | Locals(N+1) | Stacks(N+1) | ... |
       |...|              ^            |            ^          |              ^            | ... |
       |...|              | VPos(N-1)  |            | VPos(N)  |              | VPos(N+1)  | ... |
```

include/common/coredump.h
```diff
@@ -0,0 +1,287 @@
...
+  // AST::CustomSection collectCoreInstances(Runtime::StackManager &StackMgr) {
+  //   spdlog::info("Constructing CoreInstances");
+  //
+  //   AST::CustomSection CoreInstances;
+  //   CoreInstances.setName("coreinstances");
+  //   auto Frames = StackMgr.getAllFrames();
+  //
+  //   // struct CoreInstance {
+  //   //   const Runtime::Instance::ModuleInstance Moduleidx;
+  //   //   std::vector<uint8_t> Memories;
```
hydai, line 271: If we simplify the situation, let's say we disable the multi-memory proposal, and you will only have the memories with index 0.

include/common/coredump.h
```diff
@@ -0,0 +1,287 @@
...
+  //   //   std::vector<uint8_t> Globals;
```
hydai, line 272: You can use `getGlobalNum()` to get the total number of globals.

include/common/coredump.h
```diff
@@ -0,0 +1,287 @@
...
+
+  AST::CustomSection collectProcessInformation(
+      const Runtime::Instance::ModuleInstance *CurrentInstance) {
+    spdlog::info("Constructing Core");
+    // AST node
+    AST::CustomSection Core;
+    Core.setName("core");
+
+    // Insert in data and set Name of section
+    auto Name = CurrentInstance->getModuleName();
```
hydai, line 92: According to the spec, the process information is:

```
core         ::= customsec(process-info)
process-info ::= 0x0 executable-name:name
```

So, we should put the executable name, not the module name.
The executable name should be the name of the wasm file. For now, you can use `a.wasm` as a placeholder.

include/common/coredump.h
```diff
@@ -0,0 +1,287 @@
...
+  //
+  //   return Globals;
+  // }
+  // This custom section establishes an index space of modules
+  // that can be referenced in the coreinstances custom section
+  // NOTE This is for multi-module coredumps which are not supported by wasmgdb
+  // as of now
+  // coremodules := customsec(vec(coremodule)) coremodule := 0x0
+  // module-name:name
+  // AST::CustomSection collectCoreModules(Runtime::StackManager &StackMgr) {
```
hydai, line 232: According to the spec,

```
coremodules ::= customsec(vec(coremodule))
coremodule ::= 0x0 module-name:name
```

Even though we support a single module, we should use `getModuleName()` as the `module-name`, not the address of the module.

include/common/coredump.h
```diff
@@ -0,0 +1,287 @@
...
+
+    Mod.getCustomSections().emplace_back(Core);
+    // Mod.getCustomSections().emplace_back(CoreModules);
+    // Mod.getCustomSections().emplace_back(CoreInstances);
+    Mod.getCustomSections().emplace_back(CoreStack);
+    // Mod.getDataSection() = std::move(DataSection);
+    Mod.getMemorySection() = std::move(MemSec);
+    // TODO globals works with DWARF data, is this section needed?
+    //  Mod.getGlobalSection() = std::move(Globals);
+    const AST::Module Final = Mod;
```
hydai, line 72: This copy is unnecessary. What's the point of this?

hydai (changes requested):
Besides the previous review, there are some things that need to be fixed before we merge this PR.

hydai:
> * Removal of initialisation of coredump class in the VM/passing it to executor did not make a difference to the tests/failing builds nor was it necessary even for VM to have initialised it imo. For both debug and release the same issues persisted upon removing that initialisation. Not to mention coredump in itself does not require any sort of initialisation the ways the `Statistics` class did(which is initially how I tried to model coredump). Coredump collects post mortem, statistics seems to collect during runtime if I am right.

So, we don't even need the `CoreDump` class. It seems like this class can be replaced by its internal functions directly. There is no member inside the `CoreDump` class.

> * Issue on debug builds: there was a need to add the `wasmedgeSerialiser`(iirc) in the cmake list for wasi socket tests. Or else it would recognise it as an undefined symbol.

The reason is that you introduce the Serializer in the `CoreDump` class, and use it in the executor. That's why we need to update the cmakelist to fix the undefined symbol issue.

hydai:
Summary:

To get your PR merged, there are some requirements that need to be fulfilled:

1. CIs must be green, which means the code compiles and tests pass.
2. Remove unnecessary debug logs.
3. The issues mentioned above should be resolved, such as logic or implementation errors.
4. Tests should be added to verify the executions to ensure the behavior is correct.

The implementation assumes the following situations:
1. Multi-memory proposal is disabled, which means there will be a single memory (index 0).
2. Multi-module support is disabled, which means there will be a single module.

sid-008:
@hydai  Can I include the tests within `ExecutorTest.cpp` itself or would it be better to create a new file itself?

include/common/coredump.h
sid-008: While collecting these values is there any way to also get the binary encoding of their types? All values are of type `ValVariant` I think for these at least if I write code to match for the following types it shouldn't really be an issue.

value ::= 0x01       => ∅
        | 0x7F n:i32 => n
        | 0x7E n:i64 => n
        | 0x7D n:f32 => n
        | 0x7C n:f64 => n

sid-008:
> So, we don't even need the `CoreDump` class. It seems like this class can be replaced by its internal functions directly. There is no member inside the `CoreDump` class.

Would you suggest refactoring to just make them member functions of the executor?

hydai:
> @hydai Can I include the tests within `ExecutorTest.cpp` itself or would it be better to create a new file itself within the Executor folder?

We are going to have the CoreDump added to the `Executor`. It's fine to include these tests within `ExecutorTest.cpp`.

> Would you suggest refactoring to just make them member functions of the executor?

Yes.

hydai:
I am not available during the weekend, so please make sure you put the most updates before Friday night. That will be my last review for your PR before the Monday deadline.

include/common/coredump.h
sid-008: Yeah, i don't think v128 is supported in wasmgdb. Regarding the stack trace as well: Funcidx maps perfectly to `targetIndex` for a frame's `From` iterator.

sid-008:
> I am not available during the weekend, so please make sure you put the most updates before Friday night. That will be my last review for your PR before the Monday deadline.

Got it. Is there anyone else I can contact in case I have doubts?

sid-008:
> I don't think so. We will have the Mid-Autumn Festival. And most of us are OOO until 9/18.

Ah alright then. I'll push as much as i can by tomo end of day

include/common/coredump.h
sid-008: Does the arity play a role in how many values are in the value stack as well?

hydai: > Does the arity play a role in how many values are in the value stack as well? Is there a simpler way to get both locals and stack values for a particular frame than this?

No. Arity means the number of the return values. So you cannot use it for calculating the range of the stack. You should use VPos and LocalsN to calculate just like I mentioned above.

sid-008: Yep got it, the locals and stack variables have been captured correctly now. The issue is now in the binary encoding within the coredump which I am figuring out.

include/common/coredump.h
sid-008: Regarding this, I think the coredump-rewriter(my reference for coredump) itself maybe gets this wrong. Upon inspecting the memory section of the original binary that generates the trap, I can see that: `0	min=17 max=none` where 0 is the index. For some reason coredump-rewriter sets this to `initial: 0, maximum: Some(17)`

include/executor/executor.h
```diff
@@ -125,6 +125,7 @@ class HostFuncHandler {
 /// Executor flow control class.
 class Executor {
 public:
+  /// TODO add another argument to this for coredump builder to initialise it.
```
hydai, line 128: We don't need this anymore since there is no need to initialize the core dump builder.

include/executor/executor.h
```diff
@@ -137,6 +138,9 @@ class Executor {
     if (Stat) {
       Stat->setCostLimit(Conf.getStatisticsConfigure().getCostLimit());
     }
+    if (Conf.getRuntimeConfigure().isEnableCoredump()) {
+      spdlog::info("Coredump has been enabled in executor");
```
hydai, line 142: Why do we need this log?

include/runtime/instance/module.h
```diff
@@ -93,6 +95,22 @@ class ModuleInstance {
     return ModName;
   }
 
+  const std::vector<DataInstance *> &getOwnedDataInstances() const {
+    return DataInsts;
+  }
+
+  const std::vector<MemoryInstance *> getMemoryInstances() const {
```
hydai, line 102: Why is this type different from the others?

include/runtime/stackmgr.h
```diff
@@ -74,6 +75,11 @@ class StackManager {
     return Span<Value>(ValueStack.end() - N, N);
   }
 
+  Span<Value> getRangeSpan(uint32_t Start, uint32_t End) {
```
hydai, line 78: The name of `End` makes no sense. The constructor of `Span<T>` is `(BeginOffset, Length)`.

lib/executor/engine/engine.cpp
```diff
@@ -109,6 +109,10 @@ Expect<void> Executor::execute(Runtime::StackManager &StackMgr,
     // Control instructions.
     case OpCode::Unreachable:
       spdlog::error(ErrCode::Value::Unreachable);
+      if (Conf.getRuntimeConfigure().isEnableCoredump()) {
+        spdlog::info("Coredump Generation taking place");
```
hydai, line 113: Please remember to remove this line once this PR is ready.

lib/executor/executor.cpp
```diff
@@ -206,5 +207,190 @@ Executor::asyncInvoke(const Runtime::Instance::FunctionInstance *FuncInst,
...
+  std::vector<Byte> &Version = Mod.getVersion();
+  Version.insert(Version.begin(), {0x01, 0x00, 0x00, 0x00});
+
+  Mod.getCustomSections().emplace_back(Core);
+  // Mod.getCustomSections().emplace_back(CoreModules);
+  // Mod.getCustomSections().emplace_back(CoreInstances);
+  Mod.getCustomSections().emplace_back(CoreStack);
+  // Mod.getDataSection() = std::move(DataSection);
+  Mod.getMemorySection() = std::move(MemSec);
+  // TODO globals works with DWARF data, is this section needed?
```
hydai, line 240: I still don't know why the `Globals` is related to `DWARF data`. But we must store the globals section.

lib/executor/executor.cpp
```diff
@@ -206,5 +207,190 @@ Executor::asyncInvoke(const Runtime::Instance::FunctionInstance *FuncInst,
...
+
+// Data section
+// TODO malformed section
+AST::DataSection Executor::collectDataSection(Runtime::StackManager &StackMgr) {
+  spdlog::info("Collecting Memories");
+
+  AST::DataSection Datasec;
+
+  const auto *CurrentInstance = StackMgr.getModule();
+  auto &DataInstances = CurrentInstance->getOwnedDataInstances();
```
hydai, line 304: IIRC, the data instances should be empty after the instantiation. You may need to re-construct the data instances from the AST. But it's fine to leave a TODO here and implement it later.

lib/executor/executor.cpp
```diff
@@ -206,5 +207,190 @@ Executor::asyncInvoke(const Runtime::Instance::FunctionInstance *FuncInst,
...
+  auto Frames = StackMgr.getAllFrames();
+  std::reverse(Frames.begin(), Frames.end());
+  auto Size = Frames.size();
+
+  // TODO get thread info
+  auto &Content = CoreStack.getContent();
+  Content.push_back(0x00);
+  Content.push_back(0x04);
+
+  std::string str = "main";
```
hydai, line 342: Please follow the coding style. It should be `Str`. And use `ThreadName` is better.

lib/executor/executor.cpp
```diff
@@ -206,5 +207,190 @@ Executor::asyncInvoke(const Runtime::Instance::FunctionInstance *FuncInst,
...
+  Content.insert(Content.end(), str.begin(), str.end());
+  for (size_t i = 0; i < Size; i++) {
```
hydai, line 344: Should be `I`.

lib/executor/executor.cpp
```diff
@@ -206,5 +207,190 @@ Executor::asyncInvoke(const Runtime::Instance::FunctionInstance *FuncInst,
...
+//                     locals:vec(value) stack:vec(value)
+AST::CustomSection Executor::collectCoreStack(Runtime::StackManager &StackMgr) {
+  spdlog::info("Constructing CoreStack");
+
+  AST::CustomSection CoreStack;
+  CoreStack.setName("corestack");
+
+  auto Frames = StackMgr.getAllFrames();
+  std::reverse(Frames.begin(), Frames.end());
+  auto Size = Frames.size();
```
hydai, line 335: Use `FramesSize` instead.

lib/executor/executor.cpp
```diff
@@ -206,5 +207,190 @@ Executor::asyncInvoke(const Runtime::Instance::FunctionInstance *FuncInst,
...
+// frame       ::= 0x0 funcidx:u32 codeoffset:u32
+//                     locals:vec(value) stack:vec(value)
+AST::CustomSection Executor::collectCoreStack(Runtime::StackManager &StackMgr) {
+  spdlog::info("Constructing CoreStack");
+
+  AST::CustomSection CoreStack;
+  CoreStack.setName("corestack");
+
+  auto Frames = StackMgr.getAllFrames();
+  std::reverse(Frames.begin(), Frames.end());
```
hydai, line 334: You should never apply reversing to all frames. Use `rbegin()` and `rend()` instead.

include/runtime/stackmgr.h
```diff
@@ -100,6 +106,10 @@ class StackManager {
     return Vec;
   }
 
+  Span<Frame> getAllFrames() {
```
hydai, line 109: Since you are visiting all frames without any modification, use `Span<const Frame>` instead.

lib/executor/executor.cpp
```diff
@@ -206,5 +207,190 @@ Executor::asyncInvoke(const Runtime::Instance::FunctionInstance *FuncInst,
...
+
+    auto Funcidx = Frames[i].From->getTargetIndex(); // function index
+    // TODO which to use?
+    //  auto Funcidx = Item.From->getTargetIndex(); // function index
+    auto Codeoffset = 0;
+    // auto Codeoffset = 0; // only if codeoffset is unknown?
+    auto LocalsN = Frames[i].Locals;
+    std::cout << Frames[i].From->getOffset() << "\n";
+
+    auto Locals = StackMgr.getRangeSpan(Frames[i].VPos, Frames[i].VPos);
```
hydai, line 356: This is not correct. Please read the previous review of the stack layout.

lib/executor/executor.cpp
```diff
@@ -206,5 +207,190 @@ Executor::asyncInvoke(const Runtime::Instance::FunctionInstance *FuncInst,
...
+    auto Stack =
+        StackMgr.getTopSpan(StackMgr.size() - LocalsN); // TODO is this correct?
```
hydai, line 358: All frames are using the same stack. `getTopSpan` will not give you the correct value stack. Please follow the previous review about the stack layout and do the proper calculation.

hydai (changes requested):

include/runtime/stackmgr.h
sid-008: noted

lib/executor/engine/engine.cpp
sid-008: noted

lib/executor/executor.cpp
sid-008: I've tried with using the globals section and without it. For the examples I've run for some reason `info globals` still works(the same is true for coredump-rewriter, it does not include a globals section). Regardless I'll include the globals section to store it.

lib/executor/executor.cpp
sid-008: That is correct, the data instances are empty after instantiation. I don't think wasmgdb relies on the coredump's data section, and rather reconstructs it from the original binary(I think). Even in coredump-rewriter the data section is filled with long runs of zeroes

lib/executor/executor.cpp
sid-008: noted

sid-008:
- I have added the globals section support. One potential issue is that the segment being inserted to construct the section requires its `Expression` to be set as well and I don't know how I'd obtain that from an already instantiated global, for now I have included a dummy instruction. It seems like wasmgdb uses the global information from the source binary anyways.

- Corestack _seems_ to be working correctly in theory. But there seems to be some issue with regards to the serialisation of values and the width of its bits. Do give me feedback on that and how I can go about serializing it correctly

- The related tests have changed a lot and require some refactoring
- There are a few smaller things like adding a const etc that will be taken care of. I have marked them as TODO for now.

hydai:
> * Corestack _seems_ to be working correctly in theory. But there seems to be some issue with regards to the serialisation of values and the width of its bits. Do give me feedback on that and how I can go about serializing it correctly

You have the `ValVariant`; expect all of them to be `i64` as a temporary solution to serialize them. I mentioned this above; since the v128 (SIMD instructions) is not supported by the coredump proposal, let's make every value `i64` with a `XXX` comment to note this behavior.

> * The related tests have changed a lot and require some refactoring

I don't see any tests in your current PR.

> * There are a few smaller things like adding a const etc that will be taken care of. I have marked them as TODO for now.

If you are talking about `Span<const T>`, you should add it right now. Modifying a read-only variable is not allowed.

> I think this is the last review for the weekend. I'll be close to finished once the corestack serialization issue is taken care of and tests have been pushed.

Please note that the GSoC committee will review your PR and implementation status during the final evaluation after the mentor's final evaluation deadline. I will continue to review your PR after the Mid-Autumn Festival vacation. Please take the time to finish the remaining items or issues I marked above. Unless all of them are fixed, this PR will never be merged.

sid-008:
> Please note that the GSoC committee will review your PR and implementation status during the final evaluation after the mentor's final evaluation deadline. I will continue to review your PR after the Mid-Autumn Festival vacation. Please take the time to finish the remaining items or issues I marked above. Unless all of them are fixed, this PR will never be merged.

Yes, understood. I'll make sure to push the tests. I think tests can be done similarly to how loader/serialiser tests are, so i am working on it now.

> If you are talking about Span<const T>, you should add it right now. Modifying a read-only variable is not allowed.
Noted

> You have the ValVariant; expect all of them to be i64 as a temporary solution to serialize them. I mentioned this above; since the v128 (SIMD instructions) is not supported by the coredump proposal, let's make every value i64 with a XXX comment to note this behavior.
Noted as well.

If I were to continue working on this post-GSoC there's a few refactors I _think_ we can make to get rid of some redundant getter functions. I'll try to see if I can get around to it once I am done with the priority items such as tests and the above marked items. Thanks, have a great festival!

sid-008:
A small note on corestack:
- wasmgdb seems to parse `corestack` itself in a different way as opposed to the spec. [https://github.com/xtuc/wasm-coredump/blob/main/lib/parser/src/coredump.rs](https://github.com/xtuc/wasm-coredump/blob/main/lib/parser/src/coredump.rs)
- besides the `frame := 0x0 instanceidx:u32 funcidx:u32 codeoffset:u32 locals:vec(value) stack:vec(value)` before the locals value vector, a variable is needed to denote number of locals in the frame and the number of stack values. 
- After the  `thread-info` the number of Frames is to be encoded along with another value of which I am unsure. Currently investigating.

hydai:
> A small note on corestack:
> 
> * wasmgdb seems to parse `corestack` itself in a different way as opposed to the spec. https://github.com/xtuc/wasm-coredump/blob/main/lib/parser/src/coredump.rs we need to encode in `wasm u32`(this is according to the spec) in a few places, but for some fields `Byte` is fine
> * what I mean is; besides the `frame := 0x0 instanceidx:u32 funcidx:u32 codeoffset:u32 locals:vec(value) stack:vec(value)` before the locals value vector, a value is needed to denote number of locals in the frame and the number of stack values.
> * After the  `thread-info` the number of Frames is to be encoded along with another value of which I am unsure. Currently investigating. May be a misinterpretation on my end

I think it follows the spec; please check the wasm spec for presenting a vector; it should be a value for size followed by the vector elements.

hydai:
Since there are no more updates, this PR is replaced by #3860. I am closing it.

Issue #3523, opened 2024-07-02
OSPP Workspace: Support llm.c as a WasmEdge plugin


junaire:
### Summary

# Motivation

The WasmEdge Runtime is poised to offer robust inference support for AI models and Large
Language Models (LLMs) such as llama3 and phi-3-mini. Recognizing the critical importance
of training new models, we believe that leveraging WebAssembly (Wasm) can effectively
handle both inference and training tasks. Our objective is to build a WasmEdge plugin using
llm.c as the backend and to develop a set of WASI extensions (such as WASI-NN) for
developers.

# Deliverables
* A new WasmEdge plugin can be used for training GPT2.
* A Rust SDK of the WasmEdge plugin, allowing developers to create related
applications using Rust.
* Enhancement to llm.c, that makes it more modular and easier to integrate in
downstream projects.

# Milestones
- [x] (M1) [2 weeks] Initial PR for the WasmLLM plugin
* Create a PR, add a new WasmEdge plugin, and temporarily return an error when the host function is called.
- [x] (M2) [2 weeks] Fork llm.c, tweak the build system and code
* Tweak the build system of llm.c, make it possible to build it as a library.
* Refactor llm.c, move the training logic into a separate function, and export it to downstream users. Also, make necessary parameters configurable.
- [x] (M3) [1 week] Integrate llm.c into WasmEdge
* Integrate llm.c and all its dependencies into WasmEdge build systems using cmake.
- [x] (M4) [2 weeks] Implement the host functions
* Implement all host functions we defined, and write tests.
- [x] (M5) [2 weeks] Develop the Rust binding
* Develop the Rust binding, write tests, add documentation.
- [x] (M6) [2 week] Support Training GPT2 on GPU

# References
* https://summer-ospp.ac.cn/org/prodetail/248830516
* #3368 

CC @hydai

junaire:
I have created a llm.c fork in https://github.com/junaire/llm.c/tree/jun/wasmedge to support our plugin, maybe you can take a look so we can discuss the design @hydai

hydai:
I create wasmedge/llm.c via forking the upstream. You can submit your PR there: https://github.com/WasmEdge/llm.c

junaire:
> I create wasmedge/llm.c via forking the upstream. You can submit your PR there: https://github.com/WasmEdge/llm.c

[PR](https://github.com/WasmEdge/llm.c/pull/1) submitted, I feel like there are a lot of things that we should discuss.

[...]

PR #3536, opened 2024-07-06
[WASI-LLM] Implement wasi_llm module

junaire:
Related #3523

### DIFF
--- a/plugins/wasi_llm/wasillmfunc.cpp
+++ b/plugins/wasi_llm/wasillmfunc.cpp
@@ -1,0 +1,81 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2019-2024 Second State INC
+
+#include "wasillmfunc.h"
+#include "common/spdlog.h"
+
+#include <string>
+#include <string_view>
+
+namespace WasmEdge {
+namespace Host {
+
+Expect<WASILLM::ErrNo>
+WasiLLMModelCreate::bodyImpl(const Runtime::CallingFrame &Frame,
+                             uint32_t CheckPointPath,
+                             uint32_t CheckPointPathLen) {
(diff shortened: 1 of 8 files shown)


junaire:
Could you plz take a look at this? @hydai

plugins/wasi_llm/wasillmfunc.cpp
```diff
@@ -0,0 +1,81 @@
...
+#include <string_view>
+
+namespace WasmEdge {
+namespace Host {
+
+Expect<WASILLM::ErrNo>
+WasiLLMModelCreate::bodyImpl(const Runtime::CallingFrame &Frame,
+                             uint32_t CheckPointPath,
+                             uint32_t CheckPointPathLen) {
+  (void)Frame;
```
hydai, line 17: Use `maybe_unused` instead.
https://en.cppreference.com/w/cpp/language/attributes/maybe_unused

hydai (changes requested):

plugins/wasi_llm/wasillmfunc.cpp
junaire: It's inappropriate to use `[[maybe_unused]]` attribute here, since what I really want to do is simply suppress the compiler warnings temporarily, they are all **used**. Considering we're gonna remove these very soon, using the old C style `(void)x` is good enough. What's more, using `[[maybe_unused]]` to suppress the warning needs to modify the declaration of the function, which makes the diff harder to read when we change it later.
```c++
WasiLLMModelCreate::bodyImpl([maybe_unused]] const Runtime::CallingFrame &Frame,
                             [maybe_unused]] uint32_t CheckPointPath,
                             [maybe_unused]] uint32_t CheckPointPathLen) {
```

hydai (approved):

PR #3548, opened 2024-07-08
[ci] Fix CI error due to checkout action upgrade

junaire:
checkout action recently drop nodejs16 support, which cause CI failure in some runners running ancient distro, like CentOS7


0yi0:
Thanks for your contribution. But a slightly earlier PR #3548 was fixing the same issue, so I'm closing this as a dup.
#3548 didn't exhaustively check the workflow files. Feel free to open another PR/issue if you find other files having the same issue.

PR #3549, opened 2024-07-09
[Installer] Deprecate manylinux2014 with manylinux_2_28

0yi0:
This PR 
- `sed -i 's/manylinux2014/manylinux_2_28/g'` on the files:
```
.github/workflows/test-install-script.yml
utils/install.py
utils/install.sh.old
utils/install_v2.sh
```
- CentOS 7 -> AlmaLinux 8


SAtacker:
Looks like the CI could use a newer version of node.js

hydai:
@0yi0 
We cannot merge this until the manylinux2014 are all replaced in 0.13.5 and 0.14.0. Since we are still maintaining the 0.13.5, the installation will fail if we merge this PR.

0yi0:
Do we need to update installer with the newer release(s) or after the newer release(s)?

Just found that we're testing on 0.13.4 or older versions. Are we not testing latest version on purpose?

hydai:
I think we could add more matrices, such as `manylinux_2_28` + `0.14.1` or something similar.
So the installer can support manylinux2014 for 0.13.5 and 0.14.0 and manylinux2_28 for 0.14.1. WDYT?

0yi0:
Since this PR might take some time before merging, I moved CI nodejs version to a separate PR #3550

0yi0:
We still don't have manylinux_2_28 x86_64 for 0.14.1 beta, so I'll convert this into an issue.

PR #3578, opened 2024-07-19
[Runtime] Create limited page size of memory instance if configured.

q82419:
If the page size limit is set in configuration, and the `min` in memory type is smaller than the limited page size value when creating the memory instance, modify the `min` in memory type and create with the limited size.


juntao:

hydai (approved):

PR #3579, opened 2024-07-19
changed .dylib to .framework

836hardik-agrawal:
According to what i understand. I have tried to solve the problem as directed by @hydai by looking the cmake files.

### DIFF
--- a/cmake/WASINNDeps.cmake
+++ b/cmake/WASINNDeps.cmake
@@ -85,8 +85,8 @@ function(wasmedge_setup_tflite_lib)
     )
   elseif(APPLE)
     set(WASMEDGE_TENSORFLOW_DEPS_TFLITE_LIB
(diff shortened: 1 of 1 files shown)


juntao:

lazyperson1020:
@836hardik-agrawal I think  we need to change the code in such a manner that the framework is built in the /System/Library/Frameworks for the macos.

836hardik-agrawal:
@lazyperson1020 .I checked the cmake files,but not able to get what it meant by changing code in manner framework is built in 
System/Library/Frameworks for the macos.

hydai:
I don't think this is the correct way to implement this feature.

1. We want to provide an option for users that they can choose `dylib` or the new `framework` format.
2. This only replaces the `dylib` with `framework`, which is nonsense.
3. We want to provide `.framework` for all `dylib`, not the `wasi-nn` related ones.

836hardik-agrawal:
> Hi @836hardik-agrawal Are you going to implement this feature?

Sorry not able to understand and not able to do it .

Issue #3483, opened 2024-06-17
LFX Workspace: Support piper as a new backend of the WASI-NN WasmEdge plugin


PeterD1524:
### Summary

# Motivation

WasmEdge supports PyTorch, TensorFlow Lite, llama.cpp, and more NN backend. Dealing with the text-to-voice is a big thing that we want to achieve. To make it possible, we would like to integrate [piper](https://github.com/rhasspy/piper), A fast, local neural text-to-speech system in C++ as a new [WASI-NN](https://github.com/second-state/wasmedge-wasi-nn) backend.

# Details

Expected Outcome: A new plugin provides a piper [WASI-NN](https://github.com/second-state/wasmedge-wasi-nn) backend, a test suite for validating the plugin, documents, and examples for explaining how to use the plugin.

# Milestones

- [ ] (M1) [1 weeks] Piper plugin, Plans, Research
  - Improve draft work at https://github.com/PeterD1524/WasmEdge/tree/wasi_nn_piper
  - Create pull request for [wasmedge-wasi-nn](https://github.com/second-state/wasmedge-wasi-nn) to support the new Piper backend
  - Get familiar with WasmEdge development workflow
- [ ] (M2) [1 weeks] test suite, documents, and examples

### Appendix

- https://github.com/WasmEdge/WasmEdge/issues/3381
- https://github.com/PeterD1524/WasmEdge/tree/wasi_nn_piper

PeterD1524:
PR (WIP): https://github.com/WasmEdge/WasmEdge/pull/3499

PeterD1524:
PR for [WASI-NN Rust SDK](https://github.com/second-state/wasmedge-wasi-nn/tree/ggml/rust): https://github.com/second-state/wasmedge-wasi-nn/pull/15

hydai:
Thanks @PeterD1524 
@dm4 is going to review the wasmedge-wasi-nn PR.

Is #3499 ready to review, or is it still in the WIP stage?

PeterD1524:
#3499 is it still in the WIP stage.

PeterD1524:
Week 1 Summary

- Created main PR
  - https://github.com/WasmEdge/WasmEdge/pull/3499
  - added backend structure
- Created PR for wasmedge-wasi-nn to support the new Piper backend
  - https://github.com/second-state/wasmedge-wasi-nn/pull/15

PeterD1524:
Week 2 Summary

- Main PR
[...]

PR #3583, opened 2024-07-22
[WASI-NN] piper: fix piper dependencies

PeterD1524:
Hi, I am working on [LFX Workspace: Support piper as a new backend of the WASI-NN WasmEdge plugin #3483](https://www.github.com/WasmEdge/WasmEdge/issues/3483)

The dependencies of WASI-NN Piper plugin:

Currently the dependency piper unconditionally creates (or downloads) shared libraries of its dependencies:

libwasmedgePluginWasiNN.so -> libpiper.a -> libpiper_phonemize.so -> (libespeak-ng.so, libonnxruntime.so)

As a result, libwasmedgePluginWasiNN.so cannot work on its own.
This PR attempts to solve these problems.

The final result will be:

libwasmedgePluginWasiNN.so -> libpiper.a -> libpiper_phonemize.a -> (libespeak-ng.a, system libonnxruntime.so)

The plugin will static link to most of its dependencies and it will dynamic link to a system provided libonnxruntime.so.
Users will have to install onnxruntime themselves.


juntao:

.github/workflows/build-extensions.yml
```diff
@@ -116,6 +116,7 @@ jobs:
           bash utils/wasi-nn/install-openvino.sh
           bash utils/wasi-nn/install-pytorch.sh
           bash utils/wasi-nn/install-neuralspeed.sh
+          bash utils/wasi-nn/install-onnxruntime-ubuntu.sh
```
0yi0, line 119: I think we don't need `-ubuntu` suffix here? @hydai how do you think?

hydai: We should remove the suffix because this should also be used in the manylinux(CentOS) distributions.

hydai (approved):
LGTM

PR #3560, opened 2024-07-13
added linelint workflow

arujjval:
closes #3519 

changes to PR #3559  as per requested

PR #3559, opened 2024-07-13
Added .github/workflows/lint.yml

arujjval:
closes #3519

PR #3558, opened 2024-07-12
Added .github/workflows/lint.yml

arujjval:
closes #3519

PR #3592, opened 2024-07-24
added linelint workflow

### LINKED ISSUE #3519: Create a workflow detect EOF and fix it for PR automatically
### Summary

Seems linelint[1] can help us to do this.

### Appendix

1. https://github.com/marketplace/actions/linelint

arujjval:
closes #3519 

#### Notes
Previous discussions about this PR were in #3558 #3559 #3560


juntao:

dannypsnl:
In the log we can see,
```
[EOF Rule] File ".clang-tidy" has lint errors
[EOF Rule] File ".clang-tidy" lint errors fixed
```
However, there are no changes are proposed to GitHub, you need to fill that gap.

arujjval:
@dannypsnl This is fixing eof errors in our repo as we have made a PR. This is what was requested in the issue #3592.

hydai:
@dannypsnl 
I changed my mind. Let's use this linelint workflow only for detection. I don't think using autofix is a good idea now if a malicious hacker hijacks the linelint workflow, and it will allow the workflow to inject or modify some code into our codebase directly.

@0yi0 WDYT?

.github/workflows/linelint.yml
```diff
@@ -0,0 +1,13 @@
...
+
+jobs:
+  linelint:
+    runs-on: ubuntu-latest
+    name: Check if all files end in newline
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Linelint
+        uses: fernandrone/linelint@master
```
0yi0, line 12: Already mentioned in the old PRs, please use a tag instead of master.

hydai:
We should also add the following limitation to the workflow file to keep it safe:

```
permissions:
  contents: read
```

.github/workflows/linelint.yml
```diff
@@ -0,0 +1,13 @@
+on: pull_request
+name: linelint
+
+jobs:
+  linelint:
+    runs-on: ubuntu-latest
+    name: Check if all files end in newline
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
```
0yi0, line 10: `actions/checkout@v2` is a very legacy version, please use `v4` or add a reason for it.

0yi0 (changes requested):

0yi0:
@hydai No autofix sounds good to me. But if we're not running autofix, should CI report failure on lint errors?

hydai:
Definitely, we should let CI fail if the changed files do not contain EOF. Just like what we did in the misc linter.

0yi0:
@arujjval dannypsnl means your workflows only fix it in the workflows **without** any modifications to the code base (the PR or the reporitory). So `autofix` being `true` or `false` didn't matter in the PR because no changes were committed.

arujjval:
So I have to remove autofix, and only check incoming changes for EOF error?

0yi0:
@arujjval I'm wondering if we can keep `autofix: true` in `.linelint`, and use `autofix: false` in `linelint.yml`, so that when `linelint.yml` reports error, user can use your `.linelint` config directly to fix their issues.

If this works, `.linelint` is already good and need no modifications. You only need to update `linelint.yml`, and then we can merge.

I hope I explained it well. Please let met know if you have questions about this.

arujjval:
I am not able to make this lint run only on changed modified files. Instead of running a command, I am "using" the repo of linelint. One thing we can do is install linelint in the repo and then running script in workflow. Can we that or is there any other way round?

dannypsnl:
> I am not able to make this lint run only on changed modified files.

I believe you can add
```sh
git diff <branch> --name-status
```
for example, this PR merge 2 commits into master, so
```sh
git diff master --name-status
```
let sed out the files.

arujjval:
I am able to get modified files and I am passing it while running linelint, but it is still running on all files. What can be the potential issues?
Screenshot from 2024-07-30 12-14-47

0yi0:
I checked the source code of this action, it doesn't seem to provide input fields.

It runs `linelint` like this
```
ENTRYPOINT ["/linelint", "."]
```

It's probably the same as the warning at the first line in your screenshot.

> Unexpected input(s) 'files', valid inputs are ['entryPoint', 'args']

I'm not sure if that means you can change `['entryPoint', 'args']` from input though.
If yes, you can try to set `args` to the list of files.

arujjval:
@0yi0 turns out we can change `args` from input. And it seems to work. But the issue is it is not ignoring other files. Here's the result after running with args and without args

Screenshot from 2024-07-30 16-26-12
with args

Screenshot from 2024-07-30 16-25-41
without args

`args` seems to add files for lint check but it does not exclude other files. Or these files might be special case. Maybe we can correct lint error of these files (they are just 4) and then try.

hydai:
Hi @arujjval 
I'm sorry I missed this PR. Could you remind me of its current state? Is there any help you need to continue working on it?

hydai:
Closing due to inactive. Thanks.

Issue #3039, opened 2023-11-18
fd_sync on base dir fails


yagehu:
### Summary

`fd_sync` on a preopened base directory returns 76 (notcapable). On some other runtimes (WAMR, Node), it is 0 (ok).

I wonder if it's possible to customize the capability associated with the base directory. However, I believe the capability system is being scrapped completely.

### Reproduction steps

```wat
(module
  (type (func))
  (type (func (param i32) (result i32)))
  (type (func (param i32)))
  (import "wasi_snapshot_preview1" "fd_sync" (func $fd_sync (type 1)))
  (import "wasi_snapshot_preview1" "proc_exit" (func $proc_exit (type 2)))
  (func $_start
    (call $fd_sync
      (i32.const 3)
    )
    call $proc_exit
  )
  (memory 2)
  (export "memory" (memory 0))
  (export "_start" (func $_start))
)
```

1. Compile this WAT to WASM: `wat2wasm fd_sync.wat > fd_sync.wasm`
2. Run the binary with wasmedge CLI, mounting a directory: `wasmedge --dir=. fd_sync.wasm`
3. Inspect the exit code: `echo $?` which is 76.

### Components

CLI

### WasmEdge Version or Commit you used

0.13.5

### Operating system information

Fedora 39

### Hardware Architecture

amd64

hydai:
Could you check the spec of WASI to identify the expected behavior?
If the spec doesn't have a definition for this, then we will not treat this as a bug.

yagehu:
The [spec](https://github.com/WebAssembly/WASI/blob/main/legacy/preview1/docs.md#-fd_syncfd-fd---result-errno) says `fd_sync` "is similar to `fsync` in POSIX." The [POSIX fsync spec](https://pubs.opengroup.org/onlinepubs/9699919799/functions/fsync.html) permits fsyncing a directory. Coupled with the fact that capability/rights is [being scrapped](https://github.com/WebAssembly/wasi-filesystem/issues/31), I think it's reasonable to treat this as a bug.

freexploit:
Is this related to this error?:
```
The args are as follows.
/wasi_example_main.wasm
50000000
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 76, kind: Uncategorized, message: "Capabilities insufficient" }', src/main.rs:47:39
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
[...]

PR #3613, opened 2024-08-01
Fix fd_sync Error for Preopened Base Directory

MeenuyD:
Fix `fdSync` to Handle Directories in Linux Implementation.
I have updated the `fdSync` function in `inode-linux.cpp` to correctly handle synchronization requests for directories. The function now checks if the file descriptor is a directory and treats it as syncable, in addition to performing the `fsync` operation for regular files. This change ensures compatibility with other runtimes and aligns the behavior with the WASI specification.
#3039

### DIFF
--- a/lib/host/wasi/inode-linux.cpp
+++ b/lib/host/wasi/inode-linux.cpp
@@ -549,11 +549,15 @@ WasiExpect<void> INode::fdSeek(__wasi_filedelta_t Offset,
 }
 
 WasiExpect<void> INode::fdSync() const noexcept {
-  if (auto Res = ::fsync(Fd); unlikely(Res != 0)) {
-    return WasiUnexpect(fromErrNo(errno));
-  }
+    struct stat fdStat;
+    if (fstat(Fd, &fdStat) == 0 && S_ISDIR(fdStat.st_mode)) {
+        return {};
+    }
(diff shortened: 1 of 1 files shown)


juntao:

hydai (changes requested):
The coding style is not compatible with our standard, please fix it.

MeenuyD:
> The coding style is not compatible with our standard, please fix it.

How should I fix this to compatible with standard

MeenuyD:
Could you please help me in this

lib/host/wasi/inode-linux.cpp
```diff
@@ -549,6 +549,10 @@ WasiExpect<void> INode::fdSeek(__wasi_filedelta_t Offset,
 }
 
 WasiExpect<void> INode::fdSync() const noexcept {
+  struct stat fdStat;
+  if (fstat(Fd, &fdStat) == 0 && S_ISDIR(fdStat.st_mode)) {
+    return {};
+  }
```
dannypsnl, line 555: This logic is: if we get a directory, we ignore it. I don't think this is expected behaviour.

dannypsnl (changes requested):

hydai:
Closing due to inactive.

PR #3632, opened 2024-08-07
[CI] Fix mirrors on manylinux2014(CentOS 7)

hydai:
* For the installer tests, we replace the mirrors.
* For manylinux2014, we use the pre-installed gh client inside the
  docker image instead.
* For non-EOL manylinux, it's fine to use the package manager to install
  the gh client.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -73,8 +73,8 @@ jobs:
     with:
       version: ${{ needs.create_release.outputs.version }}
       matrix:
-        "[{'name':'manylinux2014 x86_64','runner':'ubuntu-latest','docker_tag':'manylinux2014_x86_64','asset_tag':'manylinux2014_x86_64'},
-        {'name':'manylinux2014 aarch64','runner':'linux-arm64-v2','docker_tag':'manylinux2014_aarch64','asset_tag':'manylinux2014_aarch64'}]"
+        "[{'name':'manylinux2014 x86_64','runner':'ubuntu-latest','docker_tag':'manylinux2014_x86_64-plugins-deps','asset_tag':'manylinux2014_x86_64'},
+        {'name':'manylinux2014 aarch64','runner':'linux-arm64-v2','docker_tag':'manylinux2014_aarch64-plugins-deps','asset_tag':'manylinux2014_aarch64'}]"
       release: true
     secrets: inherit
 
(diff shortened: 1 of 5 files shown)


juntao:

hydai:
I tried to rebuild the docker images for manylinux2014, but I had no luck.
The log shows some of the required packages are not available.

```
20.96 Loaded plugins: ovl
21.01 Could not retrieve mirrorlist http://mirrorlist.centos.org?arch=x86_64&release=7&repo=sclo-sclo error was
21.01 14: curl#6 - "Could not resolve host: mirrorlist.centos.org; Unknown error"
21.01
21.01
21.01  One of the configured repositories failed (Unknown),
21.01  and yum doesn't have enough cached data to continue. At this point the only
21.01  safe thing yum can do is fail. There are a few ways to work "fix" this:
21.01
21.01      1. Contact the upstream for the repository and get them to fix the problem.
21.01
21.01      2. Reconfigure the baseurl/etc. for the repository, to point to a working
21.01         upstream. This is most often useful if you are using a newer
21.01         distribution release than is supported by the repository (and the
21.01         packages for the previous distribution release still work).
21.01
21.01      3. Run the command with the repository temporarily disabled
21.01             yum --disablerepo=<repoid> ...
21.01
21.01      4. Disable the repository permanently, so yum won't use it by default. Yum
21.01         will then just ignore the repository until you permanently enable it
21.01         again or use --enablerepo for temporary usage:
21.01
21.01             yum-config-manager --disable <repoid>
21.01         or
21.01             subscription-manager repos --disable=<repoid>
21.01
21.01      5. Configure the failing repository to be skipped, if it is unavailable.
21.01         Note that yum will try to contact the repo. when it runs most commands,
21.01         so will have to try and fail each time (and thus. yum will be be much
21.01         slower). If it is a very temporary problem though, this is often a nice
21.01         compromise:
21.01
21.01             yum-config-manager --save --setopt=<repoid>.skip_if_unavailable=true
21.01
21.01 Cannot find a valid baseurl for repo: centos-sclo-sclo/x86_64
```

hydai:
If we want to drop the manylinux2014 directly, it will be easier.
However, it will be more challenging if we decide to salvage these assets.

IMO, I prefer to delete all of the manylinux2014 from 0.14.1, update the installer to ensure that nobody can access the manylinux2014 from 0.14.1, and use manylinux2_28 instead.

q82419 (approved):

PR #3626, opened 2024-08-05
[CMake] Fix version checking in source tarball

Biswa96:
In source tarball, .git directory does not exist. So, git describe
    command fails with the following error.
    
    fatal: not a git repository (or any of the parent directories): .git
    
    This commit checks if the current source directory is a git repository.
    Otherwise, cmake does not run the git describe command.

PR #3635, opened 2024-08-08
[CMake] Fix the .git Detection Issue

hydai:
In #3626, .git detection was introduced to avoid the error triggered by the git describe command. However, the GIT_VERSION_NOT_FOUND variable is defined and set within the git describe command call. After applying #3626, GIT_VERSION_NOT_FOUND will never be defined and set if the git command exists but the .git folder does not.

In this PR, we define and initialize GIT_VERSION_NOT_FOUND to 1. This will ensure that CPACK_PACKAGE_VERSION is set to "0.0.0-unreleased" if `git describe` is not working or the .git folder does not exist.

@q82419 
We need to merge this PR into rc.2.


juntao:

0yi0 (approved):

Issue #2761, opened 2023-08-28
The result is different from other wasm runtimes while executing the given wasm file


luxinyi0105:
## Description

While executing [the given testcase](https://github.com/luxinyi0105/wasm_runtime_test/raw/main/08-28.zip) with WasmEdge and some other runtime tools, the results are different.

### Current State

```sh
$ wasmedge mutated_file.wasm
checksum = 61D81D2D
```
It outputs the checksum result.

### Expected

I tried to use some other wasm runtimes, such as wasmer, wasmtime, and wasmi, to execute the same testcase. All runtime tools mentioned will trigger runtime exceptions, which are caused by “out of bounds memory access”.

Here are the running results of different runtime tools.
![runtime result](https://github.com/WasmEdge/WasmEdge/assets/92167273/e6496efd-a6a2-4f27-b33f-dee5a3e4214d)

Based on the above situation, we infer that there may be incorrect exception handling in WasmEdge, resulting in execution results that are different from other runtime tools.

## Environment
- Hardware Architecture: x86_64
- Operating system: Ubuntu 22.04
- Tools Version: wasmedge 0.13.3

## Extra info

The given testcase was simply mutated a wasm file, which was obtained by compiling C program generated with `Csmith` using `Emscripten Compiler`(Emcc). 

The original C program is `c_file.c`, the compilation results with Emscripten is `wasm_file.wasm`, and its wat format is `wat_file.wat`.

We mutated the wat file to change its `local.set 1271` in line 5694 with `i32.load`. The result after mutation is `mutated_file.wat`, and its wasm format is `mutated_file.wasm`.

sparker-arm:
I was just browsing the code and wondered if anyone had any missed OOB issues... It doesn't look like the AOT compiler inserts any bounds checks, an example is [here](https://github.com/WasmEdge/WasmEdge/blob/647b19e22d991ef1e526f684d5f92d48564286fb/lib/aot/compiler.cpp#L3849). I don't think creating an inbounds GEP is necessarily legal or sufficient. If there's nothing else in the runtime to catch these issues then wasmedge is severely broken!

hydai:
Thanks for this issue.

It seems like the AOT has this issue. The interpreter mode will get OOB. However, the AOT will pass and print the checksum.


![CleanShot 2023-09-08 at 01 08 29](https://github.com/WasmEdge/WasmEdge/assets/2776756/ce388d5d-b115-4b96-af3e-82f88b5907cf)

[...]

PR #3638, opened 2024-08-09
Fix: Correction of out of bounds error

Daksh-10:
**Do not merge**

According to #2761,
We do not encounter error while in WasmEdge, I tried the below code, which will check if the loaded memory goes out of total memory. 

I need to write the test cases according to the changes done. 

I'll remove `spdlog::error("YOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOO")` afterwards. 

I was thinking should I only check for i32, i64 ,f32, f64 load commands because it will be easier to do so. Otherwise it will take a lot of time to write new tests.

### DIFF
--- a/lib/validator/formchecker.cpp
+++ b/lib/validator/formchecker.cpp
@@ -20,6 +20,7 @@ namespace {
 auto logOutOfRange(ErrCode Code, ErrInfo::IndexCategory Cate, uint32_t Idx,
                    uint32_t Bound) {
   spdlog::error(Code);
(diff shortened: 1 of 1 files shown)


juntao:

dannypsnl:
Just remind, not ready PR can use draft PR.

Daksh-10:
There is nothing wrong with the code, Instr.getTargetIndex() does provide the address of instruction, so we don't need to change it.

PR #3641, opened 2024-08-12
Add a new wasi_nn backend related to burn.rs

CaptainVincent:
As our show cases, currently two models are supported: 

SqueezeNet and Whisper.

But the plugin currently supports only one model at a time.

ex.

```bash
cmake -GNinja -Bbuild -DCMAKE_BUILD_TYPE=Release -DWASMEDGE_BUILD_AOT_RUNTIME=OFF -DWASMEDGE_PLUGIN_WASI_NN_RUST_MODEL=whisper

or 

cmake -GNinja -Bbuild -DCMAKE_BUILD_TYPE=Release -DWASMEDGE_BUILD_AOT_RUNTIME=OFF -DWASMEDGE_PLUGIN_WASI_NN_RUST_MODEL=squeezenet
```


juntao:

CaptainVincent:
The missing wasi-nn headers in the wasiNNTest for the burn plugin require this part to be rewritten, which will take some time. Considering the priority, I keep only the release CI for now.

plugins/wasi_nn_rust/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
+if(CMAKE_BUILD_TYPE STREQUAL "Debug")
```
q82419, line 1: Add the liscense.

plugins/wasi_nn_rust/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
...
+message(STATUS "WasmEdge Wasi-NN Rust plugin model: ${WASMEDGE_PLUGIN_WASI_NN_RUST_MODEL}")
+set(CARGO_FEATURES "--features=${WASMEDGE_PLUGIN_WASI_NN_RUST_MODEL}")
+
+set(RS_SO ${CMAKE_CURRENT_BINARY_DIR}/${TARGET_DIR}/libwasmedgePluginWasiNN${CMAKE_SHARED_LIBRARY_SUFFIX})
+
+set(WASMEDGE_LIB_DIR ${CMAKE_CURRENT_BINARY_DIR}/../../lib/api)
+
+add_custom_target(wasi_nn_rust ALL
+  COMMAND WASMEDGE_LIB_DIR=${WASMEDGE_LIB_DIR} LD_LIBARAY_PATH=${WASMEDGE_LIB_DIR} CARGO_TARGET_DIR=${CMAKE_CURRENT_BINARY_DIR} ${CARGO_CMD} ${CARGO_FEATURES}
+  COMMAND cp ${RS_SO} ${CMAKE_CURRENT_BINARY_DIR}
```
q82419, line 18: Use `${CMAKE_COMMAND} -E copy` here.

plugins/wasi_nn_rust/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
...
+  COMMAND rm -rf ${CMAKE_CURRENT_BINARY_DIR}/${TARGET_DIR}
```
q82419, line 19: Use `${CMAKE_COMMAND} -E remove` here.

plugins/wasi_nn_rust/Cargo.toml
```diff
@@ -0,0 +1,38 @@
+[package]
+name = "wasi_nn_rust"
+version = "0.0.1"
+edition = "2021"
```
q82419, line 4: Should we use `2023` here?

plugins/wasi_nn_rust/src/lib.rs
```diff
@@ -0,0 +1,382 @@
+mod helper;
+mod models;
+
+pub enum ErrNo {
```
dm4, line 4: Is it possible to re-use [error.rs](https://github.com/second-state/wasmedge-wasi-nn/blob/ggml/rust/src/error.rs) from wasmedge-wasi-nn?

plugins/wasi_nn_rust/CMakeLists.txt
CaptainVincent: Done.

plugins/wasi_nn_rust/CMakeLists.txt
CaptainVincent: Done.

plugins/wasi_nn_rust/CMakeLists.txt
CaptainVincent: Done.

plugins/wasi_nn_rust/Cargo.toml
CaptainVincent: It seems that the 2023 edition is not available. The latest edition of Rust that can be used is 2024, but it is currently still in [beta](https://doc.rust-lang.org/edition-guide/rust-2024/index.html). Here, the previous usable version, 2021, is kept to resolve this issue.

plugins/wasi_nn_rust/src/lib.rs
CaptainVincent: I tried it out, but I found that there is no explicit method to convert it to i32. Therefore, I still need to manually write the conversion function in the plugin layer (the code on both sides will still decoupled in the future). And it also lacks a Success enum. Given these reasons, it is less concise than declaring a new enum directly, so I prefer to keep the existing approach.

dm4: Sounds reasonable.

q82419 (approved):

PR #3660, opened 2024-08-15
[CI] Build CUDA related plugins

dm4:


juntao:

0yi0 (commented):
I don't see why need array and for loops here?

dm4:
> I don't see why need array and for loops here?

I was thinking that for convenience for other plugins needing CUDA in the future, it might be easier to modify. Or do you think keeping it simple here is better?

dm4:
@hydai Do you think other things will soon use CUDA as well? If not, I'll just implement it simply here.

hydai:
Due to the frequent updates of the llama.cpp. I think it should be fine to have stable diffusion here. WDYT?

0yi0 (approved):

PR #3599, opened 2024-07-25
[CI] move wasi-nn neuralspeed backend to the last target to avoid failing others

hydai:
Because the WASI-NN Neural Speed CI has a random segfault, it will block other backends. As a workaround, make it the last target.
This will be resolved after #3596 is fixed.

PR #3677, opened 2024-08-21
WIP: [CI] Refactor ubuntu plugins

0yi0:
This aims to
- Reduce copy-pasta in `release.yml` and `build-extensions.yml`.
- Fix the issue happened in #3599, that once a job fails when using for-loop instead of matrix in workflow jobs will cause all subsequent jobs unable to run.


juntao:

0yi0:
Reopen accidentally closed PR ;_;

hydai:
> @hydai PTAL
> 
> I want to make sure
> 
> 1. Do we need to test all plugins on every combinations `ubuntu-<latest|20.04>-<clang|gcc>-x86_64`? Currently we have
>    
>    * `ubuntu-20.04-clang-x86_64` (release only)
>    * `ubuntu-22.04-clang-x86_64` (build and test)
>    * `ubuntu-22.04-gcc-x86_64` (build and test)

I am wondering if we should have every combination because the release process needs `20.04` and the testing process needs `the latest(22.04)`. However, the toolchain should be reduced, let's we should provide only one toolchain for the release, but all toolchains for testing to verify if anything related to the compilers will fail our building process. 

> 2. Is this PR able to merge with only partial refactor as it is now (9 plugins to be exact)?

Should we wait for all plugins to be refactored? Since we are currently in the 0.14.1 release process, I expect this PR to be merged after 0.14.1.

0yi0:
> > 2. Is this PR able to merge with only partial refactor as it is now (9 plugins to be exact)?
> 
> Should we wait for all plugins to be refactored? Since we are currently in the 0.14.1 release process, I expect this PR to be merged after 0.14.1.

I expect this to be merged after `0.14.1`, too.

I'm not sure about refactoring all plugins in one PR because the number of plugins are still grossing during this PR working in progress. If we can merge a partial refactor, we can ask new plugins to stop adding to the old workflows and at the same time still allowing adding new plugins; otherwise we might need to freeze the number of plugins for a short time (like after `0.14.1` to the next `rc`).

hydai:
Works for me. Let's do the partial refactoring.

PR #3688, opened 2024-08-23
[WASI-NN] Support MLX as WASI-NN backend

grorge123:


juntao:

plugins/wasi_nn/MLX/mlx/base.cpp
```diff
@@ -0,0 +1,64 @@
...
+      delete OldModule;
+    }
+  }
+  return this;
+}
+void Module::apply(std::string Key, mx::array Value) {
+  std::vector<std::string> SplitKey = splitString(Key, '.');
+  if (SplitKey.size() == 1) {
+    if (Parameters.find(Key) == Parameters.end()) {
+      spdlog::error("Unsupported weight: {}", Key);
```
hydai, line 33: Please ensure there is a prefix `[WASI-NN] MLX: ...` and use `sv`.

plugins/wasi_nn/MLX/mlx/base.h
```diff
@@ -0,0 +1,63 @@
...
+class Module {
+public:
+  virtual ~Module() {
+    for (auto Module : Submodules) {
+      delete Module.second;
+    }
+  }
+  std::string Name;
+  std::unordered_map<std::string, mx::array> Parameters{};
+  std::unordered_map<std::string, Module *> Submodules{};
```
hydai, line 23: Could we use a smart pointer here instead of the raw pointer?

plugins/wasi_nn/MLX/mlx/transformer.cpp
```diff
@@ -0,0 +1,14 @@
+#include "transformer.h"
+#include <mlx/ops.h>
+
+namespace WasmEdge::Host::WASINN::MLX {
+namespace mlx::core::nn {
+mx::array MultiHeadAttention::createAdditiveCausalMask(int N, mx::Dtype DType) {
+  auto Indices = mx::arange(N);
+  // mask = indices[:, None] < indices[None]
+  auto Mask = reshape(Indices, {N, 1}) < reshape(Indices, {1, N});
+  Mask = astype(Mask, DType) * -1e9;
```
hydai, line 10: What does `-1e9` mean?

hydai (changes requested):

plugins/wasi_nn/MLX/mlx/transformer.cpp
grorge123: It is usually set INF for the mask tensor. MLX implementation uses -1e9 instead since it is as good as INF and softmax(full(1e9)) != nan.

hydai: Let's have some comments for this.

plugins/wasi_nn/mlx.cpp
```diff
@@ -0,0 +1,344 @@
...
+    auto Weight = Builders[Idx];
+    const std::string BinModel(reinterpret_cast<char *>(Weight.data()),
+                               Weight.size());
+    spdlog::info("[WASI-NN] MLX BinModel: {}"sv, BinModel.size());
+    if (BinModel.size() == 0) {
+      Env.NNGraph.pop_back();
+      return ErrNo::InvalidArgument;
+    }
+    std::string ModelFilePath;
+    if (BinModel.substr(0, 8) == "preload:"sv) {
```
hydai, line 185: If the model does not use `nn-preload,` we will no longer accept it. Most LLM models are larger than the Wasm memory limitation (4GB), and it's not possible to pass these kinds of models via arguments. We can show the error message to ask users to pass the model through the `nn-preload` here.

hydai (changes requested):

plugins/wasi_nn/mlx.cpp
hydai: Never mind, let's ignore this.

hydai (approved):

PR #3691, opened 2024-08-24
Added stdin, stdout, stderr to Environ::init

### LINKED ISSUE #2936: feat: allow redirect of stdin, stdout, and stderr in C SDK
### Summary

For some use cases, developers may want to separate the STDIN, STDOUT, and STDERR from the default FD(0, 1, 2).
To make developers happy, we would like to have a set of APIs to redirect STDIO to several standalone FDs.

### Details

This should extend the WASI-related API like `init` with three extended parameters, which indicate the FD of STDIN, STDOUT, and STDERR.

For reference, we may take some ideas from nodejs: https://github.com/nodejs/node/pull/33544/files

### Appendix

_No response_

Wck-iipi:
1. Expands Environ::init() to take 3 more additional parameters StdinFd, StdoutFd, StderrFd.

Fixes #2936

### DIFF
--- a/include/host/wasi/environ.h
+++ b/include/host/wasi/environ.h
@@ -56,7 +56,8 @@ public:
   ~Environ() noexcept;
 
   void init(Span<const std::string> Dirs, std::string ProgramName,
-            Span<const std::string> Args, Span<const std::string> Envs);
+            Span<const std::string> Args, Span<const std::string> Envs,
+            int StdinFd = 0, int StdoutFd = 1, int StderrFd = 2);
 
   void fini() noexcept;
 
--- a/lib/host/wasi/environ.cpp
+++ b/lib/host/wasi/environ.cpp
@@ -71,7 +71,8 @@ static inline constexpr const auto kReadOnly = "readonly"sv;
 } // namespace
 
 void Environ::init(Span<const std::string> Dirs, std::string ProgramName,
-                   Span<const std::string> Args, Span<const std::string> Envs) {
+                   Span<const std::string> Args, Span<const std::string> Envs,
+                   int StdinFd, int StdoutFd, int StderrFd) {
   {
     // Open dir for WASI environment.
     std::vector<std::shared_ptr<VINode>> PreopenedDirs;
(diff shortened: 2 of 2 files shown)


juntao:

Wck-iipi:
@hydai Is this in correct direction? Are there more things I should do?

dannypsnl (approved):

PeterD1524:
I think this just allows renumbering STDIN, STDOUT, and STDERR?
Correct me if I'm wrong but I think this is not the goal of the issue.

If I understand correctly,

The original:

1. maps WASI STDIN (fd 0) to system STDIN (`STDIN_FILENO`, 0)
1. maps WASI STDOUT (fd 1) to system STDOUT (`STDOUT_FILENO`, 1)
1. maps WASI STDERR (fd 2) to system STDERR (`STDERR_FILENO`, 2)

This commit:

1. maps `StdinFd` to system STDIN (STDIN_FILENO, 0)
1. maps `StdoutFd` to system STDOUT (STDOUT_FILENO, 1)
1. maps `StderrFd` to system STDERR (STDERR_FILENO, 2)

The feat issue wants (or at least https://github.com/nodejs/node/pull/33544/files mentioned in the issue is doing):
allows redirecting WASI STDIN, STDOUT, and STDERR to user specified system fd

1. maps WASI STDIN (fd 0) to `StdinFd`
1. maps WASI STDOUT (fd 1) to `StdoutFd`
1. maps WASI STDERR (fd 2) to `StderrFd`

Maybe add some test in `test/host/wasi/wasi.cpp` to verify the behavior?

hydai (changes requested):
This is **100% incorrect**.  

We want to allow users to map the host FDs to the guest FDs (stdFd, 0, 1, 2).  
This feature will also cause a regression in behavior, and there are no tests for it.

Wck-iipi:
@PeterD1524 You were right about the Span, I shouldn't have used that. Regarding windows and unix support, I am currently using the approach that if init gets parameters then the user would be using unix, else if it is null, the value would be `VINode:stdIn`, which is compatible with Windows (Please tell me if this approach is naive or unsecure).
Regarding struct I thought it would be easier for user to enter numbers rather than creating a whole `INode` to pass as a parameter for the user(in Environ::init), therefore I used array of `std::optional<int>`, which later gets checked for null value.
Please correct me if I am wrong anywhere, thanks. If not, I will write tests for this.

PeterD1524:
@Wck-iipi The function is exposed on both Windows and UNIX, so you have to make sure people don't call it on Windows. But this is an internal API so I don't know how it is going to be used. Maybe you should ask other people for their opinions.

I think you can write test directly and run it to see if there are any problems without waiting for our reply.

hydai:
I am closing this due to inactivity. Please feel free to reopen it if you plan to continue working on this feature.

PR #3716, opened 2024-09-02
[Plugin] Support `translate` and `language` option for WASI-NN whisper backend.

q82419:
whisper backend.

### DIFF
--- a/plugins/wasi_nn/whispercpp.cpp
+++ b/plugins/wasi_nn/whispercpp.cpp
@@ -6,6 +6,7 @@
 
 #ifdef WASMEDGE_PLUGIN_WASI_NN_BACKEND_WHISPER
 #define DR_WAV_IMPLEMENTATION
+#include "simdjson.h"
 #include <examples/dr_wav.h>
 
 #include <algorithm>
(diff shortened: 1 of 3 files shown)


plugins/wasi_nn/whispercpp.cpp
```diff
@@ -196,6 +287,31 @@ Expect<ErrNo> load(WasiNNEnvironment &Env, Span<const Span<uint8_t>> Builders,
...
+      whisper_lang_id(GraphRef.SpokenLanguage.c_str()) == -1) {
+    spdlog::error("[WASI-NN] Whisper backend: Error: unknown language {}."sv,
+                  GraphRef.SpokenLanguage);
+    Env.NNGraph.pop_back();
+    return ErrNo::InvalidArgument;
+  }
+
+  // Check the translate option.
+  if (!whisper_is_multilingual(GraphRef.WhisperCtx)) {
+    if (GraphRef.SpokenLanguage != "en"sv || GraphRef.Translate) {
```
hydai, line 301: I didn't get it. Why do we disable the translate option if the translate is true?

hydai (changes requested):

hydai (approved):

PR #3750, opened 2024-09-10
[Plugin] llmc: Support training GPT2 on GPU (CUDA)

junaire:
Related #3523 
Depends on https://github.com/WasmEdge/llm.c/pull/5

### DIFF
--- a/plugins/wasmedge_llmc/llmc_fwd.h
+++ b/plugins/wasmedge_llmc/llmc_fwd.h
@@ -13,7 +13,7 @@ struct DataLoader;
 
 GPT2 *gpt2_create(const char *checkpoint_path);
 
-void gpt2_free(GPT2 *model);
+void gpt2_destroy(GPT2 *model);
 
 DataLoader *dataloader_create(const char *filename_pattern, size_t B, size_t T,
                               int process_rank, int num_processes,
--- a/plugins/wasmedge_llmc/CMakeLists.txt
+++ b/plugins/wasmedge_llmc/CMakeLists.txt
@@ -1,8 +1,6 @@
 # SPDX-License-Identifier: Apache-2.0
 # SPDX-FileCopyrightText: 2019-2024 Second State INC
 
-# TODO: Fetch llm.c source.
-
 wasmedge_add_library(wasmedgePluginWasmEdgeLLMC
   SHARED
   llmc_func.cpp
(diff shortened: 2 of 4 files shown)


juntao:

junaire:
i'll appreciate it if someone can show me how to test this cuda thing.

hydai:
Hi @junaire 
I am sorry that we don't have a CUDA-enabled CI runner to verify this feature automatically. The current process is that we will take the workflow assets and try to run them on our private machines. If you don't have such devices, that's fine; I will try to download and verify that in my own environment.

junaire:
> Hi @junaire I am sorry that we don't have a CUDA-enabled CI runner to verify this feature automatically. The current process is that we will take the workflow assets and try to run them on our private machines. If you don't have such devices, that's fine; I will try to download and verify that in my own environment.

ahhh... i thought github public runners already have GPU support?
https://github.blog/changelog/2024-07-08-github-actions-gpu-hosted-runners-are-now-generally-available/

hydai:
> ahhh... i thought github public runners already have GPU support? https://github.blog/changelog/2024-07-08-github-actions-gpu-hosted-runners-are-now-generally-available/

Well, if you check the details, you will find out that the GPU runners are `Larger Runners`, which means it's not free.

https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#per-minute-rates-for-x64-powered-larger-runners

junaire:
i tested this locally and it works, you can try to verify this on your own, note that the following patch is needed since https://github.com/WasmEdge/llm.c/pull/5 is not merged it yet:

```diff
diff --git a/plugins/wasmedge_llmc/CMakeLists.txt b/plugins/wasmedge_llmc/CMakeLists.txt
index 14908556..6ccb881d 100644
--- a/plugins/wasmedge_llmc/CMakeLists.txt
+++ b/plugins/wasmedge_llmc/CMakeLists.txt
@@ -19,7 +19,8 @@ endif()
 
 FetchContent_Declare(
   llmc
-  GIT_REPOSITORY https://github.com/WasmEdge/llm.c
+  GIT_REPOSITORY https://github.com/junaire/llm.c
+  GIT_TAG jun/cuda
 )
 FetchContent_MakeAvailable(llmc)
```

plugins/wasmedge_llmc/llmc_fwd.h
```diff
@@ -13,7 +13,7 @@ struct DataLoader;
 
 GPT2 *gpt2_create(const char *checkpoint_path);
 
-void gpt2_free(GPT2 *model);
+void gpt2_destroy(GPT2 *model);
```
dannypsnl, line 16: Is the renaming required?

junaire: it's not renaming, we're using the wrong api

dannypsnl: So that definition is external, I see.

hydai:
Okay, the current issue is that we don't have NVCC on Ubuntu and Manylinux(2014, 2_28). So, the CI failed. The dependencies should be installed, or we need to use another docker base image to handle this plugin.

junaire:
> Okay, the current issue is that we don't have NVCC on Ubuntu and Manylinux(2014, 2_28). So, the CI failed. The dependencies should be installed, or we need to use another docker base image to handle this plugin.

no, I think it's because I incorrectly used `set` instead of `option` in the cmake. i already updated the code, please try again.

hydai (approved):
I see. LGTM.

junaire:
the failure of ci seems unrelated, what can I do about it?

hydai:
Do nothing, I am merging this.

PR #3752, opened 2024-09-10
[CI] Workaround for jobs not cancellable

0yi0:
TLDR: `if: always()` cause jobs being non-cancellable. Seems like `if: !failure()` results in the same behavior here.

See https://github.com/actions/runner/issues/1846

> To cancel the workflow run, the server re-evaluates if conditions for all currently running jobs. If the condition evaluates to true, the job will not get canceled. For example, the condition if: always() would evaluate to true and the job continues to run. When there is no condition, that is the equivalent of the condition if: success(), which only runs if the previous step finished successfully.

Docs quoted from: https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/canceling-a-workflow#steps-github-takes-to-cancel-a-workflow-run

Force cancel: https://docs.github.com/en/rest/actions/workflow-runs?apiVersion=2022-11-28#force-cancel-a-workflow-run
```sh
gh api \
  --method POST \
  -H "Accept: application/vnd.github+json" \
  -H "X-GitHub-Api-Version: 2022-11-28" \
  /repos/WasmEdge/WasmEdge/actions/runs/RUN_ID/force-cancel
```

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -96,7 +96,6 @@ jobs:
     needs:
       - prep
       - paths-filter
-    if: ${{ needs.paths-filter.outputs.ci-image-base == 'true' }}
 
     name: CI image base
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


juntao:

0yi0:
NOTE: Still need to wait up to 5 minutes on self-hosted machines for cancellation.
> After the 5 minutes cancellation timeout period, the server will force terminate all jobs and steps that don't finish running or fail to complete the cancellation process.

hydai (dismissed):

hydai (approved):

Issue #2939, opened 2023-10-03
WASI preview2 proposals tracking issue


hydai:
### Summary

This issue replaced #2806.

This issue will be marked as finished once all preview2 proposals are implemented and verified.

WasmEdge would love to have native support for the WASI preview2, including the following WASI proposals:

- [ ] [WASI-IO, including WASI-POLL](https://github.com/WebAssembly/wasi-io)
- [ ] [WASI-SOCKETS](https://github.com/WebAssembly/wasi-sockets)
- [ ] [WASI-CLOCK](https://github.com/WebAssembly/wasi-clocks)
- [ ] [WASI-HTTP](https://github.com/WebAssembly/wasi-http)
- [ ] [WASI-RANDOM](https://github.com/WebAssembly/wasi-random)
- [ ] [WASI-FILESYSTEM](https://github.com/WebAssembly/wasi-filesystem)
- [x] [WASI-NN](https://github.com/WebAssembly/wasi-nn)

### Details

To ensure we can ship the preview2 proposals smoothly and are not blocked by the current plugin limitation, we will merge all proposals, except for WASI-NN, into WasmEdge Core.

The following information will be replaced by [a GitHub Project Board](https://github.com/orgs/WasmEdge/projects/6)

## WASI-IO
Status: In Progress
Assignee: @dannypsnl 
PR: https://github.com/WasmEdge/WasmEdge/pull/2934

## WASI-Sockets
Status: In Progress
Assignee: @ LFsWang 
PR: TBD

## WASI-CLOCK
Status: In Progress
Assignee: @dannypsnl 
PR: https://github.com/WasmEdge/WasmEdge/pull/2766

## WASI-HTTP
Status: Todo
Assignee:  TBD
PR: TBD

## WASI-RANDOM
Status: In Reviewing
Assignee: @LFsWang
PR: https://github.com/WasmEdge/WasmEdge/pull/2875

## WASI-FILESYSTEM
Status: Todo.
Assignee: TBD
PR: TBD

## WASI-NN
Status: Done.
Assignee: @dm4, @hydai 

### Appendix

https://github.com/orgs/bytecodealliance/projects/10/views/1
https://github.com/WebAssembly/WASI/blob/main/Proposals.md
https://github.com/bytecodealliance/wasmtime/tree/main/crates/wasi/src/preview2

dannypsnl:
NOTE: @hydai's idea that we can use external symbols to exchange data across plugin, instead of internalize everything.

[...]

PR #3753, opened 2024-09-11
[WASI] preview2 plugins, resource type instantiation, and related types export from plugin

dannypsnl:
### preview2 interfaces

- provide structure 2024/09
- provide content 2024/10-
    - refers to #2939 to get more details

- wasi:io/error
- wasi:io/streams
- wasi:cli/environment
- wasi:cli/exit
- wasi:cli/stdin
- wasi:cli/stdout
- wasi:cli/stderr
- wasi:clocks/wall-clock
- wasi:filesystem/types
- wasi:filesystem/preopens
- wasi:http/types

### What else provided in this PR

1. fix resource type internal fields 2024/10
2. fix component exported core module behaviour, it should not always require to expose host functions 2024/10
3. host function now works with `u8` 2024/10
4. `resource.drop` 2024/10-11
5. internal representation of types 2024/11

### TODO here
- [x] 2024/10/15 core module import section instantiation will crash (reason: when `FunctionInstance` is not a host function, I insert a deftype for it but no owner for the deftype)
- [x] 2024/10/17 provide resource implementation, one can code as below
    ```cxx
    class IoError : ResourceType {
    public:
      IoError() : ResourceType(new ComponentInstance("error")) {
        Impl->addHostFunc("...", ...);
        // ...
      }
    };
    ```
- [x] 2024/10-11 complete all resources content that standard required (at least to make example run)
- [x] 2024/11 fix record has no owner problem, currently, below code will not insert `Datetime` properly
    ```cxx
    class Datetime : public Record {
    public:
      Datetime()
          : Record({LabelValType("seconds", PrimValType::U64),
                    LabelValType("nanoseconds", PrimValType::U32)}) {}
    };

    WallClockModule::WallClockModule()
        : ComponentInstance("wasi:clocks/wall-clock@0.2.0") {
      addExport("datetime", Datetime());
    }
    ```

    This is a weird, since `addExport` will use right value `DefType&&` and `move` the type, then insert into the internal map of component instance. In `Datetime` case, which is a `Record`, the `variant` hierarchy is `Record <: DefValType <: DefType`, so I guess there might some problems in `variant`, if I use it then has to use the wrapping stack explicitly.

- [x] 2025/1-2025/2 refactoring `InterfaceType`
- [x] 2025/3 implements new resource machinery
- [ ] 2024/12-2025/ wasi:http


juntao:

dannypsnl:
TEST FILE: https://github.com/second-state/component-model-tests/tree/main/add

With these components, instantiation now stuck at `[warning] incomplete alias sort target export: type`.

There is an assertion failure: `Assertion failed: (Func->isHostFunction()), function exportFunction, file module.h, line 228.` needs to figure out how it happens.

PR #3825, opened 2024-10-11
[WASI-NN] piper: extend the json_input functionality to allow setting various parameters at runtime

### LINKED ISSUE #3817: feat: Allow user-specified piper parameters in audio speech request
### Summary

The following `Piper` parameters listed below are defined as fields of the [RunConfig struct](https://github.com/WasmEdge/WasmEdge/blob/master/plugins/wasi_nn/piper.h#L25-L67) in the piper plugin:

- OutputType
- SpeakerId
- NoiseScale
- LengthScale
- NoiseW
- SentenceSilenceSeconds
- JsonInput
- PhonemeSilenceSeconds

From the perspective of users' view, these parameters should be allowed to set in requests that are then passed to the server. According to the [example](https://github.com/second-state/WasmEdge-WASINN-examples/blob/master/wasmedge-piper/src/main.rs#L3-L11), however, the parameters are set by the `build_from_bytes` interface ONLY while creating a Graph instance. Is it possible to introduce `Metadata` (like used in `ggml` plugin) to allow to set the parameters at runtime.

### Details

NA.

### Appendix

_No response_

PeterD1524:
Fixes #3817.

The added parameters for `json_input` are:

- `output_type`
- `noise_scale`
- `length_scale`
- `noise_w`
- `sentence_silence`
- `phoneme_silence`

<details>

<summary>The schema of json_input after this PR</summary>

```json
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "properties": {
    "text": {
      "description": "Text input for speech synthesis",
      "type": "string"
    },
    "speaker_id": {
      "description": "Override the default speaker id, takes precedence over speaker",
      "type": "number"
    },
    "speaker": {
      "description": "Override the default speaker by name",
      "type": "string"
    },
    "output_type": {
      "description": "Override the default type of output to produce",
      "enum": ["raw", "wav"]
    },
    "noise_scale": {
      "description": "Override the default amount of noise to add during audio generation",
      "type": "number"
    },
    "length_scale": {
      "description": "Override the default speed of speaking (1 = normal, < 1 is faster, > 1 is slower)",
      "type": "number"
    },
    "noise_w": {
      "description": "Override the default variation in phoneme lengths",
      "type": "number"
    },
    "sentence_silence": {
      "default": 0.2,
      "description": "Override the default seconds of silence to add after each sentence",
      "type": "number"
    },
    "phoneme_silence": {
      "additionalProperties": {
        "type": "number"
      },
      "description": "Override the default seconds of extra silence to insert after a single phoneme, this is a mapping from single codepoints to seconds"
    }
  },
  "required": ["text"]
}
```

</details>

In Rust you'll call `GraphBuilder::build_from_bytes` with the config `{"json_input": true}` to initialize the default parameters and allow json input.
Later you can call `GraphExecutionContext::set_input` with json input like `{"text": "hello", "noise_scale": 2.0, "speaker_id": 1}` to temporarily override the default parameters.
Each time you call `GraphExecutionContext::compute`, the most recent `set_input` override is combined with the default parameters to be used within this single `compute`. The default parameters never change after initialization.


juntao:

hydai (approved):

PR #3848, opened 2024-10-23
[Installer] v2: use 0.14.1 by default

hydai:


0yi0:
Maybe we should check the version being installed is correct?

hydai:
> Maybe we should check the version being installed is correct?

Sure.

hydai:
> Maybe we should check the version being installed is correct?

Fixed.

0yi0 (approved):

SAtacker (approved):

PR #3850, opened 2024-10-23
[Voting Passed] ci(commitlint): introduce conventional commits

hydai:
Previously, WasmEdge maintainers manually checked the commit messages to see if their title meets the following format, `[Type] optionalScope: subject`. However, it's not possible to avoid fat finger merging some PR that doesn't meet this standard. So, I propose to use commitlint and the conventional commits spec for automatic detection.

If the commit message doesn't meet the standard, then the CI will fail to notify the contributors that their commit messages need to be fixed.

An example of failure will be like this:

Google Chrome 2024-10-23 18 49 17

Reference:
* Conventional commits: https://www.conventionalcommits.org/en/v1.0.0/

### DIFF
--- a/.commitlintrc.js
+++ b/.commitlintrc.js
@@ -1,0 +1,3 @@
+module.exports = {
+  extends: ["@commitlint/config-conventional"],
+};
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,7 +10,7 @@ updates:
       interval: "weekly"
     commit-message:
       # Prefix for PR title and commit messages
-      prefix: "[CI] dependabot"
+      prefix: "ci(dependabot):"
     groups:
       # Group updates, actions that match are grouped into a single PR;
       # dependabot should still raise a PR for each of the remaining actions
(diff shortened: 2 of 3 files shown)


juntao:

hydai:
@0yi0 Please DO NOT merge this until the vote passes.

I am calling maintainers for the vote.

/vote

juntao:
Can we improve this in the future to have an LLM checking the commit message? We can specify an arbitrary template format in the prompt and ask the LLM to check against it? 

It could also suggest a commit message in the comments. ;)

hydai:
> Can we improve this in the future to have an LLM checking the commit message? We can specify an arbitrary template format in the prompt and ask the LLM to check against it?
> 
> It could also suggest a commit message in the comments. ;)

Definitely, I would like to have an agent that can analyze the commit contents and format better commit messages for them. What kind of agents would be a better way to present this?
1. Like the code review agent that replies in the thread
2. Like the linter CI that shows this information in the workflow log (which may not be obvious to see)

0yi0 (approved):

0yi0 (approved):

PR #3870, opened 2024-11-12
[WASI] win: Use `ReadFile/WriteFile` instead of `ReadFileEx/WriteFileEx`

### LINKED ISSUE #3811: bug: `INode::fdWrite` sometimes returns 0 incorrectly on Windows
### Summary

While trying to replicate a read-slowdown bug observed in a Rust project, I found that sometimes `File::write_all` was reporting `WRITE_ALL_EOF` failure despite writing all its data.

On examination, it seems that `File::write` was sometimes returning `Ok(0)` despite data having been written, and this appears to be bubbling up from 

### Current State

Given this repro case in Rust:
```rust
// https://github.com/WasmEdge/WasmEdge/issues/3811
use std::fs::File;
use std::io::{Error, Write};

fn inspect_failure(e: &Error, attempts: usize, buffer_size: usize) {
    assert_eq!(e.kind(), std::io::ErrorKind::WriteZero);
    let expected_size: u64 = (attempts * buffer_size).try_into().unwrap();
    let filesize = std::fs::metadata("tmp.txt").unwrap().len();
    println!("Unexpected failure seen; after {attempts} loops of writing a {buffer_size} string ({expected_size} expected bytes), filesize was {filesize}.");
    assert_eq!(expected_size, filesize);
}

fn main() -> std::io::Result<()> {
    const BUF: &[u8] =
        "This textline will be repeated multiple times as part of a test\n".as_bytes();
    assert_eq!(BUF.len(), 64);
    let mut output = File::create("tmp.txt")?;

    for attempts in 1..=1024 {
        output
            .write_all(BUF)
            .inspect_err(|err| inspect_failure(err, attempts, BUF.len()))?;
    }
    Ok(())
}
```

the program fails more often than it passes.

Note that `File::write_all`'s implementation is to repeatedly call `File::write` with the given buffer until it is empty, and it throws an error if `File::write` reports a 0-byte success. (If the buffer is already empty, `File::write` is not called.)

Per Rust's `Write::Write` trait documentation:
```
    /// If this method consumed `n > 0` bytes of `buf` it must return [`Ok(n)`].
    /// If the return value is `Ok(n)` then `n` must satisfy `n <= buf.len()`.
    /// A return value of `Ok(0)` typically means that the underlying object is
    /// no longer able to accept bytes and will likely not be able to in the
    /// future as well, or that the buffer provided is empty.
```

I did have a simpler repro which open-coded `File::write_all` and confirmed that the failure was always coming through on the first loop, so there was no OS interrupts or partial writes being seen, which makes sense for such a small buffer.

I believe this to be a bug on the WasmEdge side because, per my repro, the data _was_ written by the call that returned `Ok(0)`, so I suspect the bug is that in `INode::fdWrite`; specifically, there's a code-path in [this block](https://github.com/WasmEdge/WasmEdge/blob/0.14.1/lib/host/wasi/inode-win.cpp#L1210-L1227) where `NWritten += NumberOfBytesWrite` is not called, even though bytes were written, and no error was returned to the caller. I can't immediately see the faulty code path, since it seems like the only way to not call that addition is to set Result. And in this case, there's only one IOV.

I _suspect_ the fix is to _aways_ call `NWritten += NumberOfBytesWrite` after `GetOverlappedResult`, irrespective of the value of the result, so that we do not accidentally signal EOF due to a (valid) partial write.

Looking at the code, the failure-path loop that calls `GetOverlappedResult` on the remaining cancelled operations may wipe out the number of bytes written, so I think we should be adding `NWritten` immediately after `GetOverlappedResult` in all cases, before looking at the status code. (This probably applies to all 8 calls to `GetOverlappedResult` in this file, I think).

### Expected State

All 65kb of data should be written and no errors raised.

### Reproduction steps

```pwsh
cargo new wasmedge-3811
cd wasmedge-3811
# Populate src/main.rs with the above repro case however you see fit
$Env:CARGO_TARGET_WASM32_WASIP1_RUNNER="wasmedge --dir .:."
cargo run --target wasm32-wasip1
```

### Screenshots

_No response_

```console
> cargo run --target wasm32-wasip1 
   Compiling wasmedge-3811 v0.1.0 (C:\Users\paulh\Rust\wasmedge-3811)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.61s
     Running `wasmedge --dir .:. target\wasm32-wasip1\debug\wasmedge-3811.wasm`
Unexpected failure seen; after 1 loops of writing a 64 string (64 expected bytes), filesize was 64.
Error: Error { kind: WriteZero, message: "failed to write whole buffer" }
error: process didn't exit successfully: `wasmedge --dir .:. target\wasm32-wasip1\debug\wasmedge-3811.wasm` (exit code: 1)
```

Note that other failure counts are possible, e.g.,
```console
Unexpected failure seen; after 25 loops of writing a 64 string (1600 expected bytes), filesize was 1600.
```

### Components

CLI, Rust SDK

0.14.1

Windows 11 (Microsoft Windows [Version 10.0.22631.4169])

### Hardware Architecture

x86_64

_No response_

ibmibmibm:
`GetOverlappedResult` occasionally succeeds but returns zero inlpNumberOfBytesTransferred`. After switching to
`ReadFile/WriteFile`, this issue no longer occurs.

* Fix #3811

### DIFF
--- a/include/system/winapi.h
+++ b/include/system/winapi.h
@@ -386,10 +386,6 @@ using OVERLAPPED_ = struct _OVERLAPPED {
 };
 using LPOVERLAPPED_ = OVERLAPPED_ *;
 
-using LPOVERLAPPED_COMPLETION_ROUTINE_ = VOID_(WASMEDGE_WINAPI_WINAPI_CC *)(
-    DWORD_ dwErrorCode, DWORD_ dwNumberOfBytesTransfered,
-    LPOVERLAPPED_ lpOverlapped) noexcept;
-
 using REPARSE_DATA_BUFFER_ = struct _REPARSE_DATA_BUFFER {
   ULONG_ ReparseTag;
   USHORT_ ReparseDataLength;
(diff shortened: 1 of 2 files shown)


juntao:

TBBle (approved):
Per https://github.com/WasmEdge/WasmEdge/issues/3811#issuecomment-2470672421, this appears to have fixed #3811 as intended.

hydai (approved):

PR #3893, opened 2024-11-27
[Loader] fix nested-component loading

dannypsnl:


juntao:

lib/loader/ast/section.cpp
```diff
@@ -217,26 +218,38 @@ Expect<void> Loader::loadSection(AST::TagSection &Sec) {
...
+    }
+    auto NestedComp = std::make_shared<AST::Component::Component>();
+    NestedComp->getMagic() = WasmMagic;
+    NestedComp->getVersion() = {Ver[0], Ver[1]};
+    NestedComp->getLayer() = {Ver[2], Ver[3]};
+
+    auto Offset = FMgr.getOffset();
+
+    if (auto Res =
+            loadComponent(*NestedComp, ExpectedSize - (Offset - StartOffset));
```
q82419, line 244: I think there may be a situation that `ExpectedSize` < `Offset` - `StartOffset`?
Consider a situation that the section size was loaded and set as size `10` in `Loader::loadSectionContent<T>()`, and `StartOffset` is `50` when entering this function, and the `Offset` is `70` after above loading. Then the evaluate of `ExpectedSize - (Offset - StartOffset)` will underflow in a `uint32_t`.

q82419 (changes requested):

lib/loader/ast/section.cpp
dannypsnl: Indeed, I push a fix for this, there are two place use the same idea. (nested module and component)

lib/loader/ast/section.cpp
```diff
@@ -217,26 +218,43 @@ Expect<void> Loader::loadSection(AST::TagSection &Sec) {
...
+                          FMgr.getLastOffset(), ASTNodeAttr::Component);
+    }
+    auto NestedComp = std::make_shared<AST::Component::Component>();
+    NestedComp->getMagic() = WasmMagic;
+    NestedComp->getVersion() = {Ver[0], Ver[1]};
+    NestedComp->getLayer() = {Ver[2], Ver[3]};
+
+    auto Offset = FMgr.getOffset();
+
+    if (ExpectedSize < Offset - StartOffset) {
```
q82419, line 243: I think this is suit for `unlikely`.

lib/loader/ast/section.cpp
```diff
@@ -259,9 +277,15 @@ Expect<void> Loader::loadSection(AST::CoreModuleSection &Sec) {
     CoreMod.getVersion() = Ver;
 
     auto Offset = FMgr.getOffset();
-    ExpectedSize -= (Offset - StartOffset);
 
-    if (auto Res = loadModuleInBound(CoreMod, ExpectedSize); !Res) {
+    if (ExpectedSize < Offset - StartOffset) {
```
q82419, line 281: I think this is suit for `unlikely`.

q82419 (changes requested):

q82419 (approved):

PR #3929, opened 2024-12-19
[WASI-NN] GGML: Support llama 3.2 vision (mllama) (PART 4)

q82419:
- [x] #3955 
- [x] #3956 
- [x] #3957
- [x] Rebase to latest llama.cpp

This is the part 4 of mllama implementation in WASI-NN GGML.
(base on #3957)

### DIFF
--- a/.github/workflows/build-extensions.yml
+++ b/.github/workflows/build-extensions.yml
@@ -159,15 +159,10 @@ jobs:
           $vsPath = (vswhere -latest -property installationPath)
           Import-Module (Join-Path $vsPath "Common7\Tools\Microsoft.VisualStudio.DevShell.dll")
           Enter-VsDevShell -VsInstallPath $vsPath -SkipAutomaticLocation -DevCmdArguments "-arch=x64 -host_arch=x64 -winsdk=10.0.22621.0"
-          $llvm = "LLVM-17.0.6-win64-MultiThreadedDLL.zip"
-          curl -sLO https://github.com/WasmEdge/llvm-windows/releases/download/llvmorg-17.0.6/LLVM-17.0.6-win64-MultiThreadedDLL.zip -o $llvm
-          Expand-Archive -Path $llvm
-          $llvm_dir = "$pwd\\LLVM-17.0.6-win64-MultiThreadedDLL\\LLVM-17.0.6-win64\\lib\\cmake\\llvm"
           $cmake_sys_version = "10.0.22621.0"
-          cmake -Bbuild -GNinja "-DCMAKE_SYSTEM_VERSION=$cmake_sys_version" -DCMAKE_MSVC_RUNTIME_LIBRARY=MultiThreadedDLL "-DLLVM_DIR=$llvm_dir" -DWASMEDGE_BUILD_TESTS=ON -DWASMEDGE_USE_LLVM=OFF -DWASMEDGE_BUILD_TOOLS=OFF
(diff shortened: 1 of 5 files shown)


juntao:

q82419:
@hydai  @dm4 

Please help to check that the behavior of running various models is reasonable.

plugins/wasi_nn/wasinnfunc.cpp
```diff
@@ -661,15 +701,23 @@ WasiNNFinalizeExecCtx::bodyImpl(const Runtime::CallingFrame &Frame,
...
+  switch (Env.NNContext[ContextId].getBackend()) {
+  case WASINN::Backend::GGML:
+    return WASINN::GGML::unload(Env, ContextId);
+  // case WASINN::Backend::Whisper:
+  //   return WASINN::Whisper::unload(Env, ContextId);
+  default:
+    spdlog::error("[WASI-NN] finalize_execution_context: Only GGML and "sv
+                  "Whisper backends support finalize_execution_context."sv);
+    return WASINN::ErrNo::InvalidArgument;
+  }
```
PeterD1524, line 720: I think these should call finalizeExecCtx instead of unload.

q82419: I forgot to uncomment this. Will fix it.

Both `finalize_execution_context` and `unload` should be called.
The `finalize_execution_context` is to finalize the execution context by context ID, which is initialized by a graph.
The `unload` is to finalize the graph by graph ID.
Some resources are hold by the graph and shared with the generated contexts. For the resource management, we should guarantee that the graph should not be deleted if the contexts still exist.
If a graph is unloaded, users will get error if they use the corresponding contexts to infer, and the graph will be fully released after the all corresponding contexts are finalized.
If a context is finalized, it will not affact the graph, and users can still use that graph to initialize another context.

PeterD1524: I think you mean that for users to release all resources, both `finalize_execution_context` and `unload` should be called, right?

But this part of code is in `Expect<WASINN::ErrNo> WasiNNFinalizeExecCtx::bodyImpl` at `plugins/wasi_nn/wasinnfunc.cpp`, so it should call `WASINN::GGML::finalizeExecCtx(Env, ContextId)` instead of `WASINN::GGML::unload(Env, ContextId)`.

q82419: I think the work flow is like this:

1. Load and create a graph, get a graph ID (assume `g_1` created)
2. Init a execution context from a graph ID, and get a context ID (assume `c_1` created and link to `g_1`)
3. Maybe several context IDs initialized by a graph ID (assume `c_2` and `c_3` created and link to `g_1`)
4. Computation
5. Now I want to finalize `c_2` context. Call `finalize_execution_context(c_2)`.
6. Computation
7. Now I want to fully delete graph `g_1`. Currently `c_1` and `c_3` contexts still alive.
8. Call `finalize_execution_context(c_1)` and `finalize_execution_context(c_3)`.
9. Call `unload(g_1)`.

PeterD1524: Yes, but with the current code when you call `finalize_execution_context(c_2)` it actually does `unload(c_2)`.
Same for `finalize_execution_context(c_1)`, it does `unload(c_1)`. `finalize_execution_context(c_3)`, it does `unload(c_3)`.
From a static analysis perspective, the function `WASINN::GGML::finalizeExecCtx` is never referenced and called. So is `WASINN::Whisper::finalizeExecCtx`.

q82419: Oh, sorry that its wrong implementation.
Fix at https://github.com/WasmEdge/WasmEdge/pull/3979 .

hydai:
@q82419 
Could you please attach the link to the upstream PR?

hydai:
No longer need this PR.  
We already support gemma3-vision, which is good enough to fully replace this model.

PR #3936, opened 2024-12-24
[WASI-NN] ChatTTS: fix unload function which incorrectly used graph id parameter as context id

PeterD1524:
Can be merged after test workflow added.

### DIFF
--- a/plugins/wasi_nn/wasinn_chattts.cpp
+++ b/plugins/wasi_nn/wasinn_chattts.cpp
@@ -305,9 +305,8 @@ Expect<WASINN::ErrNo> compute(WasiNNEnvironment &Env,
 }
 
 Expect<WASINN::ErrNo> unload(WASINN::WasiNNEnvironment &Env,
-                             uint32_t ContextId) noexcept {
-  auto &CxtRef = Env.NNContext[ContextId].get<Context>();
-  auto &GraphRef = Env.NNGraph[CxtRef.GraphId].get<Graph>();
+                             uint32_t GraphId) noexcept {
+  auto &GraphRef = Env.NNGraph[GraphId].get<Graph>();
   if (GraphRef.EnableDebugLog) {
     spdlog::info("[WASI-NN] Neural speed backend: start unload."sv);
   }
@@ -320,6 +319,7 @@ Expect<WASINN::ErrNo> unload(WASINN::WasiNNEnvironment &Env,
     GraphRef.ChatTTSModule = nullptr;
     Py_Finalize();
   }
+  Env.NNGraph.erase(Env.NNGraph.begin() + GraphId);
   return WASINN::ErrNo::Success;
 }
 


juntao:

grorge123:
Thank you for your PR. I am not sure if erasing the graph will break the graph that ID is larger than we unloaded.
In the original design, assume the unload function will be called only when we want to release wasi-nn. All the graphs shared the same Python interpreter, such that once we call `Py_Finalize` will release all the graph's resources. I think the better solution is to assume there is only one graph to avoid the above problem.

PeterD1524:
@grorge123 Thanks for pointing out. I did not notice the Py_Finalize.

If I understand correctly, unload should release only the resource used by the graph specified by the graph id.

After digging into this, I realized:

1. The WasmEdge process hangs forever if Py_Finalize is not called.

   I think this is why the original design called Py_Finalize in unload.

   For example, remove this line and it will hang forever:

   https://github.com/second-state/WasmEdge-WASINN-examples/blob/56d497af0ca77b07188d047cf9961a422d02cfb9/wasmedge-chatTTS/src/main.rs#L54

   I think this is not ideal.

2. numpy does not work with mutiple Py_Initialize and Py_Finalize.

   chattts depends on numpy.
   Therefore, it impossible to load a new graph after calling unload with the original design.

   Proof of Concept:

   ```c++
   #define PY_SSIZE_T_CLEAN
   #include <Python.h>

   #include <iostream>

   int main() {
     for (auto I = 0; I < 2; ++I) {
       std::cout << I << " start\n";
       Py_Initialize();
       auto *Module = PyImport_ImportModule("numpy");
       if (Module == nullptr) {
         PyErr_Print();
       }
       Py_Finalize();
       std::cout << I << " end\n";
     }
   }
   ```

   Output:

   ```
   0 start
   0 end
   1 start
   Traceback (most recent call last):
     File "/lib/python3.11/site-packages/numpy/__init__.py", line 130, in <module>
       from numpy.__config__ import show as show_config
     File "/lib/python3.11/site-packages/numpy/__config__.py", line 4, in <module>
       from numpy.core._multiarray_umath import (
     File "/lib/python3.11/site-packages/numpy/core/__init__.py", line 24, in <module>
       from . import multiarray
     File "/lib/python3.11/site-packages/numpy/core/multiarray.py", line 10, in <module>
       from . import overrides
     File "/lib/python3.11/site-packages/numpy/core/overrides.py", line 8, in <module>
       from numpy.core._multiarray_umath import (
   SystemError: Objects/structseq.c:476: bad argument to internal function
   1 end
   ```

   Related:
   https://stackoverflow.com/questions/7676314/py-initialize-py-finalize-not-working-twice-with-numpy

3. The process should exit normally without hanging even if Py_Finalize is not called.

   I tried doing the work of wasi-nn chattts all in a single main function and it exited immediately without calling Py_Finalize.

4. The minimal work I can get to make the WasmEdge process hang is PyImport_ImportModule("torch").

   chattts depends on torch.
   No further Python/C API calls are needed and the process will hang if not calling Py_Finalize.
   Therefore, I think the problem is related to torch.

5. After some experiments, I think the cause is calling Py_Initialize in a thread and not releasing the GIL.

   If I understand correctly, WasmEdge creates a thread to run wasm and the plugin are run in the created thread. I tried Py_Initialize and import torch in a thread in a simple program and it hangs as well.

   I looked at htop and saw torch created several threads, so I guessed that some of torch's threads are waiting to acquire the GIL to do some final clean up. (I haven't looked into torch's implementation yet.)

   But they can't because the GIL has been held by the thread we created and the GIL was never released. I tried releasing the GIL in thread we created and the program stops hanging.

   Proof of concept:

   ```c++
   #define PY_SSIZE_T_CLEAN
   #include <Python.h>

   #include <iostream>
   #include <thread>

   int main() {
     auto Thread = std::thread{[]() {
       Py_Initialize();
       // outputs 1: means this thread is holding the GIL
       std::cout << PyGILState_Check() << "\n";
       auto *Module = PyImport_ImportModule("torch");
       if (Module == nullptr) {
         PyErr_Print();
       }
       Py_CLEAR(Module);
       // uncomment below to release the GIL and avoid hanging
       // PyEval_SaveThread();
       // if the GIL is not released here,
       // other threads cannot acquire the GIL after this thread ends
     }};
     Thread.join();
     std::cout << "exiting...\n";
     // hangs forever
   }
   ```

Based on these observations, I suggest that we release the GIL by calling [PyEval_SaveThread](https://docs.python.org/3/c-api/init.html#c.PyEval_SaveThread) as soon as we call Py_Initialize (if !Py_IsInitialized()).

If we need to call any Python/C API, we acquire the GIL with [PyGILState_Ensure](https://docs.python.org/3/c-api/init.html#c.PyGILState_Ensure). When we don't need Python/C API anymore (or returning from the function), we release the GIL with [PyGILState_Release](https://docs.python.org/3/c-api/init.html#c.PyGILState_Release).

Example from https://docs.python.org/3/c-api/init.html#non-python-created-threads:

```c++
PyGILState_STATE gstate;
gstate = PyGILState_Ensure();

/* Perform Python actions here. */
result = CallSomeFunction();
/* evaluate result or handle exception */

/* Release the thread. No Python API allowed beyond this point. */
PyGILState_Release(gstate);
```

I still don't know what caused the different behavior between running in main thread and in another new thread.

I'll update this PR with this approach later.

PeterD1524:
I realized that we should not erase the graph from Env.NNGraph.
This will invalidate graph ids greater than the current graph id.

Therefore, ggml plugin and whisper plugin are doing it incorrectly:
https://github.com/WasmEdge/WasmEdge/blob/c5e1fcea9e9b438f5bad71bf851c09d97b71fb2f/plugins/wasi_nn/wasinn_ggml.cpp#L1635
https://github.com/WasmEdge/WasmEdge/blob/c5e1fcea9e9b438f5bad71bf851c09d97b71fb2f/plugins/wasi_nn/wasinn_whisper.cpp#L642

hydai:
Hi @PeterD1524 
You can find that #3929 is going to fix the graph id issue.

PeterD1524:
> Hi @PeterD1524 You can find that #3929 is going to fix the graph id issue.

Cool. Thanks.

PeterD1524:
Because the commits in [#3929](https://www.github.com/WasmEdge/WasmEdge/pull/3929) is already handling the original intent of this PR, I am closing this.

I'll open a another PR which will focus on the Py_Finalize and GIL problem.

PR #3940, opened 2024-12-27
[WASI-NN] ChatTTS: fix GIL problem and do not call Py_Finalize

PeterD1524:
As pointed out in https://github.com/WasmEdge/WasmEdge/pull/3936#issuecomment-2563290843, ChatTTS's dependency torch create threads which probably requires the GIL.
If we did not release the GIL, those threads will hang forever.
Calling `Py_Finalize` can shutdown the Python interpreter but it will also release other resources we still want to keep.

This PR:
Call `PyEval_SaveThread` to release the GIL if `Py_Initialize`d.

Added a new class `GIL`.
When constructed, it will ensure holding the GIL by calling `PyGILState_Check`.
When destructed, it will release the GIL if the GIL was acquired previously by calling `PyGILState_Release`.
This class should be used before calling Python/C API.

This class is to avoid the following pattern:
```c++
PyGILState GState = PyGILState_Ensure();
...use the Python API here...
PyGILState_Release(GState);
```
The ...use the Python API here... part may throw exceptions or return early not calling (or forget to call) `PyGILState_Release`. If we have many early returns we will need many `PyGILState_Release` too.

idea from https://abandonedwig.info/blog/2009/10/06/managing-python-gil-via-raii.html

Removed the `Py_Finalize` in unload because numpy does not allow multiple Python interpreter initializations and `Py_Finalize` will release the resources of all graphs.

### DIFF
--- a/plugins/wasi_nn/wasinn_chattts.cpp
+++ b/plugins/wasi_nn/wasinn_chattts.cpp
@@ -37,7 +37,11 @@ Expect<WASINN::ErrNo> load(WASINN::WasiNNEnvironment &Env,
   // Create Model class
   if (!Py_IsInitialized()) {
     Py_Initialize();
+    if (PyGILState_Check()) {
+      PyEval_SaveThread();
+    }
   }
+  GIL Lock;
   if (GraphRef.ChatTTSModule == nullptr) {
     GraphRef.ChatTTSModule = PyImport_ImportModule("ChatTTS");
     if (GraphRef.ChatTTSModule == nullptr) {
@@ -123,6 +127,7 @@ Expect<WASINN::ErrNo> setInput(WasiNNEnvironment &Env, uint32_t ContextId,
       Env.NNGraph.pop_back();
       return ErrNo::InvalidEncoding;
     }
+    GIL Lock;
     // Handle Refine Text Params
     PyObject *PromptObj = nullptr;
     if (Doc.at_key("prompt").error() == simdjson::SUCCESS) {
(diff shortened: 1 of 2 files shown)


juntao:

PeterD1524:
This is for future reference.

The `GIL` class is the same as the `gil_scoped_acquire` class in pybind11 if `PYBIND11_SIMPLE_GIL_MANAGEMENT` is defined:
https://github.com/pybind/pybind11/blob/cf020a1de25861d4b982081890c43e7eab1ff5e9/include/pybind11/gil.h#L185-L196

There is another complex version of `gil_scoped_acquire` if `PYBIND11_SIMPLE_GIL_MANAGEMENT` is not defined:
https://github.com/pybind/pybind11/blob/cf020a1de25861d4b982081890c43e7eab1ff5e9/include/pybind11/gil.h#L29-L136
I think the simple one is sufficient.

All `GIL` destructors should run in the reverse order of all constructor runs (like a stack) to work properly, and the number of destructors should be equal to the number of constructors.
The copy constructor and copy assigment operator are deleted to avoid such thing: 
[pybind/pybind11#4183](https://www.github.com/pybind/pybind11/pull/4183)

This does not prevent new, new[], delete, delete[] which can change the execution order of destructors.
Deleting those operators is not enough.
Users might want to associate a `GIL` with an object and they will use the `GIL` as a member of a class.
This class is still new-able:
```c++
class C {
  GIL G;
};

int main() {
  Py_Initialize();
  PyEval_SaveThread();
  auto A = std::make_unique<C>();
  auto B = std::make_unique<C>();
  std::swap(A, B); // order changed
}
```

hydai:
I see. I think the simple way is enough for the current use case. The reason is that AI-related frameworks are changing so quickly. Maybe we will have a pure C/C++ implementation of TTS that will provide better performance. Then, it can replace this backend.

hydai (approved):

PR #3944, opened 2024-12-31
[WASI-NN] ggml: update options types and embeddings generation

dm4:
- Update option types according to llama.cpp
  - In llama.cpp, the types of context parameters and common parameters are actually different.
  - For instance, `cparams.n_ctx` is `uint32_t` while `params.n_ctx` is `int32_t`.
  - This is because some parameters are initially set in common params and are only assigned to context params after evaluation.
  - The ggml plugin mainly follows the types of common params.
  - Ref:
    - https://github.com/ggerganov/llama.cpp/blob/master/common/common.h
    - https://github.com/ggerganov/llama.cpp/blob/master/include/llama.h
- Do not append SEP when getting embeddings
  - Ref: https://github.com/ggerganov/llama.cpp/blob/7eee341/examples/embedding/embedding.cpp#L147-L154
- Bump llama.cpp version and WASI-NN plugin version


juntao:

hydai (approved):

Issue #2799, opened 2023-09-10
wasi-crypto `secretkey_export` on rsa with `ENCODING_PKCS8` get pkcs1 encoded


Puellaquae:
## Description

This issue and #2266 is caused by the same reason.

`secretkey_export` on rsa with `ENCODING_PKCS8` get pkcs1 encoded

## Environment

- Hardware Architecture: (e.g. x86\_64 or aarch64)
- Operating system: (e.g. Ubuntu 20.04 or CentOS 7.6)

The following information is optional. Please provide them only if you have built from source.
- C++ Compiler version:
- CMake version:
- CMake flags: (e.g. `-DCMAKE_BUILD_TYPE=Release -DWASMEDGE_BUILD_AOT_RUNTIME=ON`)

## Steps to Reproduce

wasi binding use https://github.com/WebAssembly/wasi-crypto/blob/main/implementations/bindings/rust/src/raw.rs

```rust
const NONE_OPTS: raw::OptOptions = raw::OptOptions {
    tag: raw::OPT_OPTIONS_U_NONE.raw(),
    u: raw::OptOptionsUnion { none: () },
};

let (pk, sk) = unsafe {
    let kp = raw::keypair_generate(raw::ALGORITHM_TYPE_SIGNATURES, RSA_PKCS1_2048_SHA256, NONE_OPTS)?;
    (raw::keypair_publickey(kp)?, raw::keypair_secretkey(kp)?)
};

let res = unsafe {
    let arr = raw::secretkey_export(sk, raw::SECRETKEY_ENCODING_PKCS8)?; // get encoding pkcs1 actually
    let len = raw::array_output_len(arr)?;
    let mut buf = vec![0; len];
    raw::array_output_pull(arr, buf.as_mut_ptr(), buf.len())?;
    buf
};
```

alabulei1:
Hi @sonder-joker 

Could you please take a look at this issue? This issue affects WasmEdge-quickjs supports the Node cryoto API. Thanks.

PR #3963, opened 2025-01-10
fix: secretkey_export on rsa with ENCODING_PKCS8

deveshidwivedi:
This PR addresses the issue where RSA secret key export using `PKCS8` incorrectly returned `PKCS1` encoding. This update ensures proper `PKCS8` encoding during the export process.

Fixes: #2799


hydai:
It seems that this modification cannot pass the unit test. Please run the test suite in your own environment to ensure that your PR is working before triggering the CI again.

deveshidwivedi:
I updated the tests to use the pkcs8 format, but they're still failing. I noticed that `wasi_crypto` is missing from my build directory even after generating the build files and running tests. I'll check what I'm doing wrong and fix this.

hydai (approved):
LGTM

PR #3964, opened 2025-01-11
[WASI-NN] ggml: Use the cached image embed instead to reduce costs

hydai:
In this PR, we changed the logic for handling the image embed for the llava and qwen2vl models.

1. When the payload is provided as a base64 string, it will always recompute to ensure we never miss any changes.
2. However, once the payload is provided from a file, we will check if there is an existing computed image embed. If so, we will not recompute it to reduce time costs. The cached image embed will be consumed unless the `compute` function is called.
3. If users want to reload a new image embed, please call the `compute` function to consume it. Then, the `setInput` function will generate the image embed again.
4. The reason we have to apply this workaround is that llama.cpp has disabled all GPU backends for the Clip model. It is quite slow because the entire workload is running solely on the CPU.


juntao:

dm4 (approved):
LGTM. However, the description of this pull request on GitHub is a bit out of date. It might be a good idea to update the description to align with the current implementation logic.

hydai:
Updated. Let's merge it once the CI passes

PR #3969, opened 2025-01-13
fix nested vm call

turuslan:
WasmEdge stores references to current VM in thread local variables.
`enterFunction` calls `prepare` which overwrites these variables.
Problem occurs when VM 1 calls host, which calls VM 2
```
enterFunction(this=executor_1)          // `This = executor_1`
  ...host...                            // wasm instance 1 calls host
    WasmEdge_ExecutorInvoke(executor_2) // host calls wasm instance 2
      enterFunction(this=executor_2)    // `This = executor_2`
    ~Executor(this=executor_2)          // `executor_2` is destroyed
  ...(this=executor_2)                  // `This` still references destroyed `executor_2`
```
PR adds code to save and recover these variables.


juntao:

dannypsnl:
Can you provide which issue is related with these changes, or explain in what case you will face this problem?

turuslan:
`Executor::enterFunction` overwrites `Executor::This` when entering compiled wasm function.
`ProxyHelper::proxy` reads `Executor::This` when entering host function.

When host function creates second executor to call nested wasm,
second executor will overwrite `Executor::This`.
When that host function returns control back to first executor,
`Executor::This` is not restored and contains dangling reference to second executor.

---

Attaching code for reproduction.
[nested.zip](https://github.com/user-attachments/files/18406499/nested.zip)

Expected output
```
wasmFn
  hostFnCheck executor=0x123e04d00
  hostFnOverwrite
    wasmFnOverwrite
  hostFnCheck executor=0x123e04d00
```

When second executor still exists
```
wasmFn
  hostFnCheck executor=0x123e04d00
  hostFnOverwrite
    wasmFnOverwrite
Assertion failed: (!FrameStack.empty()), function getModule, file stackmgr.h, line 211.
```

When second executor was destroyed
```
wasmFn
  hostFnCheck executor=0x123e04d00
  hostFnOverwrite
    wasmFnOverwrite
[error] execution failed: out of bounds memory access, Code: 0x408
```

---

We use wasmedge for our blockchain implementation.
Nested execution happens during wasm code upgrade,
when old wasm code asks host to execute new wasm code,
to check if it works.

Another possible is `program.wasm` dynamically loading `plugin.wasm`.

---

This PR saves thread local variables before modifying then, and restores them on return.

hydai:
Hi @turuslan 
Could you rebase this PR or allow the maintainers to modify your PR, which is an option on the PR page?

CleanShot 2025-01-15 at 13 31 11

It doesn't allow me to update the branch and try to rerun the CI for it like above, thanks.

lib/executor/helper.cpp
```diff
@@ -124,6 +124,8 @@ Executor::enterFunction(Runtime::StackManager &StackMgr,
     Span<ValVariant> Args = StackMgr.getTopSpan(ArgsN);
     std::vector<ValVariant> Rets(RetsN);
 
+    SavedThreadLocal SavedThreadLocal;
```
q82419, line 127: Wondering why only save the local stack when entering a AOT compiled function?

turuslan: Currently only `Executor::prepare` overwrites these thread local variables.
So saving them before calling `Executor::prepare`.

hydai (approved):
LGTM. I am going to merge this. Just want to know if we can have a minimum test case so that we can add it into the unit test to verify the behavior.

ibmibmibm (approved):

turuslan:
> minimum test case
Here's sample code [nested.zip](https://github.com/user-attachments/files/18406499/nested.zip).

Should I convert it to GTest and place in "test/api" directory?
Should I add wasm code to "test/api/apiTestData" directory or embed them as array literal?
My code also adds `String` and `Ptr` wrappers to simplify sample code.

hydai:
> Should I convert it to GTest and place in "test/api" directory?

If you can do that, I would greatly appreciate it. According to your sample code, it could be placed in `test/api` since you are using the API method. I think `test/executor` is for those examples that use the `executor` instance directly.

> Should I add wasm code to "test/api/apiTestData" directory or embed them as array literal?

Please embed them as array literals.

> My code also adds `String` and `Ptr` wrappers to simplify sample code.

It should be fine if you define the same things inside the unit test files.

lib/executor/helper.cpp
hydai: Yes, the current workflow will only overwrite these thread-local variables inside the AOT-compiled condition before calling `prepare`, which will modify these values.

turuslan:
Added test.
Rebased to fresh master (can't "allow edits" for my organization fork).

hydai:
The tests on Windows are failed due to the compilation issue.
The CI of Ubuntu20.04 arm64 and Fedora Rawhide are unrelated, you can ignore these two workflows.

hydai (approved):

hydai:
Thank you @turuslan

PR #4014, opened 2025-02-10
[WASI-NN] ggml: check the projection model before loading 

dm4:
* [WASI-NN] ggml: check the projection model before loading
* [WASI-NN] ggml: apply clang-tidy
* [WASI-NN] ggml: bump llama.cpp b4681

### DIFF
--- a/plugins/wasi_nn/wasinn_ggml.cpp
+++ b/plugins/wasi_nn/wasinn_ggml.cpp
@@ -55,7 +55,7 @@ namespace {
   return Error;
 
 // Llama logging callback.
-void LlamaLogCallback(ggml_log_level LogLevel, const char *LogText,
+void llamaLogCallback(ggml_log_level LogLevel, const char *LogText,
                       void *UserData) {
   Graph &GraphRef = *reinterpret_cast<Graph *>(UserData);
   if (!GraphRef.EnableLog) {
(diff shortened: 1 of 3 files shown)


plugins/wasi_nn/wasinn_ggml.cpp
```diff
@@ -1762,7 +1763,13 @@ Expect<ErrNo> setInput(WasiNNEnvironment &Env, uint32_t ContextId,
   if (Base64ImagePos.has_value() || CxtRef.Conf.ImagePath != ""sv) {
     // Prompt with image input. Check is llava or mllama case.
 
-    // First check the projection model is loaded.
+    // First check the projection model is given.
+    if (GraphRef.MMProjModelPath == ""sv) {
+      RET_ERROR(ErrNo::InvalidArgument,
+                "setInput: projection model path is not given."sv)
```
hydai, line 1769: I think we should also have a reasoning output, such as, "The given model does not support image input."

hydai (changes requested):

plugins/wasi_nn/wasinn_ggml.cpp
dm4: I just updated it to `setInput: the given model does not support image input, so a projection model is required.`

hydai (dismissed):

hydai:
You need to bump the WASI-NN plugin version together.

hydai (approved):

PR #4020, opened 2025-02-13
[WASI-NN] ggml: add TTS speaker profile support

dm4:
- Support tts speaker profile with `tts-speaker-file`.
  - Speaker profile example: https://github.com/edwko/OuteTTS/tree/main/outetts/version/v1/default_speakers


hydai (changes requested):
Since there is a lot of special information required by TTS, I would like to have a standalone cpp file for handling the TTS feature. What do you think? Something like `wasinn_ggml_tts.cpp`

PR #4053, opened 2025-03-10
[WASI] preview2 plugins, resource type instantiation, and related types export from plugin

dannypsnl:
### preview2 interfaces

- provide structure 2024/09
- provide content 2024/10-
    - refers to #2939 to get more details

- wasi:io/error
- wasi:io/streams
- wasi:cli/environment
- wasi:cli/exit
- wasi:cli/stdin
- wasi:cli/stdout
- wasi:cli/stderr
- wasi:clocks/wall-clock
- wasi:filesystem/types
- wasi:filesystem/preopens
- wasi:http/types

### What else provided in this PR

1. fix resource type internal fields 2024/10
2. fix component exported core module behaviour, it should not always require to expose host functions 2024/10
3. host function now works with `u8` 2024/10
4. `resource.drop` 2024/10-11
5. internal representation of types 2024/11

### TODO here
- [x] 2024/10/15 core module import section instantiation will crash (reason: when `FunctionInstance` is not a host function, I insert a deftype for it but no owner for the deftype)
- [x] 2024/10/17 provide resource implementation, one can code as below
    ```cxx
    class IoError : ResourceType {
    public:
      IoError() : ResourceType(new ComponentInstance("error")) {
        Impl->addHostFunc("...", ...);
        // ...
      }
    };
    ```
- [x] 2024/10-11 complete all resources content that standard required (at least to make example run)
- [x] 2024/11 fix record has no owner problem, currently, below code will not insert `Datetime` properly
    ```cxx
    class Datetime : public Record {
    public:
      Datetime()
          : Record({LabelValType("seconds", PrimValType::U64),
                    LabelValType("nanoseconds", PrimValType::U32)}) {}
    };

    WallClockModule::WallClockModule()
        : ComponentInstance("wasi:clocks/wall-clock@0.2.0") {
      addExport("datetime", Datetime());
    }
    ```

    This is a weird, since `addExport` will use right value `DefType&&` and `move` the type, then insert into the internal map of component instance. In `Datetime` case, which is a `Record`, the `variant` hierarchy is `Record <: DefValType <: DefType`, so I guess there might some problems in `variant`, if I use it then has to use the wrapping stack explicitly.

- [x] 2025/1-2025/2 refactoring `InterfaceType`
- [x] 2025/3 implements new resource machinery
- [x] 2025/3 instantiation: with function, with type
- [x] 2025/3 import type bound
- [ ] 2024/12-2025/ wasi:http


juntao:

dannypsnl:
NOTE: Rename branch, replace #3753.

plugins/wasi_io/env.h
```diff
@@ -0,0 +1,21 @@
...
+#include <vector>
+
+namespace WasmEdge {
+namespace Host {
+
+class OutputStream;
+
+class WasiIOEnvironment {
+public:
+  std::vector<OutputStream> OutputStreamList;
```
dannypsnl, line 17: We don't have to manage resource in plugin.

dannypsnl:
update, current error message is

```
[2025-03-25 17:03:05.129] [error] component: ``, access function instance: 0/0
[2025-03-25 17:03:05.129] [error] while exporting component function `handle`.
```

This is because an export section of a component also happened before instantiation, and it is very funny that hello-wasi-http example has an export that uses an imported function in the final component.

dannypsnl:
Sort of,
1. need to fix build on different platforms
2. there are some concepts need to be carefully re-understood and rewrote, but they can be another PR

plugins/wasi_cli/base.h
```diff
@@ -0,0 +1,25 @@
...
+#include "env.h"
+
+#include "common/errcode.h"
+#include "runtime/hostfunc.h"
+
+namespace WasmEdge {
+namespace Host {
+
+template <typename T>
+class WasiCli : public Runtime::Component::HostFunction<T> {
```
q82419, line 15: Comment:
Maybe add a namespace `Preview2`.
Since there will be preview 3 in the future.

plugins/wasi_cli/env.h
```diff
@@ -0,0 +1,16 @@
...
+
+#include "plugin/plugin.h"
+
+#include <cstdint>
+#include <vector>
+
+namespace WasmEdge {
+namespace Host {
+
+class WasiCliEnvironment {};
```
q82419, line 13: If there's no necessary shared data, the environment class is not needed.

lib/runtime/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: 2019-2024 Second State INC
+
+wasmedge_add_library(wasmedgeRuntime
```
q82419, line 4: We supposed that `wasmedgeRuntime` to be a header-only library.
Would you consider this?

q82419 (commented):
I've viewed the plugins and runtime.

1. Seems like the preview 2 functions are empty.
2. Seems like the `wasi-http` only added the http-types. The host functions are not connected.
3. Please consider the `wasmedgeRuntime` to be a header-only library.

plugins/wasi_cli/base.h
dannypsnl: Agree

dannypsnl:
> 2. Seems like the `wasi-http` only added the http-types. The host functions are not connected.

Because wasi-http program should provide callback, which use types in wasi-http.

q82419:
I'll squash and rebase the commits to the branch `proposal/component`.
Due to some unknown issues that causes CI fail, I'll close this PR after the codes ready on that branch.

q82419:
Comment:

I've squashed this PR and rebased onto the branches `proposal/component` and `proposal/wasi_preview2`.
I'll keep extract the partitions of that branches when refactoring.
Keep this PR for code tracing.

q82419:
@dannypsnl 

Can you help to explain why the [code here](https://github.com/WasmEdge/WasmEdge/blob/master/lib/loader/ast/component/component_type.cpp#L494-L500):

The `Loader::loadExportDecl` will load the `core:exportdecl`, which is:
```
core:exportdecl  ::= n:<core:name> d:<core:importdesc>                    => (export n d)
```
Where the `core:name` is a utf-8 bytes vector.

But the code above use the [`Loader::loadExportName`](https://github.com/WasmEdge/WasmEdge/blob/master/lib/loader/ast/component/component_type.cpp#L367-L380), seems like it has a `0x00` prefix byte.

dannypsnl:
> @dannypsnl
> 
> Can you help to explain why the [code here](https://github.com/WasmEdge/WasmEdge/blob/master/lib/loader/ast/component/component_type.cpp#L494-L500):
> 
> The `Loader::loadExportDecl` will load the `core:exportdecl`, which is:
> 
> ```
> core:exportdecl  ::= n:<core:name> d:<core:importdesc>                    => (export n d)
> ```
> 
> Where the `core:name` is a utf-8 bytes vector.
> 
> But the code above use the [`Loader::loadExportName`](https://github.com/WasmEdge/WasmEdge/blob/master/lib/loader/ast/component/component_type.cpp#L367-L380), seems like it has a `0x00` prefix byte.

I don't remember, maybe due to proposal update the binary.md recently?

q82419:
@dannypsnl 

You implemented a new `AST::Component::FunctionType` for the component function and host function.
With lifting/lowering to `Component::FuncType`, as the [binary definition](https://github.com/WebAssembly/component-model/blob/main/design/mvp/Binary.md#type-definitions), the functype is:

```
functype      ::= 0x40 ps:<paramlist> rs:<resultlist>     => (func ps rs)
paramlist     ::= lt*:vec(<labelvaltype>)                 => (param lt)*
resultlist    ::= 0x00 t:<valtype>                        => (result t)
                | 0x01 0x00
```

Seems like it only accepts at most one result value.
But in [loader](https://github.com/WasmEdge/WasmEdge/blob/master/lib/loader/ast/component/component_type.cpp#L264-L268), you loaded the value type vector in `0x01` case of result list.
Is it the wrong implementation?

q82419:
> @dannypsnl
> 
> You implemented a new `AST::Component::FunctionType` for the component function and host function. With lifting/lowering to `Component::FuncType`, as the [binary definition](https://github.com/WebAssembly/component-model/blob/main/design/mvp/Binary.md#type-definitions), the functype is:
> 
> ```
> functype      ::= 0x40 ps:<paramlist> rs:<resultlist>     => (func ps rs)
> paramlist     ::= lt*:vec(<labelvaltype>)                 => (param lt)*
> resultlist    ::= 0x00 t:<valtype>                        => (result t)
>                 | 0x01 0x00
> ```
> 
> Seems like it only accepts at most one result value. But in [loader](https://github.com/WasmEdge/WasmEdge/blob/master/lib/loader/ast/component/component_type.cpp#L264-L268), you loaded the value type vector in `0x01` case of result list. Is it the wrong implementation?

Update:

https://github.com/WebAssembly/component-model/commit/cdad9c390b1e9e6f88421e495b3169c95641eb61

Seems like changed last year. Considering reserve the multi-return implementation.

q82419:
Due to far away from current component model implementation, close this.

Issue #3949, opened 2025-01-04
bug: Missing Subtyping Depth Validation


Q1IQ:
### Summary

WasmEdge does not enforce the subtyping depth limit when compiling WebAssembly modules, which leads to successful execution of modules that should otherwise fail validation due to excessive subtyping depth. This behavior is inconsistent with other engines like V8 and WebKit, which enforce a subtyping depth limit of 64.

### Current State

Run the following WebAssembly module:  
[program.wasm.txt](https://github.com/user-attachments/files/18305948/program.wasm.txt)

Observed output:
```
$ /path/to/wasmedge run --enable-all /path/to/test/program.wasm main
[2025-01-04 14:23:15.820] [warning] GC proposal is enabled, this is experimental.
[2025-01-04 14:23:15.856] [warning] component model is enabled, this is experimental.
1
```

### Expected State

The module should fail validation due to the subtyping depth exceeding 64, aligning with the behavior observed in V8 and WebKit. An appropriate error message should indicate that the type depth exceeds the allowed limit, similar to:
```
subtype depth for Type section's 64th signature exceeded the limits of 63
```

### Reproduction steps

1. Download [wasmedge-0.14.1-release](https://github.com/WasmEdge/WasmEdge/releases/download/0.14.1/WasmEdge-0.14.1-ubuntu20.04_x86_64.tar.gz).
2. Run the command:
```bash
./wasmedge run --enable-all ./program.wasm main
```

### Screenshots

![DESCRIPTION](LINK.png)

### Components

CLI

### WasmEdge Version or Commit you used

0.14.1

### Operating system information

Ubuntu 20.04

### Hardware Architecture

amd64

hydai:
@q82419  
There is a change in [the spec](https://github.com/WebAssembly/gc/pull/360/files) that indicates we should have some limitations for the GC proposal. I think we should also impose the same limitation to comply with the spec.

harshita9104:
Hello @hydai ,

I hope this message finds you well!
This issue interests me , could you please assign it to me !
Thank you!

hydai:
Hi @harshita9104,  
Since you are leaving multiple comments under various issues, I will only assign one of them to you. Please let me know which issue you are going to work on.

DeeBi9:
@hydai Can I pick this up!

hydai:
Hi @DeeBi9,  
[...]

PR #4054, opened 2025-03-11
fix: enforce subtyping depth limit in WebAssembly module validation

deveshidwivedi:
This PR enforces the subtyping depth limit of 64 (as per the WebAssembly specification) during the validation of WebAssembly modules. Previously, WasmEdge did not enforce this limit, allowing modules with excessive subtyping depth to pass validation.
Changes:
- Added a maximum subtyping depth constant (`MAX_SUBTYPE_DEPTH` = 63).
- Implemented recursive subtype depth calculation with cycle detection.
- Added a validation check in the type validator to enforce depth constraints and prevent invalid subtyping structures.
Fixes: [#3949](https://github.com/WasmEdge/WasmEdge/issues/3949)


juntao:

hydai:
Hi @deveshidwivedi,  
Please also add a test case for this bug fix to ensure that it resolves the mentioned issue.

core-burner:
I had done this on my branch. Adding error type instead of using InvalidSubType would be more coherent for the user. It's just 2 lines 0x228.

lib/validator/validator.cpp
```diff
@@ -850,5 +861,27 @@ Expect<void> Validator::validateConstExpr(AST::InstrView Instrs,
...
+  }
+
+  Visited.insert(TypeIdx);
+  uint32_t MaxDepth = 0;
+
+  const auto &TypeVec = Checker.getTypes();
+  const auto &Type = *TypeVec[TypeIdx];
+
+  for (auto SuperIdx : Type.getSuperTypeIndices()) {
+    if (SuperIdx < TypeVec.size()) {
```
dannypsnl, line 876: If `SuperIdx >= TypeVec.size()`, does any error report needs to be made?

dannypsnl (changes requested):

lib/validator/validator.cpp
deveshidwivedi: Yes, it fails validation with `ErrCode::Value::InvalidSubType`

lib/validator/validator.cpp
```diff
@@ -850,5 +861,27 @@ Expect<void> Validator::validateConstExpr(AST::InstrView Instrs,
   return Checker.validate(Instrs, Returns);
 }
 
+uint32_t Validator::calculateSubtypeDepth(uint32_t TypeIdx, std::set<uint32_t>& Visited) const {
```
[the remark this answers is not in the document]
deveshidwivedi: updated

include/validator/validator.h
```diff
@@ -75,9 +76,13 @@ class Validator {
...
   static inline const uint32_t LIMIT_MEMORYTYPE = 1U << 16;
   /// Proposal configure
-  const Configure Conf;
+  const Configure &Conf;
   /// Formal checker
   FormChecker Checker;
+
+  static constexpr uint32_t MAX_SUBTYPE_DEPTH = 63;
+
+  uint32_t calculateSubtypeDepth(uint32_t TypeIdx, std::set<uint32_t>& Visited) const;
```
dannypsnl, line 85: The signature is mismatched, and `Depth` can be put to the final and with default value `0`, so caller has no need to know it's there.

include/validator/validator.h
```diff
@@ -75,9 +76,13 @@ class Validator {
 
   static inline const uint32_t LIMIT_MEMORYTYPE = 1U << 16;
   /// Proposal configure
-  const Configure Conf;
+  const Configure &Conf;
```
dannypsnl, line 79: Does this change is required? This will turn a member to a reference, two semantics are very different.

dannypsnl (changes requested):

include/validator/validator.h
deveshidwivedi: I initially changed to `const Configure &Conf;` thinking it would be a performance optimization to avoid copying the `Configure` object.  However, I realize that this changes the semantics, as `Conf` is now a reference rather than a copy.

include/validator/validator.h
deveshidwivedi: Sorry for missing it earlier, I’ve updated the function signature.

dannypsnl (approved):

include/validator/formchecker.h
```diff
@@ -54,7 +54,7 @@ class FormChecker {
   void addTag(const uint32_t TypeIdx);
 
   std::vector<VType> result() { return ValStack; }
-  auto &getTypes() { return Types; }
+  auto &getTypes() const { return Types; }
```
hydai, line 57: What's the reason for changing this function signature?

include/validator/validator.h
```diff
@@ -78,6 +78,10 @@ class Validator {
   const Configure Conf;
   /// Formal checker
   FormChecker Checker;
+
+  static constexpr uint32_t MAX_SUBTYPE_DEPTH = 63;
+
+  Expect<void> calculateSubtypeDepth(uint32_t TypeIdx, std::set<uint32_t>& Visited, uint32_t Depth = 0) const;
```
hydai, line 84: Do we need to have this private function inside the class?
What if define it inside the details namespace of the implementation. We only need to pass the reference of TypeIdx and Checker into this function.

lib/validator/validator.cpp
```diff
@@ -165,32 +165,38 @@ Expect<void> Validator::validate(const AST::SubType &Type) {
...
-  // In current version, the length of type index vector will be <= 1.
-  if (Type.getSuperTypeIndices().size() > 1) {
-    spdlog::error(ErrCode::Value::InvalidSubType);
-    spdlog::error("    Accepts 1 super type currently."sv);
-    return Unexpect(ErrCode::Value::InvalidSubType);
-  }
-  for (auto Index : Type.getSuperTypeIndices()) {
-    if (Index >= TypeVec.size()) {
+// Validate subtype depth when GC proposal is enabled
+if (Conf.hasProposal(Proposal::GC) && !Type.getSuperTypeIndices().empty()) {
```
hydai, line 169: You change the behavior when the GC is disabled. What's the expected behavior if we don't enable GC but the indices' size > 1

hydai (changes requested):

include/validator/formchecker.h
deveshidwivedi: This is because, the `calculateSubtypeDepth()` method is a `const` method and it calls `Checker.getTypes()`, the `getTypes()` method must also be const to be callable from within a const context.

`auto &getTypes() { return Types; }` couldn't be called from the `calculateSubtypeDepth()` method.

dannypsnl: I think the point is others place that invokes `getTypes()`, we would like to know their behaviour didn't get changed implicitly.

include/validator/validator.h
deveshidwivedi: I understand, defined it inside the details namespace instead.

lib/validator/validator.cpp
deveshidwivedi: Updated to log an error, missed it in the previous implementation

deveshidwivedi:
Changes made:
- Moved `calculateSubtypeDepth` function to `details` namespace  
- Added error logging to handle cases where subtypes with multiple supertypes are invalid when GC is disabled
- Removed `const` from `FormChecker` to ensure expected behavior

Please let me know if the structure of code or any other aspect needs adjustment.

include/validator/validator.h
```diff
@@ -80,5 +80,9 @@ class Validator {
   FormChecker Checker;
 };
 
+namespace details {
+Expect<void> calculateSubtypeDepth(uint32_t TypeIdx, FormChecker &Checker);
+} // namespace details
+
 } // namespace Validator
-} // namespace WasmEdge
+} // namespace WasmEdge
```
dannypsnl, line 88: EOF

dannypsnl (dismissed):

include/validator/validator.h
```diff
@@ -80,5 +80,9 @@ class Validator {
   FormChecker Checker;
 };
 
+namespace details {
+Expect<void> calculateSubtypeDepth(uint32_t TypeIdx, FormChecker &Checker);
+} // namespace details
```
dannypsnl, line 85: This namespace no need to show in header actually.

deveshidwivedi: When I ran the tests it said:
```code
/home/deveshi1/wasmedge/WasmEdge/lib/validator/validator.cpp:8:1: note: ‘std::set’ is defined in header ‘<set>’; did you forget to ‘#include <set>’?
````

Comment (deveshidwivedi) [reply include/validator/validator.h:85]: Got it :)

File: include/validator/validator.h [hunk 10]
```diff
@@ -81,4 +81,4 @@ class Validator {
 };
 
 } // namespace Validator
-} // namespace WasmEdge
+} // namespace WasmEdge
```
dannypsnl, line 84: EOF still there, maybe you forget to push the commit?

dannypsnl (dismissed):
The big picture LGTM.

include/validator/validator.h
deveshidwivedi: Apologies, updated now

hydai (approved):

deveshidwivedi:
Hi @hydai @dannypsnl, I'm having trouble finding the right file to add tests for `validator.cpp`. Could you point me in the right direction?

hydai:
> Hi @hydai @dannypsnl, I'm having trouble finding the right file to add tests for `validator.cpp`. Could you point me in the right direction?

Hi @deveshidwivedi 
Due to the life time of the configuration and ASTModule, it will be easier to have this unit test inside the test/executor/executor.cpp.

dannypsnl:
@deveshidwivedi Had you run the tests locally? It seems not worked.

deveshidwivedi:
> @deveshidwivedi Had you run the tests locally? It seems not worked.

Yes I did run the tests locally, I'm trying to fix these currently.

hydai:
Hi @deveshidwivedi 
Are you going to work on this PR? If not, I will close it, thanks.

q82419:
> Hi @deveshidwivedi Are you going to work on this PR? If not, I will close it, thanks.

@hydai As you mentioned the [spec](https://github.com/WebAssembly/gc/pull/360/files) in issue #3949 ,
This spec describes the limitation values of instance entries or sizes in V8.
The core spec allows runtimes to define their limitations, so this may not be considered as a bug.

hydai:
> @hydai As you mentioned the [spec](https://github.com/WebAssembly/gc/pull/360/files) in issue #3949 , This spec describes the limitation values of instance entries or sizes in V8. The core spec allows runtimes to define their limitations, so this may not be considered as a bug.

V8 and WebKit implement the same limitation. I think we should have either this limitation or a warning so that people can recognize when the program is stuck due to the subtyping depth issue.

PR #4098, opened 2025-05-05
fix: deterministic NaN propagation in f64.max (prefer first NaN)

### LINKED ISSUE #4057: bug: WasmEdge gives different output compared to other runtimes
### Summary

I have generated a wasm binary as a test case which I ran with wasmtime, wasmer, and wasmedge runtimes. The wasmedge runtime is producing a different output compared to the other runtimes.

### Current State

Output of wasmedge:
```340279770772509196068006945519824797696```

### Expected State

Output of wasmtime and wasmer runtimes:
```340282365653287863253591949940055080960```

### Reproduction steps

Test Case:

```
(module
  (func (;0;) (export "main") (result v128)
    f64.const -nan:0xffff000000000 (;=-nan;)
    f64.const -nan:0xf800000000000 (;=-nan;)
    f64.max
    f64x2.splat
  )
)
```
 
1. Run `wat2wasm file.wat`
2. Run `wasmedge --reactor file.wasm main`

### Screenshots

_No response_

_No response_

### Components

CLI

wasmedge version 0.14.1-271-ga89c176f

Linux

### Hardware Architecture

x86_64

I have only used the default options.

### PR DESCRIPTION
**Pull request by koliangyu99**
fix: deterministic NaN propagation in f64.max (prefer first NaN)
Closes #4057

### Summary
This PR makes WasmEdge’s f64.max deterministic for NaN inputs by aligning interpreter behavior with AOT and Wasmtime:

- Both operands NaN → return the first NaN

- One operand NaN → return the non‑NaN value

### Background
Different runtimes pick different NaN payloads when f64.max sees NaNs. Wasmtime chooses the first NaN; Wasmer and AOT mode in WasmEdge chose the second. Interpreter mode in WasmEdge also picked the second, causing inconsistency.

### Changes
In include/executor/engine/binary_numeric.ipp (line 191):
### From:
```
  if (std::isnan(Z1) || std::isnan(Z2)) {
    if (std::isnan(Z2)) {
      Z1 = Z2;
    }
    // Set the most significant bit of the payload to 1.
    if constexpr (sizeof(T) == sizeof(uint32_t)) {
```

### To:
```
  if (std::isnan(Z1) && std::isnan(Z2)) {
  } else if (std::isnan(Z1)) {
    Z1 = Z2;
    return {};
  } else if (std::isnan(Z2)) {
    return {};
  }
  if (std::isnan(Z1) || std::isnan(Z2)) {
    // Set the most significant bit of the payload to 1.
    if constexpr (sizeof(T) == sizeof(uint32_t)) {
```

### 1. Both‑NaN case (nan_max.wat)

```
(module
  (func (export "main") (result i64)
    ;; first NaN
    f64.const -nan:0xffff000000000
    ;; second NaN
    f64.const -nan:0xf800000000000
    f64.max
    ;; reinterpret the f64 result bits as i64
    i64.reinterpret_f64
  )
)
```
### Outout (nan_max.wat):
```
wat2wasm local_test/nan_max.wat -o nan_max.wasm
./tools/wasmedge/wasmedge --reactor ../local_test/nan_max.wasm main \ | xargs printf '0x%016x\n'# => 0xfffffff000000000
```

### 2. One‑NaN case (one_nan.wat)
```
(module
  (func (export "main") (result i64)
    ;; First: NaN, Second: 2.0
    f64.const -nan:0xffff000000000
    f64.const 2.0
    f64.max
    ;; reinterpret the result bits to i64
    i64.reinterpret_f64
  )
)
```
### Outout (one_nan.wat):
```
wat2wasm local_test/one_nan.wat -o one_nan.wasm
./tools/wasmedge/wasmedge --reactor ../local_test/one_nan.wasm main \ | xargs printf '0x%016x\n'
# => 0x4000000000000000
```
### Follow‑Up:
I’ll add the new spec tests (float/nan_max.wat and float/one_nan.wat) in a follow‑up PR once this lands.

### DIFF
--- a/include/executor/engine/binary_numeric.ipp
+++ b/include/executor/engine/binary_numeric.ipp
@@ -188,10 +188,14 @@ TypeF<T> Executor::runMaxOp(ValVariant &Val1, const ValVariant &Val2) const {
   T &Z1 = Val1.get<T>();
   const T &Z2 = Val2.get<T>();
   const T kZero = 0.0;
+  if (std::isnan(Z1) && std::isnan(Z2)) {
+  } else if (std::isnan(Z1)) {
+    Z1 = Z2;
+    return {};
+  } else if (std::isnan(Z2)) {
+    return {};
+  }
   if (std::isnan(Z1) || std::isnan(Z2)) {
-    if (std::isnan(Z2)) {
-      Z1 = Z2;
-    }
     // Set the most significant bit of the payload to 1.
     if constexpr (sizeof(T) == sizeof(uint32_t)) {
       uint32_t I32;
--- a/include/executor/engine/binary_numeric.ipp
+++ b/include/executor/engine/binary_numeric.ipp
@@ -188,14 +188,10 @@ TypeF<T> Executor::runMaxOp(ValVariant &Val1, const ValVariant &Val2) const {
   T &Z1 = Val1.get<T>();
   const T &Z2 = Val2.get<T>();
   const T kZero = 0.0;
-  if (std::isnan(Z1) && std::isnan(Z2)) {
-  } else if (std::isnan(Z1)) {
-    Z1 = Z2;
-    return {};
-  } else if (std::isnan(Z2)) {
-    return {};
-  }
   if (std::isnan(Z1) || std::isnan(Z2)) {
+    if (!std::isnan(Z2)) {
+      Z1 = Z2;
+    }
     // Set the most significant bit of the payload to 1.
     if constexpr (sizeof(T) == sizeof(uint32_t)) {
       uint32_t I32;

### CODE REVIEW

**Comment by juntao** [comment]

File: include/executor/engine/binary_numeric.ipp [hunk 1]
```diff
@@ -188,10 +188,14 @@ TypeF<T> Executor::runMaxOp(ValVariant &Val1, const ValVariant &Val2) const {
   T &Z1 = Val1.get<T>();
   const T &Z2 = Val2.get<T>();
   const T kZero = 0.0;
+  if (std::isnan(Z1) && std::isnan(Z2)) {
+  } else if (std::isnan(Z1)) {
+    Z1 = Z2;
+    return {};
+  } else if (std::isnan(Z2)) {
+    return {};
+  }
```
q82419, line 197: I think the following is better:

```
  if (std::isnan(Z1) || std::isnan(Z2)) {
    if (!std::isnan(Z2)) {
      Z1 = Z2;
    }
    if (std::isnan(Z1)) {
      // Set the most significant bit of the payload to 1.
      if constexpr (sizeof(T) == sizeof(uint32_t)) {
        uint32_t I32;
        std::memcpy(&I32, &Z1, sizeof(T));
        I32 |= static_cast<uint32_t>(0x01U) << 22;
        std::memcpy(&Z1, &I32, sizeof(T));
      } else if constexpr (sizeof(T) == sizeof(uint64_t)) {
        uint64_t I64;
        std::memcpy(&I64, &Z1, sizeof(T));
        I64 |= static_cast<uint64_t>(0x01U) << 51;
        std::memcpy(&Z1, &I64, sizeof(T));
      }
    }
} else ...
```

The above code can quickly sellect the non-NAN one and set the most significant bit of the payload for the both NAN case.
This doesn't need another if-statement.

q82419 (changes requested):

PeterD1524:
I think it should be:
One operand NaN → return the NaN value

koliangyu99:
Thanks @q82419 for the suggestion. I’ve updated the NaN‑handling block to use a single if (isnan).

koliangyu99:
Thanks for the feedback, @PeterD1524! I found the definition from [link](https://webassembly.github.io/spec/core/exec/numerics.html?highlight=numerics#)

it indicate: If either z₁ or z₂ is a NaN, then return an element of nansₙ{z₁, z₂}.
That clause covers both “one NaN” and “two NaNs” cases. When exactly one input is NaN, the set nansₙ{z₁, z₂} contains only that one NaN, so fmax must return that NaN. So you are right. Thanks for pointing it out.

image

q82419:
> Thanks for the feedback, @PeterD1524! I found the definition from [link](https://webassembly.github.io/spec/core/exec/numerics.html?highlight=numerics#)
> 
> it indicate: If either z₁ or z₂ is a NaN, then return an element of nansₙ{z₁, z₂}. That clause covers both “one NaN” and “two NaNs” cases. When exactly one input is NaN, the set nansₙ{z₁, z₂} contains only that one NaN, so fmax must return that NaN. So you are right. Thanks for pointing it out.
> 
> image

As above, your expression is still not correct.
I think it should be:

```
if (std::isnan(Z1) || std::isnan(Z2)) {
  if (!std::isnan(Z1)) {
    Z1 = Z2;
  }
  ...
}
```

Therefore the Z1 will be selected if Z1 is NAN (no matter Z2 is NAN nor not), and Z2 will be selected if Z1 is not NAN.

koliangyu99:
Thanks for the review, @q82419! I’ve force-pushed the commit that flips the inner check exactly as you suggested. So now we return Z1 when it’s NaN, and Z2 when Z1 isn’t. Let me know if this matches the spec or if there’s anything else to adjust.

q82419:
> Thanks for the review, @q82419! I’ve force-pushed the commit that flips the inner check exactly as you suggested. So now we return Z1 when it’s NaN, and Z2 when Z1 isn’t. Let me know if this matches the spec or if there’s anything else to adjust.

Sure. You can follow the rule: If either z₁ or z₂ is a NaN, then return an element of nansₙ{z₁, z₂}.
Therefore, if both z₁ and z₂ are NaN, return z₁; otherwize, return the NaN one.

Suppose the if statement:
```
if (std::isnan(Z1) || std::isnan(Z2)) {
  if (!std::isnan(Z1)) {
    Z1 = Z2;
  }
  ...
}
```

If the `std::isnan(Z1) || std::isnan(Z2)` is true, it will enter into the statement, and in this case the z₁ or z₂ or both is NaN.
Therefore, if z₁ is not NaN, then z₂ must be NaN, return z₂; if z₁ is NaN, return z₁ no matter z₂ is NaN or not.
So we can simply check that if z₁ is NaN or not here: if z₁ is not NaN, copy the z₂ value to z₁ and set the most significant bit.

q82419:
@koliangyu99 

Since currently your code is not correct, please let us know if you want to update your changes, or we may quickly fix with our commit. Thanks!

PR #4133, opened 2025-06-02
feat(WASI-NN,ggml): bump llama.cpp b5575

hydai:
There are lots of breaking changes, please refer to the upstream commit: https://github.com/ggml-org/llama.cpp/pull/13866/files


juntao:

------

## [CMakeLists.txt](https://github.com/WasmEdge/WasmEdge/blob/5e4f7240ca7b7acb67dca9183629a5503040b820/CMakeLists.txt)

#### Potential issues

1. **Version Handling Issue**: The version handling logic using `GIT_VERSION_NOT_FOUND` and its comparison with 0 can lead to incorrect behavior if `execute_process` fails but does not return a numeric result, leading to potential incorrect package versioning.

2. **Redundant Plugin Option for WASI-NN Metal Backend**: The option `WASMEDGE_PLUGIN_WASI_NN_GGML_LLAMA_METAL` is incorrectly named and seems like it should be related to the GGML backend for the WASI-NN plugin, but it is labeled as LLAMA, which might cause confusion about its functionality.

3. **Potential Issue with LLVM Static Linking**: The logic that forces static linking of LLVM when `WASMEDGE_LINK_TOOLS_STATIC` is enabled could lead to compatibility issues on systems where LLVM's static libraries are not available or are incompatible with the current build environment.

#### Summary of changes

-**Key changes:**
  - Bumped WASI-NN library version to `0.1.22` and set its soversion to `0`.
  
- **Most important changes:**
  - Updated the version numbers for WASI-NN libraries.
  - Changed the `wasi_edge` dependency in the WasmEdge C API project.

## [cmake/WASINNDeps.cmake](https://github.com/WasmEdge/WasmEdge/blob/5e4f7240ca7b7acb67dca9183629a5503040b820/cmake%2FWASINNDeps.cmake)

#### Potential issues

1. **Unconditional `elseif()` Block**: The `elseif(BACKEND STREQUAL "tensorflowlite")` block has an unconditional `elseif()` which will always cause a syntax error or logic issue.
   ```cmake
   elseif()
     message(FATAL_ERROR "Unsupported architecture: ${CMAKE_SYSTEM_PROCESSOR}")
   endif()
   ```

2. **Missing CUDA Detection for `ggml-cuda`**: The CUDA detection in the `wasmedge_setup_llama_target` function does not verify if CUDA is available before setting `GGML_CUDA ON`. This could lead to build failures if CUDA is not installed.
   ```cmake
   if(WASMEDGE_PLUGIN_WASI_NN_GGML_LLAMA_CUBLAS)
     message(STATUS "WASI-NN GGML LLAMA backend: Enable GGML_CUDA")
     set(GGML_CUDA ON)
     # We need to set GGML_USE_CUDA for clip from llava.
     add_compile_definitions(GGML_USE_CUDA)
   else()
     message(STATUS "WASI-NN GGML LLAMA backend: Disable GGML_CUDA")
     set(GGML_CUDA OFF)
   endif()
   ```

3. **Hardcoded Version in TensorFlow Setup**: The version of TensorFlow is hardcoded multiple times, leading to potential inconsistencies if the version needs updating.
   ```cmake
   set(WASMEDGE_TENSORFLOW_DEPS_TF_HASH "266465acd642a9d2d80e56c93aa0a255597bfb3034a826bb2225e61f2bebe2e2" PARENT_SCOPE)
   ```
   This could be refactored to a single location or use version variables.

#### Summary of changes

-`GIT_TAG` for `llama.cpp` has been updated from `b5463` to `b5575`.

## [lib/wasi_nn_rpc/CMakeLists.txt](https://github.com/WasmEdge/WasmEdge/blob/5e4f7240ca7b7acb67dca9183629a5503040b820/lib%2Fwasi_nn_rpc%2FCMakeLists.txt)

#### Potential issues

1. **Versioning Issue**: The `VERSION` and `SOVERSION` properties for the target `wasiNNRPC` are set using cache variables (`WASMEDGE_WASI_NN_RPC_VERSION` and `WASMEDGE_WASI_NN_RPC_SOVERSION`), which are not defined in the provided code snippet. This will likely cause a build error.

2. **Redundant Protoc Commands**: The `execute_process` command runs `protoc` twice for generating C++ and gRPC code separately. While this is generally acceptable, it could be optimized by combining these into a single call to reduce build time and maintainability:
   ```cmake
   execute_process(
     COMMAND protoc --proto_path=${CMAKE_CURRENT_SOURCE_DIR} --cpp_out=${CMAKE_CURRENT_BINARY_DIR} --grpc_out=${CMAKE_CURRENT_BINARY_DIR} --plugin=protoc-gen-grpc=${GRPC_CPP_PLUGIN} ${CMAKE_CURRENT_SOURCE_DIR}/wasi_ephemeral_nn.proto
   )
   ```

3. **Hardcoded Compiler Flags**: The use of `${PROTOBUF_CFLAGS}` and `${GRPCPP_CFLAGS}` in `target_compile_options` suggests these are intended to be compiler flags, but they are not defined within the snippet. This could lead to undefined behavior or build errors if these variables are empty or incorrectly set. Consider using CMake's native target properties for such configurations.

#### Summary of changes

- - Added versioning variables `WASMEDGE_WASI_NN_VERSION` and `WASMEDGE_WASI_NN_SOVERSION`.
- Updated target properties to use the new versioning variables.
- Corrected variable names in `set_target_properties`.

## [plugins/wasi_nn/CMakeLists.txt](https://github.com/WasmEdge/WasmEdge/blob/5e4f7240ca7b7acb67dca9183629a5503040b820/plugins%2Fwasi_nn%2FCMakeLists.txt)

#### Potential issues

1. **Issue 1:** The `string(TOLOWER ${BACKEND} BACKEND)` line modifies the `BACKEND` variable in-place, which can lead to incorrect logic if the original case of `BACKEND` is needed later in the CMake script.
   - **Explanation:** Modifying loop variables inside the loop can cause unexpected behavior and make the code harder to debug.

2. **Issue 2:** The `target_sources(wasmedgePluginWasiNN PRIVATE ...)` block for the "mlx" backend does not check if all the listed files actually exist before adding them, which could lead to build failures.
   - **Explanation:** Adding non-existent source files results in build errors that are difficult to diagnose without proper checks.

3. **Issue 3:** The `if(WASMEDGE_BUILD_WASI_NN_RPC)` block conditionally adds include directories and link libraries but does not handle the case where these options might be disabled, potentially leading to incomplete dependencies.
   - **Explanation:** Missing handling for when `WASMEDGE_BUILD_WASI_NN_RPC` is false can result in missing necessary includes and links, causing runtime or compile-time issues.

#### Summary of changes

- - Introduced `WASMEDGE_WASI_NN_VERSION` with the value "0.1.23" to specify the library version.
- Introduced `WASMEDGE_WASI_NN_SOVERSION` with the value "0" to set the shared object version.
- Updated the handling of `WASI_NN_VERSION_LIST` and extracted `WASI_NN_VERSION_MAJOR`.

## [plugins/wasi_nn/wasinn_ggml.cpp](https://github.com/WasmEdge/WasmEdge/blob/5e4f7240ca7b7acb67dca9183629a5503040b820/plugins%2Fwasi_nn%2Fwasinn_ggml.cpp)

#### Potential issues

N/A

#### Summary of changes

- - Added `json-partial.h` and `mtmd-helper.h` to the include list.
- Modified `mtmd_helper_bitmap_init_from_buf` and `mtmd_helper_bitmap_init_from_file` calls to include a `GraphRef.VisionContext.get()` parameter.
- Ensured error handling remains consistent for invalid image bitmap creation.

dm4 (dismissed):

dm4 (approved):

PR #4134, opened 2025-06-02
fix(validator): enforce subtyping depth limit during validation

PhantomInTheWire:
### Enforce Subtyping Depth Limit in WebAssembly Validation

This PR resolves [#3949](https://github.com/wasmedge/wasmedge/issues/3949), which identified that WasmEdge incorrectly allows execution of WebAssembly modules with excessive sub-typing depth, creating behavior inconsistencies with V8 and WebKit implementations.

#### Summary

Implements enforcement of the **maximum subtyping depth limit of 63** as specified in the WebAssembly GC proposal. The validator now rejects any module containing type hierarchies that exceed this depth during validation.

**Specification Reference:**
> The maximum depth of a defined subtype hierarchy is 63 (where a type defined with no supertype has depth 0).

*Source: https://github.com/WebAssembly/gc/blob/main/document/js-api/index.bs#L1443*


juntao:

------

## [include/validator/formchecker.h](https://github.com/WasmEdge/WasmEdge/blob/623d7741345714886b0fa600a6a28871206f91ca/include%2Fvalidator%2Fformchecker.h)

#### Potential issues

1. **Potential Uninitialized Member Issue**: The `FormChecker` class has a member variable `Mems` of type `uint32_t` which is initialized directly in the constructor initializer list, but it might not be set to zero if the default constructor is used.
    * The use of `= 0` in the default constructor is unnecessary and redundant.
2. **Unused Include**: The header file includes `<cstdint>`, but there are no occurrences of `uint32_t` or other types from this header being used directly; they might be used indirectly through AST or common headers, so it's not strictly an unused include but worth mentioning for optimization.

3. **Potential Const-Correctness Issue**: The member functions `getTypes()`, `getFunctions()`, `getTables()`, and `getMemories()` return non-const references to private members, which could allow modification of these internal states unintentionally if not handled correctly.
    * These should likely be marked as `const` to prevent accidental modifications.

#### Summary of changes

1. The `getTypes()` method in the `FormChecker` class was modified to add the `const` qualifier, ensuring that calling this method does not alter the state of the object.
2. This change enforces immutability for the types retrieved from the `FormChecker`, making the code more robust and preventing unintended modifications.
3. The patch aims to maintain a consistent and safe interface by clearly indicating that accessing the types should be a read-only operation.

## [lib/validator/validator.cpp](https://github.com/WasmEdge/WasmEdge/blob/623d7741345714886b0fa600a6a28871206f91ca/lib%2Fvalidator%2Fvalidator.cpp)

#### Potential issues

**Major Issues:**

1. **Potential Stack Overflow:** The recursive function `calculateSubtypeDepthRecursiveHelper` does not have a base case that returns immediately, which could lead to stack overflow for deep subtype hierarchies.

2. **Unhandled Exceptions:** The code does not handle exceptions thrown by `spdlog::error` or other functions that might throw, potentially leading to undefined behavior if an exception is raised.

3. **Uninitialized Checker State:** The `Checker.reset(true);` call in the `validate` function initializes the checker state, but there's no guarantee that this state will be properly managed for all validation paths, which could lead to inconsistent states or errors during validation.

#### Summary of changes

- **Enforced Subtype Depth Limit**: Added a limit of 63 on subtype depth during validation.
- **Cycle Detection in Subtype Hierarchy**: Introduced cycle detection to prevent infinite recursion in subtype hierarchy checks.
- **Refactored Validation Process**: Modified the `validate` function for subtypes to include subtype depth checking.

## [test/CMakeLists.txt](https://github.com/WasmEdge/WasmEdge/blob/623d7741345714886b0fa600a6a28871206f91ca/test%2FCMakeLists.txt)

#### Potential issues

1. The `FetchContent_Declare` block lacks error handling for the case when fetching content fails.
2. The `add_subdirectory` calls assume that all subdirectories exist and are valid, which could cause issues if any are missing or improperly configured.
3. The use of `${PROCESSOR_COUNT}` in the coverage setup might not be defined, potentially leading to errors during build configuration.

#### Summary of changes

1. **Addition of Validator Subdirectory**: The most significant change is the addition of a new subdirectory "validator" using `add_subdirectory(validator)`. This indicates that a new component or module related to validation has been introduced in the project.

2. **Placement in CMakeLists.txt**: The new subdirectory is added after other core components like 'common', 'spec', 'loader', and 'executor' but before 'component' and 'thread'. This suggests that the validator might be a crucial part of the build process, closely related to these core modules.

3. **No Other Changes in File**: The patch shows only one line added with no other modifications or deletions, implying that this is a targeted addition for introducing the validator component.

## [test/validator/CMakeLists.txt](https://github.com/WasmEdge/WasmEdge/blob/623d7741345714886b0fa600a6a28871206f91ca/test%2Fvalidator%2FCMakeLists.txt)

#### Potential issues

1. **CMake Indentation Issue**: The `wasmedge_add_executable` and `target_link_libraries` commands are not properly aligned, which can lead to confusion in the build process.
2. **Missing Dependencies Check**: There is no check to ensure all dependencies (e.g., `wasmedgeTestSpec`, `wasmedgeVM`) exist before linking them.
3. **Hardcoded Test Name**: The test name `wasmedgeValidatorSubtypeTests` appears multiple times, which can be error-prone if changes are needed in the future.

#### Summary of changes

- Added a new test executable `wasmedgeValidatorSubtypeTests` targeting validation subtype tests.
- Linked the test with necessary libraries: `wasmedgeTestSpec` and `wasmedgeVM`.
- Specified Apache 2.0 license for the new file and added copyright information for Second State INC from 2019 to 2024.

## [test/validator/ValidatorSubtypeTest.cpp](https://github.com/WasmEdge/WasmEdge/blob/623d7741345714886b0fa600a6a28871206f91ca/test%2Fvalidator%2FValidatorSubtypeTest.cpp)

#### Potential issues

1. **Potential Memory Leak**: The `WasmEdge::Loader` and `WasmEdge::Validator` objects are created using smart pointers but are not explicitly destroyed before the test cases end, which could lead to potential memory leaks.
2. **Magic Number Usage**: The code contains hardcoded byte sequences (magic numbers) like `0x00, 0x61, 0x73, 0x6d` without explanation or constants, making it less readable and maintainable.
3. **Inconsistent Error Handling**: Some test cases handle the validation result with `EXPECT_FALSE(ValidationResult)`, while others directly check `EXPECT_EQ`, leading to inconsistent error handling patterns.

#### Summary of changes

### Key Changes

- **Added `ValidatorSubtypeTest` Class:**
  - A new test class `ValidatorSubtypeTest` has been introduced to manage WasmEdge Validator's subtyping tests.

- **Helper Function for Encoding LEB128:**
  - Implemented a helper function `encodeLEB128` to encode unsigned 32-bit integers in the Little Endian Base 128 (LEB128) format.

- **Test Cases for Subtype Validation:**
  - Several test cases have been added:
    - **MaxSubtypeDepthExceeded**: Validates that exceeding the maximum subtype depth results in an error.
    - **InvalidTypeIndex** and **InvalidSupertypeIndex**: Checks validation with invalid indices in type definitions.
    - **ErrorPropagationRecursive** and **ErrorPropagationInTypeHierarchy**: Tests for recursive errors within the type hierarchy.
    - **ExactMaxSubtypeDepth**: Validates exactly at the maximum allowed subtype depth.

These changes focus on ensuring that the WasmEdge Validator correctly enforces subtyping limits during module validation, detecting and reporting invalid subtype structures.

dannypsnl:
@PhantomInTheWire can you rebase and remove merge commits?

PhantomInTheWire:
done

lib/validator/validator.cpp
```diff
@@ -165,32 +217,45 @@ Expect<void> Validator::validate(const AST::SubType &Type) {
...
-      return Unexpect(ErrCode::Value::InvalidSubType);
-    }
-    auto &SuperType = TypeVec[Index]->getCompositeType();
-    if (!AST::TypeMatcher::matchType(Checker.getTypes(), SuperType, CompType)) {
-      spdlog::error(ErrCode::Value::InvalidSubType);
-      spdlog::error("    Super type not matched."sv);
-      return Unexpect(ErrCode::Value::InvalidSubType);
+
+  // Validate subtype depth when GC proposal is enabled
+  if (Conf.hasProposal(Proposal::GC) && !Type.getSuperTypeIndices().empty()) {
```
dannypsnl: Maybe you can use 
```cpp
if (!Conf.hasProposal(...::GC)) {
  ...
} else {
  ...
}
```
so
1. you no need to repeat proposal condition
2. show with the proposal enable or not, we have different treatment

PhantomInTheWire: That makes sense. I'll refactor it.

core-burner:
It'd be really nice to add the correct error type instead of InvalidSubType if this is a final implementation.
It's just 3 lines.

PhantomInTheWire:
do you mean something like this @core-burner?

```cpp
@@ -48,14 +48,14 @@ Expect<void> calculateSubtypeDepth(uint32_t TypeIdx, FormChecker &Checker) {
     const auto &TypeVec = Checker.getTypes();
 
     if (Index >= TypeVec.size()) {
-      return Unexpect(ErrCode::Value::InvalidSubType);
+      return Unexpect(ErrCode::Value::InvalidFuncTypeIdx);
     }
 
     const auto &Type = *TypeVec[Index];
 
     for (const auto SuperIdx : Type.getSuperTypeIndices()) {
       if (SuperIdx >= TypeVec.size()) {
-        return Unexpect(ErrCode::Value::InvalidSubType);
+        return Unexpect(ErrCode::Value::InvalidFuncTypeIdx);
       }
 
       if (auto Res = Calculate(SuperIdx, Depth + 1, VisitedSet); !Res) {
@@ -231,11 +231,11 @@ Expect<void> Validator::validate(const AST::SubType &Type) {
     if (!Type.getSuperTypeIndices().empty()) {
       for (const auto &Index : Type.getSuperTypeIndices()) {
         if (unlikely(Index >= TypeVec.size())) {
-          spdlog::error(ErrCode::Value::InvalidSubType);
+          spdlog::error(ErrCode::Value::InvalidFuncTypeIdx);
           spdlog::error(
               "    Super type not found. Index {} is out of bounds for defined types (size: {})."sv,
               Index, TypeVec.size());
-          return Unexpect(ErrCode::Value::InvalidSubType);
+          return Unexpect(ErrCode::Value::InvalidFuncTypeIdx);
         }
 
         // Calculate and check subtype depth
@@ -252,9 +252,9 @@ Expect<void> Validator::validate(const AST::SubType &Type) {
         auto &SuperType = TypeVec[Index]->getCompositeType();
         if (!AST::TypeMatcher::matchType(Checker.getTypes(), SuperType,
                                          CompType)) {
-          spdlog::error(ErrCode::Value::InvalidSubType);
+          spdlog::error(ErrCode::Value::TypeCheckFailed);
           spdlog::error("    Super type not matched."sv);
-          return Unexpect(ErrCode::Value::InvalidSubType);
+          return Unexpect(ErrCode::Value::TypeCheckFailed);
         }
       }
     }
```

core-burner:
Uh it has been a while but it should be 2 parts. Anyway, glad you have done it and it looks great too.

If you wanted, you could:

add to enum.inc
```
// Sub type matching or validation failed
E(InvalidSubTypeDepthGT64, 0x0227, "type depth greater than 64")
// @}
```

add to validator.cpp

```
    const size_t Depth = calculateSubtypeDepth(Type);
    if (Depth > 64) {
      spdlog::error(ErrCode::Value::InvalidSubTypeDepthGT64);
      spdlog::error("Type depth {} exceeds limit of 64", Depth);
      spdlog::error(ErrInfo::InfoAST(ASTNodeAttr::Type_Sub));
      return Unexpect(ErrCode::Value::InvalidSubTypeDepthGT64);
    }
```

Instead of
TypeCheckFailed and InvalidSubType

Also validator.h could maybe use
Expect<void> calculateSubtypeDepth(uint32_t TypeIdx, FormChecker &Checker)

Is this guard really needed? Could use it with types other than GC?
(Conf.hasProposal(Proposal::GC))

Regardless, this solution is better than mine, I think I'll use it as well :D

PhantomInTheWire:
@dannypsnl im not sure if i should add specific error codes like this, especially since the spec doesn't have anything on the kind of error codes runtimes can have

dannypsnl:
I suggest don't add error code to avoid collision from others proposals, instead, you can print more to stderr.

PhantomInTheWire:
I agree, this could cause chaos, I have just added more debug statements.

lib/validator/validator.cpp
```diff
@@ -19,6 +20,66 @@ using namespace std::literals;
...
 namespace Validator {
 
+namespace details {
+static constexpr uint32_t MaxSubtypeDepth = 63;
+
+namespace {
+Expect<void> calculateSubtypeDepth(uint32_t TypeIdx, FormChecker &Checker) {
+  std::set<uint32_t> VisitedNodes;
+
+  std::function<Expect<void>(uint32_t, uint32_t, std::set<uint32_t> &)>
```
dannypsnl: Why not just extract out this function? Long lambda isn't good to read.

dannypsnl (changes requested):
Except this I'm ok with rest code, @hydai if test pass you can make a check.

lib/validator/validator.cpp
PhantomInTheWire: Makes sense, I've refactored it into a separate helper function.

lib/validator/validator.cpp
```diff
@@ -165,32 +229,51 @@ Expect<void> Validator::validate(const AST::SubType &Type) {
     }
   }
 
-  // In current version, the length of type index vector will be <= 1.
-  if (Type.getSuperTypeIndices().size() > 1) {
-    spdlog::error(ErrCode::Value::InvalidSubType);
-    spdlog::error("    Accepts 1 super type currently."sv);
-    return Unexpect(ErrCode::Value::InvalidSubType);
-  }
-  for (auto Index : Type.getSuperTypeIndices()) {
```
hydai, line 174: According to the original logic, this loop will still run if `Type.getSuperTypeIndices().size() == 1`. What is the reason for removing it entirely when disabling the GC proposal?

hydai (changes requested):

lib/validator/validator.cpp
PhantomInTheWire: I overlooked that nuance and unintentionally introduced a regression thanks for catching it. I’ll fix it and add targeted tests to ensure this case is properly covered so we don’t run into this again.

lib/validator/validator.cpp
```diff
@@ -165,32 +229,49 @@ Expect<void> Validator::validate(const AST::SubType &Type) {
...
-    if (Index >= TypeVec.size()) {
+
+  for (const auto &Index : Type.getSuperTypeIndices()) {
+    if (unlikely(Index >= TypeVec.size())) {
       spdlog::error(ErrCode::Value::InvalidSubType);
       spdlog::error(
-          ErrInfo::InfoForbidIndex(ErrInfo::IndexCategory::DefinedType, Index,
-                                   static_cast<uint32_t>(TypeVec.size())));
+          "    Super type not found. Index {} is out of bounds for defined types (size: {})."sv,
```
hydai: And you also changed the original error message.

lib/validator/validator.cpp
```diff
@@ -165,32 +229,49 @@ Expect<void> Validator::validate(const AST::SubType &Type) {
...
+          Index, TypeVec.size());
       return Unexpect(ErrCode::Value::InvalidSubType);
     }
+
+    // Conditional Check: Subtype depth is only validated with GC
+    if (Conf.hasProposal(Proposal::GC)) {
```
hydai: The logic is completely different between your previous version and the current one. Could you please check again if this is your intention?

hydai (changes requested):

lib/validator/validator.cpp
PhantomInTheWire: yes if i kept the if else block and two separate validation paths for gc and non-gc there would have been a lot of code duplication

lib/validator/validator.cpp
PhantomInTheWire: my bad, fixed it.

lib/validator/validator.cpp
hydai: I see.

test/executor/ExecutorTest.cpp
```diff
@@ -253,6 +253,311 @@ TEST(Coredump, generateCoredump) {
...
+    Leb.push_back(Byte);
+  } while (Value != 0);
+  return Leb;
+}
+
+// Generates a Wasm module with a chain of subtypes.
+// num_total_types: The total number of types to define. Type k+1 subtypes type
+// k. The maximum depth of the chain will be num_total_types - 1.
+static std::vector<WasmEdge::Byte>
+generateWasmWithSubtypeChain(int num_total_types) {
```
hydai: Please follow the naming style.

test/executor/ExecutorTest.cpp
```diff
@@ -253,6 +253,311 @@ TEST(Coredump, generateCoredump) {
...
+                                   FuncNamesSubSizeLeb.end());
+  NameSectionPayloadContent.insert(NameSectionPayloadContent.end(),
+                                   FuncNamesSubsectionContent.begin(),
+                                   FuncNamesSubsectionContent.end());
+
+  // 6.3. Type Names Subsection (ID 0x04)
+  std::vector<WasmEdge::Byte> TypeNamesSubsectionContent;
+  TypeNamesSubsectionContent.push_back(
+      static_cast<WasmEdge::Byte>(num_total_types));
+  for (int i = 0; i < num_total_types; ++i) {
```
hydai: Ditto.

test/executor/ExecutorTest.cpp
```diff
@@ -253,6 +253,311 @@ TEST(Coredump, generateCoredump) {
...
+    TypeNamesSubsectionContent.push_back(static_cast<WasmEdge::Byte>(i));
+    std::string TypeNameStr = "t" + std::to_string(i);
+    TypeNamesSubsectionContent.push_back(
+        static_cast<WasmEdge::Byte>(TypeNameStr.length()));
+    for (char c : TypeNameStr) {
```
hydai: Ditto.

test/executor/ExecutorTest.cpp
PhantomInTheWire: done

test/executor/ExecutorTest.cpp
PhantomInTheWire: done

test/executor/ExecutorTest.cpp
PhantomInTheWire: done

hydai (approved):
LGTM.

@q82419 PTAL, if everything is fine, then we can merge it.

hydai:
Hi @dannypsnl 
Please also take a look, because there is a request changed made by you, you have to check it again. Thanks

PhantomInTheWire:
I'm not sure why macos 14 arm 64 build fails, since it seems to be passing on the run I did on my fork and im on a m3 macbook air myself, we should probably try running it again

https://github.com/PhantomInTheWire/WasmEdge/actions/runs/15556151521/job/43797309949?pr=2

hydai (changes requested):
Please squash all your commits before we process the merge request.

PhantomInTheWire:
all commits squashed and pushed.

hydai:
Hi @PhantomInTheWire 
Please ignore the commitlint check; we are fixing the related false positive detection issue.

lib/validator/validator.cpp
```diff
@@ -165,32 +229,49 @@ Expect<void> Validator::validate(const AST::SubType &Type) {
     }
   }
 
-  // In current version, the length of type index vector will be <= 1.
-  if (Type.getSuperTypeIndices().size() > 1) {
-    spdlog::error(ErrCode::Value::InvalidSubType);
-    spdlog::error("    Accepts 1 super type currently."sv);
-    return Unexpect(ErrCode::Value::InvalidSubType);
+  if (!Conf.hasProposal(Proposal::GC)) {
```
q82419: The length of type index vector checking should apply for all proposals includes GC.
Because for current GC proposal, the length should be < 1.

ref: https://webassembly.github.io/gc/core/valid/types.html#xref-syntax-types-syntax-subtype-mathsf-sub-xref-syntax-types-syntax-subtype-mathsf-final-y-ast-xref-syntax-types-syntax-comptype-mathit-comptype

test/executor/ExecutorTest.cpp
```diff
@@ -253,6 +253,311 @@ TEST(Coredump, generateCoredump) {
   EXPECT_TRUE(FindCoredump);
 }
 
+class ValidatorSubtypeTest : public testing::Test {
```
q82419: I prefer to add another test folder, not in the executor test.

lib/validator/validator.cpp
q82419: Actually the proposal checking condition is not necessary.
With the subtypes which have super type indices, the binary format syntax are accept only when the GC proposal turning on.

lib/validator/validator.cpp
```diff
@@ -19,6 +20,69 @@ using namespace std::literals;
 namespace WasmEdge {
 namespace Validator {
 
+namespace details {
```
q82419: Use an anonymous namespace instead.

lib/validator/validator.cpp
```diff
@@ -19,6 +20,69 @@ using namespace std::literals;
...
+static constexpr uint32_t MaxSubtypeDepth = 63;
+
+namespace {
```
q82419: Not to use a nested detail namespace.

lib/validator/validator.cpp
```diff
@@ -19,6 +20,69 @@ using namespace std::literals;
...
+    spdlog::error(ErrCode::Value::InvalidSubType);
+    spdlog::error("Cycle detected in subtype hierarchy for type {}.", Index);
+    return Unexpect(ErrCode::Value::InvalidSubType);
+  }
+
+  if (Depth >= MaxSubtypeDepth) {
+    spdlog::error(ErrCode::Value::InvalidSubType);
+    spdlog::error("subtype depth for Type section's {}th signature exceeded "
+                  "the limits of {}",
+                  TypeIdx + 1, MaxSubtypeDepth);
```
q82419: Not to plus one for the index.

lib/validator/validator.cpp
```diff
@@ -19,6 +20,69 @@ using namespace std::literals;
...
+static constexpr uint32_t MaxSubtypeDepth = 63;
+
+namespace {
+Expect<void>
+calculateSubtypeDepthRecursiveHelper(uint32_t Index, uint32_t Depth,
+                                     std::set<uint32_t> &VisitedSet,
+                                     FormChecker &Checker, uint32_t TypeIdx) {
+  if (VisitedSet.count(Index)) {
+    spdlog::error(ErrCode::Value::InvalidSubType);
+    spdlog::error("Cycle detected in subtype hierarchy for type {}.", Index);
```
q82419: Add 4 space prefix for the message.

lib/validator/validator.cpp
```diff
@@ -19,6 +20,69 @@ using namespace std::literals;
...
+    return Unexpect(ErrCode::Value::InvalidSubType);
+  }
+
+  if (Depth >= MaxSubtypeDepth) {
+    spdlog::error(ErrCode::Value::InvalidSubType);
+    spdlog::error("subtype depth for Type section's {}th signature exceeded "
```
q82419: Add 4 space prefix for the message.

lib/validator/validator.cpp
```diff
@@ -19,6 +20,69 @@ using namespace std::literals;
...
 namespace Validator {
 
+namespace details {
+static constexpr uint32_t MaxSubtypeDepth = 63;
+
+namespace {
+Expect<void>
+calculateSubtypeDepthRecursiveHelper(uint32_t Index, uint32_t Depth,
+                                     std::set<uint32_t> &VisitedSet,
+                                     FormChecker &Checker, uint32_t TypeIdx) {
```
q82419: `const FormChecker`

lib/validator/validator.cpp
```diff
@@ -19,6 +20,69 @@ using namespace std::literals;
...
+            SuperIdx, Depth + 1, VisitedSet, Checker, TypeIdx);
+        !Res) {
+      spdlog::error("    When checking super type index {}.", SuperIdx);
+      return Unexpect(Res.error());
+    }
+  }
+
+  return {};
+}
+Expect<void> calculateSubtypeDepth(uint32_t TypeIdx, FormChecker &Checker) {
```
q82419: `const FormChecker`

lib/validator/validator.cpp
```diff
@@ -19,6 +20,69 @@ using namespace std::literals;
...
+    spdlog::error("subtype depth for Type section's {}th signature exceeded "
+                  "the limits of {}",
+                  TypeIdx + 1, MaxSubtypeDepth);
+    return Unexpect(ErrCode::Value::InvalidSubType);
+  }
+
+  VisitedSet.insert(Index);
+  const auto &TypeVec = Checker.getTypes();
+
+  if (unlikely(Index >= TypeVec.size())) {
```
q82419: Not necessary to check the defined type bound.
Because all types and their super type indices added into the form checker are valid.

lib/validator/validator.cpp
```diff
@@ -19,6 +20,69 @@ using namespace std::literals;
...
+    spdlog::error(
+        "    Type index {} is out of bounds for defined types (size: {}).",
+        Index, TypeVec.size());
+    return Unexpect(ErrCode::Value::InvalidSubType);
+  }
+
+  const auto &Type = *TypeVec[Index];
+
+  for (const auto SuperIdx : Type.getSuperTypeIndices()) {
+    if (unlikely(SuperIdx >= TypeVec.size())) {
```
q82419: Not necessary to check the defined type bound.
Because all types and their super type indices added into the form checker are valid.

q82419 (changes requested):

PhantomInTheWire:
I've removed ValidatorSubtypeNoGcTest as they shouldn't have been there in the first place ([discussion](https://github.com/WasmEdge/WasmEdge/pull/4134#discussion_r2136428134)). I initially added them thinking I had regressed a functionality, but after reviewing the spec, it turns out subtyping shouldn't be performed at all without GC. This was confirmed when [PR #4182](https://github.com/WasmEdge/WasmEdge/pull/4182) caused all of those tests to fail.

PhantomInTheWire:
@q82419 Thanks for the review! I've addressed all the suggested refactors in the latest commit. LMK if anything else needs tweaking.

lib/validator/validator.cpp
PhantomInTheWire: can't believe i missed that, fixed it.

lib/validator/validator.cpp
PhantomInTheWire: my bad, fixed.

test/executor/ExecutorTest.cpp
PhantomInTheWire: done, although there is a catch(https://github.com/WasmEdge/WasmEdge/pull/4134#issuecomment-2981749661)

lib/validator/validator.cpp
PhantomInTheWire: The initial code was a bit confusing due to a comment stating <= 1. I've fixed it.

lib/validator/validator.cpp
```diff
@@ -165,32 +213,41 @@ Expect<void> Validator::validate(const AST::SubType &Type) {
     }
   }
 
-  // In current version, the length of type index vector will be <= 1.
```
q82419, line 168: Please don't remove this comment.

q82419:
I think the other parts are okay.

lib/validator/validator.cpp
PhantomInTheWire: my bad. fixed.

q82419 (approved):

hydai:
Thanks @PhantomInTheWire

PR #4119, opened 2025-05-23
feat(wasi-nn): bump ggml to b5463; bump wasi-nn plugin to 0.1.22

dm4:
- Bump ggml to b5463
- Bump wasi-nn plugin to 0.1.22
- Remove deprecated `dump-kv-cache`
- Use the new `media_marker` in mtmd

PR #4099, opened 2025-05-06
feat(WASI-NN/MLX): support quantized gemma3 model

grorge123:
Support mlx-community quantized [model](https://huggingface.co/mlx-community/gemma-3-4b-it-4bit).

PR #4135, opened 2025-06-04
feat(plugin): update 0.14.1_plugin_lts branch 

hydai:
This PR contains:

d5f5ee4e chore(cmake): seperate the version between wasi_nn and wasi_nn_rpc
afb98853 feat(WASI-NN,ggml): bump llama.cpp b5575
92fdba8f feat(wasi-nn): bump ggml to b5463; bump wasi-nn plugin to 0.1.22 (#4119)
f0aff4d6 feat(WASI-NN/MLX): support quantized gemma3 model (#4099)


juntao:

------

## [CMakeLists.txt](https://github.com/WasmEdge/WasmEdge/blob/d5f5ee4e866260eba9605463b7f0336a4c63c563/CMakeLists.txt)

#### Potential issues

-1. The `CMAKE_MSVC_RUNTIME_LIBRARY` variable is set to "MultiThreadedDLL" which may cause issues with static linking.
- 2. The `find_package(PkgConfig)` command is used to find gRPC and protobuf but the corresponding `.pc` files are not present in the specified repositories.
- 3. The `CPACK_PACKAGE_CONTACT` variable contains an incorrect email address, it should be `"beststeve@secondstate.io"`.

#### Summary of changes

- - **Removed WASI-NN version and soversion settings**: The patch removes the lines that defined `WASMEDGE_WASI_NN_VERSION` and `WASMEDGE_WASI_NN_SOVERSION`, indicating these configurations are no longer needed.
- **Updated commit message context**: Although not directly shown in the code diff, the commit message suggests this change is part of updating the `0.14.1_plugin_lts` branch, implying broader updates related to this version.
- **Maintained CAPI version and soversion settings**: The patch retains the configurations for `WASMEDGE_CAPI_VERSION` and `WASMEDGE_CAPI_SOVERSION`, ensuring these remain consistent with prior settings.

## [cmake/WASINNDeps.cmake](https://github.com/WasmEdge/WasmEdge/blob/d5f5ee4e866260eba9605463b7f0336a4c63c563/cmake%2FWASINNDeps.cmake)

#### Potential issues

1. **Issue 1:**  
   **Problem:** In `wasmedge_setup_tf_variables`, the `elseif()` statement for Android is incomplete and lacks a condition after `elseif()`. This will cause a syntax error in CMake.
   
   **Explanation:** The line `elseif()` should specify a condition, such as another architecture check.

2. **Issue 2:**  
   **Problem:** In `wasmedge_setup_tf_target`, `target_link_libraries` links against `${WASMEDGE_TENSORFLOW_DEPS_TF_LIB}` which is not set for the `ANDROID` platform.
   
   **Explanation:** The variable `${WASMEDGE_TENSORFLOW_DEPS_TF_LIB}` is only set for `APPLE` and `UNIX`. For Android, this variable remains unset, causing a build error.

3. **Issue 3:**  
   **Problem:** In `wasmedge_setup_llama_target`, the target properties like `POSITION_INDEPENDENT_CODE` are set on `ggml-cuda` only if `WASMEDGE_PLUGIN_WASI_NN_GGML_LLAMA_CUBLAS` is enabled, but similar settings might be necessary for other targets conditionally.
   
   **Explanation:** This conditional setting might lead to inconsistencies across different configurations and should be reviewed or extended based on requirements.

#### Summary of changes

- - **Updated GIT_TAG for llama.cpp**: Changed the tag from `b5361` to `b5575`, likely incorporating recent fixes or features in the llama.cpp repository.
- **No other significant changes detected**: The rest of the patch does not include notable modifications to functionality or configuration.

## [lib/wasi_nn_rpc/CMakeLists.txt](https://github.com/WasmEdge/WasmEdge/blob/d5f5ee4e866260eba9605463b7f0336a4c63c563/lib%2Fwasi_nn_rpc%2FCMakeLists.txt)

#### Potential issues

1. **Repeated `protoc` Execution**: The `execute_process` command runs `protoc` twice with the same `.proto` file, which is inefficient and could lead to duplicate files or unexpected behavior.
2. **Hardcoded `wasi_ephemeral_nn.proto` Path**: The path to `wasi_ephemeral_nn.proto` is hardcoded in the `execute_process` command, reducing flexibility and making the code less maintainable.
3. **Missing Dependency Check for gRPC++ Reflection**: The commented-out line `# target_link_libraries(wasiNNRPC PUBLIC grpc++_reflection)` suggests a potential dependency that could be necessary for debugging but is not included by default.

#### Summary of changes

-**Changes in the Patch**

1. **Update Library Name**: The library name was changed from `wasiNNRPC` to `wasiNNRPC`. This is a change that reflects a semantic shift or naming convention update.

2. **Add Version and SOversion Properties**:
   - A new variable `WASMEDGE_WASI_NN_RPC_VERSION` with the value `"0.1.0"` was added.
   - Another new variable `WASMEDGE_WASI_NN_RPC_SOVERSION` with the value `"0"` was also added.
   - These variables were set as properties for the target `wasiNNRPC`. The properties include version and SOversion, which are essential in specifying the library's compatibility levels.

3. **Install Destination**: The installation directory for the library has been updated from `${CMAKE_INSTALL_LIBDIR}` to a new location, indicating a potential change in where the library will be installed after being built.

4. **Target Visibility**: The visibility of the target `wasiNNRPC` has been set to `default`, which means it will not be hidden by default during linking. This is important for ensuring that the library can be used effectively.

These changes are significant as they align with semantic versioning principles and specify necessary metadata for the library, affecting how it can be installed and linked into applications.

## [plugins/wasi_nn/CMakeLists.txt](https://github.com/WasmEdge/WasmEdge/blob/d5f5ee4e866260eba9605463b7f0336a4c63c563/plugins%2Fwasi_nn%2FCMakeLists.txt)

#### Potential issues

1. **Variable Naming Issue**: The variable `WASMEDGE_PLUGIN_WASI_NN_BACKEND` is used in a foreach loop without being defined or initialized within the provided code snippet, which could lead to undefined behavior if it's not set elsewhere.

2. **Hardcoded Compiler Options**: The compiler option `-DWASMEDGE_PLUGIN` is hardcoded as a public compile option. This might not be flexible enough for different build environments and configurations, and should consider using configuration checks or variables.

3. **Version Handling**: The version parsing logic assumes that `WASMEDGE_WASI_NN_VERSION` will always have three components separated by dots. If the version format changes (e.g., to "0.1"), this could cause issues when trying to access specific indices (`WASI_NN_VERSION_PATCH`).

#### Summary of changes

-**Changes:**
1. **Updated `WASMEDGE_WASI_NN_VERSION` and `WASMEDGE_WASI_NN_SOVERSION`:** These variables are now set to "0.1.23" and "0" respectively, representing the version of the WASI-NN plugin.
2. **Added dependencies for WASI-NN:** The patch adds the necessary include directory and sets up the WASINNDeps target.
3. **Handled version parsing:** A string is parsed to separate the major version from the minor version.

**Most important changes:**
- Version update for WASI-NN plugin
- Added dependencies for WASI-NN library
- Improved dependency setup

## [plugins/wasi_nn/MLX/mlx/base.cpp](https://github.com/WasmEdge/WasmEdge/blob/d5f5ee4e866260eba9605463b7f0336a4c63c563/plugins%2Fwasi_nn%2FMLX%2Fmlx%2Fbase.cpp)

#### Potential issues

1. **Logical Error in `toQuantized` Method**: The condition `if (Prefix.empty() && Name.empty() ? "" : ".")` is incorrectly used to append a dot, which can lead to incorrect prefix formation. It should be `(Prefix.empty() || Name.empty() ? "" : ".")`.

2. **Potential Memory Leak in `toQuantized` Method**: The method does not check if `V->hasQuantize()` returns false but continues processing and potentially overwriting `V`. If quantization is unsupported, it might lead to unintended behavior or memory leaks.

3. **String Concatenation Issue in `getWeigts` Method**: The condition `NewPrefix + (NewPrefix.empty() ? "" : ".")` will always append a dot if `NewPrefix` is not empty, which is likely incorrect. It should be `(Prefix.empty() && Name.empty() ? "" : ".")`.

#### Summary of changes

- The patch updates the `toQuantized` and `getWeigts` methods of a neural network module, adding quantization support.
- Added a new parameter `Prefix` to specify the prefix for quantization scales.
- Modified the `Parameters` map when applying weights to ensure compatibility with the new method.
- Added a new parameter `Prefix` to the method signature in `getWeigts`.
- Removed the commented-out code related to weight shapes and group sizes.

## [plugins/wasi_nn/MLX/mlx/base.h](https://github.com/WasmEdge/WasmEdge/blob/d5f5ee4e866260eba9605463b7f0336a4c63c563/plugins%2Fwasi_nn%2FMLX%2Fmlx%2Fbase.h)

#### Potential issues

-3 minor bugs or issues found:

1. **Misplaced `using namespace std::literals::string_view_literals;`**: This line should be placed before the include directives to avoid any potential conflicts with other namespace using statements.

2. **Potential naming conflict**: The variable `Module` is already defined in the global scope, which could lead to unexpected behavior or confusion. Consider renaming it to something unique.

3. **Unused function parameters**: The functions `update` and `apply` are not used anywhere in the codebase, so they can be removed or commented out to clean up the code.

#### Summary of changes

- **New Method Added:**
  - `toQuantized(int GroupSize = 64, int Bits = 4, const std::string &Prefix = "", const std::unordered_map<std::string, mx::array> &Parameters = {})`
    - This method provides quantization capabilities to the module with specified parameters.
  
- **Method Changed:**
  - `virtual std::shared_ptr<nn::Module> toQuantized(int GroupSize = 64, int Bits = 4)`
    - Removed the optional `Prefix` and `Parameters` from the function signature. The `Prefix` parameter is now fixed to an empty string.
  
- **Method Deprecated:**
  - `virtual bool hasQuantize() { return false; }`
    - Added a deprecated method `hasQuantize()` that returns `false`. This indicates that quantization is currently not supported by the module.

## [plugins/wasi_nn/MLX/mlx/embedding.cpp](https://github.com/WasmEdge/WasmEdge/blob/d5f5ee4e866260eba9605463b7f0336a4c63c563/plugins%2Fwasi_nn%2FMLX%2Fmlx%2Fembedding.cpp)

#### Potential issues

1. **Unsafe `shared_from_this()` usage**: The method `toQuantized` calls `std::dynamic_pointer_cast<Embedding>(shared_from_this())`. This can throw `std::bad_weak_ptr` if the object is not managed by a `std::shared_ptr` or if `shared_from_this()` is called before any shared ownership has been established.

2. **Unnecessary use of `transpose`**: The `asLinear` function unnecessarily transposes the weight matrix every time it's called, which can be inefficient. Consider storing the transpose in `Parameters` during initialization if this operation is frequently used.

3. **Unhandled exception from `Parameters.at()`**: Both `forward` and `asLinear` methods use `Parameters.at("weight")`, which throws an exception if "weight" does not exist. This should be handled to avoid runtime errors, especially since the keys in `Parameters` might change or be misspelled.

#### Summary of changes

- - **Extended `toQuantized` Method Parameters**: The `toQuantized` method now includes additional parameters: a `const std::string &` and a `const std::unordered_map<std::string, mx::array> &`, which were previously missing.
- **Formatting Changes**: Improved formatting of the `toQuantized` method signature for better readability by adding line breaks.
- **No Functional Change in Existing Parameters**: The existing functionality using `GroupSize` and `Bits` remains unchanged.

## [plugins/wasi_nn/MLX/mlx/embedding.h](https://github.com/WasmEdge/WasmEdge/blob/d5f5ee4e866260eba9605463b7f0336a4c63c563/plugins%2Fwasi_nn%2FMLX%2Fmlx%2Fembedding.h)

#### Potential issues

1. The `Embedding` constructor uses `mx::random::normal` which may not be defined in the included headers; ensure `mx::random::normal` is correctly accessible.
2. The method `forward` is declared as virtual but lacks a definition or pure declaration, leading to potential linker errors or incomplete type issues.
3. The use of `std::unordered_map` in the parameter list without including `<unordered_map>` will result in compilation failure; include the missing header.

#### Summary of changes

- **Key Changes:**
- **Enhanced `toQuantized` Method:** Added parameters `Prefix` and `Parameters` to the `toQuantized` method, allowing for more customization in quantization.
- **Added `hasQuantize` Method:** Introduced a virtual method `hasQuantize` that returns `true`, indicating the module supports quantization.
- **Formatting Adjustments:** Improved code formatting by aligning and splitting lines for better readability.

## [plugins/wasi_nn/MLX/mlx/linear.cpp](https://github.com/WasmEdge/WasmEdge/blob/d5f5ee4e866260eba9605463b7f0336a4c63c563/plugins%2Fwasi_nn%2FMLX%2Fmlx%2Flinear.cpp)

#### Potential issues

-There is one potential issue with the `forward` method where the input array `Input` might not be properly transposed before being used in the matrix multiplication operation. The correct transpose should be performed using `transpose(Parameters.at("input")).transpose()`.

#### Summary of changes

- - **Added New Parameters**: The `toQuantized` method now includes additional parameters: a `const std::string &` and a `const std::unordered_map<std::string, mx::array> &`, though they are not currently used in the function.
- **Formatting Change**: The function signature for `toQuantized` is split across multiple lines for improved readability.
- **No Functional Change**: Despite the added parameters, the core functionality of converting a `Linear` module to a quantized version remains unchanged.

## [plugins/wasi_nn/MLX/mlx/linear.h](https://github.com/WasmEdge/WasmEdge/blob/d5f5ee4e866260eba9605463b7f0336a4c63c563/plugins%2Fwasi_nn%2FMLX%2Fmlx%2Flinear.h)

#### Potential issues

1. Missing include for `mx::array`  
2. Missing comment on the constructor parameter `EnableBias`  
3. Missing check for `toQuantized()` method

#### Summary of changes

-Added a `toQuantized` method that allows for quantizing the linear layer with customizable parameters.
-Updated the `forward` method to return a shared pointer to a quantized module.
-Implemented a virtual `hasQuantize` method to indicate whether the module is quantizable.

## [plugins/wasi_nn/MLX/model/gemma3/gemma3.cpp](https://github.com/WasmEdge/WasmEdge/blob/d5f5ee4e866260eba9605463b7f0336a4c63c563/plugins%2Fwasi_nn%2FMLX%2Fmodel%2Fgemma3%2Fgemma3.cpp)

#### Potential issues

-1. Missing include guard (`#ifndef WASMEDGE_HOST_WASINN_MLX_GEMMA3_MODEL_H_`).
- 2. Missing `#include <cassert>`.
- 3. Unchecked return value of `Model::forward(const mx::array &InputIds, const mx::array &PixelValues, const mx::array &Mask)` when the model is trained without mask.
- 4. Lack of error checking and handling in `Model::fromPretrained(const std::string &ModelPath)`.

#### Summary of changes

- - **Quantization Moved Post-Sanitize**: The quantization logic was moved after the model weights are sanitized, ensuring that sanitized weights are used during quantization.
- **Updated Logging Message**: The logging message for quantization now includes the backend information (`[WASI-NN] MLX backend:`) and uses a string view (`"sv"`).
- **Key Replacement in Sanitize**: Added key replacement to remove the prefix `"model"` from weight keys, likely to align with expected format after sanitization.

## [plugins/wasi_nn/wasinn_ggml.cpp](https://github.com/WasmEdge/WasmEdge/blob/d5f5ee4e866260eba9605463b7f0336a4c63c563/plugins%2Fwasi_nn%2Fwasinn_ggml.cpp)

#### Potential issues

N/A

#### Summary of changes

- **Key Changes:**  
  - Removed inclusion of `clip.h`.
  - Added `json-partial.h` for partial JSON parsing.
  - Removed `json.hpp` and added `mtmd-helper.h` for multimedia model processing.
  - Updated error messages to include additional context in error reports.

## [plugins/wasi_nn/wasinn_ggml.h](https://github.com/WasmEdge/WasmEdge/blob/d5f5ee4e866260eba9605463b7f0336a4c63c563/plugins%2Fwasi_nn%2Fwasinn_ggml.h)

#### Potential issues

1. **Memory Management Issue**: The `llama_model_ptr` and `llama_context_ptr` pointers in the `Graph` struct are raw pointers, which can lead to memory leaks if not properly managed (e.g., no explicit destructor to free these resources).

2. **Redundant Configuration Copies**: The `LocalConfig` is redundantly stored in both the `Graph` and `Context` structs. This duplication can cause inconsistencies and increases complexity.

3. **Magic Number Issue**: The `EmbdNormalizeType::None` enum value is set to `-1`, which is not a typical way to define an invalid or default state in enums. It's better to use an unsigned type for enum values and start from zero for valid states, indicating an uninitialized or invalid state explicitly.

#### Summary of changes

-Added inclusion of `llava.h` header file.
-Updated include paths for other libraries (`common.h`, `llama-cpp.h`, `llama.h`).
-Removed unnecessary includes that are no longer required.

## [plugins/wasi_nn/wasinn_mlx.cpp](https://github.com/WasmEdge/WasmEdge/blob/d5f5ee4e866260eba9605463b7f0336a4c63c563/plugins%2Fwasi_nn%2Fwasinn_mlx.cpp)

#### Potential issues

1. **Typographical Error in Enum Definition:**
   - Issue: Typo in enum name `AnswerSataus` should be `AnswerStatus`.
   - Impact: Compilation error due to undefined type.

2. **Incorrect Error Message in Quantization Field Retrieval:**
   - Issue: Incorrect message "Unable to retrieve the group size from quantization option." is used for both `group_size` and `bits`.
   - Impact: Misleading debug/error information for users.

3. **Potential File Descriptor Leak:**
   - Issue: File streams (`std::ifstream`, `std::ofstream`) are not guaranteed to close if an error occurs before reaching the explicit `.close()` call.
   - Impact: Resource leak, potential exhaustion of file descriptors in long-running processes.

#### Summary of changes

- • Added support for quantization options by parsing `quantization` from the document and setting `GraphRef.IsQuantized`.
• Enhanced logging to include detailed information about quantization parameters (`q_bits`, `group_size`) when loading and computing models.
• Updated the `toQuantized` method call with an additional parameter to handle weights, specifically in the context of loading a pre-quantized model.

dm4 (approved):

PR #4137, opened 2025-06-04
chore(docker): bump the llvm from 12 to 18 on ubuntu 20.04

hydai:
The mac is using llvm@18. I think we should use the same version of the llvm for all assets.

### DIFF
--- a/utils/docker/Dockerfile.ubuntu-base
+++ b/utils/docker/Dockerfile.ubuntu-base
@@ -21,17 +21,21 @@ FROM base AS deps-20
 RUN curl -sSf https://apt.kitware.com/kitware-archive.sh | sh
 RUN apt-get install -y cmake
 
+RUN wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add - \
+        && echo "deb http://apt.llvm.org/focal llvm-toolchain-focal-18 main" | tee /etc/apt/sources.list.d/llvm.list
+
 RUN apt-get install -y \
-        llvm-12-dev \
-        liblld-12-dev \
-        clang-12
+        llvm-18-dev \
+        liblld-18-dev \
+        libpolly-18-dev \
+        clang-18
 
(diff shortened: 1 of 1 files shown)


juntao:

------

## [utils/docker/Dockerfile.ubuntu-base](https://github.com/WasmEdge/WasmEdge/blob/5806518c7cecd22fc07d6eba07965b0231ee11f2/utils%2Fdocker%2FDockerfile.ubuntu-base)

#### Potential issues

1. The `apt-get update && apt-get upgrade -y` command is redundant since the Dockerfile uses `FROM ubuntu:20.04 AS base`, which already has a recent version of Debian and its packages.
2. The `RUN apt-get install -y curl dpkg-dev g++ gcc git ninja-build software-properties-common wget zlib1g-dev` command installs a set of common build tools that are typically included in the `base` image, so it can be removed.
3. The commented-out lines for different versions of Ubuntu (20.04, 22.04, and 24.04) create separate stages (`deps-20`, `deps-22`, and `deps-24`) that only install the necessary packages for each version of Ubuntu. These stages can be combined into a single stage that checks the `UBUNTU_VER` environment variable to determine which dependencies to install.

#### Summary of changes

-**Key Changes:**

1. **Updated Dockerfile for Ubuntu 20.04**
   - Added a new section for dependencies on Ubuntu 20.04
   - Updated the `FROM` line to specify Ubuntu 20.04

2. **Added LLVM 18 Installation**
   - Added a new section to download and install LLVM 18 from a snapshot repository
   - Installed `llvm-18-dev`, `liblld-18-dev`, `libpolly-18-dev`, and `clang-18`

3. **Updated Environment Variables for Clang**
   - Changed the environment variables for CC and CXX to point to LLVM 18

4. **Dockerfile for Ubuntu 22.04**
   - Added a new section for dependencies on Ubuntu 22.04
   - Updated the `FROM` line to specify Ubuntu 22.04

PR #4149, opened 2025-06-05
Add conditional check for stdbool.h before C23

### LINKED ISSUE #4147: feat: check c version before `#include <stdbool.h>` in wasmedge.h
### Summary

The C API header `wasmedge.h` includes `stdbool.h`[1].
However, this header `stdbool.h` is provided by C99 but deprecated in C23. We should add a conditional check in the header to ensure that it is not included after C23.

[1]: https://github.com/WasmEdge/WasmEdge/blob/master/include/api/wasmedge/wasmedge.h#L35

### Details

- [ ] Add a conditional check for the C standard
- [ ] Only include `stdbool.h` before C23

### Appendix

https://en.cppreference.com/w/c/header/stdbool

GautamBytes:
## Summary
Adds conditional inclusion of `stdbool.h` to prevent deprecation issues in C23.

## Changes
- Replace `#include <stdbool.h>` with conditional check
- Only include `stdbool.h` for C standards before C23 (`__STDC_VERSION__ < 202311L`)

Fixes #4147


juntao:

------

## [.github/workflows/build-extensions.yml](https://github.com/WasmEdge/WasmEdge/blob/20e9bd4e151725801655df52f01574e7e2ec8441/.github%2Fworkflows%2Fbuild-extensions.yml)

#### Potential issues

N/A

#### Summary of changes

- Changed the workflow name from "Test WasmEdge extensions" to "Extensions".
- Renamed the job "Build and Test" to "Plugins".
- No other significant changes were made to the file.

## [.github/workflows/build.yml](https://github.com/WasmEdge/WasmEdge/blob/20e9bd4e151725801655df52f01574e7e2ec8441/.github%2Fworkflows%2Fbuild.yml)

#### Potential issues

1. **Missing `if` condition for paths**: The code does not explicitly check if the specified files or directories are changed before running certain jobs, which could lead to unnecessary builds.

2. **Version fetching logic might fail**: If no matching tags exist, `git describe --match '[0-9].[0-9]*' --tag` will return a commit hash instead of a version string, potentially causing downstream issues.

3. **Permissions mismatch for 'secrets'**: The `build_on_ubuntu` job has `secrets: inherit`, but it's unclear if the necessary secrets are defined in the caller workflow or the reusable workflow being called.

#### Summary of changes

- The GitHub Actions workflow name was changed from "Test WasmEdge Core" to "Core".
- The `runner` for the `manylinux_2_28 aarch64` job and the `ubuntu-20.04, aarch64` job has been updated from `'linux-arm64-v2'` to `'ubuntu-24.04-arm'`.
- The matrix configuration was slightly adjusted in terms of specifying jobs for different platforms and architectures but no significant logical changes were observed.

## [.github/workflows/build_for_openwrt.yml](https://github.com/WasmEdge/WasmEdge/blob/20e9bd4e151725801655df52f01574e7e2ec8441/.github%2Fworkflows%2Fbuild_for_openwrt.yml)

#### Potential issues

1. The GitHub Actions workflow file uses deprecated versions of actions (`actions/checkout@v4`, `actions/setup-python@v5`, `actions/cache/restore@v4`, `actions/upload-artifact@v4`) which may not receive updates or support in the future.
2. The workflow does not handle potential failures in the `make download` and `make -j$(nproc)` commands, which could result in incomplete builds if dependencies are missing or network issues occur.
3. The `docker run` command uses a deprecated `--privileged` flag, which may pose security risks and is generally discouraged; consider using more granular capabilities instead.

#### Summary of changes

- **Update to latest Python setup action:** The patch updates the `actions/setup-python` from version `@v4` to `@v5`, ensuring compatibility with recent changes and improvements in the GitHub Actions ecosystem.
- **Git safe directory configuration:** A new job step was added to configure the current working directory as a safe directory using `git config --global --add safe.directory $(pwd)`. This is important for security and ensures that subsequent git operations within the workflow are not blocked by Git's directory safety checks.
- **Job name adjustment:** The job name "Restore OpenWrt" was moved to a new line, which helps in better readability and organization of the workflow configuration.

## [.github/workflows/commitlint.yml](https://github.com/WasmEdge/WasmEdge/blob/20e9bd4e151725801655df52f01574e7e2ec8441/.github%2Fworkflows%2Fcommitlint.yml)

#### Potential issues

1. The `ubuntu-24.04` runner does not exist, it should be replaced with an available version like `ubuntu-22.04`.
2. The fetch depth in the `checkout` action is set to 0, which might cause unnecessary large cloning; use a specific number if possible.
3. The `git --version`, `node --version`, and `npm --version` commands are not required and can be removed for efficiency.

#### Summary of changes

- **Updated OS Version**: Changed the runner OS from `ubuntu-22.04` to `ubuntu-24.04`.
- **Added GitHub Permissions**: Included permissions section for content read access.
- **Reordered Command Execution**: Moved `npx commitlint --version` command within "Install commitlint" step, ensuring version check happens after installation.

## [.github/workflows/docker.yml](https://github.com/WasmEdge/WasmEdge/blob/20e9bd4e151725801655df52f01574e7e2ec8441/.github%2Fworkflows%2Fdocker.yml)

#### Potential issues

1. **Inconsistent Use of Double and Single Quotes**: The code switches between double quotes (`"`) and single quotes (`'`) for string literals, which can lead to inconsistencies and potential bugs.

2. **Unnecessary Conditional Check Redundancy**: In the `paths-filter` job, the conditional checks for `${{ github.event_name == 'pull_request' }}` are redundant as the outputs are already set based on this condition.

3. **Potential Security Risk with Secrets**: The Docker login credentials are exposed in the logs if the job fails, which is a potential security risk; consider using `secrets.GITHUB_TOKEN` for authentication instead.

#### Summary of changes

The patch modifies the CI configuration file to update the host runner and container for a specific job. The key changes are:

1. Changed `host_runner` from `linux-arm64-v2` to `ubuntu-24.04-arm`.
2. Retained `docker_tag`, `targets`, and `host_container` unchanged.
3. The modification affects only the `manylinux_2_28 aarch64` job.

These changes are crucial for updating the CI environment for building on aarch64 architecture, potentially addressing compatibility or performance improvements with the new runner.

## [.github/workflows/release.yml](https://github.com/WasmEdge/WasmEdge/blob/20e9bd4e151725801655df52f01574e7e2ec8441/.github%2Fworkflows%2Frelease.yml)

#### Potential issues

1. The `workflow_dispatch` trigger's `inputs.logLevel` default value is incorrect; it should not have quotes around "info".
2. The `push` trigger's tag pattern `[0-9]+.[0-9]+.[0-9]+*` is incorrect and will not match valid semantic versioning tags (should use `[0-9]+\\.[0-9]+\\.[0-9]+` instead).
3. In the `create_release` job, the `gh release create` command uses `--verify-tag`, which may fail if the tag does not verify correctly; it's safer to handle this check separately or remove the flag.

#### Summary of changes

1. **Update in job configuration for `ubuntu-20.04` aarch64**:
   - Changed the runner from 'linux-arm64-v2' to 'ubuntu-24.04-arm'.

2. **Update in job configuration for `manylinux_2_28` aarch64**:
   - Changed the runner from 'linux-arm64-v2' to 'ubuntu-24.04-arm'.

## [.github/workflows/reusable-build-extensions.yml](https://github.com/WasmEdge/WasmEdge/blob/20e9bd4e151725801655df52f01574e7e2ec8441/.github%2Fworkflows%2Freusable-build-extensions.yml)

#### Potential issues

1. The `inputs.release` boolean input type is not supported by GitHub Actions, causing the workflow to fail.
2. The `if: ${{ !inputs.release }}` condition in the `build_on_ubuntu_latest` job will always evaluate to true because the `release` input is not a valid boolean.
3. The `process.env[plugin.plugin.split('-')[0]]` expression may cause runtime errors if `plugin.plugin` is undefined or does not contain a hyphen, leading to unpredictable behavior during workflow execution.

#### Summary of changes

- **Runner Change:** Changed the runner from 'linux-arm64-v2' to 'ubuntu-24.04-arm'.
- **Docker Tag Update:** Updated the docker_tag to match new arm architecture.
- **Asset Tag Adjustment:** Adjusted the asset tag for consistency.

## [.github/workflows/reusable-build-on-ubuntu.yml](https://github.com/WasmEdge/WasmEdge/blob/20e9bd4e151725801655df52f01574e7e2ec8441/.github%2Fworkflows%2Freusable-build-on-ubuntu.yml)

#### Potential issues

1. In the "Build (${{ matrix.compiler }}, ${{ matrix.build_type }})" step, the conditions for clang++ and other compilers are identical.
2. The "Upload package tarball" step installs `gh` CLI with a hardcoded architecture check using `dpkg --print-architecture`, which may not be available in all environments (e.g., Alpine).
3. The "Create and upload coverage report to Codecov" step does not include error handling for cases where the codecov.xml file might not exist.

#### Summary of changes

- Added architecture (`matrix.arch`) to the artifact name in the `name` field of the upload step.
- Changed the path from a generic "Linux" identifier to the specific architecture.
- No longer uploads package tarball when release is not true, due to added conditional check `if: ${{ inputs.release }}`.

## [.github/workflows/test-install-script.yml](https://github.com/WasmEdge/WasmEdge/blob/20e9bd4e151725801655df52f01574e7e2ec8441/.github%2Fworkflows%2Ftest-install-script.yml)

#### Potential issues

1. **Version Incompatibility:** The `actions/checkout@v4` used in the macOS jobs is not a valid version for the `actions/checkout` action, as it currently supports only `@v2` and `@v3`.
2. **Shell Specification Issue:** The `shell: zsh {0}` syntax in the macos-arm64 job steps is incorrect; use `run: bash utils/install.sh -D` instead.
3. **Path Sensitivity Issue**: The use of backticks for command substitution in `git ls-remote ... | cut -d '/' -f 3 | awk '{ if ($1 ~ /-/) print; else print $0"_" ; }'` is outdated and should be replaced with `$()` for better readability and compatibility.

#### Summary of changes

Here are the key changes in short bullet points:

- Changed `host_runner` from `linux-arm64-v2` to `ubuntu-24.04-arm`
- Added `python_package` entry for `python2 python3`

These are the most significant changes, as they affect the build and testing environment for the aarch64 architecture on manylinux 2_28.

## [.github/workflows/test-python-install-script.yml](https://github.com/WasmEdge/WasmEdge/blob/20e9bd4e151725801655df52f01574e7e2ec8441/.github%2Fworkflows%2Ftest-python-install-script.yml)

#### Potential issues

1. The `linux` job uses a mix of Ubuntu and manylinux_2_28 aarch64, but the `macos` job doesn't have similar matrix configurations for both amd64 and arm64, which might lead to inconsistencies.

2. The script uses hardcoded paths like `/usr/lib/wasmedge/` in multiple places without checking if they exist or are writable, potentially causing failures on different environments.

3. In the `Latest Release Check` step, both git and curl commands fetch the latest version, but the git command uses a complex pipeline to filter tags, which might not always match the expected format from the curl output, leading to false negatives in version comparison.

#### Summary of changes

### Top 3 Key Changes:

1. **Update Host Runner for `manylinux_2_28 aarch64`**:
   - Changed `host_runner` from `linux-arm64-v2` to `ubuntu-24.04-arm`.

2. **Ensure Docker Image Consistency**:
   - Docker image remains the same, ensuring compatibility with the updated host runner.

3. **Maintain Package Manager Configuration**:
   - The package manager `yum` is unchanged for this configuration.

## [cmake/Helper.cmake](https://github.com/WasmEdge/WasmEdge/blob/20e9bd4e151725801655df52f01574e7e2ec8441/cmake%2FHelper.cmake)

#### Potential issues

1. **Inconsistent Indentation**: The code uses inconsistent indentation, which can make it harder to read and maintain.
2. **Redundant Check for LLVM Version**: In the section handling `WASMEDGE_LLVM_LINK_STATIC_COMPONENTS`, there are redundant checks for `LLVM_VERSION_MAJOR`. It is sufficient to handle specific versions in a single condition.
3. **Potential Path Issue with ZLIB**: The code assumes that `ZLIB_LIBRARIES` points directly to the library file, which may not always be the case; it could point to an imported target or a list of libraries. This might cause issues when constructing paths for static linking components.

#### Summary of changes

Here are the top 3 most important changes:

1. **Disabled Shallow Clone**:
   - Changed `GIT_SHALLOW` from `TRUE` to `FALSE`.
2. **Updated stb_image Source Download Behavior**:
   - This change will now download the entire git history, not just the latest commit.
3. **Potential Impact on Build Time and Size**:
   - With shallow cloning disabled, the build process may take longer and require more disk space due to the complete repository history being cloned.

## [cmake/WASINNDeps.cmake](https://github.com/WasmEdge/WasmEdge/blob/20e9bd4e151725801655df52f01574e7e2ec8441/cmake%2FWASINNDeps.cmake)

#### Potential issues

1. **Potential Null Reference in `wasmedge_setup_tf_variables`**: In the function, there is a potential null reference error when checking for unsupported architectures.
    ```cmake
    elseif()
      message(FATAL_ERROR "Unsupported architecture: ${CMAKE_SYSTEM_PROCESSOR}")
    ```
   This will cause an issue if no matching architecture is found. The `elseif` block should be followed by specific conditions or removed.

2. **Uninitialized Variable in TensorFlow Setup**: In the function `wasmedge_setup_tf_lib`, there's a section for Android where nothing happens.
    ```cmake
    elseif(ANDROID)
    ```
   This will cause an issue since it doesn't set any variable, potentially leading to undefined behavior.

3. **Redundant FetchContent_Declare in TensorFlow Setup**: The `FetchContent_Declare` command is used redundantly if the content has already been declared.
    ```cmake
    FetchContent_Declare(
      wasmedge_tensorflow_lib_tflite
      URL "https://github.com/second-state/WasmEdge-tensorflow-deps/releases/download/${WASMEDGE_DEPS_VERSION}/WasmEdge-tensorflow-deps-TFLite-${WASMEDGE_DEPS_VERSION}-${WASMEDGE_TENSORFLOW_SYSTEM_NAME}.tar.gz"
      URL_HASH "SHA256=${WASMEDGE_TENSORFLOW_DEPS_TFLITE_HASH}"
    )
    ```
   The `FetchContent_Declare` should be wrapped with a condition to avoid re-declaration.

#### Summary of changes

Here are the top three changes in the patch:

- **Updated `GIT_TAG`**: Changed from `b5593` to `b5640` for fetching the llama repository.
- The rest of the lines remain unchanged.
- No other significant functional or structural changes were made.

## [include/api/wasmedge/wasmedge.h](https://github.com/WasmEdge/WasmEdge/blob/20e9bd4e151725801655df52f01574e7e2ec8441/include%2Fapi%2Fwasmedge%2Fwasmedge.h)

#### Potential issues

1. **Issue: Missing Conditional Check for `stdbool.h` Before C23**: The code does not include a check to determine if the compiler version supports C23 before including `<stdbool.h>`. This can lead to compatibility issues with older compilers that do not support C23 features.

2. **Issue: Inconsistent Use of `extern "C"` in Header Guards**: The use of `extern "C"` is wrapped inside `#ifdef __cplusplus` but not inside the header guard (`#ifndef WASMEDGE_C_API_H`). This can lead to conflicts if another part of the code includes this header differently.

3. **Issue: Unnecessary Conditional Inclusion for `<stdbool.h>`**: The conditional inclusion of `<stdbool.h>` is redundant when `__STDC_VERSION__` is defined and is less than 201710L, which indicates the support for `_Bool` type in C99. It should be updated to check for versions before C23 more accurately.

These issues need to be addressed to ensure compatibility and correct functionality across different compiler versions and environments.

#### Summary of changes

The most important changes in the patch are:

1. **Conditional check for `stdbool.h`**: The inclusion of `<stdbool.h>` is now conditioned on the absence of C++ and when the standard version is less than C23 (`__STDC_VERSION__ < 202311L`). This ensures that `stdbool.h` is only included when necessary.

2. **Preprocessor directives for conditional compilation**: The patch adds preprocessor directives to control the inclusion of `<stdbool.h>` based on the programming language and standard version in use, avoiding unnecessary inclusions and potential conflicts with C++ headers.

3. **Code structure preservation**: The overall structure of the file remains unchanged, maintaining readability and ease of understanding while ensuring backward compatibility for standards earlier than C23.

## [include/ast/component/canonical.h](https://github.com/WasmEdge/WasmEdge/blob/20e9bd4e151725801655df52f01574e7e2ec8441/include%2Fast%2Fcomponent%2Fcanonical.h)

#### Potential issues

1. The code does not include a conditional check for `stdbool.h` before C23, which could cause compatibility issues on systems that do not support the `<stdbool.h>` header in older C standards.

2. The `Span<const CanonOpt>` type is used without being defined or included, which will lead to a compilation error.

3. The formatter specialization for `WasmEdge::AST::Component::Canon` lacks proper namespace handling, potentially causing issues when formatting types from different namespaces.

#### Summary of changes

### Key Changes:

1. **Encapsulation and Access Modifiers:**
   - Moved private member variables into `private` sections of classes.
   - Changed some class members from public to private with corresponding accessor methods.

2. **Added Documentation Comments:**
   - Included detailed comments explaining the purpose and structure of each enum, class, and their methods.
   - Added TODO comments for future implementation steps.

3. **Code Cleanup and Refactoring:**
   - Removed unnecessary `#include <optional>` directive.
   - Streamlined some code sections by removing redundant comments or merging related parts together.

These changes aim to improve code readability, maintainability, and adherence to best practices in C++ development.

## [include/ast/component/instance.h](https://github.com/WasmEdge/WasmEdge/blob/20e9bd4e151725801655df52f01574e7e2ec8441/include%2Fast%2Fcomponent%2Finstance.h)

#### Potential issues

1. **Potential Issue with `std::string_view` in `getName()`:** The use of `std::string_view` for `getName()` is problematic since it returns a view that becomes invalid if the internal string is modified or destroyed.

2. **Lack of Initialization Guarantee:** Some member variables are not guaranteed to be initialized, which could lead to undefined behavior (e.g., `Name` and `Idx` in `InstantiateArgImpl`).

3. **Potential Type Mismatch with C++17 and Earlier:** The use of `std::vector<InlineExportImpl<SortType>> &&Es)` may cause issues if the code is compiled with a standard earlier than C++20 due to the lack of guaranteed rvalue reference support in containers before C++20.

These three issues are major concerns that need immediate attention.

#### Summary of changes

- **Initialization Changes:** The default constructor for both `CoreInstantiate` and `Instantiate` classes now initializes the index variable using parentheses instead of curly braces.

- **Member Variable Renaming:** `ComponentIndex` in class `Instantiate` is renamed to `ComponentIdx`.

- **Rvalue Reference:** In the parameter list of the main constructor for both `CoreInstantiate` and `Instantiate` classes, `std::vector<...>` is changed to `std::vector<...> &&`, indicating rvalue references.

## [include/ast/component/start.h](https://github.com/WasmEdge/WasmEdge/blob/20e9bd4e151725801655df52f01574e7e2ec8441/include%2Fast%2Fcomponent%2Fstart.h)

#### Potential issues

1. The header file `stdbool.h` is missing, which may lead to compilation issues in environments where the C++ compiler does not implicitly include it or if `_Bool` is not defined.
2. There is a comment with an invalid syntax `(result (value))ʳ`, which might be intended to represent something but currently doesn't convey clear information.
3. The class `Start` lacks proper member initialization, potentially leading to uninitialized member variables, although the actual impact depends on how objects of this class are created and used elsewhere in the codebase.

#### Summary of changes

- Moved member variables `FuncIdx`, `Args`, and `Result` from public to private section.
- Removed redundant declarations of member variables that were previously present in both sections.
- Improved encapsulation by restricting direct access to member variables, making the class more robust.

## [include/ast/component/type.h](https://github.com/WasmEdge/WasmEdge/blob/20e9bd4e151725801655df52f01574e7e2ec8441/include%2Fast%2Fcomponent%2Ftype.h)

#### Potential issues

Based on the provided source code and its context, here are the three major issues:

1. **Inclusion of `stdbool.h` is missing**: The subject mentions "Add conditional check for stdbool.h before C23", but there's no inclusion or conditional check for `stdbool.h` in the code.

2. **Lack of type definition for `InterfaceType`**: The code uses `InterfaceType`, but it hasn't been defined anywhere in the provided snippet, which will result in a compilation error.

3. **Incomplete `TupleTy` constructor**: The class `TupleTy` is missing its constructors, causing an error when attempting to instantiate this type.

#### Summary of changes

### Key Changes
1. **Renaming Primitive Types**:
   - `Float32` and `Float64` have been renamed to `F32` and `F64`, respectively.
   - Added a new primitive type, `ErrorContext`.

2. **Refactoring and Renaming Core Module Types**:
   - The enum class `ModuleType` has been renamed to `CoreModuleType`.
   - Correspondingly, `ModuleDecl` has been renamed to `CoreModuleDecl`.

3. **Destructor and Callback Handling in ResourceType**:
   - Adjusted the handling of destructor and callback functions within the `ResourceType` class.
   - Introduced a new boolean field `DtorSync` to indicate whether the destructor is synchronous or asynchronous.

These changes focus on renaming, adding a new primitive type, and refactoring parts of the core module structure for better clarity and functionality.

## [include/common/enum.inc](https://github.com/WasmEdge/WasmEdge/blob/20e9bd4e151725801655df52f01574e7e2ec8441/include%2Fcommon%2Fenum.inc)

#### Potential issues

1. **Macro Usage Without Include Guard**: The use of macros `A`, `O`, etc., without checking for the inclusion of required headers (e.g., `stdbool.h` or specific proposals) can lead to undefined behavior if the environment does not provide these definitions.
2. **Potential Redefinition of Macros**: Defining macros like `A`, `O`, and others in a header file without an include guard might result in multiple definitions if this header is included more than once, which can lead to redefinition errors.
3. **Undefined Behavior Without Guards**: The code relies on the existence of certain macros (`Line_FB`, `Line_FC`, etc.) but does not check for their availability or define fallback behavior if they are not present. This could result in undefined behavior.

#### Summary of changes

Here are the top 3 most important changes:

- **Renaming and Consolidation of Component Model AST Nodes:**
  - Renamed several component model AST nodes for clarity.
  - Added new types for better organization and specificity (e.g., `Comp_Sort`, `Comp_Type_Record`).

- **Restructuring of Error Codes:**
  - Reorganized error codes into specific phases (load, validation, instantiation).
  - Adjusted error code values to avoid conflicts.

- **Removal of Duplicated Code Sections:**
  - Removed duplicate component model load and validation phase sections that were previously present in the file.

## [include/common/types.h](https://github.com/WasmEdge/WasmEdge/blob/20e9bd4e151725801655df52f01574e7e2ec8441/include%2Fcommon%2Ftypes.h)

#### Potential issues

1. **Missing include for `stdbool.h` in C23**: The code does not include `<stdbool.h>` which is required to use `_Bool` type and `true`/`false` constants, potentially leading to compatibility issues with certain compilers or standards.

2. **Redundant template parameter in SIMDArray definition**: In the definition of `SIMDArray`, the conditional compilation block has a redundant `template <typename Ty, size_t TotalSize>` when `TotalSize` is not used outside of the `std::enable_if_t`. It should be removed or utilized properly.

3. **Undefined behavior in ValType constructors**: The constructor `ValType(TypeCode C)` does not handle all possible values of `TypeCode`, leading to potential undefined behavior when an unsupported type code is passed. This can result in `assumingUnreachable()` being called unexpectedly, causing program termination.

#### Summary of changes

- **Default Constructors Added**: The `ValComp` structure now has default constructors for copy and move operations.
- **Constructor Changes in Enum**: The `Enum` structure has a new constructor that accepts an initializer list or a vector of strings.
- **No Destructor Change**: Although there was no change to the destructor, it's important to note that it remains a virtual function.

## [include/loader/loader.h](https://github.com/WasmEdge/WasmEdge/blob/20e9bd4e151725801655df52f01574e7e2ec8441/include%2Floader%2Floader.h)

#### Potential issues

1. **Missing `#include <stdbool.h>` or `<cstdbool>` Check**: The code assumes the availability of `stdbool.h` but doesn't conditionally check for its presence, which could lead to compilation issues on non-C23 compliant environments.
2. **Potential Nullptr Dereference in `loadExecutable`**: There's a risk of dereferencing `IntrinsicsTable` if it's not provided or is null, potentially causing undefined behavior.
3. **Lack of Error Handling in Template Functions**: The template functions like `NodeAttrFromAST<>()`, `loadVec()`, and others could benefit from more robust error handling to manage unexpected inputs or states better.

These issues need addressing for enhanced portability and robustness.

#### Summary of changes

- Added support for various new component types in `NodeAttrFromAST`.
- Modified the `Loader` class to include optional parameters for module loading.
- Updated vector loading logic in `loadVec` method for better error handling.

## [include/runtime/component/hostfunc.h](https://github.com/WasmEdge/WasmEdge/blob/20e9bd4e151725801655df52f01574e7e2ec8441/include%2Fruntime%2Fcomponent%2Fhostfunc.h)

#### Potential issues

1. **Lack of `#include <stdbool.h>` Check**: The code does not include a check for the availability of `<stdbool.h>`, which is necessary for supporting `C23` and later standards.

2. **Undefined `ValComp` Type**: The types `List<T>`, `Record<Types...>`, `Tuple<Types...>`, `Option<T>`, `Enum`, and `Result<V, E>` are used without being defined in the provided code.

3. **Potential Misuse of `std::get`**: The use of `std::get<ValVariant>(V)` assumes that `V` is a variant containing `ValVariant`, which might not be true if `ValInterface` is not correctly typed or initialized.

#### Summary of changes

### Key Changes

1. **Added Specializations for `convert` and `emplace` Templates:**
   - Included specializations for various integer types (`uint8_t`, `uint16_t`, `uint32_t`, `uint64_t`, `int8_t`, `int16_t`) in both the `convert` and `emplace` template structs.

2. **Added Support for Complex Types:**
   - Introduced specializations for `Record<Types...>`, `Tuple<Types...>`, `Option<T>`, `Enum`, `Result<V, E>`, and `WasmEdge::Component::Variant<Types...>` in both the `convert` and `emplace` template structs.

3. **Expanded Functionality of `ValInterface`:**
   - Enhanced the `emplace` function for various data types to support emplacing values directly into `ValInterface`.

## [include/runtime/instance/component.h](https://github.com/WasmEdge/WasmEdge/blob/20e9bd4e151725801655df52f01574e7e2ec8441/include%2Fruntime%2Finstance%2Fcomponent.h)

#### Potential issues

1. **Missing Include Guard for `stdbool.h`**: The code does not check if `stdbool.h` is available before using it, which can cause compilation issues in environments where this header is not present (e.g., C++23).

   ```cpp
   #if __cplusplus >= 201709L
   // stdbool.h is included by default since C++17
   #else
   #include <stdbool.h>
   #endif
   ```

2. **Potential Uninitialized Variable Access**: In the `typeConvert` function for `FuncType`, if `Ty.getReturnTypes().size()` is 0, `RL.emplace<std::vector<LabelValType>>();` will be called unnecessarily, leading to potential uninitialized variable access.

3. **Unsafe Type Conversion in `typeConvert`**: The type conversion functions do not handle all possible types defined in `AST::Component::TypeCode`, which could lead to unexpected behavior if an unsupported type is encountered.

#### Summary of changes

- Renamed `PrimValType::Float32` to `PrimValType::F32`
- Renamed `PrimValType::Float64` to `PrimValType::F64`

## [include/validator/validator_component.h](https://github.com/WasmEdge/WasmEdge/blob/20e9bd4e151725801655df52f01574e7e2ec8441/include%2Fvalidator%2Fvalidator_component.h)

#### Potential issues

1. **Unconditional Header Inclusion**: The code does not include `<stdbool.h>`, which is necessary for the `_Bool` type used in C99 and later standards.
2. **Redundant `Expect<void>` Return Types**: All visitor methods return `Expect<void>`, but they do not utilize the `Unexpected` part of the `std::expected` structure, making them less useful than simple `void`.
3. **Uninitialized Members**: In some cases, members are accessed without initializing or validating their existence first (e.g., in `AliasTargetVisitor`), leading to potential runtime errors.

#### Summary of changes

- Renamed `ModuleDeclVisitor` to `CoreModuleDeclVisitor`
- Changed the type handled by `operator()` in `CoreDefTypeVisitor` from `const ModuleType&` to `const CoreModuleType&`.
- Updated the loop in `CoreDefTypeVisitor::operator()` to iterate over `const CoreModuleDecl &D`.

## [lib/executor/instantiate/component/instantiate_component_alias.cpp](https://github.com/WasmEdge/WasmEdge/blob/20e9bd4e151725801655df52f01574e7e2ec8441/lib%2Fexecutor%2Finstantiate%2Fcomponent%2Finstantiate_component_alias.cpp)

#### Potential issues

1. **Potential Null Pointer Dereference**: The code does not check if `ModInst`, `FuncInst`, `TableInst`, `MemInst`, or `CInst` are null before dereferencing them, which could lead to a crash.
2. **Inconsistent Error Handling**: Some cases in the switch statements log warnings but do not return an error, leading to inconsistent error handling and potential silent failures.
3. **Unused Variable `Expect<void>`**: The function returns `Expect<void>`, but it does not handle or propagate any errors explicitly within its body.

#### Summary of changes

Sure, here are the most important changes:

1. **Added conditional handling for `Table` CoreSort:**
   - Added code to handle `Table` CoreSort by obtaining table exports and adding them as a core table instance.
   - Replaced warning message with actual implementation.

2. **Refactored Code Structure:**
   - Encased the logic for `Case::Table` in braces (`{}`) to allow local variable declarations within the switch case.

3. **Comment Removed and Replaced by Implementation:**
   - Removed the comment indicating "incomplete core alias sort: table" and replaced it with functional code that processes the Table instance correctly.

## [lib/executor/instantiate/component/instantiate_component_canon.cpp](https://github.com/WasmEdge/WasmEdge/blob/20e9bd4e151725801655df52f01574e7e2ec8441/lib%2Fexecutor%2Finstantiate%2Fcomponent%2Finstantiate_component_canon.cpp)

#### Potential issues

1. **Missing header include for C++20 features**: The code uses `<sstream>` which is not part of the standard library for handling strings and I/O in modern C++. It should be replaced with `<string_view>`.

```cpp
#include <sstream> // Consider replacing or removing if not needed
```

2. **Potential undefined behavior due to uninitialized variables**: The `FuncType` variable in the `LiftTrans` constructor is used before being initialized.

```cpp
FuncType = convert(Comp, DefinedType);
spdlog::info("lifted: {}"sv, FuncType); // Ensure `FuncType` is initialized here or earlier.
```

3. **Potential memory leak in unique_ptr assignment**: The use of `std::unique_ptr` in the `Executor::instantiate` method may lead to memory leaks if exceptions are thrown.

```cpp
CompInst.addFunctionInstance(lifting(CompInst, std::get<FuncType>(AstFuncType), FuncInst, Mem, ReallocFunc)); // Ensure no exceptions thrown.
```
Ensure that all operations involving `std::unique_ptr` are exception-safe.

#### Summary of changes

### Key Changes:

1. **Addition of `PrimValType::ErrorContext` Handling**:
   - Introduced a new case for handling the type `PrimValType::ErrorContext`.
   - Added a warning message using `spdlog::warn`.

2. **Update Error Attribute from `ASTNodeAttr::Sec_Canon` to `ASTNodeAttr::Comp_Sec_Canon`**:
   - Changed error reporting in multiple locations to use `ASTNodeAttr::Comp_Sec_Canon` instead of `ASTNodeAttr::Sec_Canon`.

3. **Correction of PrimValType Enumeration Cases**:
   - Updated the cases for `PrimValType::F32`, and `PrimValType::F64`.

## [lib/executor/instantiate/component/instantiate_component_export.cpp](https://github.com/WasmEdge/WasmEdge/blob/20e9bd4e151725801655df52f01574e7e2ec8441/lib%2Fexecutor%2Finstantiate%2Fcomponent%2Finstantiate_component_export.cpp)

#### Potential issues

1. The code does not check for the inclusion of `<stdbool.h>` before C23, which is required for compatibility with older C standards.
2. Missing error handling for unsupported `SortCase` values in the switch statement, potentially leading to undefined behavior.
3. Inconsistent logging levels (using both `error` and `warn`) without a clear rationale, which may affect debugging and maintenance clarity.

#### Summary of changes

1. Changed the error message logging to use `Comp_Sec_Export` instead of `Sec_CompExport`.
2. The patch does not contain any changes related to "Add conditional check for stdbool.h before C23" as specified in your request.
3. No other significant changes were made in this file according to the provided patch.

## [lib/executor/instantiate/component/instantiate_component_import.cpp](https://github.com/WasmEdge/WasmEdge/blob/20e9bd4e151725801655df52f01574e7e2ec8441/lib%2Fexecutor%2Finstantiate%2Fcomponent%2Finstantiate_component_import.cpp)

#### Potential issues

1. **Missing Include Guard**: The source code lacks include guards or `#pragma once` to prevent multiple inclusions of the header file, which could lead to redefinition errors.
2. **Unused Includes**: The `<string_view>` and `<variant>` headers are included but not used in the provided code snippet, leading to unnecessary dependencies.
3. **Incomplete Functionality**: The `Executor::instantiate` function contains multiple TODOs indicating incomplete functionality, which should be addressed for complete implementation.

These issues are prioritized based on their potential impact on compilation and code maintainability.

#### Summary of changes

- **Renaming an Enumeration Value**: The patch changes the enumeration value `ASTNodeAttr::Sec_CompImport` to `ASTNodeAttr::Comp_Sec_Import`. This is a significant change as it alters how certain AST (Abstract Syntax Tree) nodes are identified and referenced.
  - **Error Reporting**: The modified code updates the error reporting mechanism by changing the enumeration used in the call to `spdlog::error()`, which could impact debugging and logging.

- **Consistency with Naming Conventions**: The change ensures consistency with potential naming conventions or refactoring efforts that may have occurred elsewhere in the project, making the code more maintainable.

## [lib/loader/CMakeLists.txt](https://github.com/WasmEdge/WasmEdge/blob/20e9bd4e151725801655df52f01574e7e2ec8441/lib%2Floader%2FCMakeLists.txt)

#### Potential issues

1. **Missing CMake Minimum Version**: The script does not specify the minimum required version of CMake, which is crucial for ensuring compatibility with all features and commands used.

2. **Unsupported Standard Library Feature**: The usage of `std::filesystem` is conditionally linked without checking if the compiler supports C++17 or later, where this feature was introduced.

3. **Potential Missing Header Inclusion**: There is no explicit inclusion check for `<filesystem>` in the source files that use it, which may cause compilation errors on systems where it's not available by default.

#### Summary of changes

**Key Changes:**

1. **Reordering and Splitting of Source Files:**
   - `component_import_export.cpp` is split into two separate files: `component_import.cpp` and `component_export.cpp`.
   - The order of some source files in the `wasmedgeLoader` target is changed.

2. **Addition of New Source Files:**
   - Three new source files are added to the `wasmedgeLoader` target:
     - `ast/component/component_section.cpp`
     - `ast/component/component_instance.cpp`
     - `ast/component/component_sort.cpp`

3. **Removal of Previously Included Source File:**
   - The previously included `component_import_export.cpp` is removed from the list.

## [lib/loader/ast/component/component.cpp](https://github.com/WasmEdge/WasmEdge/blob/20e9bd4e151725801655df52f01574e7e2ec8441/lib%2Floader%2Fast%2Fcomponent%2Fcomponent.cpp)

#### Potential issues

1. **Missing Conditional Check for `stdbool.h`**: The code does not include a conditional check for `stdbool.h` before C23, which could lead to compatibility issues on platforms where `<stdbool.h>` is not available or behaves differently.

2. **Inconsistent Error Handling**: The error handling in the `loadComponent` function uses both `logLoadError` and a lambda function `ReportError`, leading to inconsistent error reporting styles.

3. **Potential Infinite Loop in `loadComponent`**: The while loop condition `(!Bound.has_value() || *Bound > Offset - StartOffset)` could potentially lead to an infinite loop if `FMgr.readByte()` always succeeds, and there is no increment or exit condition based on the content read.

#### Summary of changes

- The key change is the addition of a conditional check for `stdbool.h` before C23. This likely means the code now includes or handles cases where `stdbool.h` might not be available in future C standards.
- Replaced repetitive calls to `Comp.getSections().emplace_back()` with a single call at the beginning of each section case block, streamlining the code and reducing redundancy.
- Updated error reporting to use consistent logging levels (e.g., changed `ASTNodeAttr::Component` to `ASTNodeAttr::Module`) for better uniformity and potentially more accurate error context.

## [lib/loader/ast/component/component_alias.cpp](https://github.com/WasmEdge/WasmEdge/blob/20e9bd4e151725801655df52f01574e7e2ec8441/lib%2Floader%2Fast%2Fcomponent%2Fcomponent_alias.cpp)

#### Potential issues

1. **Missing `#include <stdbool.h>` for C23:** The code does not include `<stdbool.h>`, which is necessary in C23 to use `true` and `false`. This could lead to compilation issues or incorrect behavior.

2. **Unchecked Return Values:** The return value of `FMgr.readByte()` is used directly without checking if it's valid before assigning it to `Flag`.

3. **Potential Memory Leak:** In `loadAliasTarget`, if an error occurs after emplace but before returning, the allocated memory in `AliasTarget` might not be properly handled or deallocated.

#### Summary of changes

1. **Conditional Check for `stdbool.h`**: The code includes conditional compilation checks to ensure compatibility with C23, which deprecates the use of `<stdbool.h>`. This ensures that the code can be compiled and run in environments where `<stdbool.h>` is not available or deprecated.

2. **Namespace Changes**: The namespace `using namespace AST::Component;` has been removed from the top of the file, making the code more explicit about where `Alias`, `AliasTarget`, and other related components are defined.

3. **Error Reporting Enhancements**: The error reporting mechanism has been improved for better diagnostics. The `logLoadError` function is now used consistently to log errors with appropriate context information.

GautamBytes:
@hydai PR is ready for review now. Thanks!

hydai (changes requested):
Please fix the CI first.  
I think you should also run `clang-format` in your own environment to avoid lint issues.

bindings/java/wasmedge-java/wasmedge-jni/lib/common.h
```diff
@@ -51,7 +51,7 @@ void setJavaDoubleValue(JNIEnv *env, WasmEdge_Value val, jobject jobj);
 
 void setJavaStringValue(JNIEnv *env, char *val, jobject jobj);
 
-enum WasmEdge_ValType *parseValueTypes(JNIEnv *env, jintArray jValueTypes);
+WasmEdge_ValType *parseValueTypes(JNIEnv *env, jintArray jValueTypes);
```
hydai: What is the reason for modifying this unrelated file?

hydai (changes requested):

bindings/java/wasmedge-java/wasmedge-jni/lib/common.h
GautamBytes: This fix was necessary because the JNI function was declared as

```c
enum WasmEdge_ValType *parseValueTypes(...);
```

but in `wasmedge.h` we’ve typedef’d the enum as `WasmEdge_ValType`, so the compiler complained about a tag/type mismatch and the Java binding wouldn’t build. Changing it to

```c
WasmEdge_ValType *parseValueTypes(...);
```

aligns the declaration with the typedef and resolves that compile error.

hydai: The current situation is that JNI is completely outdated, and someone is going to work on it. Therefore, you can ignore the Java binding-related workflows.

GautamBytes: Okay sure , reverted the change!

PR #4154, opened 2025-06-09
ci(commitlint): use label for showing the commitlint result

hydai:


juntao:

------

## [.github/workflows/commitlint.yml](https://github.com/WasmEdge/WasmEdge/blob/c241cffba72ad46afa98a9e22cda7045911f01ba/.github%2Fworkflows%2Fcommitlint.yml)

#### Potential issues

1. **Ubuntu Version Issue**: `ubuntu-24.04` is not a valid Ubuntu version; use an existing and supported version like `ubuntu-22.04`.

2. **Commit Range Calculation**: `${{ github.event.pull_request.commits }}` should be used to calculate the commit range correctly, but it might lead to incorrect ranges if the number of commits is zero or negative; consider using a more robust method.

3. **Label Removal Error Handling**: The error handling for label removal does not differentiate between errors due to non-existent labels and other issues; consider enhancing the error handling to provide more clarity.

#### Summary of changes

1. **Label and Comment Management**: The patch introduces a new system to manage GitHub labels (commitlint-pass, commitlint-fail) and comments for each PR based on the commitlint validation result.
2. **Fail Job on Failure**: It adds a step to fail the job if commitlint validation fails, providing clear guidance on how to fix commit messages.
3. **Update Run Environment**: The runner's OS version is updated from ubuntu-22.04 to ubuntu-24.04.

hydai:
@0yi0 
Would you consider configuring the workflow to leave a comment with the commitlint result under the PR? It will be updated when the workflow produces a new result.

If so, I can add it at the same PR.

hydai:
`commitlint check okay`:

CleanShot 2025-06-10 at 10 46 57


`commitlint check failed`:

Google Chrome 2025-06-10 10 44 57

0yi0 (commented):
Consider replace all `always()` with `!cancelled()` according to [the docs of `always()`](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#always).

hydai:
> Consider replace all `always()` with `!cancelled()` according to [the docs of `always()`](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#always).

Sounds great. Done.

0yi0 (approved):

PR #4165, opened 2025-06-12
ci(commitlint): use updateComment instead of re-creating a new one

hydai:


juntao:

------

## [.github/workflows/commitlint.yml](https://github.com/WasmEdge/WasmEdge/blob/0bb788ff200c4ec27a1a97f76c72c5b42886e205/.github%2Fworkflows%2Fcommitlint.yml)

#### Potential issues

1. **Potential Error in Fetching Commits Range**: The expression `${{ github.event.pull_request.commits }}` might not work as intended, especially if the number of commits is 1, which could lead to invalid range syntax.

2. **Potential for Unnecessary API Calls**: In both success and failure cases, it checks all comments each time a new label/comment needs to be added or updated, which can be inefficient for large PRs with many comments.

3. **Non-Atomic Operation Risk**: Removing labels and adding them back in separate steps can lead to inconsistencies if the workflow is interrupted between these actions.

#### Summary of changes

### Key Changes:
1. **Refactored Comment Handling:**
   - Updated code to use `updateComment` instead of re-creating new comments for commitlint status updates.
   - Added logic to find existing comments and update them if they exist, or create a new comment otherwise.

2. **Removed Redundant Step:**
   - Removed the "Remove previous commitlint comments" step as it is no longer necessary with the updated approach.

3. **Added 'ci' Label to Valid Types List:**
   - Updated the list of valid types in the failure comment message to include `ci`.

hydai:
In the previous PR #4154, this workflow attempted to delete the previous result comment and create a new one. However, other workflows, such as the review bot and the `codecov` bot, do not do this; instead, they update the original comments.

After applying this PR, the output will be:

CleanShot 2025-06-12 at 15 52 49

cc @0yi0

0yi0 (approved):

PR #4168, opened 2025-06-12
ci(commitlint): add missing permissions and github tokens

hydai:
Still failed, check https://github.com/WasmEdge/WasmEdge/actions/runs/15611065731/job/43972967501?pr=4134

I changed how this workflow is triggered. It should have the same permissions as the issue labeler workflow.

### DIFF
--- a/.github/workflows/commitlint.yml
+++ b/.github/workflows/commitlint.yml
@@ -1,6 +1,11 @@
-name: Commit Lint
+name: "Commit Lint"
 
-on: [pull_request]
+on:
+  pull_request_target:
+    types:
+      - opened
+      - synchronize
+      - reopened
 
 permissions:
   contents: read
@@ -9,6 +14,7 @@ jobs:
     permissions:
       contents: read
       pull-requests: write
+      issues: write
     runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


juntao:

------

## [.github/workflows/commitlint.yml](https://github.com/WasmEdge/WasmEdge/blob/4191774a722ce952f3164b4cfa8956042b6dac75/.github%2Fworkflows%2Fcommitlint.yml)

#### Potential issues

1. The workflow uses `ubuntu-24.04` as the runner, which may not exist; consider using a supported version like `ubuntu-latest`.
2. In the "Install commitlint" step, install dependencies in one command to avoid redundant package installations.
3. The workflow does not handle errors from `npx commitlint`, which may lead to false negatives or positives.

#### Summary of changes

1. **Added `pull_request_target` trigger with specific event types:**
   - Changed the trigger from `pull_request` to `pull_request_target`.
   - Added types: `opened`, `synchronize`, and `reopened`.

2. **Updated permissions for job:**
   - Included additional permission: `issues: write`.

3. **Added GitHub token in the script:**
   - Injected `${{ secrets.GITHUB_TOKEN }}` as a parameter to the action.
   - This allows the workflow to authenticate API requests using GitHub's default token.

These changes ensure that the Commit Lint job runs on specific pull request events, has necessary permissions for required actions, and is authenticated with the GitHub token.

Issue #4171, opened 2025-06-12
bug: commitlint failed on the forked repo


hydai:
### Summary

Please check the failed one: https://github.com/WasmEdge/WasmEdge/actions/runs/15613481927/job/43981202301
This is a PR from a forked repo.

Please also check the passed one: https://github.com/WasmEdge/WasmEdge/actions/runs/15613564901/job/43980374056?pr=4160
This is a PR from this repo.

### Current State

```txt
Run npx commitlint --from 7ba2412f326c41e2cf7930052963a79c159cf29f~1 --to 7ba2412f326c41e2cf7930052963a79c159cf29f --verbose
file:///home/runner/work/WasmEdge/WasmEdge/node_modules/@commitlint/cli/lib/cli.js:132
        throw err;
        ^

Error: fatal: ambiguous argument '7ba2[4](https://github.com/WasmEdge/WasmEdge/actions/runs/15613481927/job/43981202301#step:6:5)12f326c41e2cf7930052963a79c159cf29f~1..7ba2412f326c41e2cf7930052963a79c159cf29f': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'

    at Transform.transform [as _transform] (/home/runner/work/WasmEdge/WasmEdge/node_modules/git-raw-commits/index.js:100:36)
    at Transform._write (node:internal/streams/transform:171:8)
    at writeOrBuffer (node:internal/streams/writable:[5](https://github.com/WasmEdge/WasmEdge/actions/runs/15613481927/job/43981202301#step:6:6)72:12)
    at _write (node:internal/streams/writable:501:10)
    at Writable.write (node:internal/streams/writable:510:10)
    at Socket.ondata (node:internal/streams/readable:1009:22)
    at Socket.emit (node:events:53[6](https://github.com/WasmEdge/WasmEdge/actions/runs/15613481927/job/43981202301#step:6:7):35)
    at addChunk (node:internal/streams/readable:561:12)
    at readableAddChunkPushByteMode (node:internal/streams/readable:512:3)
    at Readable.push (node:internal/streams/readable:3[9](https://github.com/WasmEdge/WasmEdge/actions/runs/15613481927/job/43981202301#step:6:10)2:5)

Node.js v20.19.2
```

### Expected State

The workflow should show `pass` or `fail`.

### Reproduction steps

1. Just create a PR from a forked repo.

### Screenshots

[...]

PR #4174, opened 2025-06-13
ci(commitlint): EXPECTED FAILED

hydai:
This is for verifying #4171 
DONT MERGE


juntao:

------

## [.github/workflows/commitlint.yml](https://github.com/WasmEdge/WasmEdge/blob/be716e7e467910dfcb63fdd9706aca75225b6077/.github%2Fworkflows%2Fcommitlint.yml)

#### Potential issues

1. **Incorrect Node.js installation command**: The script installs the latest version of Node.js from the nodesource repository, which might not be compatible with the desired setup.

2. **Unnecessary `continue-on-error` in commitlint step**: The `continue-on-error: true` directive should be removed as it masks potential errors and could lead to false positives or negatives in the commitlint validation process.

3. **Potential race condition in label management**: The script removes labels before adding new ones, which could lead to a brief period where no labels are applied if there's an error during removal; consider checking for existing labels first.

#### Summary of changes

- **Use GitHub Actions `GITHUB_SHA` instead of `head.sha`**: The most important change is replacing the `head.sha` reference with `$(git -C "${GITHUB_WORKSPACE}" rev-parse HEAD)` to ensure it uses the current commit SHA from the local workspace, which is more reliable in the context of GitHub Actions.

- **No additional changes**: Other parts of the script remain unchanged. This focus ensures minimal disruption while fixing the potential issue related to using `head.sha`.

- **Retains existing conditional logic and error handling**: The change keeps the `if` condition for PR targets and continues-on-error behavior, preserving the original intention and error handling strategy.

PR #4184, opened 2025-06-16
test: remove `O0` for universal wasm tests

q82419:


juntao:

------

## [test/api/APIAOTCoreTest.cpp](https://github.com/WasmEdge/WasmEdge/blob/18141b116e90c32a2ff8e00924287c5e49edc4ce/test%2Fapi%2FAPIAOTCoreTest.cpp)

#### Potential issues

1. **Memory Leak**: `WasmEdge_ConfigureContext *ConfCxt` is created but never deleted, leading to a memory leak.
2. **Exception Safety**: The code does not handle potential exceptions from `std::filesystem::u8path` operations, which could lead to undefined behavior if an exception is thrown.
3. **Redundant Validations**: In the `T.onValidate`, `T.onInstantiate` lambdas, `WasmEdge_VMValidate(VM)` is called twice unnecessarily.

#### Summary of changes

- Removed the line `WasmEdge_ConfigureCompilerSetOptimizationLevel(ConfCxt, WasmEdge_CompilerOptimizationLevel_O0)` which sets the optimization level to O0 for universal wasm tests.
- The removal implies that the default optimization level will now be used instead of explicitly setting it to the lowest level (O0).
- This change might affect the performance and size of the compiled WebAssembly modules in the test cases.

## [test/api/APIUnitTest.cpp](https://github.com/WasmEdge/WasmEdge/blob/18141b116e90c32a2ff8e00924287c5e49edc4ce/test%2Fapi%2FAPIUnitTest.cpp)

#### Potential issues

**Issue 1: Missing Null Check for `hexToFile`:** The function `hexToFile` does not check if the provided file path is null, which can lead to undefined behavior.

**Issue 2: Potential Memory Leak in `createExternModule`:** In the `createExternModule` function, there is a possibility of memory leaks because the created host functions (`HostFunc`) are not being deleted explicitly before returning.

**Issue 3: Uninitialized Variable in `externAdd`, `externSub`, etc.:** The functions such as `externAdd`, `externSub`, and others do not initialize the output variable `Out[0]` if input validation fails, leading to undefined behavior.

#### Summary of changes

1. Removed `WasmEdge_ConfigureCompilerSetOptimizationLevel` setting to O0 for universal WASM tests.
2. The order of operations in the test remains unchanged, maintaining compilation process integrity.
3. No other changes were made to the test, ensuring minimal disruption to existing functionality.

## [test/llvm/LLVMcoreTest.cpp](https://github.com/WasmEdge/WasmEdge/blob/18141b116e90c32a2ff8e00924287c5e49edc4ce/test%2Fllvm%2FLLVMcoreTest.cpp)

#### Potential issues

1. **Redundant Compiler Configure Settings**: The `setDumpIR(true)` in both NativeCoreTest and CustomWasmCoreTest classes is redundant since it doesn't affect the test outcomes, which might confuse future maintainers or mislead into thinking IR dumping is a part of the test's intent.

2. **Uninitialized Variable `Conf`**: In `TEST(Configure, ConfigureTest)`, the variable `Conf` is used uninitialized in the first block, which could lead to undefined behavior; it should be initialized before passing to `LLVM::Compiler`.

3. **Hardcoded Paths for Temporary Files**: The use of hardcoded paths (`std::filesystem::temp_directory_path()`) for temporary files in multiple tests can cause issues if these paths collide, and they lack proper cleanup in some cases (e.g., test failures or interruptions), potentially leading to file leaks.

#### Summary of changes

- **Removed O0 Optimization Level Setting**: The primary change is the removal of setting `OptimizationLevel::O0` in three test cases: `NativeCoreTest`, `CustomWasmCoreTest`, and `JITCoreTest`. This change was made to remove unnecessary optimization level configuration, allowing the compiler to use its default settings.

- **Consistency Across Tests**: By removing this explicit setting, the tests now follow a more consistent approach, not enforcing an optimization level that might be redundant or non-standard in these contexts.

q82419:
Seems like issues occur on Windows. Close this.

Issue #4142, opened 2025-06-05
feat: update the wasi-nn openvino for fixing the misaligned layout issue


hydai:
### Summary

The input layout must be fixed to NHWC in the OpenVINO backend.

Previously, we implemented [the NHWC conversion inside the plugin](https://github.com/WasmEdge/WasmEdge/blob/master/plugins/wasi_nn/wasinn_openvino.cpp#L104-L113).

After updating the examples, we are now able to ensure that the application sends the correct NHWC layout, so we don't need to make this tricky for users.

### Details

Original issue:

> @hydai 
> can you or someone take care of the other example (openvino-mobilenet-raw) and wasmedge?
i have not used wasi-nn with wasmedge.

> _Originally posted by @yamt in https://github.com/second-state/WasmEdge-WASINN-examples/issues/190#issuecomment-2943087996_

---

Please also refer to the following issues for the detailed information:

1. https://github.com/bytecodealliance/wasmtime/issues/10867#issuecomment-2927998250
2. https://github.com/bytecodealliance/wasmtime/issues/4464

hydai:
@LFsWang Can you please follow up this issue?

LFsWang:
ok

PR #4185, opened 2025-06-16
fix(wasi_nn): update input tensor shape and simplify model compilation

LFsWang:
#4142 
https://github.com/second-state/WasmEdge-WASINN-examples/pull/190

No longer adjust the input shape from `NHWC` to `NCHW` again.


juntao:

------

## [plugins/wasi_nn/wasinn_openvino.cpp](https://github.com/WasmEdge/WasmEdge/blob/4b249e392bc1b56a24b04f293858116e738a2f58/plugins%2Fwasi_nn%2Fwasinn_openvino.cpp)

#### Potential issues

1. **Hardcoded Input Shape**: In the `setInput` function, the input tensor shape is hardcoded to `{1, 3, 224, 224}` which may not be flexible enough for all use cases.

   ```cpp
   ov::Shape InputShape = {1, 3, 224, 224};
   ```

2. **Lack of Error Handling**: The code does not handle the case where `Env.newGraph` or `Env.newContext` returns an invalid graph or context ID.

   ```cpp
   uint32_t GId = Env.newGraph(Backend::OpenVINO);
   ContextId = Env.newContext(GraphId, Env.NNGraph[GraphId]);
   ```

3. **Unchecked Tensor Data Type**: The `setInput` function does not validate if the input tensor data type matches the expected `F32`, which could lead to runtime errors.

   ```cpp
   ov::element::Type InputType = ov::element::f32;
   ```

#### Summary of changes

- Changed input tensor shape from `{1, 224, 224, 3}` to `{1, 3, 224, 224}`, which aligns with the NCHW format.
- Removed preprocessing steps for setting layout and resizing, simplifying model compilation.
- Directly compiled `GraphRef.OpenVINOModel` instead of using `PPP.build()`, reducing redundancy.

## [test/plugins/wasi_nn/wasi_nn.cpp](https://github.com/WasmEdge/WasmEdge/blob/4b249e392bc1b56a24b04f293858116e738a2f58/test%2Fplugins%2Fwasi_nn%2Fwasi_nn.cpp)

#### Potential issues

### Major Issues

1. **Potential Memory Safety Issue**: The code uses `readEntireFile` function to read file contents into a vector, but it does not check if the file size exceeds the maximum possible size for a standard `std::vector<uint8_t>`. This could lead to an out-of-bounds memory allocation.
   - **Fix**: Add a check on the file size before attempting to allocate memory.

2. **Potential Null Pointer Dereference**: The function `createModule` dynamically casts a unique pointer but does not check if the cast was successful, which can result in a null pointer dereference if the cast fails.
   - **Fix**: Verify that the dynamic cast is successful by checking the pointer before proceeding.

3. **Potential Integer Overflow**: The code uses `std::vector<uint8_t>` to store file contents and perform arithmetic operations on the indices without checking for potential integer overflows, especially when handling large files or memory pointers.
   - **Fix**: Ensure that all arithmetic operations involving memory sizes and pointers are safe from overflow by using appropriate checks.

#### Summary of changes

### Key Changes

1. **Update Input Tensor Shape Conversion**:
   - The input tensor is now converted from NHWC to NCHW format. This change aligns with the OpenVINO model's expectation, which requires input tensors in NCHW format.

2. **Legacy Tensor Handling**:
   - Introduced `TensorDataLegecy` to handle the legacy input tensor data and perform necessary transformations before using it in the test case.

3. **Code Comments Added for Clarity**:
   - Additional comments explaining the conversion process and referencing relevant external sources have been added for better understanding of the changes made.

hydai:
Looks like the tests should be updated too.

plugins/wasi_nn/wasinn_openvino.cpp
```diff
@@ -101,23 +101,11 @@ Expect<WASINN::ErrNo> setInput(WASINN::WasiNNEnvironment &Env,
...
-    ov::preprocess::PrePostProcessor PPP(GraphRef.OpenVINOModel);
-    PPP.input()
-        .tensor()
-        .set_shape(InputShape)
-        .set_element_type(InputType)
-        .set_layout(InputLayout);
-    PPP.input().preprocess().resize(
-        ov::preprocess::ResizeAlgorithm::RESIZE_LINEAR);
-    PPP.input().model().set_layout("NCHW");
-    PPP.output().tensor().set_element_type(ov::element::f32);
```
yamt, line 117: this postprocessor removal would allows wasmedge to run models with non-fp32 outputs.
eg. https://github.com/openvinotoolkit/open_model_zoo/tree/cd7ebe313b69372763e76b82e5d24935308fece4/models/public/license-plate-recognition-barrier-0007

yamt (commented):
thank you.

Issue #4193, opened 2025-06-17
bug: replace the current sbom workflow with REST API


hydai:
### Summary

WasmEdge is using [advanced-security/sbom-generator-action](https://github.com/advanced-security/sbom-generator-action) to generate SBOMs during the [release process](https://github.com/WasmEdge/WasmEdge/blob/master/.github/workflows/release.yml#L136-L164). However, this action is archived, and it recommends that users switch to the [GitHub REST API](https://docs.github.com/en/rest/dependency-graph/sboms?apiVersion=2022-11-28#export-a-software-bill-of-materials-sbom-for-a-repository) instead.

We should replace the current workflow with this new REST API.

### Current State

It works, but the action is archived.

### Expected State

Use the REST API to generate the same SBOM tarball.

### Reproduction steps

Please check the release workflows for the details.

### Screenshots

![Image](https://github.com/user-attachments/assets/4b6af2db-a134-4b0d-9526-0adc7c0c94c2)

### Components

Others

### WasmEdge Version or Commit you used

master

### Operating system information

none

### Hardware Architecture

none

### Compiler flags and options

none

PR #4196, opened 2025-06-19
ci(sbom): migrate SBOM generation to REST API

PhantomInTheWire:
fixes: https://github.com/WasmEdge/WasmEdge/issues/4193
trial run: https://github.com/PhantomInTheWire/WasmEdge/actions/runs/15766897704/job/44444967299
generated somb: [sbom.tar.gz](https://github.com/PhantomInTheWire/WasmEdge/releases/download/v0.0.1-test/sbom.tar.gz)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -149,16 +149,45 @@ jobs:
       - name: Grant the safe directory for git
         run: |
           git config --global --add safe.directory $(pwd)
-      - name: generate sbom
-        uses: advanced-security/sbom-generator-action@6fe43abf522b2e7a19bc769aec1e6c848614b517 # v0.0.2
-        id: sbom
+
+      - name: Generate SBOM
+        id: generate_sbom
+        run: |
+          echo "Generating SBOM for repository: $GITHUB_REPOSITORY"
+
+          curl --retry 5 \
+            --retry-delay 2 \
+            --retry-connrefused \
+            -L \
+            -H "Accept: application/vnd.github+json" \
+            -H "Authorization: Bearer ${GITHUB_TOKEN}" \
+            -H "X-GitHub-Api-Version: 2022-11-28" \
+            "https://api.github.com/repos/${GITHUB_REPOSITORY}/dependency-graph/sbom" \
+            -o sbom.json
+
+          if [ ! -s sbom.json ]; then
+            echo "Error: Failed to generate SBOM or received empty response"
+            exit 1
+          fi
+
+          if ! jq empty sbom.json 2>/dev/null; then
+            echo "Error: Generated SBOM is not valid JSON"
+            cat sbom.json
+            exit 1
+          fi
+
+          echo "SBOM generated successfully"
         env:
           GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
 
-      - name: Generate tar.gz file and upload release
+      - name: Create SBOM tarball and upload to release
         shell: bash
         run: |
-          tar -zcvf sbom.tar.gz ${{steps.sbom.outputs.fileName}}
-          gh release upload ${{needs.create_release.outputs.version}} sbom.tar.gz --clobber
+          SBOM_FILENAME="sbom_${VERSION}.json"
+          mv sbom.json "$SBOM_FILENAME"
+          tar -zcvf sbom.tar.gz "$SBOM_FILENAME"
+          gh release upload $VERSION sbom.tar.gz --clobber
+          echo "SBOM uploaded to release as sbom.tar.gz"
         env:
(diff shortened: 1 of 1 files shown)


juntao:

------

## [.github/workflows/release.yml](https://github.com/WasmEdge/WasmEdge/blob/8a0fe7ca739fad01f1d81893a0b6005dd7fe7a00/.github%2Fworkflows%2Frelease.yml)

#### Potential issues

N/A

#### Summary of changes

N/A

.github/workflows/release.yml
```diff
@@ -149,16 +149,51 @@ jobs:
...
-        env:
-          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
 
-      - name: Generate tar.gz file and upload release
-        shell: bash
+      - name: Generate SBOM
+        id: generate_sbom
+        run: |
+          OWNER=$(echo $GITHUB_REPOSITORY | cut -d'/' -f1)
+          REPO=$(echo $GITHUB_REPOSITORY | cut -d'/' -f2)
```
hydai: Since the OWNER and REPO are static, why do we need to parse them from an external variable?

.github/workflows/release.yml
```diff
@@ -149,16 +149,51 @@ jobs:
...
+          if [ ! -s sbom.json ]; then
+            echo "Error: Failed to generate SBOM or received empty response"
+            exit 1
+          fi
+          
+          if ! jq empty sbom.json 2>/dev/null; then
+            echo "Error: Generated SBOM is not valid JSON"
+            cat sbom.json
+            exit 1
+          fi
```
hydai: The GitHub API may fail; maybe we should implement a retry mechanism to ensure this workflow does not break due to temporary network issues.

.github/workflows/release.yml
```diff
@@ -149,16 +149,51 @@ jobs:
...
+          
+          echo "SBOM generated successfully"
+          echo "SBOM size: $(wc -c < sbom.json) bytes"
+          
+          jq . sbom.json > sbom-formatted.json
+          mv sbom-formatted.json sbom.json
```
hydai: What is the reason for formatting this JSON file?

.github/workflows/release.yml
```diff
@@ -149,16 +149,51 @@ jobs:
...
+          fi
+          
+          if ! jq empty sbom.json 2>/dev/null; then
+            echo "Error: Generated SBOM is not valid JSON"
+            cat sbom.json
+            exit 1
+          fi
+          
+          echo "SBOM generated successfully"
+          echo "SBOM size: $(wc -c < sbom.json) bytes"
```
hydai: Why do we need this information? Will it help us verify something?

.github/workflows/release.yml
```diff
@@ -149,16 +149,51 @@ jobs:
...
+          
+          jq . sbom.json > sbom-formatted.json
+          mv sbom-formatted.json sbom.json
+
+      - name: Create SBOM tarball and upload to release
         run: |
-          tar -zcvf sbom.tar.gz ${{steps.sbom.outputs.fileName}}
-          gh release upload ${{needs.create_release.outputs.version}} sbom.tar.gz --clobber
+          TIMESTAMP=$(date -u +"%Y%m%d_%H%M%S")
+          SBOM_FILENAME="sbom_${TIMESTAMP}.json"
```
hydai: What is the reason for using a timestamp instead of the release version?

.github/workflows/release.yml
PhantomInTheWire: sure I can just do `"https://api.github.com/repos/${GITHUB_REPOSITORY}/dependency-graph/sbom"`

.github/workflows/release.yml
PhantomInTheWire: done.

.github/workflows/release.yml
PhantomInTheWire: It's meant to pretty-print the output in case a human needs to inspect it. However, I realized that using `-o` with the `curl` command already formats the output, so I’ve removed the extra step.

.github/workflows/release.yml
PhantomInTheWire: It occasionally fails silently with `{"message":"Not Found","documentation_url":"https://docs.github.com/rest","status":"404"}`. Knowing the file size helps detect such silent failures. However, this only occurs when the repository doesn't exist or the dependency graph isn't enabled, cases that won't affect our workflow. So, I'll remove it.

.github/workflows/release.yml
PhantomInTheWire: changed to release version

hydai (approved):

PR #4199, opened 2025-06-22
[WIP] feat(wasi-nn): add device string retrieval and improve input tensor handling

LFsWang:
Fix example, and handle general tensor cases for openvino
https://github.com/second-state/WasmEdge-WASINN-examples/pull/195


juntao:

------

## [plugins/wasi_nn/wasinn_openvino.cpp](https://github.com/WasmEdge/WasmEdge/blob/1c716d76127212ce0e6cbb59be5e29a873840d11/plugins%2Fwasi_nn%2Fwasinn_openvino.cpp)

#### Potential issues

1. In the `setInput` function, the device string is retrieved and used to compile the model every time an input tensor is set, which is inefficient; this should be done once during initialization or context creation.
2. In the same function, the tensor type check only allows for F32, but there's no handling for other types that might be supported in the future.
3. The `compute` function does not handle cases where the input tensors have not been set or are invalid before calling `infer()`, potentially leading to runtime errors during inference.

#### Summary of changes

Here are the key changes in short bullet points, listed from most important to least:

1. **Added Device String Retrieval Functionality:**
   - A new function `GetDeviceString` has been introduced to convert a WASI-NN `Device` type to its corresponding OpenVINO device string representation (e.g., "CPU" or "GPU").
   - This function returns an error if the device type is not supported.

2. **Improved Input Tensor Handling:**
   - The input tensor's shape is now dynamically generated based on the `Dimension` data from the input tensor, rather than using a hardcoded shape.
   - This change makes the code more flexible and adaptable to different input sizes.

3. **Dynamic Device Selection for Model Compilation:**
   - The device string retrieved from `GetDeviceString` is now used when compiling the OpenVINO model, allowing for dynamic selection of the device ("CPU" or "GPU") instead of hardcoding "CPU".

These changes aim to enhance flexibility and configurability in handling devices and input tensors within the WASI-NN backend for OpenVINO.

hydai (approved):

PR #4201, opened 2025-06-23
feat: unified all installer CI workflows

hydai:


juntao:

------

## [.github/scripts/installer-test-helpers.sh](https://github.com/WasmEdge/WasmEdge/blob/9fede396e8a6bb1e277045513b156954d17ae20c/.github%2Fscripts%2Finstaller-test-helpers.sh)

#### Potential issues

1. The script assumes the existence of `~/.wasmedge/bin/wasmedge` and `~/.wasmedge/plugin/` without verifying their presence first, which could lead to errors if they do not exist.
2. The `verify_version` function uses a hardcoded path `~/.wasmedge`, which may not be suitable for all environments or users with different installation paths.
3. The script does not handle the case where the version command outputs multiple lines or no output, potentially causing false positives in version comparison.

#### Summary of changes

Here are the key changes in the patch:

1. **Introduction of `verify_version` function:** This function checks if the installed version of wasmedge matches the expected version.
2. **Introduction of `verify_plugin` function:** This function verifies the existence of a specified plugin file.
3. **Bash script setup for installer tests:** The entire patch is a bash script designed to assist with installer testing by providing functions to verify the correct installation of software versions and plugins.

## [.github/workflows/test-install-script.yml](https://github.com/WasmEdge/WasmEdge/blob/571a100d80e3a95d84f3ed7834e63f74c07632f7/.github%2Fworkflows%2Ftest-install-script.yml)

#### Potential issues

1. **Incorrect conditional in macos_amd64 job**: The condition `${{ matrix.name == 'MacOS-arm64' }}` is incorrect as `matrix.name` does not exist in the `macos_amd64` job.
2. **Shell specification inconsistency**: In `macos-arm64` job, `shell: zsh {0}` should be specified only once at the top level instead of repeating it for each step.
3. **Missing permission to write files**: The `permissions: contents: read` might prevent some steps from writing files (e.g., creating `.bashrc` or `.zshenv`). Consider adding `write` permission where necessary.

#### Summary of changes

Here are the top three most important changes:

1. **Removed `test-install-script.yml` Workflow**: The entire content of this workflow file, which included multiple jobs for testing WasmEdge installation on different Linux and macOS environments, has been removed.
2. **Deleted All CI Jobs**: This includes various job configurations for Ubuntu, manylinux, and macOS platforms with specific environment setups and steps for installing and uninstalling WasmEdge, along with plugin checks.
3. **Removed Environment Variables and Permissions**: All environment variables (`env`), permissions, and concurrent execution settings were eliminated from this file.

These changes imply a significant refactor or consolidation of CI workflows related to installers.

## [.github/workflows/test-installer-v2.yml](https://github.com/WasmEdge/WasmEdge/blob/571a100d80e3a95d84f3ed7834e63f74c07632f7/.github%2Fworkflows%2Ftest-installer-v2.yml)

#### Potential issues

1. **Inconsistent Shell Usage**: The `macos` job uses both `bash` and `zsh`, which can lead to unexpected behavior if the scripts are not compatible with both shells.

2. **Missing Error Handling**: Some steps do not handle errors properly (e.g., the `Install zsh` step), which could cause silent failures in the CI pipeline.

3. **Hardcoded Paths**: The uninstallation script is called with hardcoded options (`-q -V`) without checking if these are supported or needed for all scenarios, potentially leading to issues on different systems.

#### Summary of changes

Key Changes:

- **Removed**: The entire CI workflow file `test-installer-v2.yml` was deleted.
- **Unified Workflow Removal**: This suggests that the workflows for testing the installer were likely unified or consolidated elsewhere, making this specific workflow redundant.
- **Test Scenarios**: The removed file contained multiple test scenarios and jobs for different operating systems (CentOS 9 Stream, Ubuntu 20.04, Mac M1, Mac Intel), which are now presumably covered by another unified workflow.

These changes imply a consolidation of CI workflows to streamline the testing process, likely improving maintainability and reducing redundancy.

## [.github/workflows/test-installers.yml](https://github.com/WasmEdge/WasmEdge/blob/9fede396e8a6bb1e277045513b156954d17ae20c/.github%2Fworkflows%2Ftest-installers.yml)

#### Potential issues

1. **Missing Error Handling in `get-latest-version` Job**: The script does not handle the case where the curl command fails to retrieve the latest version, which could lead to unreliable behavior downstream.
2. **Potential Race Condition in Version Checks**: In some tests, the version is verified by sourcing environment files and then checking the version output. This process might fail if the environment setup takes time or is not properly synchronized with the verification step.
3. **Hardcoded Dependency Versions**: The script uses specific versions of dependencies (e.g., `actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683`). Hardcoding versions can make the workflow brittle to changes and updates in these dependencies.

#### Summary of changes

- **Unified Installer CI Workflows**: The patch unifies all installer CI workflows, streamlining the process for testing installers on different platforms (Linux, macOS) and using different scripts (`install.sh`, `install_v2.sh`, `install.py`).

- **Concurrent Execution with Cancellation**: Added concurrency settings to allow running jobs in parallel while canceling in-progress runs if a new one is triggered.

- **Extensive Platform Testing**: Introduced matrix testing for various Linux distributions (Ubuntu, CentOS) and macOS versions (Intel, Apple Silicon), ensuring comprehensive coverage across different environments.

## [.github/workflows/test-python-install-script.yml](https://github.com/WasmEdge/WasmEdge/blob/571a100d80e3a95d84f3ed7834e63f74c07632f7/.github%2Fworkflows%2Ftest-python-install-script.yml)

#### Potential issues

1. **Matrix name inconsistency**: The matrix for 'manylinux_2_28 aarch64' uses `ubuntu-24.04-arm` as the runner, but the corresponding image is set to `wasmedge/wasmedge:manylinux_2_28_aarch64`, which may cause confusion or errors.

2. **Incorrect plugin check**: In the 'Plugins Install check' step for Python 3, the expected output is incorrectly referenced as 'Wasmedge Tensorflowlite Plugin', while it should be 'WasmEdgeTensorflow.so'.

3. **Unnecessary commands in some steps**: Some steps include redundant or unnecessary commands, such as multiple `ls` calls and echo statements, which can lead to cluttered logs and potential misinterpretation of results.

These are the most significant issues found within the provided code.

#### Summary of changes

- **Unified CI Workflows**: The patch removes multiple redundant CI workflows related to testing Python installer scripts.
  - Consolidates various jobs and steps into a single, more streamlined workflow file.
  - Removes the `test-python-install-script` workflow file entirely.

- **Concurrency Management**: Adds concurrency group management using `${{ github.workflow }}-${{ github.head_ref || github.ref }}`.
  - Sets `cancel-in-progress: true` to cancel any in-progress jobs from previous runs when a new run is triggered.

- **Event Triggers and Permissions**:
  - Specifies event triggers for push and pull request events targeting the master branch.
  - Configures permissions to allow read access to repository contents.

.github/workflows/test-installers.yml
```diff
@@ -0,0 +1,383 @@
...
+        fetch-depth: 0
+
+    - name: Install dependencies
+      run: |
+        ${{ matrix.package_manager }} update -y
+        ${{ matrix.package_manager }} install -y git curl ${{ matrix.python_packages }}
+
+    - name: Test install.sh basic functionality
+      run: |
+        # Test latest release installation
```
0yi0, line 133: Should there be a version check for installed version? Version consistency check doesn't seem to check this.

.github/workflows/test-installers.yml
```diff
@@ -0,0 +1,383 @@
...
+        fetch-depth: 0
+
+    - name: Setup
+      run: ${{ matrix.setup_cmd }}
+
+    - name: Test install_v2.sh basic functionality
+      shell: bash
+      run: |
+        # Basic installation with GGML plugin
+        bash utils/install_v2.sh -V
```
0yi0, line 239: Is this supposed to install the latest version or `0.14.1`? If this command aims to install the latest version, maybe we could use version like from version consistency check instead of hard-coded here.

.github/workflows/test-installers.yml
hydai: The default is to download the latest release. However, we don't verify the version at this moment. Maybe we can use `gh` to get the latest version and compare it with that?

.github/workflows/test-installers.yml
hydai: It is supposed to install the latest version; however, `0.14.1` is specified inside the script because we don't want any extra dependencies when running it.

hydai:
Hi @0yi0 
I have fixed the issues you mentioned, PTAL.

PR #4001, opened 2025-01-30
[WIP][CI] Ensure the libwasmedge will not expose unnecessary symbols.

Ytemiloluwa:
This PR updates `whitelist.symbols` to explicitly define allowed symbols, preventing unnecessary symbol exposure. Fixed #3743 @hydai.

Issue #3655, opened 2024-08-14
libwasmedge exposes hundreds of C++ symbols (0.14.0 regression)


paravoid:
### Summary

Compiling 0.14.0 results into a ton of `WasmEdge::` C++ symbols being exposed. 684 to be exact. 

For example:
``` WasmEdge::PO::ArgumentParser::SubCommandDescriptor::consume_argument(WasmEdge::PO::ArgumentParser::ArgumentDescriptor&, std::basic_string_view<char, std::char_traits<char> >)@Base
 WasmEdge::PO::ArgumentParser::SubCommandDescriptor::consume_long_option(std::basic_string_view<char, std::char_traits<char> >)@Base
 ```

As far as I can tell, this is a side effect of commit 0b285f7bc01cc1797997da1e537161d31d7449c1, which commented out `CXX_VISIBILITY_PRESET` in cmake/Helper.cmake.

I assume this change was not intentional?

(Also see issue #1678 and d4f604b6a05b48a1ad4d9b4fb65c5b7314860a1b for prior work in this area).

### Reproduction steps

1. Build with options '...'
2. Execute with flags '....'
3. Execute with inputs '....'
4. Get error

### Components

CLI, C SDK

### WasmEdge Version or Commit you used

0.14.0

### Operating system information

Debian unstable

### Hardware Architecture

amd64

hydai:
This change is not intentional. I created #3656 to fix it. Because we don't have a CI check for this, could you give some ideas on what we can do to avoid this error being made again?

paravoid:
Thank you so much for the blazingly fast response! I don't have any good ideas on how to catch that kind of thing with CI :/ I'll think about it.

For what it's worth, the way I identified this is through `dpkg-gensymbols`, which happens as part of the Debian package build process.

hydai:
And I found that the ffmpeg plugin will fail if we make `CXX_VISIBILITY_PRESET` hidden. I think this is the reason that the previous contributor for the ffmpeg plugin tried to disable it. I am digging into this plugin to check if there is something wrong because the current design of the plugin architecture should never use these symbols from our codebase.

hydai:
For now, the first priority is to revert this regression. Since I don't have much time for the wasmedge-ffmpeg plugin, I will temporarily remove and disable all related CI workflows.

If anyone is interested in fixing the wasmedge-ffmpeg, please check #3658 for more details.

paravoid:
[...]

PR #4204, opened 2025-06-24
[CI] Fix symbol exposure regression and add CI check to prevent future issues

### LINKED ISSUE #3743: feat: Ensure the libwasmedge will not expose unnecessary symbols
### Summary

Please check the previous issue. #3655 
WasmEdge made a regression by exposing lots of unnecessary symbols in 0.14.0 and fixed this regression in 0.14.1.
However, we would like to avoid this problem happening again.

### Details

Thanks to the ideas provided by the author of #3655 (Faidon Liambotis). We currently have the following solutions:

1. Use `dpkg-gensymbols` to check if we expose any unexpected symbols.
2. Another way is to use an allowed list to show the explicit symbol list instead of the wildcard. Ref: https://github.com/WasmEdge/WasmEdge/issues/3655#issuecomment-2291097273

Moreover, we welcome any good ideas for fixing this. If you have any thoughts, please feel free to comment below. Thanks!

### Appendix

_No response_

thedevyashsaini:
### Summary

This PR addresses GitHub issue #3743 by implementing a comprehensive solution to prevent unintended symbol exposure from the WasmEdge library. The solution includes a robust cross-platform symbol checking script integrated into CI workflows to catch symbol exposure regressions early.

### Problem

A CMake regression (identified in issue #3655) caused hundreds of internal WasmEdge symbols to be exported from `libwasmedge`, creating potential ABI stability issues and exposing implementation details that should remain private. The original fix in PR #3656 addressed the immediate CMake issue, but there was no automated mechanism to prevent future regressions.

### Solution

This PR implements a complete symbol exposure management system:

#### 1. **Symbol Whitelist** (whitelist.symbols)
- Comprehensive list of 282 legitimate WasmEdge C API symbols
- Only symbols starting with `WasmEdge` prefix (public API) are allowed
- Serves as the authoritative source for intended public exports

#### 2. **Cross-Platform Symbol Checking Script** (check_symbols.sh)
- **Linux/macOS**: Uses `nm` to extract dynamic symbols
- **Windows**: Uses `dumpbin` to extract DLL exports  
- **Features**:
  - Robust error handling with clear diagnostic messages
  - Line ending normalization for cross-platform compatibility
  - Tool existence verification before execution
  - Detailed reporting of symbol counts and mismatches
  - Non-zero exit codes for CI integration

#### 3. **CI Integration**
- Integrated into all major build workflows:
  - Ubuntu (reusable-build-on-ubuntu.yml)
  - macOS (reusable-build-on-macos.yml) 
  - Windows (reusable-build-on-windows.yml)
- Consistent invocation: `bash .github/scripts/check_symbols.sh`
- Runs after build but before packaging to catch issues early

#### 4. **Documentation & Guidelines**
- `.github/scripts/SYMBOL_CHECKING.md`: Comprehensive technical documentation and maintenance guidelines

### Technical Details

**Symbol Extraction Logic:**
```bash
# Linux: Extract dynamic symbols with nm
nm -D --defined-only libwasmedge.so | awk '/^[0-9a-fA-F]+ [TBDW] / {print $3}' | grep -E "^WasmEdge"

# macOS: Extract symbols with nm and remove underscore prefix
nm -g libwasmedge.dylib | awk '/^[0-9a-fA-F]+ [TBDWS] _WasmEdge/ {print substr($3,2)}'

# Windows: Extract DLL exports with dumpbin (objdump fallback)
dumpbin //EXPORTS wasmedge.dll | awk '/^[[:space:]]*[0-9]+[[:space:]]+[0-9A-Fa-f]+[[:space:]]+[0-9A-Fa-f]+[[:space:]]+/ {print $4}' | grep -E "^WasmEdge"
```

**Error Detection:**
- **Unexpected symbols**: Found in library but not in whitelist → CI failure
- **Missing symbols**: In whitelist but not in library → Warning only (allows for conditional compilation)

### Addresses Original Feedback

This implementation addresses all feedback from the previously closed PR #4001:
- ✅ Removed redundant `chmod +x` commands
- ✅ Consistent script invocation across all platforms
- ✅ Improved error messages with actionable guidance
- ✅ Tool existence checks before execution
- ✅ Proper project organization in scripts

### Testing

- ✅ Verified script works on all target platforms
- ✅ Confirmed no unexpected symbols are currently exposed (282 symbols match whitelist exactly)
- ✅ Tested error conditions (missing tools, missing files, malformed whitelist)
- ✅ Validated CI integration doesn't break existing workflows

### Future Benefits

- **Regression Prevention**: Automatic detection of symbol exposure issues
- **ABI Stability**: Ensures only intended API symbols are public
- **Maintainability**: Clear guidelines for whitelist updates
- **Cross-Platform**: Consistent behavior across Linux, macOS, and Windows

### Files Changed

- **New**: check_symbols.sh - Main symbol checking script
- **New**: whitelist.symbols - Authoritative symbol whitelist
- **New**: SYMBOL_CHECKING.md - Technical documentation & Maintenance guidelines
- **Modified**: CI workflows (Ubuntu, macOS, Windows) - Added symbol checking step

### Closes

- Fixes #3743 
- Addresses the regression prevention aspect of #3655


juntao:

------

## [.github/scripts/check_symbols.sh](https://github.com/WasmEdge/WasmEdge/blob/622d446c1a41407972484fd2f614cdfd22842062/.github%2Fscripts%2Fcheck_symbols.sh)

#### Potential issues

1. **Incorrect Use of `shift`**: The script uses `shift` after setting a flag (`VERBOSE=true`), but it should be used to remove the processed argument from the list.
2. **Unchecked Command Availability**: The script assumes the availability of tools like `dumpbin`, `objdump`, and `nm` without checking them in all required contexts, which may lead to failures on some systems.
3. **Potential Path Traversal Vulnerability**: Using `mktemp -d` directly might be unsafe in certain environments; it's better to use a secure temporary directory creation method with additional checks.

#### Summary of changes

- Added a CI check script to ensure only intended API symbols are exposed by `libwasmedge` by comparing exported symbols with a whitelist.
- Introduced OS-specific logic to detect and handle library paths for different operating systems (Linux, macOS, Windows).
- Implemented symbol extraction mechanisms using platform-specific tools like `nm`, `objdump`, and `dumpbin`, including fallback options when primary tools are unavailable.

## [.github/scripts/whitelist.symbols](https://github.com/WasmEdge/WasmEdge/blob/622d446c1a41407972484fd2f614cdfd22842062/.github%2Fscripts%2Fwhitelist.symbols)

#### Potential issues

1. **Function Name Confusion**: Some function names like `WasmEdge_ConfigureCompilerGetOptimizationLevel` and `WasmEdge_ConfigureCompilerSetOptimizationLevel` might be confusing to users who are looking for the configuration options for compilers, as they might expect a single unified function.

2. **Inconsistent Naming Conventions**: The naming conventions seem inconsistent; some functions use camel case (e.g., `WasmEdge_Driver_UniTool`), while others use underscores (e.g., `WasmEdge_ConfigureCompilerSetOutputFormat`).

3. **Redundant Functions**: There are many similar functions like `WasmEdge_VMRunWasmFromASTModule`, `WasmEdge_VMRunWasmFromBuffer`, and `WasmEdge_VMRunWasmFromFile`; consolidating these might simplify the API for developers.

These are the top issues, focusing on maintainability, usability, and consistency in naming conventions.

#### Summary of changes

### Key Changes

1. **Exposed WasmEdge Symbols**: The patch includes a comprehensive list of WasmEdge symbols that were previously not exposed but now are, which addresses the symbol exposure regression issue.

2. **Added Functions for Configuration and Statistics**:
   - Added numerous configuration functions (`WasmEdge_Configure*`) to manage settings such as optimization level, output format, and memory configurations.
   - Introduced statistics functions (`WasmEdge_Statistics*`) to measure and set various performance metrics like instruction counting and time measuring.

3. **Enhanced Module Instance Management**:
   - Added multiple functions for managing module instances (`WasmEdge_ModuleInstance*`), including creating, adding components (functions, globals, memories), finding elements, and listing them.
   - These enhancements improve the flexibility and control over module instances within the WasmEdge environment.

These changes collectively ensure that the API provides more comprehensive functionality while preventing future symbol exposure issues through CI checks.

## [.github/workflows/reusable-build-on-macos.yml](https://github.com/WasmEdge/WasmEdge/blob/622d446c1a41407972484fd2f614cdfd22842062/.github%2Fworkflows%2Freusable-build-on-macos.yml)

#### Potential issues

1. The job `build_on_macos` has a permission issue: it requests `contents: write` permission but only uses actions that require `contents: read`.
2. The `check_symbols.sh` script is executed without specifying its path or ensuring it exists, which could lead to failures if the file is not present.
3. Inconsistent use of double quotes in YAML strings can cause unexpected behavior (e.g., `"$GITHUB_ENV"` vs `$GITHUB_ENV`).

#### Summary of changes

1. **New Job Added**: Introduced a new job named "Check symbol exposure" that runs the script `.github/scripts/check_symbols.sh` to verify symbol visibility.

2. **Sequential Execution**: The new check is placed before the existing "Upload artifact" step, ensuring it runs prior to uploading any artifacts.

3. **Potential Regression Fix**: By adding this verification step, the patch aims to prevent future issues related to unintended symbol exposure, which may have caused previous regressions.

## [.github/workflows/reusable-build-on-ubuntu.yml](https://github.com/WasmEdge/WasmEdge/blob/622d446c1a41407972484fd2f614cdfd22842062/.github%2Fworkflows%2Freusable-build-on-ubuntu.yml)

#### Potential issues

1. **Redundant Step**: The step "Grant the safe directory for git" is redundant and duplicate of the earlier step "Ensure git safe directory".

2. **Unnecessary Shell Specifier**: In the build step, specifying `shell: bash` is unnecessary as the default shell is already set to bash in the `defaults` section.

3. **Inconsistent Dependency Installation**: The installation of dependencies differs between release and non-release builds, which might lead to inconsistencies. Consider unifying the dependency installation process.

#### Summary of changes

- Added a new step in the CI/CD pipeline to check symbol exposure, which is critical for preventing unintended exposure of internal symbols.
- The new step uses a script located at `.github/scripts/check_symbols.sh` to validate the exported symbols.
- This step only runs when coverage tests are not being performed.

## [.github/workflows/reusable-build-on-windows.yml](https://github.com/WasmEdge/WasmEdge/blob/622d446c1a41407972484fd2f614cdfd22842062/.github%2Fworkflows%2Freusable-build-on-windows.yml)

#### Potential issues

1. **Incorrect path for uploading artifacts:** In the "Upload artifact" step, the path `build\\WasmEdge-${{ inputs.version }}-Windows.zip` uses backslashes which might cause issues on Unix-based systems; it should use forward slashes (`/`) or double backslashes (`\\`).

2. **Potential environment variable conflict:** The `product_version` is being set twice, once in the "Generate product version and package Windows installer" step and again in the subsequent upload steps, which could lead to unexpected behavior if not properly handled.

3. **Hardcoded SDK version:** The SDK version (`26100`) is hardcoded in multiple places; it would be more robust to parameterize this value for easier updates and maintenance.

#### Summary of changes

**Top 3 changes:**

1. **Added a new CI job to check symbol exposure:**
   - The new job uses `vswhere` and PowerShell to enter the Visual Studio development shell.
   - It then runs a custom script (`check_symbols.sh`) to verify symbol exposure.

2. **Introduced symbolic links for checking symbols:**
   - The patch includes changes to create necessary symbolic links before running the check, ensuring correct symbol visibility (not shown in the provided snippet).

3. **Updated CI workflow structure:**
   - The new job is added before the "Create WasmEdge package" step.
   - This ensures that symbol exposure checks are performed as part of the CI process.

.github/scripts/check_symbols.sh
```diff
@@ -0,0 +1,137 @@
...
+fi
+
+echo "Checking symbol exposure for libwasmedge..."
+echo "Library path: $LIB_PATH"
+echo "Whitelist file: $WHITELIST_FILE"
+
+# Check if library exists
+if [ ! -f "$LIB_PATH" ]; then
+    echo "Error: Library not found at $LIB_PATH"
+    echo "Make sure you have built WasmEdge first with: cmake --build build"
```
hydai: I don't think this error message will be helpful. The reason is that people may have their own locations for the `LIB_PATH`; it is not related to the building process.

.github/scripts/check_symbols.sh
```diff
@@ -0,0 +1,137 @@
...
+
+# Check if whitelist exists
+if [ ! -f "$WHITELIST_FILE" ]; then
+    echo "Error: Whitelist file not found at $WHITELIST_FILE"
+    exit 1
+fi
+
+# Verify whitelist is not empty
+if [ ! -s "$WHITELIST_FILE" ]; then
+    echo "Error: Whitelist file is empty: $WHITELIST_FILE"
```
hydai: Why do we need to check if the file is not empty?

.github/scripts/check_symbols.sh
```diff
@@ -0,0 +1,137 @@
...
+    exit 1
+fi
+
+# Create temporary directory for processing
+TEMP_DIR=$(mktemp -d)
```
hydai: Windows doesn't provide `mktemp`.

.github/scripts/check_symbols.sh
```diff
@@ -0,0 +1,137 @@
...
+    echo "  2. Symbol extraction failed"
+    echo "  3. All symbols are hidden (check build configuration)"
+    exit 1
+fi
+
+# Sort whitelist for comparison and convert line endings
+sort "$WHITELIST_FILE" | tr -d '\r' > "$TEMP_DIR/whitelist_sorted.symbols"
+
+# Check for unexpected symbols (in library but not in whitelist)
+comm -23 "$TEMP_DIR/extracted.symbols" "$TEMP_DIR/whitelist_sorted.symbols" > "$TEMP_DIR/unexpected.symbols"
```
hydai: Ditto.

docs/SYMBOL_WHITELIST_GUIDELINES.md
```diff
@@ -0,0 +1,48 @@
+# Symbol Whitelist Change Template
+
+When modifying `lib/api/whitelist.symbols`, use this commit message template:
+
+```
+[API] Update symbol whitelist: [brief description]
```
hydai: This commit message is not compatible with our commitlint style.

lib/api/whitelist.symbols
hydai, line 1: Maybe we should put this file together with check-symbol.sh.

docs/symbol_exposure_management.md
hydai, line 1: These two documents are redundant; they should be integrated into one and placed in the same folder as check-symbol.sh.

.github/scripts/SYMBOL_CHECKING.md
```diff
@@ -0,0 +1,173 @@
...
+
+This ensures a stable public API surface while preventing accidental exposure of internal implementation details.
+
+## Files
+
+- `check_symbols.sh` - Main checking script
+- `whitelist.symbols` - Approved symbol list (282 symbols)  
+- `SYMBOL_CHECKING.md` - This documentation file
+
+All symbol management components are co-located in `.github/scripts/` for easy maintenance and Windows compatibility.
```
hydai: What is the relationship between co-location and Windows compatibility?

.github/scripts/SYMBOL_CHECKING.md
```diff
@@ -0,0 +1,173 @@
...
+2. **Filtering** to only WasmEdge-prefixed symbols (`^WasmEdge`)
+3. **Comparing** against the approved whitelist
+4. **Reporting** any discrepancies with actionable guidance
+
+This ensures a stable public API surface while preventing accidental exposure of internal implementation details.
+
+## Files
+
+- `check_symbols.sh` - Main checking script
+- `whitelist.symbols` - Approved symbol list (282 symbols)
```
hydai: I don't think we need to write down the `282 symbols` here, since it may change, and we will need to update this documentation asynchronously, which means it could become outdated at certain moments if contributors don't notice this.

.github/scripts/SYMBOL_CHECKING.md
```diff
@@ -0,0 +1,173 @@
...
+4. Remove old name after deprecation period
+
+## Troubleshooting
+
+**"Library not found" error:**
+- Ensure WasmEdge is built: `cmake --build build`
+- Check if it's a static-only build (will be auto-detected)
+
+**"No symbols found" error:**
+- Library might be stripped or static
```
hydai: If it's static, this condition should have already been handled in the previous error, or is there any possibility of reaching this situation?

.github/scripts/SYMBOL_CHECKING.md
```diff
@@ -0,0 +1,173 @@
...
+
+1. **Test first**: Run `bash .github/scripts/check_symbols.sh` before committing
+2. **Add symbols alphabetically** in `whitelist.symbols`
+3. **Document changes** in commit message using the template below
+4. **Verify CI passes** on all platforms
+
+### Commit Message Template
+
+```
+[API] Update symbol whitelist: [brief description]
```
hydai: The commit message template is wrong. Please check our commitlint policy.

.github/scripts/SYMBOL_CHECKING.md
```diff
@@ -0,0 +1,173 @@
...
+
+- Added: WasmEdge_NewFunction1, WasmEdge_NewFunction2
+- Removed: WasmEdge_OldFunction1 (deprecated)
+- Reason: [explain why these changes are necessary]
+
+Verified with: bash .github/scripts/check_symbols.sh
```
hydai: There is no need to add this line, since the CI will check it.

.github/scripts/SYMBOL_CHECKING.md
```diff
@@ -0,0 +1,173 @@
...
+1. Remove function from `whitelist.symbols`
+2. Update documentation about deprecation
+3. Run symbol check to verify
+4. Consider transition period for major changes
+
+**Renaming Functions:**
+1. Add new function name to whitelist
+2. Keep old name temporarily for transition
+3. Update all references in codebase
+4. Remove old name after deprecation period
```
hydai: There is a lot of redundant text. Please make the process clear and simple to understand. If you are using AI to generate this document, please ensure it is lean and efficient.

.github/scripts/SYMBOL_CHECKING.md
```diff
@@ -0,0 +1,173 @@
...
+- Template instantiations
+- C++ class internals
+- Debug/development-only functions
+
+### Updating Process
+
+1. **Test first**: Run `bash .github/scripts/check_symbols.sh` before committing
+2. **Add symbols alphabetically** in `whitelist.symbols`
+3. **Document changes** in commit message using the template below
+4. **Verify CI passes** on all platforms
```
hydai: Please make it clear and simple instead of adding unnecessary texts.

.github/scripts/SYMBOL_CHECKING.md
```diff
@@ -0,0 +1,173 @@
...
+### 1. Symbol Whitelist (`whitelist.symbols`)
+
+This file contains the complete list of symbols that are allowed to be exported from `libwasmedge`. It includes:
+
+- All WasmEdge C API functions (e.g., `WasmEdge_VersionGet`, `WasmEdge_VMCreate`)
+- Plugin creation functions
+- Driver utility functions
+- Essential utility functions
+
+Currently maintains **282 approved symbols**.
```
hydai: Too complex and unnecessary.

hydai (changes requested):
Your document is unnecessarily repetitive. Please make it clearer and simpler.

.github/scripts/check_symbols.sh
```diff
@@ -0,0 +1,239 @@
...
+
+# Script to check that libwasmedge doesn't expose unnecessary symbols
+# This script compares the exported symbols against a whitelist to ensure
+# only intended API symbols are exposed.
+
+set -euo pipefail  # Exit on error, undefined vars, pipe failures
+
+# Default paths
+SYMBOLS_PATH=${SYMBOLS_PATH:-"$(pwd)"}
+WHITELIST_FILE="$SYMBOLS_PATH/.github/scripts/whitelist.symbols"
```
PhantomInTheWire: We should consider doing something like this:

```bash
SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
SYMBOLS_PATH="${SYMBOLS_PATH:-"${SCRIPT_DIR%/.github/scripts}"}"
WHITELIST_FILE="$SYMBOLS_PATH/.github/scripts/whitelist.symbols"
```

This removes the assumption that the script is always executed from the repository root.

.github/scripts/check_symbols.sh
```diff
@@ -0,0 +1,239 @@
...
+    mkdir -p "$TEMP_DIR"
+fi
+trap "rm -rf $TEMP_DIR" EXIT
+
+# Extract symbols from library
+echo "Extracting symbols from library..."
+if [[ "$OSTYPE" == "msys" ]] || [[ "$OSTYPE" == "cygwin" ]]; then
+    # Windows: Use dumpbin to extract exports, fallback to objdump
+    if command -v dumpbin >/dev/null 2>&1; then
+        dumpbin //EXPORTS "$LIB_PATH" | awk '/^[[:space:]]*[0-9]+[[:space:]]+[0-9A-Fa-f]+[[:space:]]+[0-9A-Fa-f]+[[:space:]]+/ {print $4}' | grep -E "^WasmEdge" | sort > "$TEMP_DIR/extracted.symbols"
```
PhantomInTheWire, line 180: non-blocking but dumpbin typically uses `/`, although cmd ignores the second slash on most systems, we should probably use the more documented `/` to avoid confusion
https://github.com/MicrosoftDocs/cpp-docs/blob/main/docs/build/reference/dumpbin-options.md

.github/scripts/check_symbols.sh
```diff
@@ -0,0 +1,239 @@
...
+{
+    sort "$TEMP_DIR/extracted.symbols" > "$TEMP_DIR/extracted_sorted.symbols"
+    
+    # Find symbols in extracted but not in whitelist (unexpected)
+    grep -Fxv -f "$TEMP_DIR/whitelist_sorted.symbols" "$TEMP_DIR/extracted_sorted.symbols" > "$TEMP_DIR/unexpected.symbols" 2>/dev/null || true
+    
+    # Find symbols in whitelist but not in extracted (missing)
+    grep -Fxv -f "$TEMP_DIR/extracted_sorted.symbols" "$TEMP_DIR/whitelist_sorted.symbols" > "$TEMP_DIR/missing.symbols" 2>/dev/null || true
+}
+
```
PhantomInTheWire: non-blocking: We're currently performing a double sort, once during the `nm` stage and again here, which is redundant and can be avoided. Additionally, the diffing logic could be simplified. Instead of using `grep -Fxv -f`, consider using `comm`, which is designed to compare two sorted files line by line, should be simpler and faster.

.github/scripts/check_symbols.sh
```diff
@@ -0,0 +1,239 @@
...
+    echo
+    echo "This might indicate:"
+    echo "1. Symbols removed from the API (update whitelist)"
+    echo "2. Conditional compilation excluding some symbols"
+    echo "3. Platform-specific symbols not available on this build"
+    echo
+fi
+
+echo "✅ Symbol exposure check passed - no unexpected symbols found!"
+exit 0
```
PhantomInTheWire: EOF

.github/workflows/reusable-build-on-windows.yml
```diff
@@ -71,6 +71,12 @@ jobs:
           tools\wasmedge\wasmedge -v
           ctest --output-on-failure
           cd -
+      - name: Check symbol exposure
+        run: |
+          $vsPath = (vswhere -latest -property installationPath)
+          Import-Module (Join-Path $vsPath Common7\Tools\Microsoft.VisualStudio.DevShell.dll)
+          Enter-VsDevShell -VsInstallPath $vsPath -SkipAutomaticLocation -DevCmdArguments "-arch=x64 -host_arch=x64 -winsdk=10.0.26100.0"
```
PhantomInTheWire, line 78: this runs in Build WasmEdge step too, we should probably follow DRY here.

.github/scripts/SYMBOL_CHECKING.md
```diff
@@ -0,0 +1,54 @@
...
+- Add WasmEdge_ValidatorCreateWithConfig to whitelist
+```
+
+Remove symbols when deprecating API functions:
+
+```
+feat(api): remove deprecated function
+
+- Remove WasmEdge_OldFunction from whitelist
+```
```
PhantomInTheWire: I think you forgot to update these commit messages.
https://github.com/WasmEdge/WasmEdge/blob/master/docs/CONTRIBUTING.md

PhantomInTheWire (changes requested):

.github/workflows/reusable-build-on-windows.yml
thedevyashsaini: While valid, this change affects the overall workflow architecture and is outside the scope of this symbol checking PR. The current duplication is minimal (3 lines) and clear. I'd prefer to address workflow refactoring in a separate PR to keep this focused on symbol checking functionality. The current approach works reliably and is easy to understand.

.github/scripts/check_symbols.sh
thedevyashsaini: Good catch on the double sort - I'll clean that up. As for `comm`, I actually used it before but switched to `grep -Fxv` for better Windows compatibility since comm isn't always available there. grep has been more portable across our supported platforms.

.github/scripts/check_symbols.sh
```diff
@@ -0,0 +1,235 @@
...
+    mkdir -p "$TEMP_DIR"
+fi
+trap "rm -rf $TEMP_DIR" EXIT
+
+# Extract symbols from library
+echo "Extracting symbols from library..."
+if [[ "$OSTYPE" == "msys" ]] || [[ "$OSTYPE" == "cygwin" ]]; then
+    # Windows: Use dumpbin to extract exports, fallback to objdump
+    if command -v dumpbin >/dev/null 2>&1; then
+        dumpbin //EXPORTS "$LIB_PATH" | awk '/^[[:space:]]*[0-9]+[[:space:]]+[0-9A-Fa-f]+[[:space:]]+[0-9A-Fa-f]+[[:space:]]+/ {print $4}' | grep -E "^WasmEdge" | sort > "$TEMP_DIR/extracted.symbols"
```
hydai, line 180: Is there any reason to have `//` here? According to the document, the command should only have a single `/`.

.github/scripts/check_symbols.sh
```diff
@@ -0,0 +1,235 @@
...
+    fi
+elif [[ "$OSTYPE" == "darwin"* ]]; then
+    # macOS: Use nm to extract symbols (no -D flag on macOS)
+    if ! command -v nm >/dev/null 2>&1; then
+        echo "Error: nm not found. Make sure Xcode Command Line Tools are installed"
+        exit 1
+    fi
+    
+    echo "Debug: Attempting to extract symbols from macOS dylib..."
+    echo "Debug: Library info: $(file "$LIB_PATH" 2>/dev/null || echo "file command not available")"
```
hydai: Can we have a verbose mode or something similar to hide unnecessary debug messages when there is no need to print this information, especially when everything is fine?

.github/scripts/SYMBOL_CHECKING.md
```diff
@@ -0,0 +1,58 @@
...
+
+Remove symbols when deprecating API functions:
+
+```
+feat(api): remove deprecated function
+
+- Remove WasmEdge_OldFunction from whitelist
+
+Signed-off-by: Your Name <your.email@example.com>
+```
```
hydai: I don't think we need this. All commit message-related formatting should be documented in the docs or contributing.md. It's unnecessary to have another location for this information.

.github/scripts/check_symbols.sh
thedevyashsaini: it really is supposed to be a single / for dumpbin.
But in our CI setup, using /EXPORTS exits with code 1 - something about how the shell handles the slash IG.
Using //EXPORTS bypasses that and makes it reach dumpbin properly.

.github/scripts/check_symbols.sh
```diff
@@ -0,0 +1,289 @@
...
+    cat "$TEMP_DIR/unexpected.symbols"
+    eprintf ""
+    eprintf "These symbols should either be:"
+    eprintf "1. Added to the whitelist if they are intended public API"
+    eprintf "2. Made static or marked with hidden visibility if they are internal"
+    eprintf ""
+    exit 1
+fi
+
+# Show missing symbols if any (warning only)
```
hydai, line 275: What is the reason for setting the warning level when detecting missing symbols?

.github/scripts/check_symbols.sh
```diff
@@ -0,0 +1,289 @@
...
+eprintf() {
+	command printf '%s\n' "$1" 1>&2
+}
+
+debug() {
+	if [ "$VERBOSE" = true ]; then
+		command printf '\e[0;36mDebug\e[0m: %s\n' "$1" 1>&2
+	fi
+}
+
```
PhantomInTheWire, line 61: non blocking but why tabs here? we are using spaces everywhere(i don't mean to start a holy war but that's just what everything else uses in this file too, so this feels random)

PhantomInTheWire (approved):

PhantomInTheWire:
everything else seems fine.

.github/scripts/check_symbols.sh
```diff
@@ -0,0 +1,289 @@
...
+eprintf() {
+	command printf '%s\n' "$1" 1>&2
+}
+
+debug() {
+	if [ "$VERBOSE" = true ]; then
+		command printf '\e[0;36mDebug\e[0m: %s\n' "$1" 1>&2
+	fi
+}
```
hydai, line 60: I found the indentation is inconsistent. Please unify it with the original style.

.github/scripts/check_symbols.sh
```diff
@@ -0,0 +1,289 @@
...
+if [[ "$OSTYPE" == "linux-gnu"* ]]; then
+    LIB_PATH="$SYMBOLS_PATH/build/lib/api/libwasmedge.so"
+elif [[ "$OSTYPE" == "darwin"* ]]; then
+    LIB_PATH="$SYMBOLS_PATH/build/lib/api/libwasmedge.dylib"
+elif [[ "$OSTYPE" == "msys" ]] || [[ "$OSTYPE" == "cygwin" ]]; then
+    LIB_PATH="$SYMBOLS_PATH/build/lib/api/wasmedge.dll"
+else
+    error "Unsupported OS: $OSTYPE"
+    exit 1
+fi
```
hydai, line 77: Is there a way for users to specify the `LIB_PATH`? You are assuming that there must already be a `build` folder. However, people can always use `cmake -Bbuild_somewhereelse` to change it.

.github/scripts/check_symbols.sh
```diff
@@ -0,0 +1,289 @@
...
+# Alternative library paths to check
+ALT_PATHS=()
+if [[ "$OSTYPE" == "linux-gnu"* ]]; then
+    ALT_PATHS+=("$SYMBOLS_PATH/build/lib/libwasmedge.so")
+elif [[ "$OSTYPE" == "darwin"* ]]; then
+    ALT_PATHS+=("$SYMBOLS_PATH/build/lib/libwasmedge.dylib")
+elif [[ "$OSTYPE" == "msys" ]] || [[ "$OSTYPE" == "cygwin" ]]; then
+    ALT_PATHS+=("$SYMBOLS_PATH/build/bin/wasmedge.dll")
+    ALT_PATHS+=("$SYMBOLS_PATH/build/lib/wasmedge.dll")
+fi
```
hydai, line 88: Why do we need this part? Will any distribution place these files under such a path?

.github/scripts/check_symbols.sh
```diff
@@ -0,0 +1,289 @@
...
+        done
+        eprintf ""
+        info "Searching for library files in build directory..."
+        if [[ "$OSTYPE" == "linux-gnu"* ]]; then
+            SEARCH_RESULT=$(find "$SYMBOLS_PATH/build" -name "*wasmedge*.so" 2>/dev/null | head -5)
+        elif [[ "$OSTYPE" == "darwin"* ]]; then
+            SEARCH_RESULT=$(find "$SYMBOLS_PATH/build" -name "*wasmedge*.dylib" 2>/dev/null | head -5)
+        elif [[ "$OSTYPE" == "msys" ]] || [[ "$OSTYPE" == "cygwin" ]]; then
+            SEARCH_RESULT=$(find "$SYMBOLS_PATH/build" -name "*wasmedge*.dll" 2>/dev/null | head -5)
+        fi
```
hydai, line 124: The logic is so weird. IIRC, our shared library must be under `<build_folder/lib/api/libwasmedge.*`. If not, we should figure it out and fix the issue instead of trying to search for it.

.github/scripts/check_symbols.sh
```diff
@@ -0,0 +1,289 @@
...
+            warn "No WasmEdge shared library files found in build directory."
+            eprintf ""
+            # Check if this is a static-only build
+            if [[ "$OSTYPE" == "linux-gnu"* ]]; then
+                STATIC_LIBS=$(find "$SYMBOLS_PATH/build" -name "*wasmedge*.a" 2>/dev/null | head -3)
+            elif [[ "$OSTYPE" == "darwin"* ]]; then
+                STATIC_LIBS=$(find "$SYMBOLS_PATH/build" -name "*wasmedge*.a" 2>/dev/null | head -3)
+            elif [[ "$OSTYPE" == "msys" ]] || [[ "$OSTYPE" == "cygwin" ]]; then
+                STATIC_LIBS=$(find "$SYMBOLS_PATH/build" -name "*wasmedge*.lib" 2>/dev/null | head -3)
+            fi
```
hydai, line 141: Checking the static libraries is fine because we do have such a use case; however, it should be under a specific path. What is the reason for using `find` to search for it?

.github/scripts/check_symbols.sh
hydai: Do you mean we must use `//` in the shell script?

.github/scripts/check_symbols.sh
```diff
@@ -0,0 +1,289 @@
...
+    # Try different nm approaches for macOS - NOTE: macOS symbols have underscore prefix
+    if nm -g "$LIB_PATH" 2>/dev/null | awk '/^[0-9a-fA-F]+ [TBDWS] _WasmEdge/ {print substr($3,2)}' | sort > "$TEMP_DIR/extracted.symbols"; then
+        debug "Successfully extracted symbols using nm -g (removed underscore prefix)"
+    elif nm -D "$LIB_PATH" 2>/dev/null | awk '/^[0-9a-fA-F]+ [TBDWS] _WasmEdge/ {print substr($3,2)}' | sort > "$TEMP_DIR/extracted.symbols"; then
+        debug "Successfully extracted symbols using nm -D (removed underscore prefix)"
+    elif nm "$LIB_PATH" 2>/dev/null | awk '/^[0-9a-fA-F]+ [TBDWS] _WasmEdge/ {print substr($3,2)}' | sort > "$TEMP_DIR/extracted.symbols"; then
+        debug "Successfully extracted symbols using nm (removed underscore prefix)"
+    elif nm "$LIB_PATH" 2>/dev/null | grep "_WasmEdge" | awk '{print substr($NF,2)}' | sort > "$TEMP_DIR/extracted.symbols"; then
+        debug "Successfully extracted symbols using simplified parsing (removed underscore prefix)"
+    else
```
hydai, line 207: What is the reason we have to try such similar commands? Does it relate to the versions of `nm`?

hydai:
Hi @thedevyashsaini,  
Just wanted to check if you are still working on this issue.

hydai:
Closing this due to inactivity. Thank you again for your time and contribution. Please feel free to reopen it if you still wish to work on it.

PR #4207, opened 2025-06-27
Feat/redirect stdio c sdk

PhantomInTheWire:
fixes: https://github.com/WasmEdge/WasmEdge/issues/2936


juntao:

------

## [examples/capi/wasi-env/.gitignore](https://github.com/WasmEdge/WasmEdge/blob/55778d9d08cdb15e027fa30a658c67c6257d1ef5/examples%2Fcapi%2Fwasi-env%2F.gitignore)

#### Potential issues

1. Missing semicolon after `return;` statement.

#### Summary of changes

- Added support for redirecting stdin, stdout, and stderr to `.wasm`, `.txt`, and `.out` files.

## [examples/capi/wasi-env/set_wasi_env.c](https://github.com/WasmEdge/WasmEdge/blob/55778d9d08cdb15e027fa30a658c67c6257d1ef5/examples%2Fcapi%2Fwasi-env%2Fset_wasi_env.c)

#### Potential issues

-There are no major bugs or issues in the provided source code.

#### Summary of changes

-Added `#include <fcntl.h>` and `#include <unistd.h>` for file operations.
-Used `open` to redirect `stdout` to a file named "stdout.txt".
-Modified `WasmEdge_ModuleInstanceInitWASI` call to use the new redirection parameters (`STDIN_FILENO`, `OutFd`, and `STDERR_FILENO`).
-Added a line to close `OutFd` after program execution.

## [include/api/wasmedge/wasmedge.h](https://github.com/WasmEdge/WasmEdge/blob/55778d9d08cdb15e027fa30a658c67c6257d1ef5/include%2Fapi%2Fwasmedge%2Fwasmedge.h)

#### Potential issues

N/A

#### Summary of changes

- **Key Changes:**  
1. Added two new functions to support redirecting stdin, stdout, and stderr for WASI instances:
   - `WasmEdge_ModuleInstanceCreateWASIWithFds`
   - `WasmEdge_ModuleInstanceInitWASIWithFds`

2. Updated the function signatures of the existing `WasmEdge_ModuleInstanceCreateWASI` and `WasmEdge_ModuleInstanceInitWASI` functions to accept new arguments:
   - `StdInFd`, `StdOutFd`, and `StdErrFd` for file descriptor redirection.

3. Updated the documentation to include information about these new functions.

These changes aim to enhance flexibility and support in WASI instances, allowing developers to redirect I/O streams as needed during execution.

## [include/host/wasi/environ.h](https://github.com/WasmEdge/WasmEdge/blob/55778d9d08cdb15e027fa30a658c67c6257d1ef5/include%2Fhost%2Fwasi%2Fenviron.h)

#### Potential issues

N/A

#### Summary of changes

- * Add new method `initWithFds` in the `Environ` class to allow redirecting stdin, stdout, and stderr.
- * This method takes additional integer parameters for file descriptors.
- * These fd values should be provided at construction time to enable redirection.

## [include/host/wasi/inode.h](https://github.com/WasmEdge/WasmEdge/blob/55778d9d08cdb15e027fa30a658c67c6257d1ef5/include%2Fhost%2Fwasi%2Finode.h)

#### Potential issues

N/A

#### Summary of changes

- **Changes**: The file `c-sdk/src/sdks/c/ino.h` was modified. Key changes include:
  - Adding a new static function `stdErr()` which returns the standard error stream.
  - Introducing a new class method `fromFd(int32_t Fd)` that allows redirecting of stdin, stdout, and stderr in a Wasi environment using file descriptors.

- **Most important change**: The addition of a new method `fromFd` enables flexibility in setting file handles for stdin, stdout, and stderr. This is particularly useful in scenarios where the program needs to manipulate these streams dynamically during execution.

## [include/host/wasi/vinode.h](https://github.com/WasmEdge/WasmEdge/blob/55778d9d08cdb15e027fa30a658c67c6257d1ef5/include%2Fhost%2Fwasi%2Fvinode.h)

#### Potential issues

-1. The `stdIn`, `stdOut`, and `stdErr` methods are missing the return type declaration for creating new file descriptors.
-2. The `bind` method does not handle the case where the bind operation would fail, which could lead to a segmentation fault.
-3. The `getAddrinfo` method is not returning any value when it fails, making the function less useful and harder to

#### Summary of changes

- **Added functions:**
  - `fromFd(int32_t Fd, __wasi_rights_t FRB, __wasi_rights_t FRI)` to allow redirection of stdin, stdout, and stderr from file descriptors.

- **Changed `canonialGuest` function:**
  - Updated the function to take a `std::string_view` instead of `const std::string&`.

- **Modified constructor calls:**
  - Added the new constructor call in each of the static functions.

## [lib/api/wasmedge.cpp](https://github.com/WasmEdge/WasmEdge/blob/55778d9d08cdb15e027fa30a658c67c6257d1ef5/lib%2Fapi%2Fwasmedge.cpp)

#### Potential issues

N/A

#### Summary of changes

-**Summary of Changes**:
- **New Function**: Added `WasmEdge_ModuleInstanceCreateWASIWithFds` to create a WasmEdge module instance with redirectable file descriptors (`StdInFd`, `StdOutFd`, `StdErrFd`).
- **Initialization Functions**: Modified existing functions to accept additional arguments for file descriptor redirections.
- **Function Signatures and Types**: Updated function signatures and return types for better clarity.

- **Main Changes**:
  1. Added new C API functions: `WasmEdge_ModuleInstanceCreateWASIWithFds`, `WasmEdge_ModuleInstanceInitWASI`, and `WasmEdge_ModuleInstanceInitWASIWithFds`.
  2. Modified existing functions to accept file descriptor arguments (`StdInFd`, `StdOutFd`, `StdErrFd`).

- **Environment Configuration**: The environment configuration in the Wasi module now supports file descriptor redirections for standard input, output, and error streams.

- **Logging**: Added logging for errors related to WASI environment initialization.

## [lib/host/wasi/environ.cpp](https://github.com/WasmEdge/WasmEdge/blob/55778d9d08cdb15e027fa30a658c67c6257d1ef5/lib%2Fhost%2Fwasi%2Fenviron.cpp)

#### Potential issues

-3 issues:

1. The code does not handle the case where the program name is empty.
2. There are no checks for exceeding the maximum allowed number of arguments for `argv`.
3. The destructor should be moved to the end of the file to avoid any dangling pointers.

#### Summary of changes

-**Key Changes:**
1. Added a new method `initWithFds` that allows redirecting stdin, stdout, and stderr.
2. Changed the constructor to use this new method with default fildes.
3. Updated the environment initialization process to use the new `initWithFds` method.
4. Removed unused code in the constructor.
5. Modified the destructor to clear any remaining environment variables.

## [lib/host/wasi/inode-linux.cpp](https://github.com/WasmEdge/WasmEdge/blob/55778d9d08cdb15e027fa30a658c67c6257d1ef5/lib%2Fhost%2Fwasi%2Finode-linux.cpp)

#### Potential issues

N/A

#### Summary of changes

- *Added a namespace for utility functions.*
- *Created `createStdNode` function to handle creating INode objects from standard file descriptors (stdin, stdout, stderr).*
- *Implemented methods `stdIn()`, `stdOut()`, and `stdErr()` to return the INode objects for these standard streams.*
- *Modified the existing `open` method to include a new argument for opening files by file descriptor.*

## [lib/host/wasi/inode-macos.cpp](https://github.com/WasmEdge/WasmEdge/blob/55778d9d08cdb15e027fa30a658c67c6257d1ef5/lib%2Fhost%2Fwasi%2Finode-macos.cpp)

#### Potential issues

N/A

#### Summary of changes

- **Important Changes**:
1. Added a namespace `WasiExpect` for managing expected exceptions.
2. Created a function `createStdNode` to handle file descriptor validation and creation of standard I/O nodes.
3. Moved the existing methods `stdIn`, `stdOut`, and `stdErr` directly to the class definition.

- **Minor Changes**:
1. Updated `reset` method in `DirHolder` to include a default case.
2. Added `stdIn`, `stdOut`, and `stdErr` methods returning instances of `INode`.
3. Moved `open` method from an existing namespace to the class definition.

## [lib/host/wasi/inode-win.cpp](https://github.com/WasmEdge/WasmEdge/blob/55778d9d08cdb15e027fa30a658c67c6257d1ef5/lib%2Fhost%2Fwasi%2Finode-win.cpp)

#### Potential issues

N/A

#### Summary of changes

- The patch introduces a new class `WasiExpect` that allows handling of WASI errors in a more structured way.
- It introduces a function `getWindowsHandle` that retrieves the Windows handle from an FD.
- The `INode::fromFd` function creates an `INode` instance from an FD.

## [lib/host/wasi/vinode.cpp](https://github.com/WasmEdge/WasmEdge/blob/55778d9d08cdb15e027fa30a658c67c6257d1ef5/lib%2Fhost%2Fwasi%2Fvinode.cpp)

#### Potential issues

-3
1. The `WASIAddrinfo` function is not implemented and should be implemented to handle address information retrieval.
2. The `sockOpen` function does not handle the `SockType` parameter.
3. The `directOpen` function does not implement the `RightsBase` and `RightsInheriting` parameters.

#### Summary of changes

- **Key Changes**:
  - Added a new function `fromFd(int32_t, __wasi_rights_t, __wasi_rights_t)` to allow redirecting the file descriptor (FD) of a `VINode`.
  - This function is implemented in the `VINode` class.
  
- **Most Important Changes**:
  - Introduced a helper function to convert a `std::shared_ptr<INode>` from a file descriptor, returning an `WasiExpect<std::shared_ptr<VINode>>` for error handling.
  - Added the `fromFd` function in the `VINode` class.

These changes enable the ability to dynamically redirect the input and output of a process in the C SDK using FDs, enhancing the flexibility and dynamic nature of the environment.

## [test/host/wasi/wasi.cpp](https://github.com/WasmEdge/WasmEdge/blob/55778d9d08cdb15e027fa30a658c67c6257d1ef5/test%2Fhost%2Fwasi%2Fwasi.cpp)

#### Potential issues

N/A

#### Summary of changes

- **Key Changes**:

1. Added support for custom file descriptors in the WasmEdge C SDK.
2. Created functions `openFileForWrite`, `openFileForRead`, and `closeFd` to handle file operations.
3. Defined helper functions `isValidFd`, `SetupTestFile`, `VerifyFileContent`, and `Cleanup` for managing test files.
4. Updated the `CustomFds` test suite to test different types of custom file descriptors: stdin, stdout, and stderr.
5. Added error handling checks in the test cases.

**Most Important Changes:**
- Introduced support for custom file descriptors.
- Enhanced the test framework with a `CustomFds` suite containing various test cases for different file descriptors.
- Added error handling and cleanup code to manage test files.

examples/capi/wasi-env/.gitignore
```diff
@@ -1 +1,4 @@
 target/
+set_wasi_env
+stdout.txt
+wasi_get_env.wasm
```
hydai: May I ask why these files were generated?

lib/api/wasmedge.cpp
```diff
@@ -2020,17 +2020,28 @@ WasmEdge_ModuleInstanceCreate(const WasmEdge_String ModuleName) {
...
                                   const uint32_t ArgLen,
                                   const char *const *Envs,
                                   const uint32_t EnvLen,
                                   const char *const *Preopens,
                                   const uint32_t PreopenLen) {
-  auto *WasiMod = new WasmEdge::Host::WasiModule();
-  WasmEdge_ModuleInstanceInitWASI(toModCxt(WasiMod), Args, ArgLen, Envs, EnvLen,
-                                  Preopens, PreopenLen);
-  return toModCxt(WasiMod);
+  return WasmEdge_ModuleInstanceCreateWASIExtended(
```
hydai: Why do you need to change the original behavior?

lib/api/wasmedge.cpp
```diff
@@ -2073,7 +2085,15 @@ WASMEDGE_CAPI_EXPORT void WasmEdge_ModuleInstanceInitWASI(
...
   auto &WasiEnv = WasiMod->getEnv();
-  WasiEnv.init(DirVec, ProgName, ArgVec, EnvVec);
+  WasiEnv.init(DirVec, ProgName, ArgVec, EnvVec, StdInFd, StdOutFd, StdErrFd);
+}
+
+WASMEDGE_CAPI_EXPORT void WasmEdge_ModuleInstanceInitWASI(
+    WasmEdge_ModuleInstanceContext *Cxt, const char *const *Args,
+    const uint32_t ArgLen, const char *const *Envs, const uint32_t EnvLen,
+    const char *const *Preopens, const uint32_t PreopenLen) {
+  WasmEdge_ModuleInstanceInitWASIExtended(Cxt, Args, ArgLen, Envs, EnvLen,
```
hydai: Ditto.

lib/host/wasi/environ.cpp
```diff
@@ -7,6 +7,11 @@
 #include "host/wasi/vfs.h"
 #include "host/wasi/vinode.h"
 
+#if WASMEDGE_OS_WINDOWS
+#include <fcntl.h>
+#include <io.h>
+#endif
+
```
hydai: If you need to separate the platform-independent code, please create a new file and include it just like we did in `wasi/inode-xxx`.

lib/host/wasi/environ.cpp
```diff
@@ -7,6 +7,11 @@
 #include "host/wasi/vfs.h"
 #include "host/wasi/vinode.h"
 
+#if WASMEDGE_OS_WINDOWS
+#include <fcntl.h>
+#include <io.h>
```
hydai: Can you identify which function or class definition requires this header?

hydai (changes requested):

examples/capi/wasi-env/.gitignore
PhantomInTheWire: The README instructs the user to generate the following:

```bash
cp ./target/wasm32-wasip1/release/wasi_get_env.wasm ../
clang set_wasi_env.c -o set_wasi_env -lwasmedge
```

In the new example, `stdout.txt` is where `stdout` is redirected:

```c
int out_fd = open("stdout.txt", O_WRONLY | O_CREAT | O_TRUNC, 0644);
```

lib/api/wasmedge.cpp
PhantomInTheWire: I updated the original functions to call the extended counterparts with default FDs to avoid duplicating initialization logic while preserving existing behavior. This keeps the API backwards compatible and follows DRY.

lib/host/wasi/environ.cpp
PhantomInTheWire: I think it was get _get_osfhandle but we don't need to worry about that anymore since these are already included in the win.h file included in environ-win.cpp

lib/host/wasi/environ.cpp
PhantomInTheWire: sure, i've moved the platform specific code to environ-xxx files

examples/capi/wasi-env/.gitignore
hydai: It makes no sense that we have to add these files to the ignore list. People can modify the command, so `set_wasi_env` can be any name they want. If you want to ignore it, using `*.txt`, `*.wasm`, and changing the name of the built object with `*.out` is better than this.

lib/host/wasi/environ-macos.cpp
```diff
@@ -11,6 +11,19 @@
...
+    __WASI_RIGHTS_FD_ADVISE | __WASI_RIGHTS_FD_FILESTAT_GET |
+    __WASI_RIGHTS_FD_READ | __WASI_RIGHTS_POLL_FD_READWRITE;
+static inline constexpr const __wasi_rights_t kStdOutDefaultRights =
+    __WASI_RIGHTS_FD_ADVISE | __WASI_RIGHTS_FD_DATASYNC |
+    __WASI_RIGHTS_FD_FILESTAT_GET | __WASI_RIGHTS_FD_SYNC |
+    __WASI_RIGHTS_FD_WRITE | __WASI_RIGHTS_POLL_FD_READWRITE;
+static inline constexpr const __wasi_rights_t kStdErrDefaultRights =
+    kStdOutDefaultRights;
+static inline constexpr const __wasi_rights_t kNoInheritingRights =
+    static_cast<__wasi_rights_t>(0);
```
hydai: Why do you duplicate these lines three times? What's the reason for not putting them into the header?

hydai (changes requested):

examples/capi/wasi-env/.gitignore
PhantomInTheWire: makes sense, i've changed the commands to use .out in readme and added *.wasm *.txt *.out to .gitignore

lib/host/wasi/environ-macos.cpp
PhantomInTheWire: Yeah, I don’t know why, but I feel nervous about adding new files, so I usually prefer modifying existing ones. fixed it. also is there a policy or guideline for creating new files, folders, test suites etc?

lib/host/wasi/environ.cpp
```diff
@@ -55,23 +55,15 @@ static inline constexpr const __wasi_rights_t kPreOpenInheritingRights =
...
-    static_cast<__wasi_rights_t>(0);
 static inline constexpr const auto kReadOnly = "readonly"sv;
 
 } // namespace
 
 void Environ::init(Span<const std::string> Dirs, std::string ProgramName,
-                   Span<const std::string> Args, Span<const std::string> Envs) {
+                   Span<const std::string> Args, Span<const std::string> Envs,
+                   uint32_t stdInFd, uint32_t stdOutFd, uint32_t stdErrFd) {
+  initStdFds(stdInFd, stdOutFd, stdErrFd);
```
hydai: What is the reason for moving the fd initialization to the very first section?

test/host/wasi/wasi.cpp
```diff
@@ -2998,6 +2999,168 @@ TEST(WasiTest, SymbolicLink) {
...
+    int TestFd = -1;
+
+#if WASMEDGE_OS_WINDOWS
+    ASSERT_EQ(_sopen_s(&TestFd, TempFilename, _O_WRONLY | _O_CREAT | _O_TRUNC,
+                       _SH_DENYNO, _S_IREAD | _S_IWRITE),
+              0);
+#else
+    TestFd =
+        open(TempFilename, O_WRONLY | O_CREAT | O_TRUNC, S_IRUSR | S_IWUSR);
+#endif
```
hydai: Can we extract these parts into a helper function?  
This would allow us to use the following syntax:

```cpp
TestFd = openFd(TempFilename);
closeFd(TestFd);
```

And the `#ifdef` sections will be hidden inside these two functions.

hydai (changes requested):
Your test cases only cover valid file descriptors as customized stdin, stdout, and stderr. What if users try to bind an invalid file descriptor to these functions? How will it be tested?

lib/host/wasi/environ.cpp
PhantomInTheWire: I think it makes more sense to initialize all basic file descriptors before processing or assigning any additional file descriptors.

test/host/wasi/wasi.cpp
PhantomInTheWire: makes sense. done.

PhantomInTheWire:
I've added new tests but the invalid custom stdin seems to be failing in windows CI, I don't have a windows VM setup right now, I'll probably have to set one up locally or on the cloud to fix it, so for now I've added a TODO. I'll try fixing it later this week.

hydai:
> I've added new tests but the invalid custom stdin seems to be failing in windows CI, I don't have a windows VM setup right now, I'll probably have to set one up locally or on the cloud to fix it, so for now I've added a TODO. I'll try fixing it later this week.

Hi, you can simply create a PR to your own forked repo to trigger the Windows CI instead of trying to ping us to approve the workflow execution. I will no longer approve the workflows, as it will also trigger other CI resources, which could occupy the hardware of others. Please show the CI pass results from your own repo to continue this review process.

hydai:
I will convert this into a draft PR at this moment. Thank you again for your contribution.

PhantomInTheWire:
Hi @hydai,

Just to clarify, I think there might have been a misunderstanding here. I was already running CI on my fork (the PR on the fork and this PR share the same branch, which is probably why you were getting pinged). The frequent pushes were due to me trying to debug the failing Windows CI.

For now, I’ve disabled the Windows-specific test(the new invalid custom stdin fd test that I just added) and marked it with a TODO since I don’t have a Windows VM set up to debug it fully, but everything else is [passing here](https://github.com/PhantomInTheWire/WasmEdge/actions/runs/16114371430/job/45464924467?pr=2).

Please let me know if there’s anything else you’d like before moving this out of draft. Thanks for your time and review!

hydai:
Thanks for the clarification. However, if there is a test that cannot pass on Windows, I am unable to merge this PR. The only reason is that it would introduce unexpected behavior for some users on Windows with this new feature. Does that make sense?

I think you could create a single, limited workflows PR to your own branch to try to fix the test case with CI, if you don't have such an environment to debug with.

PhantomInTheWire:
I totally get that. It was taking too long to debug this on ci since the ci runs are very long(even with custom workflows), I've set up a VM on azure, it'll take some time to set the dev env up on it. I'll push the fix soon hopefully.

PhantomInTheWire:
it seems like the windows CI was was hanged on the invalid stdin tests, i've modified the tests. but i am not sure this is the best way to approach this. i would appreciate some help here @hydai  https://github.com/PhantomInTheWire/WasmEdge/actions/runs/16165202026/job/45625437121?pr=2

PhantomInTheWire:
Hi @hydai, following up I don't think there is any other way to test this on windows, opening for review now

PhantomInTheWire:
Hi @hydai, @q82419 @ibmibmibm 

Just checking in on this PR as it’s been a while without updates. Let me know if there’s anything you’d like me to adjust or clarify to help move this forward. Appreciate your time and review whenever you get a chance!

include/api/wasmedge/wasmedge.h
```diff
@@ -2149,6 +2149,18 @@ WasmEdge_ModuleInstanceCreateWASI(const char *const *Args,
...
+///
+/// \param StdInFd the file descriptor for standard input.
+/// \param StdOutFd the file descriptor for standard output.
+/// \param StdErrFd the file descriptor for standard error.
+WASMEDGE_CAPI_EXPORT extern WasmEdge_ModuleInstanceContext *
+WasmEdge_ModuleInstanceCreateWASIExtended(
+    const char *const *Args, const uint32_t ArgLen, const char *const *Envs,
+    const uint32_t EnvLen, const char *const *Preopens,
+    const uint32_t PreopenLen, const uint32_t StdInFd, const uint32_t StdOutFd,
+    const uint32_t StdErrFd);
```
hydai: You are using a very unusual pattern.  
All FD are set to `uint32_t`; however, every time, you need to `static_cast` them to `int32_t` first and then proceed with the remaining steps.

What is the reason? Please explain the logic you are trying to implement in the whole PR.

include/api/wasmedge/wasmedge.h
```diff
@@ -2169,6 +2181,18 @@ WASMEDGE_CAPI_EXPORT extern void WasmEdge_ModuleInstanceInitWASI(
     const uint32_t ArgLen, const char *const *Envs, const uint32_t EnvLen,
     const char *const *Preopens, const uint32_t PreopenLen);
 
+/// Same as WasmEdge_ModuleInstanceInitWASI but with extended support for File
+/// Descriptors
+///
+/// \param StdInFd the file descriptor for standard input.
+/// \param StdOutFd the file descriptor for standard output.
+/// \param StdErrFd the file descriptor for standard error.
+WASMEDGE_CAPI_EXPORT extern void WasmEdge_ModuleInstanceInitWASIExtended(
```
hydai: I don't think "Extended" is a good idea for the name. It doesn't convey the intention of this function.

lib/host/wasi/environ.cpp
```diff
@@ -111,10 +103,6 @@ void Environ::init(Span<const std::string> Dirs, std::string ProgramName,
 
     std::sort(PreopenedDirs.begin(), PreopenedDirs.end());
 
-    FdMap.emplace(0, VINode::stdIn(kStdInDefaultRights, kNoInheritingRights));
-    FdMap.emplace(1, VINode::stdOut(kStdOutDefaultRights, kNoInheritingRights));
-    FdMap.emplace(2, VINode::stdErr(kStdErrDefaultRights, kNoInheritingRights));
-
```
hydai: What is the reason for implementing three different methods for the initialization process, whereas in the previous implementation we relied on only one?

For example, we could use `VINode::stdIn(StdInFd, ...)` instead.

Then, you wouldn't have to extract the rights to a new file.

hydai (changes requested):

include/api/wasmedge/wasmedge.h
PhantomInTheWire: I initially tried to avoid checking for negative values, but that ended up causing more issues. I've refactored the code to handle this properly. I'll add a detailed explanation in the PR description.(my bad should've added that since this is a change to the dev facing c-api)

include/api/wasmedge/wasmedge.h
PhantomInTheWire: how about `Extended` -> `WithFds`

lib/host/wasi/environ.cpp
PhantomInTheWire: I was initially doing everything in one method but the involved a lot of platform dependent code, which was then moved to 3 different files and lead to right.h being created. but you're right this makes more sense and is a lot cleaner. fixed.

PhantomInTheWire:
Hi @hydai can you take a look again when you get a chance? thanks.

lib/api/wasmedge.cpp
```diff
@@ -2020,17 +2020,28 @@ WasmEdge_ModuleInstanceCreate(const WasmEdge_String ModuleName) {
...
                                   const char *const *Envs,
                                   const uint32_t EnvLen,
                                   const char *const *Preopens,
                                   const uint32_t PreopenLen) {
-  auto *WasiMod = new WasmEdge::Host::WasiModule();
-  WasmEdge_ModuleInstanceInitWASI(toModCxt(WasiMod), Args, ArgLen, Envs, EnvLen,
-                                  Preopens, PreopenLen);
-  return toModCxt(WasiMod);
+  return WasmEdge_ModuleInstanceCreateWASIWithFds(
+      Args, ArgLen, Envs, EnvLen, Preopens, PreopenLen, 0, 1, 2);
```
hydai: I have a question. According to the original design, the default FD will be the macro defined by their own platform, e.g., `STDIN_FILENO`. However, you are trying to apply magic numbers here. That's why I mentioned that you should not wrap this with the `WithFds` edition.

lib/host/wasi/inode-win.cpp
```diff
@@ -330,6 +330,26 @@ SymlinkPriviledgeHolder SymlinkPriviledgeHolder::Holder;
 
 } // namespace
 
+namespace {
+static winapi::HANDLE_ getWindowsHandle(int32_t Fd, winapi::DWORD_ StdHandleId,
+                                        std::string_view Name) {
```
hydai: Why do we need `Name` here? It seems like it's going to be used inside the warning only.

lib/host/wasi/inode-win.cpp
```diff
@@ -330,6 +330,26 @@ SymlinkPriviledgeHolder SymlinkPriviledgeHolder::Holder;
...
+  if (Fd < 0) {
+    return winapi::GetStdHandle(StdHandleId);
+  }
+
+  auto raw = _get_osfhandle(Fd);
```
hydai: Please follow the naming style. If you are unsure of the rules, please run clang-tidy or follow the LLVM coding style guide.

lib/host/wasi/inode-win.cpp
```diff
@@ -330,6 +330,26 @@ SymlinkPriviledgeHolder SymlinkPriviledgeHolder::Holder;
...
+  auto Handle = reinterpret_cast<winapi::HANDLE_>(raw);
+
+  if (Handle != winapi::INVALID_HANDLE_VALUE_ && Handle != nullptr) {
+    return Handle;
+  }
+  spdlog::warn("    Failed to get OS handle for custom {} FD: {}. "
+               "Falling back to default.",
+               Name, Fd);
```
hydai: I would like to discuss this behavior: do we really want users to receive a default FD when the provided FD is invalid?  
If so, we must document this behavior so that people understand what is happening when the given FD is broken or invalid. Otherwise, we should report a fatal error to indicate that this is unacceptable.

cc @q82419 @ibmibmibm

hydai (changes requested):

lib/host/wasi/inode-win.cpp
PhantomInTheWire: fxied.

lib/host/wasi/inode-win.cpp
PhantomInTheWire: fixed.

lib/api/wasmedge.cpp
PhantomInTheWire: I looked at the node.js implementation and they seemed to be using 0,1,2 as well so I don't think there should be any issues how ever to avoid edge cases I think we could just use -1 as default for all of these and then it can fall back to default ones like `STDIN_FILENO` based on inode-* implementations for now.

lib/host/wasi/inode-win.cpp
PhantomInTheWire: @hydai I agree we should either document this fallback or make it a hard failure. Open to feedback on which direction makes more sense here.

lib/api/wasmedge.cpp
hydai: You definitely cannot use `-1`, as it is one of the possible bad FDs defined in the spec.  
If you can confirm that all platforms (Windows, macOS, Linux, BSD, and others) are always using 0, 1, 2 without any exceptions, then it's acceptable. Otherwise, we should keep the default macro from their own definitions instead of guessing the possible situation.

PhantomInTheWire: On POSIX systems like macOS, Linux, and BSD, the behavior is consistent: file descriptors `0`, `1`, and `2` correspond to `stdin`, `stdout`, and `stderr`. However, Windows does not implement file descriptors in the same way. Instead, it uses handles, and compatibility with file descriptors is provided via the C runtime, which maps `0`, `1`, and `2` to the appropriate standard streams.

Given this, using system-specific macros instead of hardcoded magic numbers is relatively straightforward to implement, and doing so ***may*** be slightly safer and more readable, maintainable etc. I'll make that change.

* Node.js implementation: [[wasi.js#L84](https://github.com/nodejs/node/blob/cde8f275ad6ceecc4837fa8a64ba948b39d084b3/lib/wasi.js#L84)](https://github.com/nodejs/node/blob/cde8f275ad6ceecc4837fa8a64ba948b39d084b3/lib/wasi.js#L84)
* StackOverflow discussion: [Is it possible that Linux file descriptor 0, 1, 2 not for stdin, stdout, and stderr?](https://stackoverflow.com/questions/22367920/is-it-possible-that-linux-file-descriptor-0-1-2-not-for-stdin-stdout-and-stderr)

PhantomInTheWire:
@hydai @q82419 @ibmibmibm,
update: I think we should just let it throw an error if a bad fd is passed, the fallback behaviour if needed can be introduced in another PR.
I have also updated the patch to now use macros instead of magic numbers.

PS: don't mind the force pushes it was me fighting with windows CI again on my fork that shares the same branch as thir PR.

lib/host/wasi/environ.cpp
```diff
@@ -7,6 +7,10 @@
 #include "host/wasi/vfs.h"
 #include "host/wasi/vinode.h"
 
+#if !WASMEDGE_OS_WINDOWS
+#include <unistd.h>
+#endif
+
```
hydai: I don't think this is correct, since you are trying to make things more complex.  
I would like to have two init functions:  
1. The default one, which keeps the old way  
2. The new one with specific FDs, which will return an error if the given FD is invalid.  

This will never affect the original users and will allow those who want to customize their own FDs with a specific function.

lib/host/wasi/inode-linux.cpp
```diff
@@ -155,12 +155,21 @@ void DirHolder::reset() noexcept {
     Cookie = 0;
   }
 }
+namespace {
+inline int getValidFd(int32_t Fd) { return Fd; }
+} // namespace
```
hydai: I don't understand the point; this function does nothing.

lib/host/wasi/inode-win.cpp
```diff
@@ -785,6 +815,9 @@ WasiExpect<void> INode::fdDatasync() const noexcept {
 }
 
 WasiExpect<void> INode::fdFdstatGet(__wasi_fdstat_t &FdStat) const noexcept {
+  if (Handle == nullptr) {
+    return WasiUnexpect(__WASI_ERRNO_BADF);
+  }
```
hydai: What's the reason we need to do this special check compared to other platforms?

include/common/defines.h
```diff
@@ -51,3 +51,9 @@
 #else
 #define WASMEDGE_EXPORT [[gnu::visibility("default")]]
 #endif
+
+#if WASMEDGE_OS_WINDOWS
+#define STDIN_FILENO 0  /* standard input file descriptor */
+#define STDOUT_FILENO 1 /* standard output file descriptor */
+#define STDERR_FILENO 2 /* standard error file descriptor */
+#endif
```
hydai: This is not the correct way to handle it.  
We are expecting the value from the headers, not a self-defined one. If you just want to do this, then it's like you are writing 0, 1, 2 directly.

lib/host/wasi/inode-linux.cpp
q82419: Ditto.

include/common/defines.h
PhantomInTheWire: There is no `STDIN_FILENO` macro on windows(although there should be one in something like cygwin) since fds work differently there they use handles, the internal windows implementation in inode-win.cpp checks for it and does it in a platform specific way without actually using the hardcoded magic numbers. the implementation is something like this.

```cpp
if (Fd == STDIN_FILENO)
  return GetStdHandle(STD_INPUT_HANDLE_);
if (Fd == STDOUT_FILENO)
  return GetStdHandle(STD_OUTPUT_HANDLE_);
if (Fd == STDERR_FILENO)
  return GetStdHandle(STD_ERROR_HANDLE_);
```

lib/host/wasi/inode-linux.cpp
PhantomInTheWire: my bad I forgot to clean it up.

lib/host/wasi/inode-win.cpp
PhantomInTheWire: it was needed on windows because the windows implementation was earlier returning nullptrs.(i've changed the implementation to do things better so we don't need these checks anymore)

lib/host/wasi/environ.cpp
PhantomInTheWire: sure. I've refactored the new function to gracefully propagate errors.

PhantomInTheWire:
Hey @hydai @q82419,

I’ve made a significant refactor to the error handling, but I’m unsure if I’m propagating the errors in the right way. Could you guys take a look when you get a chance? Thanks.
Also it does show a fatal error now that indicates that it is not acceptable.(that's what you meant right or am i misunderstanding something?)

include/host/wasi/inode.h
```diff
@@ -346,11 +346,9 @@ class INode
...
   INode &operator=(INode &&RHS) noexcept = default;
 
-  static INode stdIn() noexcept;
-
-  static INode stdOut() noexcept;
-
-  static INode stdErr() noexcept;
+  static WasiExpect<INode> stdIn(int32_t Fd);
+  static WasiExpect<INode> stdOut(int32_t Fd);
+  static WasiExpect<INode> stdErr(int32_t Fd);
```
hydai: Just to clarify, I prefer:
1. Keep the original functions.
2. Create a new function called `static WasiExpect<INode> fromFd(int32_t Fd)`.

The reason is that the overall logic is different when handling `stdin`, `stdout`, and `stderr` with a given FD.

include/host/wasi/vinode.h
```diff
@@ -46,12 +46,12 @@ class VINode : public std::enable_shared_from_this<VINode> {
...
-  static std::shared_ptr<VINode> stdOut(__wasi_rights_t FRB,
-                                        __wasi_rights_t FRI);
-  static std::shared_ptr<VINode> stdErr(__wasi_rights_t FRB,
-                                        __wasi_rights_t FRI);
+  static WasiExpect<std::shared_ptr<VINode>>
+  stdIn(int32_t Fd, __wasi_rights_t FRB, __wasi_rights_t FRI);
+  static WasiExpect<std::shared_ptr<VINode>>
+  stdOut(int32_t Fd, __wasi_rights_t FRB, __wasi_rights_t FRI);
+  static WasiExpect<std::shared_ptr<VINode>>
+  stdErr(int32_t Fd, __wasi_rights_t FRB, __wasi_rights_t FRI);
```
hydai: So, we will have two different kinds of function signatures:

1. The previous ones
2. A new function called `static WasiExpect<std::shared_ptr<VINode>> fromFd(int32_t Fd, __wasi...)`

lib/api/wasmedge.cpp
```diff
@@ -23,6 +23,10 @@
 #include "driver/wasiNNRPCServerTool.h"
 #endif
 
+#if !WASMEDGE_OS_WINDOWS
+#include <unistd.h>
+#endif
+
```
hydai: And we no longer need this because the new function signature will ask users to provide all given FDs.

lib/api/wasmedge.cpp
```diff
@@ -2073,7 +2090,21 @@ WASMEDGE_CAPI_EXPORT void WasmEdge_ModuleInstanceInitWASI(
...
+  }
+}
+
+WASMEDGE_CAPI_EXPORT void WasmEdge_ModuleInstanceInitWASI(
+    WasmEdge_ModuleInstanceContext *Cxt, const char *const *Args,
+    const uint32_t ArgLen, const char *const *Envs, const uint32_t EnvLen,
+    const char *const *Preopens, const uint32_t PreopenLen) {
+  WasmEdge_ModuleInstanceInitWASIWithFds(Cxt, Args, ArgLen, Envs, EnvLen,
+                                         Preopens, PreopenLen, STDIN_FILENO,
+                                         STDOUT_FILENO, STDERR_FILENO);
```
hydai: The only reason you need to use macros is because you are trying to reuse `WasmEdge_ModuleInstanceInitWASIWithFds` inside `WasmEdge_ModuleInstanceInitWASI`. If we don't reuse it, these macros are no longer required here.

lib/host/wasi/environ.cpp
hydai: Ditto @lib/api/wasmedge.cpp

lib/host/wasi/environ.cpp
```diff
@@ -72,6 +76,35 @@ static inline constexpr const auto kReadOnly = "readonly"sv;
 
 void Environ::init(Span<const std::string> Dirs, std::string ProgramName,
                    Span<const std::string> Args, Span<const std::string> Envs) {
+  auto Result = initWithFds(Dirs, std::move(ProgramName), Args, Envs,
+                            STDIN_FILENO, STDOUT_FILENO, STDERR_FILENO);
+  if (!Result) {
+    spdlog::error("Environment initialization failed: {}", Result.error());
```
hydai: Again, I prefer to keep two separate functions.  
The reason is that the previous `init` will never fail when initializing `STDIN`, `STDOUT`, and `STDERR`.  
However, when you try to handle and reuse the code, it will create a breaking change, causing the behavior to be unexpectedly different here.

lib/host/wasi/inode-linux.cpp
```diff
@@ -156,11 +156,33 @@ void DirHolder::reset() noexcept {
   }
 }
 
-INode INode::stdIn() noexcept { return INode(STDIN_FILENO); }
+namespace {
+WasiExpect<INode> createStdNode(int32_t Fd, const char *streamName) {
```
hydai: Again, I don't think we need to pass `streamName` here just for error logging.

lib/host/wasi/inode-linux.cpp
```diff
@@ -156,11 +156,33 @@ void DirHolder::reset() noexcept {
...
+  return createStdNode(Fd, "stdin");
+}
+
+WasiExpect<INode> INode::stdOut(int32_t Fd) {
+  return createStdNode(Fd, "stdout");
+}
+
+WasiExpect<INode> INode::stdErr(int32_t Fd) {
+  return createStdNode(Fd, "stderr");
+}
```
hydai: You can keep the original ones, and have a new function called `WasiExpect<INode> INode::fromFd(int32_t Fd)` for all three functions.

lib/host/wasi/vinode.cpp
```diff
@@ -28,19 +28,31 @@ VINode::VINode(INode Node, __wasi_rights_t FRB, __wasi_rights_t FRI,
...
-std::shared_ptr<VINode> VINode::stdErr(__wasi_rights_t FRB,
-                                       __wasi_rights_t FRI) {
-  return std::make_shared<VINode>(INode::stdErr(), FRB, FRI);
+WasiExpect<std::shared_ptr<VINode>>
+VINode::stdErr(int32_t Fd, __wasi_rights_t FRB, __wasi_rights_t FRI) {
+  auto NodeResult = INode::stdErr(Fd);
+  if (!NodeResult) {
+    return WasiUnexpect(NodeResult.error());
+  }
+  return std::make_shared<VINode>(std::move(*NodeResult), FRB, FRI);
```
hydai: So, the original ones are not changed, and you will have a new function here:

```cpp
WasiExpect<std::shared_ptr<VINode>>
VINode::fromFd(int32_t Fd, __wasi_rights_t FRB, __wasi_rights_t FRI) {
  auto NodeResult = INode::fromFd(Fd);
  if (!NodeResult) {
    return WasiUnexpect(NodeResult.error());
  }
  return std::make_shared<VINode>(std::move(*NodeResult), FRB, FRI);
```

PhantomInTheWire:
Thankyou @hydai for taking the time to give such detailed explanation of the expected changes. It took a while to go through all of it but I've now updated the implementation to match your requirements and squashed all the commits.

test/host/wasi/wasi.cpp
```diff
@@ -2998,6 +2999,236 @@ TEST(WasiTest, SymbolicLink) {
...
+  auto VerifyFileContent = [&](const std::string_view &Expected) {
+    std::ifstream File(TempFilename, std::ios::binary);
+    if (!File.is_open()) {
+      ADD_FAILURE() << "Failed to open temp file for verification";
+      return;
+    }
+    std::string Content((std::istreambuf_iterator<char>(File)),
+                        std::istreambuf_iterator<char>());
+    EXPECT_EQ(Content, Expected);
+    std::remove(TempFilename);
```
ibmibmibm: use `std::filesystem::remove` instead

test/host/wasi/wasi.cpp
```diff
@@ -2998,6 +2999,236 @@ TEST(WasiTest, SymbolicLink) {
...
+  };
+
+  auto Cleanup = [&](int TestFd) {
+    closeFd(TestFd);
+    std::remove(TempFilename);
```
ibmibmibm: use `std::filesystem::remove` instead

test/host/wasi/wasi.cpp
```diff
@@ -2998,6 +2999,236 @@ TEST(WasiTest, SymbolicLink) {
...
+  Mod.addHostMemory(
+      "memory", std::make_unique<WasmEdge::Runtime::Instance::MemoryInstance>(
+                    WasmEdge::AST::MemoryType(1)));
+  auto *MemInstPtr = Mod.findMemoryExports("memory");
+  ASSERT_TRUE(MemInstPtr != nullptr);
+  auto &MemInst = *MemInstPtr;
+  WasmEdge::Runtime::CallingFrame CallFrame(nullptr, &Mod);
+
+  std::array<WasmEdge::ValVariant, 1> Errno;
+  const char TempFilename[] = "wasi_custom_fd_test.tmp";
```
ibmibmibm: use `std::filesystem::u8path` instead

ibmibmibm (changes requested):

test/host/wasi/wasi.cpp
PhantomInTheWire: done.

test/host/wasi/wasi.cpp
PhantomInTheWire: done.

test/host/wasi/wasi.cpp
PhantomInTheWire: done.

lib/api/wasmedge.cpp
```diff
@@ -2076,6 +2089,43 @@ WASMEDGE_CAPI_EXPORT void WasmEdge_ModuleInstanceInitWASI(
...
+      EnvVec.emplace_back(Envs[I]);
+    }
+  }
+  if (Preopens) {
+    for (uint32_t I = 0; I < PreopenLen; I++) {
+      DirVec.emplace_back(Preopens[I]);
+    }
+  }
+  auto &WasiEnv = WasiMod->getEnv();
+  WasiEnv.initWithFds(DirVec, ProgName, ArgVec, EnvVec, StdInFd, StdOutFd,
```
hydai: `initWithFds` will return an error if the FDs are broken. However, the failure logic is not handled here.

lib/host/wasi/inode-win.cpp
```diff
@@ -330,6 +330,29 @@ SymlinkPriviledgeHolder SymlinkPriviledgeHolder::Holder;
 
 } // namespace
 
+namespace {
+static WasiExpect<winapi::HANDLE_> getWindowsHandle(int32_t Fd) {
+
+  if (Fd < 0) {
+    spdlog::error("Invalid file descriptor: {}", Fd);
+    return WasiUnexpect(__WASI_ERRNO_BADF);
```
hydai: May I ask why this error format is not the same as the one used in macOS and Linux?

lib/api/wasmedge.cpp
PhantomInTheWire: handled. should be fine now.

lib/host/wasi/inode-win.cpp
PhantomInTheWire: my bad. it should be consistent. fixed.

hydai:
Thanks.

Issue #3958, opened 2025-01-08
bug: Empty generation returned if `n_predict` is `-1` or `-2`


apepkuss:
### Summary

When the metadata field `n_predict` is set to `-1` or `-2` by following the definition given by llama.cpp, the generation is an empty string. The following snapshot shows the case of `n_predict=-2`: 

![image](https://github.com/user-attachments/assets/f90366b9-0ab2-4c06-bdb9-3b17d7993771)

### Reproduction steps

N/A

### Screenshots

N/A

### Components

Others

### WasmEdge Version or Commit you used

0.14.1 with ggml plugin b4419

### Operating system information

macOS 15.1.1

### Hardware Architecture

aarch64

hydai:
Please check the specifications again regarding this value, as the version has changed significantly.

hydai:
Fixed by #4208

PR #4208, opened 2025-06-29
fix(wasi_nn, ggml):  Empty generation returned if n_predict is -1 or -2

PhantomInTheWire:
**Resolves:**

* https://github.com/WasmEdge/WasmEdge/issues/3958

**Changes:**
`n_predict = -1` and `n_predict = -2` now correctly generate tokens during inference.

cc: @apepkuss

### DIFF
--- a/plugins/wasi_nn/wasinn_ggml.cpp
+++ b/plugins/wasi_nn/wasinn_ggml.cpp
@@ -744,7 +744,7 @@ ErrNo parseMetadata(Graph &GraphRef, LocalConfig &ConfRef,
     GraphRef.Params.sampling.dry_allowed_length =
         static_cast<int32_t>(DryAllowedLength);
   }
-  if (Doc.at_key("dry-penalty-last-n").error() == simdjson::SUCCESS) {
+  if (Doc.at_key("dry-last-n-penalty").error() == simdjson::SUCCESS) {
     int64_t DryLastNPenalty;
     auto Err = Doc["dry-last-n-penalty"].get<int64_t>().get(DryLastNPenalty);
     if (Err) {
@@ -2260,7 +2260,9 @@ ErrNo sampleOutput(Graph &GraphRef, Context &CxtRef,
   }
   // Deal with end of text token.
   const llama_vocab *Vocab = llama_model_get_vocab(GraphRef.LlamaModel.get());
-  if (llama_vocab_is_eog(Vocab, common_sampler_last(CxtRef.LlamaSampler))) {
+  // Only stop on EOS if GraphRef.Params.sampling.ignore_eos is false.
+  if (!GraphRef.Params.sampling.ignore_eos &&
+      llama_vocab_is_eog(Vocab, common_sampler_last(CxtRef.LlamaSampler))) {
     LOG_INFO(GraphRef.EnableLog, "sampleOutput: EOS token found."sv)
     return ErrNo::EndOfSequence;
   }
@@ -3140,8 +3142,10 @@ Expect<ErrNo> compute(WasiNNEnvironment &Env, uint32_t ContextId) noexcept {
 
   // Main prediction loop.
   LOG_DEBUG(GraphRef.EnableDebugLog, "compute: enter main prediction loop"sv)
-  int64_t NRemain = CxtRef.Conf.NPredict;
-  while (NRemain-- > 0) {
+  int64_t NPredict =
+      CxtRef.Conf.NPredict < 0 ? INT32_MAX : CxtRef.Conf.NPredict;
+
+  while (NPredict-- > 0) {
     ReturnCode = sampleOutput(GraphRef, CxtRef);
     if (ReturnCode != ErrNo::Success) {
       break;


juntao:

------

## [plugins/wasi_nn/wasinn_ggml.cpp](https://github.com/WasmEdge/WasmEdge/blob/bc042eda146a17c9e40ab81e35a2edc8b1d3cf86/plugins%2Fwasi_nn%2Fwasinn_ggml.cpp)

#### Potential issues

1. **Potential Integer Overflow**: The code does not check for integer overflow when converting `int64_t` to `int32_t` using `static_cast<int32_t>()`.
2. **Uninitialized Variables**: Some variables might be used without being initialized if the corresponding keys are missing in the JSON, potentially leading to undefined behavior.
3. **Error Handling Inconsistency**: The code uses a mix of returning errors and logging errors with `spdlog::error`, which can lead to inconsistencies in error handling.

The most important issue is the potential for integer overflow due to unchecked conversions from `int64_t` to `int32_t`. This could lead to data corruption or unexpected behavior.

#### Summary of changes

- **Key Changes:**
  1. **Updated Key Name**: Changed `"dry-penalty-last-n"` to `"dry-last-n-penalty"` for consistent naming.
 2. **EOS Check Conditional**: Added a conditional check `GraphRef.Params.sampling.ignore_eos` before checking for end-of-sequence (EOS) token in the `sampleOutput()` function.
 3. **Loop Condition Update**: Modified the main prediction loop to handle cases where `NPredict` is -1 or -2 by setting it to `INT32_MAX`, allowing infinite generation.

PhantomInTheWire:
Thanks for the feedback, @hydai.

I’ve pushed a revised implementation that tries to aligns closely to the handling of `n_predict` in [[llama.cpp mtmd-cli.cpp](https://github.com/ggml-org/llama.cpp/blob/master/tools/mtmd/mtmd-cli.cpp)].
Locally, both `n_predict = -1` and `n_predict = -2` now generate tokens as expected, with `-2` filling the context window more aggressively. Here’s the [output](https://pastebin.com/EiMmTX5s) for reference.

That said, I may still have gaps in understanding here, so please let me know if there’s something I’ve missed or need to adjust further.

plugins/wasi_nn/wasinn_ggml.cpp
```diff
@@ -3140,14 +3142,16 @@ Expect<ErrNo> compute(WasiNNEnvironment &Env, uint32_t ContextId) noexcept {
 
   // Main prediction loop.
   LOG_DEBUG(GraphRef.EnableDebugLog, "compute: enter main prediction loop"sv)
-  int64_t NRemain = CxtRef.Conf.NPredict;
-  while (NRemain-- > 0) {
+  const int64_t NToPredict =
+      CxtRef.Conf.NPredict < 0 ? INT32_MAX : CxtRef.Conf.NPredict;
+
+  for (int64_t I = 0; I < NToPredict; ++I) {
```
hydai:
```suggestion
  int64_t NPredict =
      CxtRef.Conf.NPredict < 0 ? INT32_MAX : CxtRef.Conf.NPredict;

  while(NPredict-- > 0) {
```

hydai:
Please fix the clang-format issue.

hydai:
Hi @dm4,
I think it's okay to proceed with the merge after the lint issue is resolved. Could you please take another look at it?

plugins/wasi_nn/wasinn_ggml.cpp
```diff
@@ -3140,14 +3142,16 @@ Expect<ErrNo> compute(WasiNNEnvironment &Env, uint32_t ContextId) noexcept {
...
+      CxtRef.Conf.NPredict < 0 ? INT32_MAX : CxtRef.Conf.NPredict;
+
+  while(NPredict-- > 0) {
     ReturnCode = sampleOutput(GraphRef, CxtRef);
     if (ReturnCode != ErrNo::Success) {
       break;
     }
   }
-  if (ReturnCode == ErrNo::EndOfSequence) {
+  if (ReturnCode == ErrNo::EndOfSequence || ReturnCode == ErrNo::ContextFull) {
```
dm4: Currently, there are some WASM examples [like this one](https://github.com/second-state/WasmEdge-WASINN-examples/blob/67f174bab59d98c1b52f7367ec0928701dc998f9/wasmedge-ggml/multimodel/src/main.rs#L161) that determine whether the `wasi-nn` plugin returns `ErrNo::ContextFull` to decide whether to increase the context size next time. Perhaps we should retain this return value?

hydai: You are right. We should keep the `ContextFull` error so users can understand that the reason for stopping is due to a full context rather than other errors.

PhantomInTheWire: I suppose an error would be better than silent fail

hydai: If you reassigned the result, this will cause the silent failure.
However, it will let users know this error happens when the correct error code returns.

PhantomInTheWire: that makes sense.

dm4 (approved):
LGTM

PR #4188, opened 2025-06-16
Add conditional check for stdbool.h before C23 

GautamBytes:
## Summary
Adds conditional inclusion of `stdbool.h` to prevent deprecation issues in C23.

## Changes
- Replace `#include <stdbool.h>` with conditional check
- Only include `stdbool.h` for C standards before C23 (`__STDC_VERSION__ < 202311L`)

Fixes #4147

PR #4225, opened 2025-07-01
fix(CAPI): no longer include stdbool.h after C23

### LINKED ISSUE #4147: feat: check c version before `#include <stdbool.h>` in wasmedge.h

hydai:
Fixes #4147

Replace #4188 due to inactivity.

### DIFF
--- a/include/api/wasmedge/wasmedge.h
+++ b/include/api/wasmedge/wasmedge.h
@@ -32,7 +32,10 @@
 #define WASMEDGE_CAPI_PLUGIN_EXPORT __attribute__((visibility("default")))
 #endif // _WIN32
 
+#if !defined(__cplusplus) &&                                                   \
+    (!defined(__STDC_VERSION__) || __STDC_VERSION__ < 202311L)
 #include <stdbool.h>
+#endif
 #include <stdint.h>
 #include <time.h>
 


juntao:

------

## [include/api/wasmedge/wasmedge.h](https://github.com/WasmEdge/WasmEdge/blob/34fbbabe66bf67776c9b93df49bd9737373ea89a/include%2Fapi%2Fwasmedge%2Fwasmedge.h)

#### Potential issues

1. **Include guard mismatch**: The file is missing a `#endif` for the `#ifdef __cplusplus` block at the end of the file.
2. **Missing inclusion of `stdint.h`**: In C23, `_Bool` and `bool` are part of the standard library, so there's no need to include `<stdbool.h>` if `__STDC_VERSION__ < 202311L`.
3. **Potential integer overflow in `WasmEdge_ResultGen`**: The comment states that the error code is 24-bit, but `uint32_t` is used, which could lead to unexpected behavior if values exceed 24 bits.

#### Summary of changes

- **Conditional inclusion of `stdbool.h`**: The patch modifies the include directive for `<stdbool.h>` to conditionally include it only when not compiling with C23 standard or higher, and not in C++ mode. This change aims to modernize the codebase by leveraging features from the latest C standard (C23) where `_Bool` and `bool` are built-in.

- **Use of preprocessor directives**: The patch introduces a compound preprocessor condition using `#if !defined(__cplusplus) && (!defined(__STDC_VERSION__) || __STDC_VERSION__ < 202311L)` to handle the conditional inclusion, ensuring compatibility across different environments and standards.

- **No other changes in include guards or library includes**: The patch does not affect any other parts of the file, maintaining the existing `#include` directives for `<stdint.h>` and `<time.h>`.

Issue #2481, opened 2023-05-09
[feat.]: Methods to delete the registered modules


chenzhuofu:
## Motivation

In WasmEdge C-SDK, there is no provisioned method to remove the modules registerd by VM.
 
I can understand the reason is that it's vm owns those modules. However it's still such a wasting behavior if u don't delete the registered module after using it.

Maybe we can create some methods like this:
```cpp
WASMEDGE_CAPI_EXPORT extern void
WasmEdge_VMDeleteRegisteredModule(const WasmEdge_VMContext *Cxt,
                               const WasmEdge_String ModuleName);
```

aetiurf:
And when we invoke this function, the VM could release its ownership and delete this named module.

hydai:
@q82419 
What do you think?

q82419:
As discussed before, the module instances may have dependencies: module A imports the functions of module B when instantiating.
Therefore, this function is not safe if delete the specific registered module, or it may cause core dump when execution.
As the steps in [module instantiation](https://webassembly.github.io/spec/core/exec/modules.html#instantiation), the [external value (the instance address in store)](https://webassembly.github.io/spec/core/exec/runtime.html#syntax-externval) should be able to be found. Therefore in WasmEdge we did the strict linking to reduce the instance searching time when executing.
I think this function is only available if we implement the dynamic WASM module instance linker.

You can use the `WasmEdge_VMCleanUp` to delete all registered modules without WASI and plug-ins.

aetiurf:
> module A imports the functions of module B when instantiating

Maybe I could call it "synchronization problem" ?

IMO, this problem should be considered by developers. That is, When a developer calls `WasmEdge_VMDeleteRegisteredModule`, he should ensure that the registered module isn't or will never become the dependency of other modules (from other threads) anymore.

q82419:
> > module A imports the functions of module B when instantiating
> 
> Maybe I could call it "synchronization problem" ?
> 
> IMO, this problem should be considered by developers. That is, When a developer calls `WasmEdge_VMDeleteRegisteredModule`, he should ensure that the registered module isn't or will never become the dependency of other modules (from other threads) anymore.
> 
> It's just like the ModuleInstance owned by developer: one can call `WasmEdge_ModuleInstanceDelete` whenver he want.

Yes, you're right, developers should ensure the life-cycle of the modules.
[...]

PR #4233, opened 2025-07-04
feat(c-api): add method to delete registered modules from VM

ShigrafS:
Closes: #2481 

Currently, the WasmEdge C API does not provide a way to delete or unregister a module that has been registered to a `WasmEdge_VMContext`. This can lead to inefficient memory/resource usage in long-running applications or when modules are dynamically managed.

This PR introduces a new API function:

```c
WASMEDGE_CAPI_EXPORT extern void
WasmEdge_VMDeleteRegisteredModule(const WasmEdge_VMContext *Cxt,
                                  const WasmEdge_String ModuleName);
```

This allows users to explicitly delete a named module registered in a VM, enabling better control over module lifecycle and resource management.

#### What’s Done

* **API Definition**
  Added the `WasmEdge_VMDeleteRegisteredModule` function declaration to the public C API header (`include/api/wasmedge/wasmedge.h`).

* **Core Implementation**
  Implemented the function in `lib/api/wasmedge.cpp` to:

  * Validate input parameters.
  * Retrieve the store context from the VM.
  * Locate and unregister the named module from the store.
  * Delete the module instance to release resources.

* **Store Manager Enhancements**
  Extended the `StoreManager` class (`include/runtime/storemgr.h`) with an `unregisterModule` method that safely removes a named module from internal storage with proper synchronization.

* **Unit Tests**
  Added new test cases in `test/api/APIVMCoreTest.cpp` to verify:

  * Successful registration and subsequent deletion of a module updates the registered module count correctly.
  * Attempting to delete a non-existent module does not affect the registered modules or cause errors.
 

#### Build Success & Test Coverage

* **Build System**
  The project was successfully built using **CMake + Ninja** on a Windows environment (`MSVC`). A clean build was performed to ensure no stale artifacts affected the result.

* **Test Execution**
  Focused test case `wasmedgeAPIVMCoreTests` (Test #14), which contains the new module deletion test cases, was run independently to validate the functionality.

  ```bash
  PS D:\Programs\wasmedge\WasmEdge\build> ctest -R wasmedgeAPIVMCoreTests --output-on-failure
  Test project D:/Programs/wasmedge/WasmEdge/build
      Start 14: wasmedgeAPIVMCoreTests
  1/1 Test #14: wasmedgeAPIVMCoreTests ...........   Passed  362.60 sec

  ```

* **Conclusion**
  The new API function behaves as expected:

  * Registered modules are correctly removed.
  * Non-existent module deletion does not cause crashes or memory faults.

  All assertions passed, confirming the correctness and stability of the feature under test.

### DIFF
--- a/include/api/wasmedge/wasmedge.h
+++ b/include/api/wasmedge/wasmedge.h
@@ -3788,7 +3788,7 @@ WASMEDGE_CAPI_EXPORT extern uint32_t WasmEdge_VMGetFunctionList(
     const WasmEdge_VMContext *Cxt, WasmEdge_String *Names,
     const WasmEdge_FunctionTypeContext **FuncTypes, const uint32_t Len);
 
-/// Unsafely delete a registered module from the VM context.
+/// Forcibly delete a registered module from the VM context.
 ///
 /// \warning This function does not check whether other modules depend on the
 /// target. Deleting a module that is still in use may cause undefined behavior
@@ -3800,8 +3800,8 @@ WASMEDGE_CAPI_EXPORT extern uint32_t WasmEdge_VMGetFunctionList(
 /// \param Cxt the WasmEdge_VMContext to delete the module from.
 /// \param ModuleName the name of the module to delete.
 WASMEDGE_CAPI_EXPORT extern void
-WasmEdge_VMUnsafeDeleteRegisteredModule(const WasmEdge_VMContext *Cxt,
-                                        const WasmEdge_String ModuleName);
+WasmEdge_VMForceDeleteRegisteredModule(const WasmEdge_VMContext *Cxt,
+                                       const WasmEdge_String ModuleName);
 
 /// Get the module instance corresponding to the WasmEdge_HostRegistration
 /// settings.
--- a/include/api/wasmedge/wasmedge.h
+++ b/include/api/wasmedge/wasmedge.h
@@ -3788,6 +3788,18 @@ WASMEDGE_CAPI_EXPORT extern uint32_t WasmEdge_VMGetFunctionList(
     const WasmEdge_VMContext *Cxt, WasmEdge_String *Names,
     const WasmEdge_FunctionTypeContext **FuncTypes, const uint32_t Len);
 
+/// Delete a registered module from the VM context.
+///
+/// \warning Developers must ensure the module is not depended by other modules.
+/// Deleting a module that is still in use may cause undefined behavior or
+/// segmentation faults.
+///
+/// \param Cxt the WasmEdge_VMContext to delete the module from.
+/// \param ModuleName the name of the module to delete.
+WASMEDGE_CAPI_EXPORT extern void
+WasmEdge_VMDeleteRegisteredModule(const WasmEdge_VMContext *Cxt,
+                                  const WasmEdge_String ModuleName);
+
 /// Get the module instance corresponding to the WasmEdge_HostRegistration
 /// settings.
 ///
--- a/include/api/wasmedge/wasmedge.h
+++ b/include/api/wasmedge/wasmedge.h
@@ -3788,17 +3788,20 @@ WASMEDGE_CAPI_EXPORT extern uint32_t WasmEdge_VMGetFunctionList(
     const WasmEdge_VMContext *Cxt, WasmEdge_String *Names,
     const WasmEdge_FunctionTypeContext **FuncTypes, const uint32_t Len);
 
-/// Delete a registered module from the VM context.
+/// Unsafely delete a registered module from the VM context.
 ///
-/// \warning Developers must ensure the module is not depended by other modules.
-/// Deleting a module that is still in use may cause undefined behavior or
-/// segmentation faults.
+/// \warning This function does not check whether other modules depend on the
+/// target. Deleting a module that is still in use may cause undefined behavior
+/// or crashes.
+///
+/// A safer deletion API may be introduced in the future once module dependency
+/// management is implemented.
 ///
 /// \param Cxt the WasmEdge_VMContext to delete the module from.
 /// \param ModuleName the name of the module to delete.
 WASMEDGE_CAPI_EXPORT extern void
-WasmEdge_VMDeleteRegisteredModule(const WasmEdge_VMContext *Cxt,
-                                  const WasmEdge_String ModuleName);
+WasmEdge_VMUnsafeDeleteRegisteredModule(const WasmEdge_VMContext *Cxt,
+                                        const WasmEdge_String ModuleName);
 
 /// Get the module instance corresponding to the WasmEdge_HostRegistration
 /// settings.
--- a/include/runtime/storemgr.h
+++ b/include/runtime/storemgr.h
@@ -98,6 +98,18 @@ public:
     return {};
   }
 
+  /// Unregister a named module from this store.
+  Expect<void> unregisterModule(std::string_view Name) {
+    std::unique_lock Lock(Mutex);
+    auto Iter = NamedMod.find(Name);
+    if (Iter == NamedMod.cend()) {
+      return Unexpect(ErrCode::Value::UnknownImport);
+    }
+    (const_cast<Instance::ModuleInstance *>(Iter->second))->unlinkStore(this);
+    NamedMod.erase(Iter);
+    return {};
+  }
+
   void addNamedModule(std::string_view Name,
                       const Instance::ModuleInstance *Inst) {
     std::unique_lock Lock(Mutex);
--- a/lib/api/wasmedge.cpp
+++ b/lib/api/wasmedge.cpp
@@ -3104,22 +3104,23 @@ WASMEDGE_CAPI_EXPORT void WasmEdge_VMCleanup(WasmEdge_VMContext *Cxt) {
   }
 }
 
-void WasmEdge_VMDeleteRegisteredModule(const WasmEdge_VMContext *Cxt,
-                                       const WasmEdge_String ModuleName) {
+void WasmEdge_VMUnsafeDeleteRegisteredModule(const WasmEdge_VMContext *Cxt,
+                                             const WasmEdge_String ModuleName) {
   if (!Cxt || !ModuleName.Buf) {
     return; // Invalid input
   }
+
   // Cast away const to match WasmEdge_VMGetStoreContext signature
   WasmEdge_StoreContext *StoreCxt =
       WasmEdge_VMGetStoreContext(const_cast<WasmEdge_VMContext *>(Cxt));
   if (!StoreCxt) {
     return; // Invalid store context
   }
+
   const WasmEdge_ModuleInstanceContext *ModInst =
       WasmEdge_StoreFindModule(StoreCxt, ModuleName);
   if (ModInst) {
     fromStoreCxt(StoreCxt)->unregisterModule(genStrView(ModuleName));
-    // Cast away const for WasmEdge_ModuleInstanceDelete
     WasmEdge_ModuleInstanceDelete(
         const_cast<WasmEdge_ModuleInstanceContext *>(ModInst));
   }
--- a/lib/api/wasmedge.cpp
+++ b/lib/api/wasmedge.cpp
@@ -3104,8 +3104,8 @@ WASMEDGE_CAPI_EXPORT void WasmEdge_VMCleanup(WasmEdge_VMContext *Cxt) {
   }
 }
 
-void WasmEdge_VMUnsafeDeleteRegisteredModule(const WasmEdge_VMContext *Cxt,
-                                             const WasmEdge_String ModuleName) {
+void WasmEdge_VMForceDeleteRegisteredModule(const WasmEdge_VMContext *Cxt,
+                                            const WasmEdge_String ModuleName) {
   if (!Cxt || !ModuleName.Buf) {
     return; // Invalid input
   }
--- a/lib/api/wasmedge.cpp
+++ b/lib/api/wasmedge.cpp
@@ -3104,6 +3104,27 @@ WASMEDGE_CAPI_EXPORT void WasmEdge_VMCleanup(WasmEdge_VMContext *Cxt) {
   }
 }
 
+void WasmEdge_VMDeleteRegisteredModule(const WasmEdge_VMContext *Cxt,
+                                       const WasmEdge_String ModuleName) {
+  if (!Cxt || !ModuleName.Buf) {
+    return; // Invalid input
+  }
+  // Cast away const to match WasmEdge_VMGetStoreContext signature
+  WasmEdge_StoreContext *StoreCxt =
+      WasmEdge_VMGetStoreContext(const_cast<WasmEdge_VMContext *>(Cxt));
+  if (!StoreCxt) {
+    return; // Invalid store context
+  }
+  const WasmEdge_ModuleInstanceContext *ModInst =
+      WasmEdge_StoreFindModule(StoreCxt, ModuleName);
+  if (ModInst) {
+    fromStoreCxt(StoreCxt)->unregisterModule(genStrView(ModuleName));
+    // Cast away const for WasmEdge_ModuleInstanceDelete
+    WasmEdge_ModuleInstanceDelete(
+        const_cast<WasmEdge_ModuleInstanceContext *>(ModInst));
+  }
+}
+
 WASMEDGE_CAPI_EXPORT uint32_t
 WasmEdge_VMGetFunctionListLength(const WasmEdge_VMContext *Cxt) {
   if (Cxt) {
--- a/test/api/APIVMCoreTest.cpp
+++ b/test/api/APIVMCoreTest.cpp
@@ -225,6 +225,43 @@ TEST(AsyncExecute, InterruptTest) {
   WasmEdge_VMDelete(VM);
 }
 
+TEST(WasmEdgeVM, DeleteRegisteredModule) {
+  WasmEdge_ConfigureContext *Conf = WasmEdge_ConfigureCreate();
+  WasmEdge_VMContext *VMCxt = WasmEdge_VMCreate(Conf, nullptr);
+
+  uint32_t originalCount = WasmEdge_VMListRegisteredModuleLength(VMCxt);
+
+  WasmEdge_String ModuleName = WasmEdge_StringCreateByCString("test_module");
+  WasmEdge_ModuleInstanceContext *ModInst =
+      WasmEdge_ModuleInstanceCreate(ModuleName);
+
+  WasmEdge_Result Res = WasmEdge_VMRegisterModuleFromImport(VMCxt, ModInst);
+  EXPECT_TRUE(WasmEdge_ResultOK(Res));
+  EXPECT_EQ(WasmEdge_VMListRegisteredModuleLength(VMCxt), originalCount + 1);
+
+  WasmEdge_VMDeleteRegisteredModule(VMCxt, ModuleName);
+  EXPECT_EQ(WasmEdge_VMListRegisteredModuleLength(VMCxt), originalCount);
+
+  // Cleanup
+  WasmEdge_StringDelete(ModuleName);
+  WasmEdge_VMDelete(VMCxt);
+  WasmEdge_ConfigureDelete(Conf);
+}
+
+TEST(WasmEdgeVM, DeleteNonExistentModule) {
+  WasmEdge_ConfigureContext *Conf = WasmEdge_ConfigureCreate();
+  WasmEdge_VMContext *VMCxt = WasmEdge_VMCreate(Conf, nullptr);
+  WasmEdge_String ModuleName =
+      WasmEdge_StringCreateByCString("nonexistent_module");
+  // Delete a module that doesn't exist
+  WasmEdge_VMDeleteRegisteredModule(VMCxt, ModuleName);
+  EXPECT_EQ(WasmEdge_VMListRegisteredModuleLength(VMCxt), 12); // Still 12
+  // Cleanup
+  WasmEdge_StringDelete(ModuleName);
+  WasmEdge_VMDelete(VMCxt);
+  WasmEdge_ConfigureDelete(Conf);
+}
+
 } // namespace
 
 GTEST_API_ int main(int argc, char **argv) {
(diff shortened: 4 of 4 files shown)


juntao:

------

## [include/api/wasmedge/wasmedge.h](https://github.com/WasmEdge/WasmEdge/blob/a1b46170da0f04843c99fd5b531481f3aef2fd05/include%2Fapi%2Fwasmedge%2Fwasmedge.h)

#### Potential issues

1. **Missing Include Guards**: The file does not include standard include guards to prevent multiple inclusions, which can lead to compilation errors if the header is included more than once in a single translation unit.

2. **Inconsistent Use of `extern` Keyword**: Some functions are declared with `extern`, while others are not. This inconsistency could lead to unexpected behavior, especially when linking. For example:
   ```cpp
   WASMEDGE_CAPI_EXPORT extern const char *WasmEdge_VersionGet(void);
   ```
   should be consistent with:
   ```cpp
   WASMEDGE_CAPI_EXPORT const char *WasmEdge_VersionGet(void);
   ```

3. **Potential Memory Leak**: The `WasmEdge_BytesCreate` function allocates memory for the `WasmEdge_Bytes` structure, but there is no corresponding free function to release this memory. This could lead to memory leaks if users of the API fail to properly manage the allocated resources.

#### Summary of changes

Here are the key changes summarized in bullet points:

- **Added a new function:** `WasmEdge_VMForceDeleteRegisteredModule` to delete registered modules from the VM context.
- **Function Signature:**
  ```cpp
  WASMEDGE_CAPI_EXPORT extern void WasmEdge_VMForceDeleteRegisteredModule(const WasmEdge_VMContext *Cxt, const WasmEdge_String ModuleName);
  ```
- **Warning Note:** The function does not check for module dependencies and may cause undefined behavior or crashes if used improperly.

## [include/runtime/storemgr.h](https://github.com/WasmEdge/WasmEdge/blob/a1b46170da0f04843c99fd5b531481f3aef2fd05/include%2Fruntime%2Fstoremgr.h)

#### Potential issues

1. **Incorrect Lock Type in `reset()`**: The `reset()` method uses a shared lock (`std::shared_lock`) instead of an exclusive lock (`std::unique_lock`), which can lead to race conditions when modifying the `NamedMod` map.

2. **Potential Null Pointer Dereference in `unregisterModule()`**: If `Iter->second` is null, calling `unlinkStore(this)` will result in a segmentation fault; there should be a check for this.

3. **Use of `const_cast` in `reset()` and `registerModule()`**: The use of `const_cast` to modify `Pair.second` and `ModInst` can lead to undefined behavior if the original object was not actually mutable, despite being declared as such; consider safer alternatives like using non-const references.

#### Summary of changes

### Key Changes

1. **Addition of `unregisterModule` Method:**
   - Introduced a new method, `Expect<void> unregisterModule(std::string_view Name)`, which allows for the removal of a named module from the store.
2. **Unlinking Store in Unregistration Process:**
   - The `unregisterModule` method unlinks the store from the module instance before removing it from the `NamedMod` map to ensure proper cleanup.
3. **Error Handling on Unknown Module:**
   - Implemented error handling to return an `Unexpect(ErrCode::Value::UnknownImport)` if the specified module name is not found in the store.

## [lib/api/wasmedge.cpp](https://github.com/WasmEdge/WasmEdge/blob/a1b46170da0f04843c99fd5b531481f3aef2fd05/lib%2Fapi%2Fwasmedge.cpp)

#### Potential issues

### Major Bugs or Issues:

1. **Uninitialized Variables**: In the function `WasmEdge_LogSetCallback`, the `Message.LoggerName` and `Message.Time` fields are not properly initialized before they might be used.
2. **Memory Leak in String Handling**: The function `WasmEdge_StringCreateByBuffer` dynamically allocates memory for a string but does not provide a corresponding delete method, potentially leading to memory leaks when the allocated buffer is no longer needed.
3. **Incorrect Type Conversion**: In `genValType`, the conversion from `WasmEdge_ValType` to `ValType` may result in incorrect values due to possible misalignment or truncation of data.

These issues should be addressed for correct functionality and memory management.

#### Summary of changes

**Top 3 most important changes:**

1. **New API Function**: Added a new function `WasmEdge_VMForceDeleteRegisteredModule` to delete registered modules from the VM by name.
2. **Input Validation**: The function checks for invalid inputs and returns early if either the VM context or module name is null.
3. **Module Deletion**: If the module exists, it unregisters the module from the store context and deletes the module instance.

## [test/api/APIVMCoreTest.cpp](https://github.com/WasmEdge/WasmEdge/blob/a1b46170da0f04843c99fd5b531481f3aef2fd05/test%2Fapi%2FAPIVMCoreTest.cpp)

#### Potential issues

1. **Memory Leak Potential**: The `WasmEdge_StringCreateByCString` function is used to create string objects, but there is no corresponding `WasmEdge_StringDelete` call for some of these objects, which could lead to memory leaks.
2. **Null Pointer Dereference Risk**: In the `TEST(WasmEdgeVM, ForceDeleteInvalidInput)` test, the code attempts to delete a registered module with an empty name and a null VM context without checking if the inputs are valid first, which could lead to undefined behavior or crashes.
3. **Inconsistent Error Handling**: The code does not consistently handle errors returned by WasmEdge functions, potentially leading to silent failures or unexpected behavior in some cases.

These issues should be addressed to ensure the code is robust and memory-safe.

#### Summary of changes

Here are the key changes made in the patch:

1. **Added a new method `WasmEdge_VMForceDeleteRegisteredModule`**: This function is used to delete a registered module from the VM, even if it is still being used.

2. **Added three test cases for the new feature**:
   - `TEST(WasmEdgeVM, ForceDeleteRegisteredModule)`: Tests the successful deletion of an existing registered module.
   - `TEST(WasmEdgeVM, ForceDeleteNonExistentModule)`: Tests handling of deletion requests for non-existent modules.
   - `TEST(WasmEdgeVM, ForceDeleteInvalidInput)`: Tests handling of invalid inputs like null VM context or empty module name.

These changes primarily involve adding a new functionality to force delete registered modules and ensuring it is robust against edge cases.

include/api/wasmedge/wasmedge.h
```diff
@@ -3788,6 +3788,17 @@ WASMEDGE_CAPI_EXPORT extern uint32_t WasmEdge_VMGetFunctionList(
     const WasmEdge_VMContext *Cxt, WasmEdge_String *Names,
     const WasmEdge_FunctionTypeContext **FuncTypes, const uint32_t Len);
 
+/// Delete a registered module from the VM context.
+///
+/// \warning Developers must ensure the module is not a dependency of other modules.
```
dannypsnl:
```suggestion
/// \warning Developers must ensure the module is not depended by other modules.
```

include/runtime/storemgr.h
```diff
@@ -98,6 +98,18 @@ class StoreManager {
...
   }
 
+  /// Unregister a named module from this store.
+  Expect<void> unregisterModule(std::string_view Name) {
+    std::unique_lock Lock(Mutex);
+    auto Iter = NamedMod.find(Name);
+    if (Iter == NamedMod.cend()) {
+      return Unexpect(ErrCode::Value::UnknownImport);
+    }
+    (const_cast<Instance::ModuleInstance *>(Iter->second))->unlinkStore(this);
```
dannypsnl, line 108: Is this casting required?

ShigrafS: Hi @dannypsnl ,

Thank you for the review and for pointing out the `const_cast` in `unregisterModule`. The `const_cast` is currently required because `NamedMod` stores `const Instance::ModuleInstance *` pointers to ensure const-correctness during map operations, but `unlinkStore` is a non-const method that modifies the `ModuleInstance` (e.g., clearing its store reference).

I checked the declaration of `unlinkStore` in `module.h`, and it’s defined as `void unlinkStore(StoreManager *Store);`, which requires a non-const `ModuleInstance`. Without `const_cast`, the compiler would reject the call due to the const qualifier on `Iter->second`.

While the `const_cast` is safe here (since `NamedMod` owns the pointers and `unlinkStore` is designed to modify the instance), I agree it’s worth exploring alternatives to improve const-correctness. Possible options include:
1. Making `unlinkStore` a `const` method if it only modifies mutable fields or doesn’t affect the logical state of `ModuleInstance`.
2. Changing `NamedMod` to store non-const `Instance::ModuleInstance *` pointers, if safe across the codebase.

For this PR, I propose keeping the `const_cast` to maintain compatibility with the current `unlinkStore` implementation. I can investigate making `unlinkStore` `const` or adjusting `NamedMod` in a follow-up PR, if you think that’s a good approach. 

Please let me know your opinion and if any change is required.

ShigrafS: Hi @dannypsnl ,

Thank you for your review. 

Thanks for catching the `const_cast`. It’s needed because `NamedMod` stores `const` pointers, but `unlinkStore` modifies the module, so we have to cast away const.

It’s safe here, but we could improve by making `unlinkStore` const or changing `NamedMod` later. For now, I’d propose to keep the cast to avoid breaking things.

Please let me know you opinion and let me know if any change is needed.

dannypsnl (approved):

ShigrafS:
I've changes the commit message to align with `commitlint` and have rebased the branch.

hydai (changes requested):
Please squash your commits into one.

dannypsnl (dismissed):

include/api/wasmedge/wasmedge.h
```diff
@@ -3788,6 +3788,18 @@ WASMEDGE_CAPI_EXPORT extern uint32_t WasmEdge_VMGetFunctionList(
...
+///
+/// \warning Developers must ensure the module is not depended by other modules.
+/// Deleting a module that is still in use may cause undefined behavior or
+/// segmentation faults.
+///
+/// \param Cxt the WasmEdge_VMContext to delete the module from.
+/// \param ModuleName the name of the module to delete.
+WASMEDGE_CAPI_EXPORT extern void
+WasmEdge_VMDeleteRegisteredModule(const WasmEdge_VMContext *Cxt,
+                                  const WasmEdge_String ModuleName);
```
hydai: I am thinking of extending this function to:

```suggestion
WasmEdge_VMDeleteRegisteredModule(const WasmEdge_VMContext *Cxt,
                                  const WasmEdge_String ModuleName, bool ForceDelete);
```

The reason is that we would like to handle the following situations. Assuming we have two modules, A and B (where B depends on A):

1. Users delete B and try to execute A -> Fine.
2. Users delete A and try to execute B -> Failed because A no longer exists.
3. Users delete both A and B -> Happy users, happy life.

Since this PR is intended to provide a way to remove an existing module and is *not* meant to handle the dependency tree for removing related modules, I would like to have an option to control the behavior if we wish to support this in the future.

When `ForceDelete` is set to `true`, we will delete the module regardless of any relationships between the target and its upstreams or downstreams. Otherwise, an error should be shown to users indicating that the module is still in use and cannot be deleted—similar to package managers. If users still want to delete the module, they should delete B first, then delete A.

At this stage, I will say that if `ForceDelete` is set to `false`, we should warn users that this is not yet implemented and then do nothing. Otherwise, simply execute the same logic you used here.

cc @ibmibmibm @q82419 @dannypsnl  WDYT?

hydai: @ShigrafS 
Does the above part make sense to you? Do you have any ideas about this?

ShigrafS: Hi @hydai ,

Thank you for the detailed feedback and the thoughtful suggestion to extend `WasmEdge_VMDeleteRegisteredModule` with a `ForceDelete` parameter. I'll update the function signature and update the code accordingly.

dannypsnl: I think the function document already warn users the result, I don't think we need the flag.

Also, if we add the flag, this PR needs to check if there are some modules depends on itself, then it cannot be removed.

hydai: Hi @dannypsnl,

It's all about the compatibility issue. If we don't care about it and don't want the extension—I mean the feature to clean up the related modules—that's totally fine. However, if we do want this feature, I would prefer to have a fixed function signature to avoid modifying it in the near future. This is a new function for the C API, and adding the current function signature then changing it later would be a breaking change.

Still, I am unsure whether we truly want it or if we just want users to manage module dependencies themselves, since C allows users to control raw pointers and assume responsibility on their own.

I am open to the decision. If you are concerned that adding the new option with a `TODO` or `NOP` behavior is a significant issue, another proposal is to keep this simple signature here and add a new function in the future, which may be very similar to this one, to handle the dependency logic.

Does this sound good to you?

Also, I would like to ask @ibmibmibm and @q82419 for their opinions on this. Since it is a new function added to the C API header, we should take it seriously before merging.

hydai: > Thank you for the detailed feedback and the thoughtful suggestion to extend `WasmEdge_VMDeleteRegisteredModule` with a `ForceDelete` parameter. I'll update the function signature and update the code accordingly.

Hi @ShigrafS 
Could you please wait a moment until we decide how to proceed with this new function? I just don't want you to modify the code and then roll it back once we have a second thought. ;-)

dannypsnl: @hydai I prefers to provide two APIs

1. `UnsafeDelete`
2. `Delete`

instead of using flag.

q82419: > I think the function document already warn users the result, I don't think we need the flag.
> 
> Also, if we add the flag, this PR needs to check if there are some modules depends on itself, then it cannot be removed.

In my opinion, the module dependency checking should be applied one day in the future.
It's like the GC mechanism in module dependencies. The `delete` function will be the API to notice the WasmEdge internal to release the module once it is not be referred anymore, and the `Force` parameter is to forcibly release the module instance.
Therefore I suppose the API with the `Force` parameter, even though it's not used currently. This is to prrevent from changing API in the future, which makes troubles.

hydai: What if we use `UnsafeDelete` for now and add a new `Delete` function with the `Force` parameter in the future?

The reason I have this idea is that we don't have such a mechanism for the modules' dependency checks, and there is no timeline to enable it at this stage. Adding a new option with no immediate effect in the near future seems odd.

Additionally, we can redirect the `UnsafeDelete(...)` function to `Delete(..., Force=true)` once the mechanism is implemented.

However, this wouldn't make sense if we decide to have the `Delete(Force)` signature in the end; it would result in a temporary `UnsafeDelete` function for a while.

dannypsnl: Sounds good

ibmibmibm: LGTM

hydai: Sounds like we have a consensus. Does this solution work for you, @q82419?

ShigrafS: Hi! 
Just popping in to check whether we’ve reached a final decision on this. Based on the latest discussion, it sounds like we’re leaning toward going with `UnsafeDelete` for now and potentially introducing `Delete(..., Force)` later when the dependency mechanism is more defined.

Just wanted to confirm if that’s the direction we’re locking in, or if there’s anything else that still needs to be clarified before moving forward. 
Please let me know!
Thanks!

q82419: > Sounds like we have a consensus. Does this solution work for you, @q82419?

I think it can be the temporary solution for the `unsafeDelete` function right now.
Once the module dependency mechanism being implemented we can discuss a new `Delete` function with the `force` flag and mark this unsafe function as deprecated.

hydai: Yup. We have reached the final decision. Let's proceed with `UnsafeDelete` for now.

ShigrafS:
@hydai @dannypsnl 

```
 Start 14: wasmedgeAPIVMCoreTests
14/29 Test #14: wasmedgeAPIVMCoreTests ...........   Passed  381.56 sec

100% tests passed, 0 tests failed out of 29
```

Build successfull. 
Test passing.

ShigrafS:
@hydai Clang-format has been applied.

lib/api/wasmedge.cpp
```diff
@@ -3104,6 +3104,28 @@ WASMEDGE_CAPI_EXPORT void WasmEdge_VMCleanup(WasmEdge_VMContext *Cxt) {
...
+  if (!Cxt || !ModuleName.Buf) {
+    return; // Invalid input
+  }
+
+  // Cast away const to match WasmEdge_VMGetStoreContext signature
+  WasmEdge_StoreContext *StoreCxt =
+      WasmEdge_VMGetStoreContext(const_cast<WasmEdge_VMContext *>(Cxt));
+  if (!StoreCxt) {
+    return; // Invalid store context
+  }
```
hydai, line 3118: This condition is not covered by your tests. Please add a new one for it.

include/runtime/storemgr.h
```diff
@@ -98,6 +98,18 @@ class StoreManager {
     return {};
   }
 
+  /// Unregister a named module from this store.
+  Expect<void> unregisterModule(std::string_view Name) {
+    std::unique_lock Lock(Mutex);
+    auto Iter = NamedMod.find(Name);
+    if (Iter == NamedMod.cend()) {
+      return Unexpect(ErrCode::Value::UnknownImport);
+    }
```
hydai, line 107: Ditto.

ShigrafS:
```
Test project D:/Programs/build/ShigrafS/WasmEdge/build
    Start 14: wasmedgeAPIVMCoreTests
1/1 Test #14: wasmedgeAPIVMCoreTests ...........   Passed  540.18 sec

Total Test time (real) = 540.26 sec
```

Tests passed.

lib/api/wasmedge.cpp
ShigrafS: Test added.

test/api/APIVMCoreTest.cpp
```diff
@@ -225,6 +225,84 @@ TEST(AsyncExecute, InterruptTest) {
...
+      WasmEdge_ModuleInstanceCreate(ModuleName);
+
+  WasmEdge_Result Res = WasmEdge_VMRegisterModuleFromImport(VMCxt, ModInst);
+  EXPECT_TRUE(WasmEdge_ResultOK(Res));
+  EXPECT_EQ(WasmEdge_VMListRegisteredModuleLength(VMCxt), originalCount + 1);
+
+  // Unsafe delete
+  WasmEdge_VMUnsafeDeleteRegisteredModule(VMCxt, ModuleName);
+  EXPECT_EQ(WasmEdge_VMListRegisteredModuleLength(VMCxt), originalCount);
+
```
PhantomInTheWire, line 245:
```suggestion
  // Verify module is no longer accessible
  WasmEdge_StoreContext *StoreCxt = WasmEdge_VMGetStoreContext(VMCxt);
  const WasmEdge_ModuleInstanceContext *FindResult = WasmEdge_StoreFindModule(StoreCxt, ModuleName);
  EXPECT_EQ(FindResult, nullptr);
```

we should check if this is truly inaccessible after deletion

ShigrafS: Tests added.

ShigrafS:
```
Test project D:/Programs/build/ShigrafS/WasmEdge/build
    Start 14: wasmedgeAPIVMCoreTests
1/1 Test #14: wasmedgeAPIVMCoreTests ...........   Passed  450.47 sec

Total Test time (real) = 450.92 sec
```
Test passed.

hydai:
Hi @ShigrafS

I need more time to consider the naming, and I am still unsure if "Unsafe" is the most accurate word in this context.  
Everywhere else, we use "Unsafe" for situations where there is no mutex protection and the code isn't thread-safe.  
However, in this case, it refers to "skip dependency checks for the deleted module." This leads to a completely different meaning between the two use cases and could confuse users.  
Aside from that, I have no other issues with this PR.

hydai:
Hi @ShigrafS 
Let's make things easier. Since we will have `Delete(... Force)` in the future, just replace `Unsafe` with `Force` directly.

ShigrafS:
Hi @hydai 
Thank you for the clarification. 
I'll be proceeding with `Force`.

hydai (approved):

ShigrafS:
@hydai 
Should I rebase the branch now?

hydai:
@ShigrafS  
Yes, could you please check why the API unit tests failed?

ShigrafS:
@hydai 
```
Total Test time (real) = 1569.29 sec

The following tests FAILED:
	 14 - wasmedgeAPIVMCoreTests (Timeout)
Errors while running CTest
Output from these tests are in: /__w/WasmEdge/WasmEdge/build/Testing/Temporary/LastTest.log
Use "--rerun-failed --output-on-failure" to re-run the failed cases verbosely.
Error: Process completed with exit code 8.
``` 

It says `Timeout`. 
It probably timed out because it exceeded the allowed time. 

Although, I'm not sure why it failed on MacOS. It doesn't provide any log.

ShigrafS:
@hydai 
Ubuntu Tests are still failing due to Timeout. 
```
97% tests passed, 1 tests failed out of 29

Total Test time (real) = 1567.86 sec

The following tests FAILED:
Errors while running CTest
	 14 - wasmedgeAPIVMCoreTests (Timeout)
Output from these tests are in: /__w/WasmEdge/WasmEdge/build/Testing/Temporary/LastTest.log
Use "--rerun-failed --output-on-failure" to re-run the failed cases verbosely.
Error: Process completed with exit code 8.
```

However, the MacOS tests are probably failing due to a different reason: 
```
97% tests passed, 1 tests failed out of 29

Total Test time (real) = 294.90 sec

The following tests FAILED:
Errors while running CTest
Output from these tests are in: /Users/runner/work/WasmEdge/WasmEdge/build/Testing/Temporary/LastTest.log
Use "--rerun-failed --output-on-failure" to re-run the failed cases verbosely.
	 14 - wasmedgeAPIVMCoreTests (Failed)
Error: Process completed with exit code 8.
```

How should we proceed with this?

ShigrafS:
@hydai 
The detailed CI tests for MacOS: 
https://github.com/ShigrafS/WasmEdge/actions/runs/16383044978

ShigrafS:
@hydai 
The Tests should pass now. 

Tested on a separate branch: https://github.com/ShigrafS/WasmEdge/actions/runs/16387504954

hydai (approved):
Thanks.

PR #4235, opened 2025-07-04
docker(ubuntu): simplify image names for plugins

0yi0:
Update `"tags-simplified"` so that images for both base and plugins have a default one.

Before this PR:
- `ubuntu-24.04-build-clang` -> `ubuntu-24.04` 

After this PR:
- `ubuntu-24.04-build-clang` -> `ubuntu-24.04` 
- `ubuntu-20.04-build-clang-plugins-deps` -> `ubuntu-20.04-plugins`

### DIFF
--- a/utils/docker/docker-bake.ubuntu.hcl
+++ b/utils/docker/docker-bake.ubuntu.hcl
@@ -76,7 +76,11 @@ function "tags-backports" {
 
 function "tags-simplified" {
   params = [target, ubuntu, toolchain]
-  result = target == "base" && toolchain == "clang" ? "ubuntu-${ubuntu}" : ""
+  result = toolchain == "clang" ? join("-", compact([
+    "ubuntu",
+    ubuntu,
+    target == "plugins" ? "plugins" : "",
+  ])) : ""
 }
 
 function "tags" {


juntao:

------

## [utils/docker/docker-bake.ubuntu.hcl](https://github.com/WasmEdge/WasmEdge/blob/a4343fc210af7b1907dccffe9b90fac787b16613/utils%2Fdocker%2Fdocker-bake.ubuntu.hcl)

#### Potential issues

1. **Incorrect Ubuntu Version Handling in "tags-simplified" Function**: The function `tags-simplified` appends the full ubuntu version twice, which is redundant and incorrect.

2. **Hardcoded Base Image Tag for CUDA Target**: The `cuda` target hardcodes the base image tag (`wasmedge/wasmedge:ubuntu-20.04-build-gcc`), which could lead to issues if this tag changes or is not available.

3. **Redundant Parameter in "tags-latest" Function**: The `tags-latest` function uses a redundant parameter `toolchain`, as it's only relevant for the latest base image with clang.

#### Summary of changes

1. **Simplified Image Naming for Plugins**:
   - The function `tags-simplified` now checks if the target is "plugins" and includes it in the image name only when relevant.

2. **Consolidated String Construction**:
   - The code uses a more concise approach to construct the result string, leveraging the `join` function with an array for clarity.

3. **Conditional Logic Adjustments**:
   - The conditional logic now explicitly handles the case where the target is "plugins", streamlining the naming convention.

These changes aim to simplify and clarify the image-naming process, especially for plugin-related targets in a Docker environment running on Ubuntu.

hydai (approved):

PR #4251, opened 2025-07-07
feat: support WasmEdge on the s390x platform

vishruth-thimmaiah:
This PR adds support for using WasmEdge on s390x hardware.


juntao:

------

## [.github/workflows/build_for_s390x.yml](https://github.com/WasmEdge/WasmEdge/blob/477c3a478de77217af32947a3f7603a007f7bfb4/.github%2Fworkflows%2Fbuild_for_s390x.yml)

#### Potential issues

-1. The `wasmedge` and `wabt` commands are not defined in the script.
- 2. The `make install` command does not ensure that WasmEdge is installed correctly or that it is accessible from the PATH.

#### Summary of changes

-1. Added a workflow for building and testing WasmEdge on s390x architecture.
- Specified the jobs and their steps, including checking out code and building with Docker.
- Used `curl` to install lld versions less than 18 as they cause segfaults.
- Installed necessary packages using `apt-get`, including LLVM for compiling Wasm.
- Included example Wasm files for testing.

## [cmake/WASINNDeps.cmake](https://github.com/WasmEdge/WasmEdge/blob/477c3a478de77217af32947a3f7603a007f7bfb4/cmake%2FWASINNDeps.cmake)

#### Potential issues

1. In `wasmedge_setup_tflite_lib`, there is no library named `libtensorflowlite_c.so`.
2. In `wasmedge_setup_bitnet_target`, the variable `ggml_ACCELERATE` is not defined.
3. In `wasmedge_setup_simdjson`, the variable `simdjson::simdjson` is not defined.

#### Summary of changes

-**Changes:**
  1. **Support GGML_BLAS on s390x platform**: Enable the WasmEdge backend with GGML BLAS support if the WASI-NN configuration option `WASMEDGE_PLUGIN_WASI_NN_GGML_LLAMA_BLAS` is set.
  2. **Disable GGML_BLAS by default**: If the option is not set, disable GGML_BLAS support.

**Most important changes first:**

1. **Enable GGML BLAS support for s390x platform**: This change allows the WasmEdge backend to use GGML BLAS on systems where it supports it.
2. **Disable GGML BLAS by default**: This ensures that the backend is not enabled by default, unless explicitly configured to do so.

These changes are crucial for improving performance and compatibility of the WASMEDge LLAMA backend on s390x platforms.

## [include/api/wasmedge/int128.h](https://github.com/WasmEdge/WasmEdge/blob/477c3a478de77217af32947a3f7603a007f7bfb4/include%2Fapi%2Fwasmedge%2Fint128.h)

#### Potential issues

-The source code uses `__int128` types for both unsigned and signed 128-bit integers, which are not standard C99 types and may not be supported on all platforms.

#### Summary of changes

-**Key changes:**
  - Added support for the s390x platform.
  - Introduced `uint128_t` and `int128_t` types specific to the s390x architecture.
  - Updated the conditional compilation check for architecture-specific code.

## [include/ast/instruction.h](https://github.com/WasmEdge/WasmEdge/blob/477c3a478de77217af32947a3f7603a007f7bfb4/include%2Fast%2Finstruction.h)

#### Potential issues

-1. The class lacks proper error handling for memory allocations and resource deallocations.
- 2. There is no meaningful validation or sanitization of input parameters to prevent buffer overflows or undefined behavior.
- 3. The `Flags` struct is not used for any meaningful purpose and could be removed for better readability and maintainability.

#### Summary of changes

- Added support for the S390x architecture.

## [include/common/endian.h](https://github.com/WasmEdge/WasmEdge/blob/477c3a478de77217af32947a3f7603a007f7bfb4/include%2Fcommon%2Fendian.h)

#### Potential issues

-1. Missing header inclusion for `endian.h`.
- 2. Incorrectly defined `WASMEDGE_ENDIAN_LITTLE_BYTE` on Windows.
- 3. Missing check for little-endian architecture in macOS and Linux.

#### Summary of changes

-**Key Changes:**
1. **Added Endian Enum**: Introduced an `Endian` enum for handling endianness, supporting little, big, and native.
2. **Byteswap Function**: Added a template function `byteswap` that handles byte swapping for integral types.
   - Uses conditional compilation with GCC, Intel Compiler, or Clang.
   - Supports little and big-endian integers up to 16 bytes.
3. **Namespace and Include**: Moved the endianness logic into a namespace and included necessary headers (`defines.h`, `int128.h`).

**Most Important Changes:**
- **Endianness Handling**: Implemented a consistent way to handle byte swapping across different platforms.
- **Byteswap Function**: Provided a reusable function for byte swapping of integral types.
- **Namespace Placement**: Moved the endianness logic into a namespace (`WasmEdge`) for better organization.

## [include/common/int128.h](https://github.com/WasmEdge/WasmEdge/blob/477c3a478de77217af32947a3f7603a007f7bfb4/include%2Fcommon%2Fint128.h)

#### Potential issues

N/A

#### Summary of changes

- The patch modifies the WasmEdge library to support running on the s390x platform, adding support for big-endian little endian byte-swapping.

## [include/common/types.h](https://github.com/WasmEdge/WasmEdge/blob/477c3a478de77217af32947a3f7603a007f7bfb4/include%2Fcommon%2Ftypes.h)

#### Potential issues

-1. The `ValType` class definition should be made more efficient by reducing redundancy and improving readability.
2. The `RefVariant` class definition should be simplified to handle references more efficiently.
3. The `ValComp` class definition should be refactored to eliminate redundant data and improve memory usage.

#### Summary of changes

- Added `Endian` enum to the codebase.
- Improved `retrieveExternRef` function to support endian conversion for reference types.
- Enhanced `EndianValue` class with methods to get endianness-little value and raw value.

## [include/executor/engine/atomic.ipp](https://github.com/WasmEdge/WasmEdge/blob/477c3a478de77217af32947a3f7603a007f7bfb4/include%2Fexecutor%2Fengine%2Fatomic.ipp)

#### Potential issues

-1. Missing include directives: `<experimental/scope.hpp>`.
- 2. In `runAtomicWaitOp`, missing error handling for memory out of bounds.
- 3. In `runAtomicLoadOp`, missing error handling for memory out of bounds.

#### Summary of changes

-**Key Changes in Summary:**

1. **Include Header Files:** The patch includes headers for handling endianness and types.
2. **Atomic Load Store Operations:** Added functions to handle atomic load and store operations.
3. **Atomic Arithmetical Operations:** Defined a template function `runAtomicOp` for executing atomic arithmetic operations.
4. **Atomic Compare-Exchange Operation:** Implemented the atomic compare-exchange operation using `std::atomic<std::atomic<T> *>`.
5. **Atomic Wait Function:** Added a function to wait on an atomic variable with an optional timeout.

The most important changes are:
- Including endianness handling and type definitions.
- Adding functions for atomic operations such as load, store, arithmetic, and compare-exchange.
- Defining a template function `runAtomicOp` for better code reusability.

These changes enhance the capabilities of the executor to handle operations on atomic objects efficiently.

## [include/executor/engine/binary_numeric_vector.ipp](https://github.com/WasmEdge/WasmEdge/blob/477c3a478de77217af32947a3f7603a007f7bfb4/include%2Fexecutor%2Fengine%2Fbinary_numeric_vector.ipp)

#### Potential issues

-There are no major coding issues found in the provided source code.

#### Summary of changes

- **Key Changes:**

1. **Byte Order Handling in Replace Lane Operation:**
   - The code now handles byte order (little-endian) for `TOut` types differently.
     - If `Endian::native == Endian::little`, the code uses the original byte order.
     - Otherwise, it swaps the bytes.

2. **Vector Narrowing Operation on Little-Endian Platforms:**
   - For little-endian platforms (`sizeof(TOut) == 1`), it ensures that the vector values are placed correctly in `Val1`.
   - For little-endian platforms (`sizeof(TOut) == 2`), it handles the vector narrowing correctly.
     - The original code was incorrect; it only copied the first two elements to `Val1`.
     - The updated code uses the correct byte order and ensures all elements are placed in `Val1`.

3. **Vector Narrowing Operation on Big-Endian Platforms:**
   - For big-endian platforms, it reverses the process.
   - If `sizeof(TOut) == 1`, it swaps the bytes before placing them in `Val1`.
   - If `sizeof(TOut) == 2`, it handles the vector narrowing correctly and places all elements in `Val1`.

4. **Byte Order Handling in Vector Narrowing:**
   - The byte order handling is consistent across different types (`TIn` and `HVTOut`) to ensure the output is correctly placed.

These changes address issues related to the platform-specific behavior of little- and big-endian operations, ensuring the code runs correctly on both platforms.

## [include/executor/engine/memory.ipp](https://github.com/WasmEdge/WasmEdge/blob/477c3a478de77217af32947a3f7603a007f7bfb4/include%2Fexecutor%2Fengine%2Fmemory.ipp)

#### Potential issues

1. Missing include for `Endian` header.
2. Incorrect return type of `runLoadLaneOp` and `runStoreLaneOp`. They should return `void` as they modify the stack.
3. Incorrect argument order in `runLoadExpandOp`. It takes a pointer to an array instead of the array itself.

#### Summary of changes

- **Added Support for WasmEdge**: Added support for the s390x platform in the executor to handle SIMD operations.
  
- **Memory Lane Handling**: The memory lane of the load and store instructions is now retrieved from the instruction's metadata. This allows for dynamic indexing based on platform-specific configurations.
  
- **Type Conversion**: The buffer data is converted to the appropriate type (`VT`) before assignment, ensuring type safety and compatibility with SIMD operations.

## [include/executor/engine/unary_numeric_vector.ipp](https://github.com/WasmEdge/WasmEdge/blob/477c3a478de77217af32947a3f7603a007f7bfb4/include%2Fexecutor%2Fengine%2Funary_numeric_vector.ipp)

#### Potential issues

1. The function `runExtractLaneOp` does not handle cases where the input vector is empty or has an odd size.
2. The `runVectorTruncSatOp` function uses an incorrect approach to saturate the output values when converting between float and int vectors.
3. There are no major coding issues in the file.

#### Summary of changes

- **Key Changes:**
  * Added inclusion of `common/endian.h` to support native-endian operations.
  * Updated vector extraction lane logic to handle both little and big-endian systems.
  * Enhanced vector extend operations for low, high, and saturate conversions.
  * Improved vector convert operations from floatx2_t to doublex2_t.
  * Added vector demote and promote operations with endian-specific handling.
  * Implemented bitwise mask operations using endian-specific logic.

## [include/executor/executor.h](https://github.com/WasmEdge/WasmEdge/blob/477c3a478de77217af32947a3f7603a007f7bfb4/include%2Fexecutor%2Fexecutor.h)

#### Potential issues

### Major Issues
1. **Missing Header Guards**:
   - The header file includes `#include "common/errcode.h"` without any header guards.

2. **Missing Namespace Declarations**:
   - The namespace declarations at the beginning of each translation unit are missing, which is considered good practice.

3. **Incorrect Function Signature for `invoke` and `asyncInvoke`**:
   - The function signatures in `executor.h` do not match their implementation in `executor.cpp`. For example, `invoke` should take a reference to the function instance instead of a pointer, and `asyncInvoke` should return an `Async` object.

### Summary
- Missing header guards.
- Missing namespace declarations at the beginning of each translation unit.
- Incorrect function signatures for `invoke` and `asyncInvoke`.

#### Summary of changes

- Added support for WasmEdge on the s390x platform.
- Moved includes from `common/defines.h` to `common/types.h`.

## [include/host/wasi/environ.h](https://github.com/WasmEdge/WasmEdge/blob/477c3a478de77217af32947a3f7603a007f7bfb4/include%2Fhost%2Fwasi%2Fenviron.h)

#### Potential issues

N/A

#### Summary of changes

- **Changes in the WasiAddrStorage struct**:
  - Added `EndianValue` for converting `AddressFamily` from native byte order to little-endian.

- **Changes in Environ class methods**:
  - Used `EndianValue` to ensure `Arguments` and `EnvironVariables` are in little-endian format before passing them to WASI.
  
- **Changes in PreStat struct**:
  - Added a new member variable `pr_name_len`, converted it from native byte order to little-endian using `EndianValue`, and used it to set the length of the path.

These changes ensure that the WasmEdge platform supports WasiAddrStorage, Environ class methods, and PreStat struct with proper endianness conversions.

## [include/host/wasi/inode.h](https://github.com/WasmEdge/WasmEdge/blob/477c3a478de77217af32947a3f7603a007f7bfb4/include%2Fhost%2Fwasi%2Finode.h)

#### Potential issues

N/A

#### Summary of changes

-**Key Changes:**  
- Added `#include "common/types.h"` to include the `types.h` header file, which likely contains definitions and utilities for type conversions.
- Updated includes from `common/defines.h` and `common/span.h` to `common/defines.h`, `common/errcode.h`, and `common/types.h` respectively.  
- Adjusted indentation to match the new includes.

## [include/host/wasi/vinode.h](https://github.com/WasmEdge/WasmEdge/blob/477c3a478de77217af32947a3f7603a007f7bfb4/include%2Fhost%2Fwasi%2Fvinode.h)

#### Potential issues

-3 issues found:

1. **`fdPread` and `fdPwrite` methods**:
   - Both `fdPread` and `fdPwrite` accept a `Span<Span<uint8_t>>` for the input buffers, which is not allowed in C++ as it implies passing multiple spans by reference. Instead, they

#### Summary of changes

- Updated WasmEdge support for s390x platform.
- Added support for converting `FsRightsBase` and `FsRightsInheriting` to little-endian values using `EndianValue`.
- Modified the `fdFdstatGet` method to set the correct endianness before calling `Node.fdFdstatGet`.

## [include/runtime/instance/data.h](https://github.com/WasmEdge/WasmEdge/blob/477c3a478de77217af32947a3f7603a007f7bfb4/include%2Fruntime%2Finstance%2Fdata.h)

#### Potential issues

1. Missing include for `Byte`.
2. Incorrect offset type in `loadValue` method.
3. Memory access out of bounds check missing.

#### Summary of changes

- **Key Changes:**
  * Added support for WasmEdge on the s390x platform.
  * Modified data loading method to use `EndianValue<uint128_t>` for efficient memory access and endianness handling.

### Most Important Changes:
- **Data Loading Optimization:** Changed data loading method to use `EndianValue<uint128_t>`, which is more efficient for big-endian systems and provides better control over memory alignment.
- **Endianness Handling:** Ensured that the loaded data is handled correctly regardless of whether it's native or big-endian.

## [include/runtime/instance/memory.h](https://github.com/WasmEdge/WasmEdge/blob/477c3a478de77217af32947a3f7603a007f7bfb4/include%2Fruntime%2Finstance%2Fmemory.h)

#### Potential issues

1. **Memory Instance Constructor**: The constructor does not check if the allocation is successful, which may lead to undefined behavior or memory corruption.
   - Issue: Ensure that `Allocator::allocate` returns a valid pointer or throw an exception.

2. **Memory Instance Destructor**: The destructor does not deallocate any memory, which can cause memory leaks.
   - Issue: Ensure that `Allocator::release` is called before the object is deleted to free up resources.

3. **MemoryInstance Data Access Methods**: The methods for accessing and manipulating data should handle edge cases such as out-of-bounds accesses.
   - Issue: Add checks in the methods to ensure the accessed memory address falls within valid bounds.

#### Summary of changes

-Added support for WasmEdge on the s390x platform.
-Included necessary header files.
-Fixed type casts and conversions to ensure correct memory operations are performed.
-Extended the `MemoryInstance` class to include methods for loading and storing data.
-Improved error handling and validation of memory operations.

## [lib/api/wasmedge.cpp](https://github.com/WasmEdge/WasmEdge/blob/477c3a478de77217af32947a3f7603a007f7bfb4/lib%2Fapi%2Fwasmedge.cpp)

#### Potential issues

N/A

#### Summary of changes

-### Important Changes
1. **Added S390X Support**: The function `to_uint128_t`, `to_int128_t`, and `to_WasmEdge_128_t` have been modified to include support for the `__s390x__` compiler flag, allowing them to handle 128-bit integer conversions on the S390X architecture.

## [lib/common/hash.cpp](https://github.com/WasmEdge/WasmEdge/blob/477c3a478de77217af32947a3f7603a007f7bfb4/lib%2Fcommon%2Fhash.cpp)

#### Potential issues

1. Missing include of `<cstdint>` and `<string>`
2. Incorrect use of `likely` macro
3. Potential division by zero in `isPrime` function

#### Summary of changes

-**Changes Summary**:  
  - **Endian Handling**: Added `WasmEdge::Endian` namespace and used `WasmEdge::byteswap` for endian swapping.
  - **Function Changes**: Modified the `read` function to use the new endian handling, supporting both little and big-endian architectures.

- **Top 3 Most Important Changes**:
  1. Added support for `WasmEdge::Endian` namespace for managing endianness.
  2. Implemented `WasmEdge::byteswap` for swapping byte order within uint64_t and uint32_t values.
  3. Modified the `read` function to handle both little and big-endian byte orders.

This change ensures compatibility with different platforms that use different byte order conventions, enhancing the robustness of the codebase.

## [lib/executor/engine/engine.cpp](https://github.com/WasmEdge/WasmEdge/blob/477c3a478de77217af32947a3f7603a007f7bfb4/lib%2Fexecutor%2Fengine%2Fengine.cpp)

#### Potential issues

N/A

#### Summary of changes

- **New Includes:** Added `#include "common/endian.h"` to support the WasmEdge architecture's endianness.
- **Endian Handling:** Modified code for handling the endianess of input parameters, ensuring compatibility with both big and little-endian systems.

## [lib/executor/engine/proxy.cpp](https://github.com/WasmEdge/WasmEdge/blob/477c3a478de77217af32947a3f7603a007f7bfb4/lib%2Fexecutor%2Fengine%2Fproxy.cpp)

#### Potential issues

-1. **Code Complexity**: The code has a high complexity due to the use of template metaprogramming for proxying intrinsics. This can lead to performance issues and difficulty in maintaining.
  
- 2. **Thread Safety**: The thread_local variables `Executor::This`, `Executor::CurrentStack`, `Executor::ExecutionContext`, `Executor::StackTrace`, and `Executor::StackTraceSize` are not properly synchronized or guarded. This could lead to race conditions and unexpected behavior.
  
- 3. **Type Safety**: The code does not enforce type safety for function arguments and results. It relies on implicit conversions and may lead to runtime errors if types do not match expectations.

#### Summary of changes

- **Changes in this patch:**
  - **Support for WasmEdge on the s390x platform:** 
    - Added materializing of arguments to prevent runtime failures.
  
- **Most important changes first:**
  - Materializing of arguments is a critical enhancement for compatibility with WasmEdge on the s390x platform.

## [lib/executor/helper.cpp](https://github.com/WasmEdge/WasmEdge/blob/477c3a478de77217af32947a3f7603a007f7bfb4/lib%2Fexecutor%2Fhelper.cpp)

#### Potential issues

-1. Incorrect code comment in `enterFunction` where it says the PC should increase after the callee return, but it actually decrements. This is a logical error that could cause unexpected behavior.
-2. In `getFuncInstByIdx`, if the module instance is nullptr, it returns null instead of throwing an exception. This might lead to runtime errors if the caller tries to use the returned value.
-3. In `getTabInstByIdx`, if the module instance is nullptr, it also returns null instead of throwing an exception. This could cause runtime errors in similar scenarios.

#### Summary of changes

- Added a check for the endianness of `Val` before accessing its value.
- Adjusted the shift values based on the current architecture's endianness.
- Used constexpr to avoid runtime type checking at compile time, improving performance.
- Added a missing brace in the else statement for clarity.

## [lib/host/wasi/clock-linux.cpp](https://github.com/WasmEdge/WasmEdge/blob/477c3a478de77217af32947a3f7603a007f7bfb4/lib%2Fhost%2Fwasi%2Fclock-linux.cpp)

#### Potential issues

-`clockResGet` and `clockTimeGet` functions do not check if the clock ID is valid, which could lead to undefined behavior. Ensure that the clock ID passed as an argument is within the expected range before calling `clock_getres` or `clock_gettime`.

#### Summary of changes

- Added include for `common/types.h` which contains type definitions required for the clock functions.
- Adjusted logic in `clockResGet` and `clockTimeGet` to correctly handle time values using the `<stdbool.h>` library's `le()` function.

## [lib/host/wasi/inode-linux.cpp](https://github.com/WasmEdge/WasmEdge/blob/477c3a478de77217af32947a3f7603a007f7bfb4/lib%2Fhost%2Fwasi%2Finode-linux.cpp)

#### Potential issues

N/A

#### Summary of changes

- - Added #include "common/types.h" at the top of the file to support WasmEdge on the s390x platform.
- - Converted `__wasi_fdstat_t` and `__wasi_filestat_t` types from network byte order to little-endian using EndianValue for consistency.
- - Modified `fdFdstatGet`, `fdFilestatGet`, and `pathFilestatGet` functions to handle file attributes in little-endian format.

## [lib/host/wasi/linux.h](https://github.com/WasmEdge/WasmEdge/blob/477c3a478de77217af32947a3f7603a007f7bfb4/lib%2Fhost%2Fwasi%2Flinux.h)

#### Potential issues

1. Missing #include "common/types.h".

#### Summary of changes

- **Key Changes:**
  - Added `#include "common/types.h"` to import necessary types.
  - Updated `fromTimespec` function to use `EndianValue`.
  
**Most Important Changes:**

1. **Importing Types:** Added an inclusion for `types.h`, which likely contains essential types used in the codebase.

2. **Updated Function Name:** Changed `fromTimespec` to `fromTimespec<EndianValue>`, suggesting a new design choice involving endian handling for timestamps.

## [lib/host/wasi/wasifunc.cpp](https://github.com/WasmEdge/WasmEdge/blob/477c3a478de77217af32947a3f7603a007f7bfb4/lib%2Fhost%2Fwasi%2Fwasifunc.cpp)

#### Potential issues

N/A

#### Summary of changes

N/A

## [lib/llvm/codegen.cpp](https://github.com/WasmEdge/WasmEdge/blob/477c3a478de77217af32947a3f7603a007f7bfb4/lib%2Fllvm%2Fcodegen.cpp)

#### Potential issues

-1. Missing include directive for `llvm/codegen.h`.
-2. Incorrect use of `LLVMModuleFlagBehaviorError` without checking its return value.
-3. Unused variable `FTy`.

#### Summary of changes

- * Added support for WasmEdge on the s390x platform.

## [lib/llvm/compiler.cpp](https://github.com/WasmEdge/WasmEdge/blob/477c3a478de77217af32947a3f7603a007f7bfb4/lib%2Fllvm%2Fcompiler.cpp)

#### Potential issues

N/A

#### Summary of changes

-**Changes summary**

1. **Supporting WasmEdge on the s390x platform**
   - Added support for round-even instructions on the s390x architecture.
   - Enhanced memory access functions to handle different endianness and alignment requirements.

2. **Optimizing vector operations**
   - Added support for nearest and float-nearest vector operations on the x86_64 and aarch64 platforms.
   - Introduced a switch-endian function to simplify handling of data in different byte orders.

3. **Enhancing integer arithmetic**
   - Implemented big-endian integer promotion for 128-bit vectors.
   - Added support for big-endian integer multiplication and division operations.

**Most important changes:**

1. Support for WasmEdge on the s390x platform, enabling native instructions.
2. Enhanced memory access functions with endianness handling and alignment checks.
3. Optimized vector operations for different platforms, including nearest and float-nearest operations.

## [lib/llvm/llvm.h](https://github.com/WasmEdge/WasmEdge/blob/477c3a478de77217af32947a3f7603a007f7bfb4/lib%2Fllvm%2Fllvm.h)

#### Potential issues

N/A

#### Summary of changes

-Add support for WasmEdge on the s390x platform (`Bswap`, `S390VPerm`).
- Improve instruction mapping and attribute handling.

## [lib/loader/ast/instruction.cpp](https://github.com/WasmEdge/WasmEdge/blob/477c3a478de77217af32947a3f7603a007f7bfb4/lib%2Floader%2Fast%2Finstruction.cpp)

#### Potential issues

N/A

#### Summary of changes

- **Removed unnecessary casting**: The patch removes unnecessary casting of `uint128_t` to `uint32_t` or `uint64_t`, simplifying the code.

test/thread/ThreadTest.cpp
```diff
@@ -209,13 +210,14 @@ TEST(AsyncExecute, GasThreadTest) {
...
-          "mt19937",
-          std::initializer_list<WasmEdge::ValVariant>{
-              UINT32_C(2504) * Index, UINT64_C(5489), UINT64_C(100000) + Index},
-          {WasmEdge::ValType(WasmEdge::TypeCode::I32),
-           WasmEdge::ValType(WasmEdge::TypeCode::I64),
-           WasmEdge::ValType(WasmEdge::TypeCode::I64)});
+      AsyncResults[Index] =
+          VM.asyncExecute("mt19937",
+                          std::initializer_list<WasmEdge::ValVariant>{
+                              static_cast<uint32_t>(2504 * Index),
```
hydai:
```suggestion
                              UINT32_C(2504) * static_cast<uint32_t>(Index),
```

test/thread/ThreadTest.cpp
```diff
@@ -177,13 +177,14 @@ TEST(AsyncExecute, ThreadTest) {
...
```
hydai:
```suggestion
                              UINT32_C(2504) * static_cast<uint32_t>(Index),
```

include/common/int128.h
```diff
@@ -23,14 +23,10 @@
 #include "endian.h"
 
 #include <cstdint>
+#include <iostream>
```
hydai: What's reason to add the iostream header?

include/ast/instruction.h
```diff
@@ -339,8 +332,8 @@ class Instruction {
     } Memories;
     // Type 8: Num.
 #if defined(__x86_64__) || defined(__aarch64__) ||                             \
-    (defined(__riscv) && __riscv_xlen == 64)
-    uint128_t Num;
+    (defined(__riscv) && __riscv_xlen == 64) || defined(__s390x__)
+    ValVariant Num;
```
hydai: What's the reason to use `ValVariant` instead of the `uint128_t`?

include/common/types.h
```diff
@@ -840,6 +840,31 @@ template <typename T> inline T &retrieveExternRef(const RefVariant &Val) {
...
+  // static_assert(std::is_integral_v<T>);
+  if constexpr (sizeof(T) == 1) {
+    return Val;
+  } else if constexpr (sizeof(T) == 2) {
+    return static_cast<T>(__builtin_bswap16(static_cast<uint16_t>(Val)));
+  } else if constexpr (sizeof(T) == 4) {
+    return static_cast<T>(__builtin_bswap32(static_cast<uint32_t>(Val)));
+  } else if constexpr (sizeof(T) == 8) {
+    return static_cast<T>(__builtin_bswap64(static_cast<uint64_t>(Val)));
+  }
```
hydai: What will happen if `type(Val) == int128_t`?

include/ast/instruction.h
vishruth-thimmaiah: That change was unintentional, I've reverted it now.

include/common/types.h
vishruth-thimmaiah: There currently isn't any instance where `type(Val) == int128_t`, but I've updated the function to support `int128` as well.

include/ast/instruction.h
```diff
@@ -62,14 +62,7 @@ class Instruction {
...
   /// Constructor assigns the OpCode and the Offset.
   Instruction(OpCode Byte, uint32_t Off = 0) noexcept
-      : Offset(Off), Code(Byte) {
-#if defined(__x86_64__) || defined(__aarch64__) ||                             \
-    (defined(__riscv) && __riscv_xlen == 64)
-    Data.Num = static_cast<uint128_t>(0);
-#else
-    Data.Num.Low = static_cast<uint64_t>(0);
-    Data.Num.High = static_cast<uint64_t>(0);
-#endif
```
hydai: What's the reason for deleting this part?

include/common/types.h
```diff
@@ -840,6 +840,29 @@ template <typename T> inline T &retrieveExternRef(const RefVariant &Val) {
...
+  if constexpr (sizeof(T) == 1) {
+    return Val;
+  } else if constexpr (sizeof(T) == 2) {
+    return static_cast<T>(__builtin_bswap16(static_cast<uint16_t>(Val)));
+  } else if constexpr (sizeof(T) == 4) {
+    return static_cast<T>(__builtin_bswap32(static_cast<uint32_t>(Val)));
+  } else if constexpr (sizeof(T) == 8) {
+    return static_cast<T>(__builtin_bswap64(static_cast<uint64_t>(Val)));
+  } else if constexpr (sizeof(T) == 16) {
+    return ((__uint128_t)__builtin_bswap64((uint64_t)Val) << 64) |
```
hydai: Use `static_cast`

hydai (changes requested):

include/runtime/instance/memory.h
```diff
@@ -282,16 +284,33 @@ class MemoryInstance {
...
           assuming(sizeof(T) == 16);
           Value = 0U;
           std::memcpy(&Value, &DataPtr[Offset], Length);
+#if !WASMEDGE_ENDIAN_LITTLE_BYTE
+          Value = (static_cast<uint128_t>(
+                       __builtin_bswap64(static_cast<uint64_t>(Value)))
+                   << 64) |
+                  static_cast<uint128_t>(
+                      __builtin_bswap64(static_cast<uint64_t>(Value >> 64)));
+#endif
```
PhantomInTheWire: we should probably consider extracting this to a lambda this would allow us to do something like this
```cpp
 Value = bswap128(Value);
 ```

File: lib/host/wasi/wasifunc.cpp [hunk 9]
```diff
@@ -402,6 +403,10 @@ Expect<uint32_t> WasiArgsGet::body(const Runtime::CallingFrame &Frame,
     return Res.error();
   }
 
+#if !WASMEDGE_ENDIAN_LITTLE_BYTE
+  Argv[Argv.size() - 1] = __builtin_bswap32(Argv[Argv.size() - 1]);
+#endif
+
```
PhantomInTheWire: why are we byte swapping only the last element?

lib/host/wasi/wasifunc.cpp
```diff
@@ -1929,24 +1954,29 @@ Expect<uint32_t> WasiSockRecvFromV1::body(const Runtime::CallingFrame &Frame,
...
     // Check for invalid address.
     if (unlikely(RiDataArr.size() != BufLen)) {
       return __WASI_ERRNO_FAULT;
     }
     WasiRiData.emplace_back_unchecked(RiDataArr);
   }
 
+  uint32_t A;
+  MemInst->loadValue(A, 0U);
+
```
PhantomInTheWire: do we use these values somewhere? if yes you should probably rename A. but these looks more like debug statements, it's best to remove dead code.

test/spec/spectest.cpp
PhantomInTheWire, line 1: We seem to repeating the SIMD vector parsing code for different lane types with only the array type and size changing. I think we can create simple helper template function to make this more cleaner. something with this kind of signature.
```cpp
template<typename T, size_t N>
static void parseSIMDLanes(WasmEdge::uint128_t& V128, 
	const simdjson::dom::array& ValueNodeArray)
````

File: lib/host/wasi/wasifunc.cpp (hunk 9 above)
Comment (vishruth-thimmaiah) [reply lib/host/wasi/wasifunc.cpp]: The elements are reversed by the `Env.argsGet` function. This is code i forgot to cleanup

File: include/runtime/instance/memory.h (hunk 8 above)
Comment (vishruth-thimmaiah) [reply include/runtime/instance/memory.h]: Should I add it as a private method under `MemoryInstance`, or in a file like `include/common/int128.h`?

Comment (PhantomInTheWire) [reply include/runtime/instance/memory.h]: putting a simple lambda in this file itself should do for now

File: test/spec/spectest.cpp
Comment (vishruth-thimmaiah) [reply test/spec/spectest.cpp:1]: I'm not sure if I should try to refactor this part into a template function. There are slight differences in how floating point values are handled, as well as using `stdoll` instead of `stdol` for f64/i64.

https://github.com/WasmEdge/WasmEdge/blob/6c710cf3a43adbbc8f63710f9260c5da22eb7f2b/test/spec/spectest.cpp#L497-L595

Comment (PhantomInTheWire) [reply test/spec/spectest.cpp:1]: it's fine try to cleanup as much as you can

File: include/executor/engine/atomic.ipp [hunk 11]
```diff
@@ -168,7 +169,16 @@ TypeT<T> Executor::runAtomicAddOp(Runtime::StackManager &StackMgr,
...
+  I Return;
+#if WASMEDGE_ENDIAN_LITTLE_BYTE
+  Return = AtomicObj->fetch_add(Value);
+#else
+  I Result;
+  do {
+    Return = AtomicObj->load();
+    Result = getLittleEndian(Return) + Value;
+  } while (!AtomicObj->compare_exchange_weak(Return, getLittleEndian(Result)));
+#endif
```
hydai: I am wondering if we can extract this pattern into a template or separate functions. The reason is that I don't want many `#if ... #else ... #endif` blocks scattered across various files, as it will be very difficult to maintain.

include/executor/engine/atomic.ipp
```diff
@@ -86,7 +87,7 @@ TypeT<T> Executor::runAtomicLoadOp(Runtime::StackManager &StackMgr,
   }
 
   I Return = AtomicObj->load();
-  RawAddress.emplace<T>(static_cast<T>(Return));
+  RawAddress.emplace<T>(static_cast<T>(getLittleEndian(Return)));
```
hydai: Also, I am wondering if we can handle this similarly to how `std::endian` and `std::byteswap` work in C++20/23. It should be easier to migrate if we are moving forward to a new C++ standard, just like we did previously with the `expected` backport.

lib/host/wasi/wasifunc.cpp
```diff
@@ -757,12 +758,14 @@ Expect<uint32_t> WasiFdPread::body(const Runtime::CallingFrame &Frame,
     // Capping total size.
     const __wasi_size_t Space =
         std::numeric_limits<__wasi_size_t>::max() - TotalSize;
-    const __wasi_size_t BufLen =
-        unlikely(IOV.buf_len > Space) ? Space : IOV.buf_len;
+    const __wasi_size_t BufLen = unlikely(getLittleEndian(IOV.buf_len) > Space)
+                                     ? Space
+                                     : getLittleEndian(IOV.buf_len);
```
PhantomInTheWire: what is the reason for calling getLittleEndian twice?

lib/host/wasi/wasifunc.cpp
```diff
@@ -859,12 +862,14 @@ Expect<uint32_t> WasiFdPwrite::body(const Runtime::CallingFrame &Frame,
...
```
PhantomInTheWire: ditto.

lib/host/wasi/wasifunc.cpp
```diff
@@ -914,12 +919,14 @@ Expect<uint32_t> WasiFdRead::body(const Runtime::CallingFrame &Frame,
...
```
PhantomInTheWire: ditto.

lib/host/wasi/wasifunc.cpp
```diff
@@ -1077,12 +1084,14 @@ Expect<uint32_t> WasiFdWrite::body(const Runtime::CallingFrame &Frame,
...
```
PhantomInTheWire: ditto.

include/runtime/instance/memory.h
```diff
@@ -327,7 +342,28 @@ class MemoryInstance {
...
+        uint64_t Val = 0;
+        if constexpr (std::is_integral_v<T>) {
+          Val = static_cast<uint64_t>(Value);
+        } else if constexpr (std::is_same_v<T, float>) {
+          Val = static_cast<uint64_t>(
+              *reinterpret_cast<const uint32_t *>(&Value));
+        } else if constexpr (std::is_same_v<T, double>) {
+          Val = static_cast<uint64_t>(
+              *reinterpret_cast<const uint64_t *>(&Value));
+        }
```
PhantomInTheWire: can this be simplified to
```cpp
uint64_t Val = 0;
std::memcpy(&Val, &Value, sizeof(T));
```

include/runtime/instance/memory.h
```diff
@@ -327,7 +342,28 @@ class MemoryInstance {
     }
     // Copy the stored data to the value.
     if (likely(Length > 0)) {
+#if WASMEDGE_ENDIAN_LITTLE_BYTE
       std::memcpy(&DataPtr[Offset], &Value, Length);
+#else
+      if constexpr (Length > 8) {
+        int128_t Val = Value;
```
PhantomInTheWire: what is the logic behind this being `int128_t` and not `uint128_t`?

include/runtime/instance/memory.h
```diff
@@ -342,6 +378,13 @@ class MemoryInstance {
   uint8_t *DataPtr = nullptr;
   const uint32_t PageLimit;
   /// @}
+  static inline uint128_t bswap128(uint128_t Val) noexcept {
```
PhantomInTheWire: we should consider making this `constexpr` so that it can be evaluated at compile time for constant values

include/runtime/instance/memory.h
```diff
@@ -282,16 +284,29 @@ class MemoryInstance {
...
         // Floating case. Do the memory copy.
-        std::memcpy(&Value, &DataPtr[Offset], sizeof(T));
+#if WASMEDGE_ENDIAN_LITTLE_BYTE
+        std::memcpy(&Value, &DataPtr[Offset], Length);
+#else
+        uint64_t LoadVal = 0;
+        std::memcpy(&LoadVal, &DataPtr[Offset], Length);
+        LoadVal = __builtin_bswap64(LoadVal) << (64 - Length * 8);
+        std::memcpy(&Value, &LoadVal, Length);
+#endif
```
PhantomInTheWire: are you sure this code is correct? i think it should be a right shift.

test/spec/spectest.cpp
```diff
@@ -95,6 +95,21 @@ SpecTest::CommandID resolveCommand(std::string_view Name) {
...
+  for (std::string_view X : ValueNodeArray) {
+    V[I] = static_cast<T>(std::stoull(std::string(X)));
+    I++;
+  }
+#if !WASMEDGE_ENDIAN_LITTLE_BYTE
+  std::reverse(V, V + 16 / N);
+#endif
+  std::memcpy(&V128, &V, 16);
+}
+
```
PhantomInTheWire: we should validate if ValueNodeArray contains exactly 16/N elements, else this could lead to undefined behaviour, if the array has fewer elements than expected

test/spec/spectest.cpp
```diff
@@ -518,15 +495,13 @@ bool SpecTest::compare(const std::pair<std::string, std::string> &Expected,
...
-#else
-      const auto VF = reinterpret_cast<floatx4_t>(V64);
-      const auto VI = reinterpret_cast<uint32x4_t>(V64);
+      float VF[4];
+      uint32_t VI[4];
+      std::memcpy(VF, &Got.first.get<float>(), 16);
+      std::memcpy(VI, &Got.first.get<uint32_t>(), 16);
+#if !WASMEDGE_ENDIAN_LITTLE_BYTE
+      std::reverse(VI, VI + 4);
+      std::reverse(VF, VF + 4);
```
PhantomInTheWire: `get<T>()` doesn't return a pointer, it returns a single value, this might work sometimes due to lucky alignment but this can copy garbage data, and read things it is not supposed to and cause undefined behaviour.

test/spec/spectest.cpp
```diff
@@ -542,15 +517,13 @@ bool SpecTest::compare(const std::pair<std::string, std::string> &Expected,
...
-#else
-      const auto VF = reinterpret_cast<doublex2_t>(V64);
-      const auto VI = reinterpret_cast<uint64x2_t>(V64);
+      double VF[2];
+      uint64_t VI[2];
+      std::memcpy(VF, &Got.first.get<double>(), 16);
+      std::memcpy(VI, &Got.first.get<uint64_t>(), 16);
+#if !WASMEDGE_ENDIAN_LITTLE_BYTE
+      std::reverse(VI, VI + 2);
+      std::reverse(VF, VF + 2);
```
PhantomInTheWire: ditto.

include/runtime/instance/memory.h
vishruth-thimmaiah: We've flipped the bytes here for big endian, so the bit shift has to be a left shift.

include/runtime/instance/memory.h
vishruth-thimmaiah: Value here is in big endian form, so we cant directly memcpy in a way similar to loadValue, which reads a little endian value.

include/executor/engine/atomic.ipp
vishruth-thimmaiah: We could define the following template:
```cpp
template <typename T, typename AtomicOp, typename BinaryOp>
T atomicFetchOp(std::atomic<T> *AtomicObj, T Value, AtomicOp Op,
                BinaryOp BinOp) {
  if constexpr (WASMEDGE_ENDIAN_LITTLE_BYTE) {
    return Op(AtomicObj, Value);
  } else {
    T Return;
    T Result;
    do {
      Return = AtomicObj->load();
      Result = BinOp(getLittleEndian(Return), Value);
    } while (
        !AtomicObj->compare_exchange_weak(Return, getLittleEndian(Result)));
    return Return;
  }
}

```
And call with:
```cpp
  I Return = atomicFetchOp(
      AtomicObj, Value,
      [](std::atomic<I> *Obj, I Val) { return Obj->fetch_add(Val); },
      std::plus<I>{});
```

Maybe we could place this in `binary_numeric.ipp` for maintainability?
If this approach looks good, I can proceed with the implementation.

include/executor/engine/atomic.ipp
vishruth-thimmaiah: For std::endian, in `endian.h` we could use:
```cpp
namespace WasmEdge {
enum class Endian {
  Little = 1,
  Big = 0,
  Native = WASMEDGE_ENDIAN_LITTLE_BYTE,
};
} // namespace WasmEdge
```
and replace the preprocessor-switches with `if constexpr(Endian::Native == Endian::Little)`

for `std::byteswap`, perhaps we separate the byte swapping logic from `getLittleEndian`?

Let me know if this is a suitable approach, and i'll implement it.

hydai: `std::endian` works for me.

I am wondering if we can implement an operator for these types instead of calling `getLittleEndian`. Would this be better?

include/executor/engine/atomic.ipp
hydai: Looks better than the `#if...#else` blocks.

hydai (changes requested):

include/executor/engine/atomic.ipp
vishruth-thimmaiah: you mean something like this?
```cpp
LE<I> Return = AtomicObj->load();
RawAddress.emplace<T>(static_cast<T>(Return));
```

hydai: Something like this:

```cpp
EndianValue<I> Return = AtomicObj->load();
// for getting little endian, le()
RawAddress.emplace<T>(static_cast<T>(Return.le()));
// for getting big endian, be()
RawAddress.emplace<T>(static_cast<T>(Return.be()));
// for getting the native value, raw()
RawAddress.emplace<T>(static_cast<T>(Return.raw()));
```

Just an idea; I'm not sure if this is a good approach or not. This should be cleaner than inserting `getLittleEndian` everywhere.

.github/workflows/build_for_s390x.yml
```diff
@@ -0,0 +1,95 @@
+name: Build and Test WasmEdge on s390x arch
```
hydai, line 1: Have you tested this workflow on your own forked repo? If so, what is the total execution time?

include/common/types.h
```diff
@@ -840,6 +841,62 @@ template <typename T> inline T &retrieveExternRef(const RefVariant &Val) {
...
+    Native = Val;
+    LittleEndian = switchEndian(Native);
+    return *this;
+  }
+
+  constexpr const T le() const noexcept { return LittleEndian; }
+  constexpr const T newLe() noexcept {
+    LittleEndian = switchEndian(Native);
+    return LittleEndian;
+  }
```
hydai: I don't see the point of this function. If you can use the constructor to perform the little-endian conversion, why do you need this new function for it?

include/common/types.h
```diff
@@ -840,6 +841,62 @@ template <typename T> inline T &retrieveExternRef(const RefVariant &Val) {
...
+
+  constexpr const T &raw() const noexcept { return Native; }
+  constexpr T &raw() noexcept { return Native; }
+
+private:
+  T Native;
+  T LittleEndian;
```
hydai: This is weird. We should only have one buffer for it. Since you have the endian information and the buffer, you can output the value with `le`, `be`, or `native` by applying the conversion inside these functions.

.github/workflows/build_for_s390x.yml
vishruth-thimmaiah: 44min - [workflow](https://github.com/vishruth-thimmaiah/WasmEdge/actions/runs/17125199004)

include/common/types.h
vishruth-thimmaiah: I also added the ability to update `Native` directly using `&raw()`, but if applying the conversion within `le()` is preferred, this can be removed.

.github/workflows/build_for_s390x.yml
vishruth-thimmaiah: I also tested a second [workflow](https://github.com/vishruth-thimmaiah/WasmEdge/actions/runs/17153985153/job/48666466813) that runs the entire test suite. would that be a better approach? The new workflow takes ~1hr30min

include/executor/engine/binary_numeric_vector.ipp
```diff
@@ -14,7 +14,11 @@ Expect<void> Executor::runReplaceLaneOp(ValVariant &Val1,
                                         const uint8_t Index) const {
   using VTOut [[gnu::vector_size(16)]] = TOut;
   VTOut &Result = Val1.get<VTOut>();
-  Result[Index] = static_cast<TOut>(Val2.get<TIn>());
+  if constexpr (Endian::native == Endian::little)
+    Result[Index] = static_cast<TOut>(Val2.get<TIn>());
+  else
+    Result[(16 / sizeof(TOut)) - 1 - Index] =
+        static_cast<TOut>(Val2.get<TIn>());
```
hydai: All your if-else blocks must be wrapped with `{}`.

include/host/wasi/environ.h
```diff
@@ -150,6 +154,7 @@ class Environ : public PollerContext {
       if (Env.size() > 1) {
         Env[1] = Env[0] + Size + UINT32_C(1);
       }
+      Env[0] = EndianValue(Env[0]).le();
```
hydai, line 163: You convert this after `Env[1]` is calculated. Could you explain the logic here? Why do we only need to consider `Env[0]` and ignore `Env[1]`?

lib/executor/engine/engine.cpp
```diff
@@ -1011,20 +1015,16 @@ Expect<void> Executor::execute(Runtime::StackManager &StackMgr,
...
 #ifdef __clang__
-      uint8x16_t Result = {Vector[Index[0] & 0xF],  Vector[Index[1] & 0xF],
-                           Vector[Index[2] & 0xF],  Vector[Index[3] & 0xF],
-                           Vector[Index[4] & 0xF],  Vector[Index[5] & 0xF],
-                           Vector[Index[6] & 0xF],  Vector[Index[7] & 0xF],
-                           Vector[Index[8] & 0xF],  Vector[Index[9] & 0xF],
-                           Vector[Index[10] & 0xF], Vector[Index[11] & 0xF],
-                           Vector[Index[12] & 0xF], Vector[Index[13] & 0xF],
-                           Vector[Index[14] & 0xF], Vector[Index[15] & 0xF]};
+      uint8x16_t Result = __builtin_shufflevector(Vector, Index);
```
hydai, line 1028: Do you have any idea about the minimum supported version of this built-in function? Since WasmEdge supports multiple LLVM versions, it would be better to check the version before using this method.

.github/workflows/build_for_s390x.yml
hydai: Sounds good enough for me. The current workflow of RISC-V takes a similar amount of time.

include/host/wasi/environ.h
vishruth-thimmaiah: This is nested in a loop defined on line 148. As `Env` is shrunk down using `subspan()` on every iteration, all the values within Env will be swapped.

lib/executor/engine/engine.cpp
vishruth-thimmaiah: This appears to be supported in clang 3.0, so it should be supported by all newer versions.

test/spec/spectest.cpp
```diff
@@ -95,6 +96,21 @@ SpecTest::CommandID resolveCommand(std::string_view Name) {
...
+                           const simdjson::dom::array &ValueNodeArray) {
+  assuming(16 / N == ValueNodeArray.size());
+  T V[16 / N];
+  size_t I = 0;
+  for (std::string_view X : ValueNodeArray) {
+    V[I] = static_cast<T>(std::stoull(std::string(X)));
+    I++;
+  }
+  if constexpr (Endian::native == Endian::big)
+    std::reverse(V, V + 16 / N);
```
hydai: Ditto.

test/host/socket/wasi_socket.cpp
```diff
@@ -16,6 +17,8 @@
 #include <string>
 #include <string_view>
 
+#define SWAP_E(Val) WasmEdge::EndianValue(Val).le()
+
```
hydai: What is the reason we need to add this MACRO? Why don't we just use the original statement?

test/host/socket/wasi_socket.cpp
```diff
@@ -124,6 +127,13 @@ void allocateAddrinfoArray(WasmEdge::Runtime::Instance::MemoryInstance &MemInst,
     }
   }
 }
+
+template <typename T>
+T getValue(WasmEdge::Runtime::Instance::MemoryInstance &MemInst, uint32_t Ptr) {
+  T Value;
+  MemInst.loadValue(Value, Ptr);
+  return Value;
+}
```
hydai: Is there any benefit to having this helper function?

test/host/wasi/wasi.cpp
```diff
@@ -19,6 +20,8 @@
 #include <string_view>
 #include <thread>
 
+#define SWAP_E(Val) WasmEdge::EndianValue(Val).le()
```
hydai: Ditto.

test/host/wasi/wasi.cpp
```diff
@@ -40,13 +43,20 @@ void writeAddress(WasmEdge::Runtime::Instance::MemoryInstance &MemInst,
...
               sizeof(__wasi_address_t));
 }
 
+template <typename T>
+T getValue(WasmEdge::Runtime::Instance::MemoryInstance &MemInst, uint32_t Ptr) {
+  T Value;
+  MemInst.loadValue(Value, Ptr);
+  return Value;
+}
+
```
hydai: Ditto.

lib/loader/ast/instruction.cpp
```diff
@@ -616,12 +616,12 @@ Expect<void> Loader::loadInstruction(AST::Instruction &Instr) {
   // Const Instructions.
   case OpCode::I32__const:
     EXPECTED_TRY(FMgr.readS32().map_error(ReportError).map([&](int32_t Num) {
-      Instr.setNum(static_cast<uint128_t>(static_cast<uint32_t>(Num)));
+      Instr.setNum(static_cast<uint32_t>(Num));
```
hydai, line 619: What is the reason for removing the original conversion?

lib/loader/ast/instruction.cpp
```diff
@@ -616,12 +616,12 @@ Expect<void> Loader::loadInstruction(AST::Instruction &Instr) {
...
     }));
     return {};
   case OpCode::I64__const:
     EXPECTED_TRY(FMgr.readS64().map_error(ReportError).map([&](int64_t Num) {
-      Instr.setNum(static_cast<uint128_t>(static_cast<uint64_t>(Num)));
+      Instr.setNum(static_cast<uint64_t>(Num));
```
hydai, line 624: Ditto.

lib/llvm/compiler.cpp
```diff
@@ -4570,6 +4615,8 @@ class FunctionCompiler {
                          LLVM::Type VectorTy) noexcept {
     auto Vector = stackPop();
     compileLoadOp(MemoryIndex, Offset, Alignment, LoadTy);
+    if constexpr (Endian::native == Endian::big)
+      Index = VectorTy.getVectorSize() - 1 - Index;
```
hydai: Ditto.

lib/llvm/compiler.cpp
```diff
@@ -3998,8 +4004,45 @@ class FunctionCompiler {
...
+        Ret = Builder.createExtractValue(Exchange, 0);
+        auto Success = Builder.createExtractValue(Exchange, 1);
+        Builder.createCondBr(Success, OkBB, AtomicBB);
+        Builder.positionAtEnd(OkBB);
+      } else {
+        Ret = Builder.createAtomicRMW(BinOp, Ptr, switchEndian(Value),
+                                      LLVMAtomicOrderingSequentiallyConsistent);
+      }
+    } else
+      Ret = Builder.createAtomicRMW(BinOp, Ptr, switchEndian(Value),
```
hydai: Ditto.

lib/llvm/compiler.cpp
```diff
@@ -4604,6 +4652,8 @@ class FunctionCompiler {
                           unsigned Alignment, unsigned Index, LLVM::Type LoadTy,
                           LLVM::Type VectorTy) noexcept {
     auto Vector = Stack.back();
+    if constexpr (Endian::native == Endian::big)
+      Index = VectorTy.getVectorSize() - Index - 1;
```
hydai: Ditto.

lib/llvm/compiler.cpp
```diff
@@ -4621,6 +4671,8 @@ class FunctionCompiler {
   }
   void compileExtractLaneOp(LLVM::Type VectorTy, unsigned Index) noexcept {
     auto Vector = Builder.createBitCast(Stack.back(), VectorTy);
+    if constexpr (Endian::native == Endian::big)
+      Index = VectorTy.getVectorSize() - Index - 1;
```
hydai: Ditto.

lib/llvm/compiler.cpp
```diff
@@ -4636,6 +4688,8 @@ class FunctionCompiler {
   void compileReplaceLaneOp(LLVM::Type VectorTy, unsigned Index) noexcept {
     auto Value = Builder.createTrunc(stackPop(), VectorTy.getElementType());
     auto Vector = Stack.back();
+    if constexpr (Endian::native == Endian::big)
+      Index = VectorTy.getVectorSize() - Index - 1;
```
hydai: Ditto.

lib/llvm/compiler.cpp
```diff
@@ -5001,15 +5067,19 @@ class FunctionCompiler {
...
+                 : Builder.createShuffleVector(
+                       F2, F1, LLVM::Value::getConstVector32(LLContext, Mask));
+    stackPush(Builder.createBitCast(V, Context.Int64x2Ty));
   }
   void compileVectorExtend(LLVM::Type FromTy, bool Signed, bool Low) noexcept {
     auto ExtTy = FromTy.getExtendedElementVectorType();
     const auto Count = FromTy.getVectorSize();
     std::vector<uint32_t> Mask(Count / 2);
+    if constexpr (Endian::native == Endian::big)
+      Low = !Low;
```
hydai: Ditto

lib/executor/helper.cpp
```diff
@@ -496,12 +504,16 @@ ValVariant Executor::unpackVal(const ValType &Type, const ValVariant &Val,
...
+      if constexpr (Endian::native == Endian::big)
+        Num >>= 24;
       if (IsSigned) {
         return static_cast<uint32_t>(static_cast<int8_t>(Num));
       } else {
         return static_cast<uint32_t>(static_cast<uint8_t>(Num));
       }
     case TypeCode::I16:
+      if constexpr (Endian::native == Endian::big)
+        Num >>= 16;
```
hydai: Ditto

lib/executor/helper.cpp
```diff
@@ -496,12 +504,16 @@ ValVariant Executor::unpackVal(const ValType &Type, const ValVariant &Val,
     uint32_t Num = Val.get<uint32_t>();
     switch (Type.getCode()) {
     case TypeCode::I8:
+      if constexpr (Endian::native == Endian::big)
+        Num >>= 24;
```
hydai: Ditto

lib/executor/engine/engine.cpp
```diff
@@ -1777,7 +1778,14 @@ Expect<void> Executor::execute(Runtime::StackManager &StackMgr,
     case OpCode::I8x16__relaxed_swizzle: {
       const ValVariant Val2 = StackMgr.pop();
       ValVariant &Val1 = StackMgr.getTop();
-      const uint8x16_t &Index = Val2.get<uint8x16_t>();
+      uint8x16_t Index = Val2.get<uint8x16_t>();
+      if constexpr (Endian::native == Endian::big) {
+#if defined(_MSC_VER) && !defined(__clang__)
+        std::for_each(Index.begin(), Index.end(), [](auto &i) { i = 15 - i; });
```
hydai: `i` -> `I`

include/runtime/instance/memory.h
```diff
@@ -16,15 +16,18 @@
...
 #include "common/errcode.h"
 #include "common/errinfo.h"
+#include "common/int128.h"
 #include "common/spdlog.h"
 #include "system/allocator.h"
 
 #include <algorithm>
 #include <cstdint>
 #include <cstring>
+#include <fmt/base.h>
```
hydai: What is the reason for including this new header?

hydai (changes requested):

lib/loader/ast/instruction.cpp
vishruth-thimmaiah: When we read the value in `engine.cpp` using `getNum()`, we aren't converting it back to uint64/uint32. This wasn't a problem on little endian, but on big endian, it returns the high bits, which are usually garbage values.

Also, the original code did not convert `float`/`double` to `uin128_t`, so I decided to change the code here to copy the same behavior.

hydai:
Please fix the compilation errors. Since executing the workflows within the WasmEdge repository requires approval, you may need to create a PR in your own forked repository to trigger most of the workflows.

test/host/socket/wasi_socket.cpp
vishruth-thimmaiah: I added this function to reduce the need to define extra variables in the main testsuite, as we call this function ~40 times

vishruth-thimmaiah:
The workflow errors should now be fixed.

test/host/socket/wasi_socket.cpp
hydai: Most of them are something like this: `*MemInst.getPointer<const uint32_t *>(0)`. It avoids calling a template `getValue` function and reduces the calling frame cost. I didn't find the `extra` variables. Instead of your new method, it will always create a new `T Value` when calling this function.

hydai:
Hi @vishruth-thimmaiah 
Please rebase to the latest master branch before we merge it.

Also, I would like to invite you to present your work at the WasmEdge community meeting on September 2nd. Please refer to [this page](https://github.com/WasmEdge/WasmEdge/issues/4335) for details.  
Are you available at that time?

vishruth-thimmaiah:
> Also, I would like to invite you to present your work at the WasmEdge community meeting on September 2nd. Please refer to [this page](https://github.com/WasmEdge/WasmEdge/issues/4335) for details.
> Are you available at that time?

Yes, I'm available and would be happy to present.

hydai:
> Yes, I'm available and would be happy to present.

Awesome, could you please leave a comment under the community meeting issue to add your agenda? You will have 15 minutes to present your work. Thanks.

hydai:
Hi @vishruth-thimmaiah,  
Since I am merging another mentee's PR first, you will need to rebase your branch again. I am sorry about that.

hydai (approved):

hydai:
Thank you, @vishruth-thimmaiah, for your wonderful work.

PR #4255, opened 2025-07-10
ci(release): fixed the incorrect GitHub permissions

hydai:
Unfortunately, we introduce a regression for the release process.

CleanShot 2025-07-10 at 15 03 21

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -51,6 +51,8 @@ jobs:
 
   create_source_tarball:
     needs: create_release
+    permissions:
+      contents: write
     uses: ./.github/workflows/reusable-create-source-tarball.yml
     with:
       version: ${{ needs.create_release.outputs.version }}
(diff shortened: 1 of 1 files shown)


juntao:

------

## [.github/workflows/release.yml](https://github.com/WasmEdge/WasmEdge/blob/8acb5cd2d29cf46fb7951f9c046c77ad38711454/.github%2Fworkflows%2Frelease.yml)

#### Potential issues

1. The `push` event's `tags` pattern is incorrect and will not match semantic version tags as intended.
2. In the `create_release` job, the `gh release create` command is using `${{ secrets.GITHUB_TOKEN }}`, which should be accessed directly without wrapping it in double quotes.
3. The `generate_sbom` job's `GITHUB_TOKEN` environment variable is not properly escaped for use with GitHub CLI commands.

#### Summary of changes

Here are the key changes in the patch:

- **Added write permissions for `contents`** to all jobs that require it. This is necessary for creating and uploading artifacts, such as source tarballs and binaries.
  - Added to the following jobs: `create_source_tarball`, `build_on_macos`, `build_on_ubuntu_20_04`, `build_on_manylinux_2_28`, `build_on_debian_static`, `build_on_alpine_static`, `build_on_windows`, `build_on_windows_msvc`, `build_on_android`, and `Build_and_Upload`.

These changes ensure that each job has the necessary permissions to write contents, which is likely required for creating and uploading release artifacts.

PR #4083, opened 2025-04-16
WIP [CI] linelint: add EOF newline check workflow and configuration

varun-r-mallya:
tentatively fixes #3519
I am opening this PR right now just to check if the CI works well. Please keep in mind that this is not the final PR

PR #4256, opened 2025-07-10
WIP: ci(linters): add EOF newline check via linelint to misc CI workflow

ShigrafS:
### 📌 Summary

This PR adds support for **EOF (End Of File) newline enforcement** using `linelint`, integrated into the existing CI via the `misc-linters.yml` workflow. The goal is to detect and prevent missing trailing newlines in modified files during pull requests, ensuring consistency and adherence to best practices.

Closes: #3519
Related earlier PRs: #3558, #3559, #3560, #3592, #4083

---

#### `.github/workflows/misc-linters.yml`

* **New CI job: `linelint`**

  * Uses `@linelint/cli` installed via `npm` (instead of the GitHub Action wrapper) to allow argument-based control.
  * Automatically detects files changed in the PR using `git diff`.
  * Skips third-party or irrelevant directories (`thirdparty/`, `dist/`, `.git/`).
  * Runs linelint only on changed files (as required by previous feedback).
  * Fails CI if any file is missing an EOF newline.

#### `.linelint.yml`

* Configures `linelint` rules:

  * Enables the `end-of-file` rule with `single-newline: true` to enforce exactly one newline at EOF.
  * `autofix: true` and `disable-autofix: false` allow local autofixing (for contributor convenience), but CI does not use autofix to ensure integrity and security.
  * Ignores irrelevant directories (`.git/`, `thirdparty/`, `dist/`).

---

### ⚠️ Design Decisions & Feedback Considered

* ✅ **Avoided use of GitHub Action `fernandrone/linelint@master`** due to:

  * Lack of support for passing file paths directly
  * Instability of using `@master` (as flagged in reviews)
* ✅ **CI permissions limited** (`contents: read`) for security (per @hydai's feedback).
* ✅ **CI fails on lint errors** but does not autofix them — this balances safety and enforcement.
* ✅ **One-liner fix possible locally**: contributors can run `linelint --fix` using the provided config to resolve issues quickly.
* ✅ **Uses modern `checkout@v4`** instead of the legacy `v2`.

---

### 📎 Follow-up Suggestions

* Optionally, we can include a pre-commit hook in the future for `linelint`.

### DIFF
--- a/.github/workflows/misc-linters.yml
+++ b/.github/workflows/misc-linters.yml
@@ -33,3 +33,37 @@ jobs:
           pip install codespell
           # exclude files which may be synchronized from other places
           git ls-files | grep -v "^thirdparty" | grep -v "/thirdparty/" | grep -v "/dist/" | xargs -t codespell --ignore-words .github/workflows/ignore_words 2>/dev/null
+  lineguard:
+    name: lineguard
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - name: Harden the runner (Audit all outbound calls)
+        uses: step-security/harden-runner@6c439dc8bdf85cadbbce9ed30d1c7b959517bc49 # v2.12.2
+        with:
+          egress-policy: audit
+
+      - name: Checkout code
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          fetch-depth: 0
+      - name: Ensure git safe directory
+        run: |
+          git config --global --add safe.directory $(pwd)
+      - name: Install lineguard
+        run: |
+          curl -L https://github.com/hydai/lineguard/releases/download/v0.1.4/lineguard-v0.1.4-x86_64-unknown-linux-gnu.tar.gz | tar xz
+          sudo mv lineguard /usr/local/bin/
+      - name: Run lineguard
+        run: |
+          BASE_REF=${{ github.base_ref }}
+          if git rev-parse --verify origin/$BASE_REF >/dev/null 2>&1; then
+            FROM_COMMIT=$(git rev-parse origin/$BASE_REF)
+            TO_COMMIT=$(git rev-parse HEAD)
+            lineguard --config .lineguardrc --fix --verbose --from $FROM_COMMIT --to $TO_COMMIT || { echo "Lineguard found issues"; exit 1; }
+          else
+            FROM_COMMIT=$(git rev-parse HEAD~1)
+            TO_COMMIT=$(git rev-parse HEAD)
+            lineguard --config .lineguardrc --fix --verbose --from $FROM_COMMIT --to $TO_COMMIT || { echo "Lineguard found issues"; exit 1; }
+          fi
--- a/.lineguardrc
+++ b/.lineguardrc
@@ -1,0 +1,8 @@
+file_extensions = ["rs", "toml", "md", "yml", "yaml", "cpp", "hpp", "c", "h", "sh", "ipp", "cmake", "txt", "py", "wxs", "spdx"]
+ignore_patterns = [
+    "**/.git/**",
+    "thirdparty/",
+]
+[checks]
+newline_ending = true
+trailing_spaces = true


juntao:

------

## [.github/workflows/misc-linters.yml](https://github.com/WasmEdge/WasmEdge/blob/a04aded2ad444b6833fa0adf2233e52d0d6197f8/.github%2Fworkflows%2Fmisc-linters.yml)

#### Potential issues

1. The `checkout` step in the `lineguard` job is missing the `fetch-depth: 0` option, which might cause issues with fetching complete history needed for `lineguard`.

2. In both jobs (`misc` and `lineguard`), the use of `$(pwd)` in the `Ensure git safe directory` steps is not portable across different operating systems (e.g., Windows).

3. The `codespell` installation step in the `check spell` job does not specify a version, which could lead to unexpected behavior if the default version changes.

#### Summary of changes

- Added a new GitHub Actions job called `lineguard` to check for EOF newline issues.
  - The job installs and runs the `lineguard` tool, which scans the code for missing end-of-file newlines between two specified commits.
  - Implemented a conditional logic to determine the `FROM_COMMIT` based on whether the base branch exists or not.

## [.lineguardrc](https://github.com/WasmEdge/WasmEdge/blob/a04aded2ad444b6833fa0adf2233e52d0d6197f8/.lineguardrc)

#### Potential issues

1. The `file_extensions` list should be sorted alphabetically to maintain consistency and readability.
2. The `ignore_patterns` list uses a trailing comma which is unnecessary and should be removed for better code quality.
3. The `trailing_spaces` check should be explicitly enabled in the `checks` section as it might not be automatically enabled by default.

#### Summary of changes

- **Added new file extensions to lint**: The patch introduces a list of 15 file extensions to be checked by the linter, including popular languages and formats like Rust (rs), TOML (toml), Markdown (md), YAML (yml/yaml), C++ (cpp/hpp), C (c/h), shell scripts (sh), and more.

- **Enabled newline ending check**: The configuration now ensures that all checked files have a newline at the end, promoting consistency in file formatting. This change is reflected by the `newline_ending = true` setting under `[checks]`.

- **Enabled trailing spaces check**: The patch also introduces a check for trailing spaces, which can help maintain clean and consistent code style across the project. This is enabled with the `trailing_spaces = true` setting.

ShigrafS:
@hydai 

I pushed a second commit that introduced an error and encountered this failure in the workflow:

```
3s
Run npm install -g @linelint/cli
npm error code E404
npm error 404 Not Found - GET https://registry.npmjs.org/@linelint%2fcli - Not found
npm error 404
npm error 404  '@linelint/cli@*' is not in this registry.
npm error 404
npm error 404 Note that you can also install from a
npm error 404 tarball, folder, http url, or git url.
npm error A complete log of this run can be found in: /home/runner/.npm/_logs/2025-07-10T15_43_46_984Z-debug-0.log
Error: Process completed with exit code 1.
```

For reference: [https://github.com/ShigrafS/WasmEdge/actions/runs/16199690580/job/45735288129](https://github.com/ShigrafS/WasmEdge/actions/runs/16199690580/job/45735288129)

Given this, we have two main options:

* Use the built-in GitHub Action (which runs linelint on **all files** instead of just changed ones):

```yaml
- name: Run linelint
  uses: actions/linelint@master
```

* Or install and run the `linelint` CLI via Go for more control and potentially only on changed files:

```yaml
- name: Install linelint
  run: |
    go install github.com/fernandrone/linelint@latest
    echo "$HOME/go/bin" >> $GITHUB_PATH
```

Please let me know which approach you prefer or if you have other suggestions!

test-linelint.txt
hydai, line 1: Is this a misaddition into this PR?

.github/workflows/misc-linters.yml
```diff
@@ -33,3 +33,44 @@ jobs:
...
+          if git rev-parse --verify origin/$BASE_REF >/dev/null 2>&1; then
+            FILES=$(git diff --name-only --diff-filter=d origin/$BASE_REF...HEAD | grep -v "^thirdparty/" | grep -v "/dist/" | tr '\n' ' ')
+          else
+            FILES=$(git diff --name-only --diff-filter=d HEAD~1...HEAD | grep -v "^thirdparty/" | grep -v "/dist/" | tr '\n' ' ')
+          fi
+          if [ -z "$FILES" ]; then
+            echo "No files changed, skipping linelint."
+            exit 0
+          fi
+          echo "files=$FILES" >> $GITHUB_OUTPUT
```
hydai: Could you please explain the logic you are using here? For example, why does the `else` condition only select `HEAD~1`?

.github/workflows/misc-linters.yml
```diff
@@ -33,3 +33,44 @@ jobs:
...
+      - name: Install linelint
+        run: |
+          npm install -g @linelint/cli
+      - name: Check for EOF newlines
+        run: |
+          # Check each file individually to handle spaces in filenames
+          for file in ${{ steps.changed-files.outputs.files }}; do
+            linelint "$file" || { echo "EOF lint error in $file"; exit 1; }
```
hydai: Can this tool be applied to multiple files at the same time?

.github/workflows/misc-linters.yml
```diff
@@ -33,3 +33,44 @@ jobs:
...
+          done
```
hydai: EOF

.linelint.yml
```diff
@@ -0,0 +1,10 @@
+autofix: true 
+ignore:
+  - .git/
+  - thirdparty/
+  - dist/
```
hydai: If this rule can be applied to the tool, why do we need to filter the changed files with this: `grep -v "^thirdparty/" | grep -v "/dist/"`?

hydai (changes requested):

.github/workflows/misc-linters.yml
ShigrafS: Primarily, we would be using `git diff origin/$BASE_REF...HEAD` to identify the files changed in the pull request.
The `else` case is meant for cases which are non-PR based. Like a direct push. This ensures that it works for changes within a single commit.

.github/workflows/misc-linters.yml
ShigrafS: Yes, `linelint` can be applied to multiple files at once. 
However running it one file at a time has two benefits: 
- To handle filenames with spaces safely.
- To show exactly which file failed, with a clear message per file.

.linelint.yml
ShigrafS: It’s a small performance optimization and helps keep the CI logs cleaner.

The extra `grep -v` in the script just filters them out earlier. This way, we avoid calling the linter on files that will be ignored anyway.

test-linelint.txt
ShigrafS: This file with errors was pushed along with the second commit. 
The purpose was to trigger `linelint` to check if it would work on new pushed commits. 

This file would be removed before this PR is merged.

ShigrafS:
> Given this, we have two main options:

> - Use the built-in GitHub Action (which runs linelint on all files instead of just changed ones):
> - Or install and run the linelint CLI via Go for more control and potentially only on changed files:

@hydai From what I understood from the previous unmerged PRs, and the link that the issue referenced, `@linelint/cli` is Go based and not npm based. 
The earlier PRs #3592 and #4083 used GitHub actions, but they run across the entire repo, not just the changed files, which reviewers previously flagged as a problem. 
This leads to two options for us. Both of which comes with their own tradeoffs. 

Please let me know which approach would you prefer, or should we explore other options as well. 
Thanks!

hydai:
Well, I am building our own version of LineLint to suit our needs.

Here is the repo: https://github.com/hydai/lineguard  
It's not ready yet; I am still fixing some issues with the Git integration.

You may try it to see if it meets the requirements we are looking for.

hydai:
I think this is ready to be used. You can also add the `.lineguardrc` file to configure the default behavior: https://github.com/hydai/lineguard/blob/master/.lineguardrc.example

.github/workflows/misc-linters.yml
```diff
@@ -33,3 +33,41 @@ jobs:
...
+          if git rev-parse --verify origin/$BASE_REF >/dev/null 2>&1; then
+            FILES=$(git diff --name-only --diff-filter=d origin/$BASE_REF...HEAD | tr '\n' ' ')
+          else
+            FILES=$(git diff --name-only --diff-filter=d HEAD~1...HEAD | tr '\n' ' ')
+          fi
+          if [ -z "$FILES" ]; then
+            echo "No files changed, skipping lineguard."
+            exit 0
+          fi
+          echo "files=$FILES" >> $GITHUB_OUTPUT
```
hydai: You don't need this part. The lineguard can use `--from <commit_hash>` and `--to <commit_hash>` to filter those changed files between these two commits.

.github/workflows/misc-linters.yml
```diff
@@ -33,3 +33,41 @@ jobs:
...
+      - name: Install lineguard
+        run: |
+          curl -L https://github.com/hydai/lineguard/releases/latest/download/lineguard-latest-linux-x86_64.tar.gz | tar xz
+          sudo mv lineguard /usr/local/bin/
+      - name: Run lineguard
+        run: |
+          lineguard --config .lineguardrc --fix ${{ steps.changed-files.outputs.files }} || { echo "Lineguard found issues"; exit 1; }
```
hydai: Since we won't apply `fix` automatically for a PR, we don't need to call `--fix` for it.

.lineguardrc
```diff
@@ -0,0 +1,14 @@
+file_extensions = ["rs", "toml", "md", "yml", "yaml"]
```
hydai: You should also include C++-related extensions here.

.lineguardrc
```diff
@@ -0,0 +1,14 @@
...
+    "dist/",
+    "*.bak",
+    "*.tmp"
+]
+[checks]
+newline_ending = true
+trailing_spaces = true
+autofix = true
+[rules]
+end-of-file = { enable = true, single-newline = true, disable-autofix = false }
```
hydai: I am not sure, but why do these sections work?

https://github.com/hydai/lineguard/blob/master/.lineguardrc.example#L14-L17

.github/workflows/misc-linters.yml
```diff
@@ -33,3 +33,32 @@ jobs:
...
+        run: |
+          curl -L https://github.com/hydai/lineguard/releases/download/v0.1.3/lineguard-v0.1.3-x86_64-unknown-linux-gnu.tar.gz | tar xz
+          sudo mv lineguard /usr/local/bin/
+      - name: Run lineguard
+        run: |
+          BASE_REF=${{ github.base_ref }}
+          if git rev-parse --verify origin/$BASE_REF >/dev/null 2>&1; then
+            lineguard --config .lineguardrc --fix --from origin/$BASE_REF --to HEAD || { echo "Lineguard found issues"; exit 1; }
+          else
+            lineguard --config .lineguardrc --fix --from HEAD~1 --to HEAD || { echo "Lineguard found issues"; exit 1; }
```
hydai: You should use `git rev-parse` to obtain the commit hash and use it for the lineguard.

.lineguardrc
```diff
@@ -0,0 +1,12 @@
...
+    "**/.git/**",
+    "thirdparty/",
+    "dist/",
+    "*.bak",
+    "*.tmp"
+]
+[checks]
+newline_ending = true
+trailing_spaces = true
+autofix = true
```
hydai: Lineguard doesn't provide this option. Could you check the document again?

.lineguardrc
```diff
@@ -0,0 +1,12 @@
+file_extensions = ["rs", "toml", "md", "yml", "yaml", "cpp", "hpp", "c", "h"]
```
hydai: Add `sh`, `ipp`, `cmake`, `txt`, `py`, `wxs`, `spdx`

hydai (changes requested):

.lineguardrc
ShigrafS: Thanks for catching this. 
It was an oversight.

.github/workflows/misc-linters.yml
```diff
@@ -33,3 +33,35 @@ jobs:
...
+      - name: Install lineguard
+        run: |
+          curl -L https://github.com/hydai/lineguard/releases/download/v0.1.3/lineguard-v0.1.3-x86_64-unknown-linux-gnu.tar.gz | tar xz
+          sudo mv lineguard /usr/local/bin/
+      - name: Run lineguard
+        run: |
+          BASE_REF=${{ github.base_ref }}
+          if git rev-parse --verify origin/$BASE_REF >/dev/null 2>&1; then
+            FROM_COMMIT=$(git rev-parse origin/$BASE_REF)
+            lineguard --config .lineguardrc --fix --from $FROM_COMMIT --to HEAD || { echo "Lineguard found issues"; exit 1; }
```
hydai: `to` should be `git rev-parse HEAD`

.lineguardrc
```diff
@@ -0,0 +1,11 @@
...
+ignore_patterns = [
+    "**/.git/**",
+    "thirdparty/",
+    "dist/",
+    "*.bak",
+    "*.tmp"
+]
+[checks]
+newline_ending = true
+trailing_spaces = true
```
hydai: EOF

.github/workflows/misc-linters.yml
```diff
@@ -33,3 +33,35 @@ jobs:
...
+          else
+            FROM_COMMIT=$(git rev-parse HEAD~1)
+            lineguard --config .lineguardrc --fix --from $FROM_COMMIT --to HEAD || { echo "Lineguard found issues"; exit 1; }
+          fi
```
hydai: EOF

.github/workflows/misc-linters.yml
hydai: And we should also apply `--verbose` for detailed information.

hydai:
I have a new idea, since I found that we can actually use a bash script to compose a simple command to handle these cases.  
Please refer to this section: https://github.com/hydai/lineguard/blob/master/README.md#shell-alternatives

However, using a tool with a configuration file may reduce the maintenance cost.  
WDYT?

cc @0yi0

ShigrafS:
I think using a bash script would be a good lightweight option to keep things simple for now. 

However, considering potential future needs and the risk of growing complexity, I’d suggest using `lineguard` with a configuration file. It provides a more scalable and maintainable solution, and would allow us to handle more complex scenarios down the line without requiring significant changes to WasmEdge repo.

hydai:
It may be pretty easy like this:

I run the command under lineguard

```bash
FROM=v0.1.0
TO=HEAD
git diff --name-only $FROM $TO | while read f; do
    [ -f "$f" ] && {
        grep -q '[[:space:]]$' "$f" && echo "$f: trailing spaces"
        [ -n "$(tail -c 1 "$f")" ] && echo "$f: missing newline"
    }
done
.github/workflows/benchmark.yml: trailing spaces
.github/workflows/benchmark.yml: missing newline
.lineguardrc.example: missing newline
CLAUDE.md: missing newline
Makefile: missing newline
README.md: trailing spaces
README.md: missing newline
docs/planning/GITHUB_WORKFLOWS_PLAN.md: trailing spaces
docs/planning/GITHUB_WORKFLOWS_PLAN.md: missing newline
docs/planning/IMPLEMENTATION_PLAN.md: trailing spaces
docs/planning/IMPLEMENTATION_PLAN.md: missing newline
docs/user/CLI_REFERENCE.md: missing newline
```

ShigrafS:
This seems easy to implement and effective, though it may be less scalable in the long run.
I'm happy to go with whatever the team prefers!

hydai:
Since we will need the configuration file at this stage, let's keep the lineguard until others have a second thought.

.lineguardrc
```diff
@@ -0,0 +1,11 @@
+file_extensions = ["rs", "toml", "md", "yml", "yaml", "cpp", "hpp", "c", "h", "sh", "ipp", "cmake", "txt", "py", "wxs", "spdx"]
+ignore_patterns = [
+    "**/.git/**",
+    "thirdparty/",
+    "dist/",
+    "*.bak",
+    "*.tmp"
```
hydai: I am not sure, but do we have these files in this repository?

hydai (changes requested):

.lineguardrc
ShigrafS: No, we don't. 
I just placed them here in case we need them in future, as these are quite commonly used. 
Should I remove them?

hydai: Yes, please. I believe these are just from the example config.

test-linelint.txt
PhantomInTheWire, line 1: is this file here just for debugging on your fork or does it have a better reason to be here?

ShigrafS: It's meant for testing. 
Would be removed before the PR is merged.

hydai:
Please squash all your commits and remove the redundant files.

ShigrafS:
@hydai 
Commits squashed and redundant files removed.

hydai (approved):

PR #4261, opened 2025-07-11
bug: uninstaller not working when the .bash_profile is not a regular file.

### LINKED ISSUE #3185: bug: uninstaller not working when the `.bash_profile` is not a regular file.
### Summary

Please check this issue:
https://github.com/second-state/LlamaEdge/issues/100

Someone is running the `installer`. However, it will run the uninstaller first to remove the previous installation. Then, it encountered the bug.

### Current State

`sed: /Users/user/.bash_profile: in-place editing only works for regular files`

I believe the error is caused by https://github.com/WasmEdge/WasmEdge/blob/master/utils/uninstall.sh#L254-L255

### Expected State

Everything is fine.

### Reproduction steps

Run the uninstaller on a machine which `.bash_profile` is not a regular file.

### Screenshots

DESCRIPTION

_No response_

### Components

Others

master

macOS 14.2.1

### Hardware Architecture

x86_64

none.

0x-avi:
this pr fixes #3185  where the BSD grep used by MacOs was causing the uninstaller to fail to remove environment variables.

> [34601ad](https://github.com/WasmEdge/WasmEdge/commit/34601adb71e205e8a4cee03eb11de08a495ac0e4) adds the fix and also removes the repetitive code using a simple for loop over all possible configs
> [86295c9](https://github.com/WasmEdge/WasmEdge/commit/86295c9f126fbcf0ce78d7fe487413816705b975) adds a test to check if the workflow correctly removes the environment variable
the code has been tested locally once and also tested on a dummy commit on github workflows and passes all checks.

### DIFF
--- a/.github/workflows/build_for_nix.yml
+++ b/.github/workflows/build_for_nix.yml
@@ -42,7 +42,7 @@ jobs:
         egress-policy: audit
 
     - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
-    - uses: cachix/install-nix-action@f0fe604f8a612776892427721526b4c7cfb23aba # v31.4.1
+    - uses: cachix/install-nix-action@c134e4c9e34bac6cab09cf239815f9339aaaf84e # v31.5.1
       with:
         nix_path: nixpkgs=channel:nixos-24.05
     - run: nix build
--- a/.github/workflows/misc-linters.yml
+++ b/.github/workflows/misc-linters.yml
@@ -33,3 +33,37 @@ jobs:
           pip install codespell
           # exclude files which may be synchronized from other places
           git ls-files | grep -v "^thirdparty" | grep -v "/thirdparty/" | grep -v "/dist/" | xargs -t codespell --ignore-words .github/workflows/ignore_words 2>/dev/null
+  lineguard:
+    name: lineguard
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - name: Harden the runner (Audit all outbound calls)
+        uses: step-security/harden-runner@6c439dc8bdf85cadbbce9ed30d1c7b959517bc49 # v2.12.2
+        with:
+          egress-policy: audit
+
+      - name: Checkout code
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          fetch-depth: 0
+      - name: Ensure git safe directory
+        run: |
+          git config --global --add safe.directory $(pwd)
+      - name: Install lineguard
+        run: |
+          curl -L https://github.com/hydai/lineguard/releases/download/v0.1.4/lineguard-v0.1.4-x86_64-unknown-linux-gnu.tar.gz | tar xz
+          sudo mv lineguard /usr/local/bin/
+      - name: Run lineguard
+        run: |
+          BASE_REF=${{ github.base_ref }}
+          if git rev-parse --verify origin/$BASE_REF >/dev/null 2>&1; then
+            FROM_COMMIT=$(git rev-parse origin/$BASE_REF)
+            TO_COMMIT=$(git rev-parse HEAD)
+            lineguard --config .lineguardrc --fix --verbose --from $FROM_COMMIT --to $TO_COMMIT || { echo "Lineguard found issues"; exit 1; }
+          else
+            FROM_COMMIT=$(git rev-parse HEAD~1)
+            TO_COMMIT=$(git rev-parse HEAD)
+            lineguard --config .lineguardrc --fix --verbose --from $FROM_COMMIT --to $TO_COMMIT || { echo "Lineguard found issues"; exit 1; }
+          fi
--- a/.lineguardrc
+++ b/.lineguardrc
@@ -1,0 +1,8 @@
+file_extensions = ["rs", "toml", "md", "yml", "yaml", "cpp", "hpp", "c", "h", "sh", "ipp", "cmake", "txt", "py", "wxs", "spdx"]
+ignore_patterns = [
+    "**/.git/**",
+    "thirdparty/",
+]
+[checks]
+newline_ending = true
+trailing_spaces = true
--- a/cmake/WASINNDeps.cmake
+++ b/cmake/WASINNDeps.cmake
@@ -343,7 +343,7 @@ function(wasmedge_setup_llama_target target)
     FetchContent_Declare(
       llama
       GIT_REPOSITORY https://github.com/ggml-org/llama.cpp.git
-      GIT_TAG        b5835
+      GIT_TAG        b5896
       GIT_SHALLOW    FALSE
     )
     FetchContent_MakeAvailable(llama)
--- a/plugins/wasi_nn/CMakeLists.txt
+++ b/plugins/wasi_nn/CMakeLists.txt
@@ -23,7 +23,7 @@ wasmedge_add_library(wasmedgePluginWasiNN
 include(WASINNDeps)
 wasmedge_setup_wasinn_target(wasmedgePluginWasiNN PLUGINLIB)
 
-set(WASMEDGE_WASI_NN_VERSION "0.1.27" CACHE STRING "WasmEdge WASI-NN library version")
+set(WASMEDGE_WASI_NN_VERSION "0.1.28" CACHE STRING "WasmEdge WASI-NN library version")
 set(WASMEDGE_WASI_NN_SOVERSION "0" CACHE STRING "WasmEdge WASI-NN library soversion")
 
 # Handle the version of the WASI-NN plugin


juntao:

------

## [.github/workflows/test-installers.yml](https://github.com/WasmEdge/WasmEdge/blob/281c3d60ff3f679bae301bbc5aca06558005602c/.github%2Fworkflows%2Ftest-installers.yml)

#### Potential issues

1. **Uninstaller Check for `.bash_profile`**: In the `test-install-sh-macos`, `test-install-v2-sh-macos`, and `test-install-py-macos` jobs, the uninstallation test does not account for scenarios where `.bash_profile` is not a regular file, which could lead to false negatives.

2. **Version Mismatch Handling**: The script does not handle version mismatches properly in the `test-install-v2-sh-macos` job when verifying installed versions against expected ones, potentially causing the tests to fail incorrectly.

3. **Shell Script Dependency on `.bashrc`**: In `test-install-py`, there is an implicit assumption that `.bashrc` exists and is sourced correctly, which may not always be the case in all environments, leading to potential failures during verification steps.

#### Summary of changes

Here are the top three most important changes in the patch:

- **Added dummy .bash_profile creation**: Before running uninstallation tests, a check is added to create an empty `.bash_profile` if it doesn't exist. This ensures that the uninstaller has a valid file to process, even if the user's system didn't have one initially.

  ```diff
  [ -f ~/.bash_profile ] || echo > ~/.bash_profile
  ```

- **Added post-uninstallation verification**: After running the uninstall script, the patch adds a check to verify that the WasmEdge source line is not present in either `.zshrc` or `.bash_profile`. If the line is found, it indicates that the uninstallation failed.

  ```diff
  for file in "$HOME/.zshrc" "$HOME/.bash_profile"; do
    [ -f "$file" ] || continue
    if grep -q ". \"$IPATH/env\"" "$file"; then
      echo "✗ Found WasmEdge source line in: $file"
      echo "✗ Uninstallation failed"
      exit 1
    fi
  done
  ```

- **Renamed job ID for clarity**: The test job ID `test-install-v2-macos` was renamed to `test-install-v2-sh-macos` to better reflect its purpose.

  ```diff
  -  test-install-v2-macos:
  +  test-install-v2-sh-macos:
  ```

These changes address the bug by ensuring that the uninstaller works correctly even when the `.bash_profile` is not present initially, and adds verification to confirm successful uninstallation.

## [utils/uninstall.sh](https://github.com/WasmEdge/WasmEdge/blob/281c3d60ff3f679bae301bbc5aca06558005602c/utils%2Funinstall.sh)

#### Potential issues

1. **Potential Issue with Permissions Check**: The script assumes that the absence of `sudo` command means it's running without root permissions, which might not always be true in non-standard environments.

2. **Hardcoded File and Directory Checks**: The script uses specific file names like `.bash_profile`, `.profile`, etc., which may not exist or have different names in various environments, potentially leading to incomplete uninstallation.

3. **Potential Infinite Loop in `ask_remove`**: If the user input is not 'Y', 'N', or any of their uppercase counterparts, it asks again in an infinite loop without breaking.

#### Summary of changes

**Key changes:**

- **Refactoring of file checking and modification logic**: The patch refactors the code to use a loop for checking and modifying `.bash_profile`, `.profile`, and shell rc files, making it more concise and easier to maintain.
- **Improved error handling**: Adds `continue` statements to skip non-existent files and checks if `line_num` is non-empty before proceeding with modifications. This prevents errors when the file or line is not found.
- **Backups before modification**: Creates a backup of each file before modifying it, ensuring that users can revert changes if needed.

0x-avi:
@SAtacker please review when you get time , also i have signed my commits after i saw the fail but i have verified they are signed how can i rerun that workflow
edit: figured it out

0x-avi:
@hydai if you get time could you review please? thanks in advance

.github/workflows/test-installers.yml
```diff
@@ -181,6 +181,9 @@ jobs:
         # Source helper functions
         . ./.github/scripts/installer-test-helpers.sh
 
+        # dummy .bashprofile creation to check uninstallation
+        [ -f ~/.bash_profile ] || echo > ~/.bash_profile
+
```
hydai, line 186: You are creating a regular file.

.github/workflows/test-installers.yml
```diff
@@ -191,6 +194,16 @@ jobs:
         source ~/.wasmedge/env 2>/dev/null || source ~/.bashrc 2>/dev/null || true
         verify_version "${LATEST_VERSION}"
         bash utils/uninstall.sh -q -V
+        # Test if the environment variable was removed
+        IPATH="$HOME/.wasmedge"
+        for file in "$HOME/.zshrc" "$HOME/.bash_profile"; do
+        [ -f "$file" ] || continue
+        if grep -q ". \"$IPATH/env\"" "$file"; then
+          echo "❌ Found WasmEdge source line in: $file"
+          echo "❌ Uninstallation failed"
```
hydai: Please follow the same error message style.

.github/workflows/test-installers.yml
```diff
@@ -274,7 +287,7 @@ jobs:
         bash utils/uninstall.sh -q -V
 
   # Test install_v2.sh on macOS
-  test-install-v2-macos:
+  test-install-v2-sh-macos:
```
hydai, line 290: I don't think we need to update this. The Installer v2 is script-based.

utils/uninstall.sh
```diff
@@ -247,12 +247,34 @@ main() {
...
-    [[ -f "${__HOME__}/.bash_profile" ]] && line_num="$(grep -n ". \"${IPATH}/env\"" "${__HOME__}/.bash_profile" | cut -d : -f 1)" &&
-        [[ "$line_num" != "" ]] && sed -i.wasmedge_backup -e "${line_num}"'d' "${__HOME__}/.bash_profile"
+    resolve_path() {
+      local file="$1"
+      if command -v realpath >/dev/null 2>&1; then
+        realpath "$file"
+      elif command -v readlink >/dev/null 2>&1; then
+        readlink "$file"
+      elif command -v python3 >/dev/null 2>&1; then
+        python3 -c 'import os,sys; print(os.path.realpath(sys.argv[1]))' "$file"
```
hydai: We should never introduce any dependencies in the uninstaller.

utils/uninstall.sh
```diff
@@ -247,12 +247,34 @@ main() {
     _shell_="${SHELL#${SHELL%/*}/}"
     _shell_rc=".""$_shell_""rc"
 
-    [[ -f "${__HOME__}/${_shell_rc}" ]] && line_num="$(grep -n ". \"${IPATH}/env\"" "${__HOME__}/${_shell_rc}" | cut -d : -f 1)" &&
-        [ "$line_num" != "" ] && sed -i.wasmedge_backup -e "${line_num}"'d' "${__HOME__}/${_shell_rc}"
-    [[ -f "${__HOME__}/.profile" ]] && line_num="$(grep -n ". \"${IPATH}/env\"" "${__HOME__}/.profile" | cut -d : -f 1)" &&
-        [[ "$line_num" != "" ]] && sed -i.wasmedge_backup -e "${line_num}"'d' "${__HOME__}/.profile"
-    [[ -f "${__HOME__}/.bash_profile" ]] && line_num="$(grep -n ". \"${IPATH}/env\"" "${__HOME__}/.bash_profile" | cut -d : -f 1)" &&
-        [[ "$line_num" != "" ]] && sed -i.wasmedge_backup -e "${line_num}"'d' "${__HOME__}/.bash_profile"
+    resolve_path() {
```
hydai, line 250: Why do we need this?

hydai (changes requested):

.github/workflows/test-installers.yml
0x-avi: does it need to be a symlink?

utils/uninstall.sh
0x-avi: because the code was repeated thrice i thought making a loop made more sense, should i revert to the original way of doing things? , if we are talking about why im using a function i thought it would be cleaner, if the question is why i used realpath or readlink, i dont know any other way of finding the actual path from a symlink

.github/workflows/test-installers.yml
hydai: Please refer to the original issue. It's a symlink.

https://github.com/LlamaEdge/LlamaEdge/issues/100#issuecomment-1918437450

0x-avi: sorry about that , created a dummy file and then symlinked to that

utils/uninstall.sh
0x-avi: @hydai should i revert it?

.github/workflows/test-installers.yml
```diff
@@ -8,6 +8,7 @@ on:
   push:
     branches:
       - master
+      - uninstaller-fix
```
PhantomInTheWire: I think this line was likely added for testing on your fork. If that’s the case, you should probably remove it.

0x-avi: Hey you are right! Will remove

0x-avi:
@PhantomInTheWire good to go!

.github/workflows/test-installers.yml
```diff
@@ -181,6 +181,15 @@ jobs:
...
 
+        # Backup contents of existing .bash_profile (if any)
+        [ -e ~/.bash_profile ] && cp ~/.bash_profile ~/.dummy_bash_source
+
+        # Remove existing .bash_profile (file or symlink)
+        rm -f ~/.bash_profile
+
+        # Create symlink pointing to dummy source
+        ln -s ~/.dummy_bash_source ~/.bash_profile
+
```
hydai: This is preparing the test case; you should not add an empty line between them. Instead, write a comment indicating your intended action and combine these three lines.

.github/workflows/test-installers.yml
```diff
@@ -310,6 +329,16 @@ jobs:
...
+
```
hydai: Ditto.

utils/uninstall.sh
```diff
@@ -247,12 +247,32 @@ main() {
...
+      cp "$real_file" "$real_file.bak"
+      sed "${line_num}d" "$real_file.bak" > "$real_file"
+      rm -f "$real_file.bak"
+    done
+
+    exit 0
+
+
+
+
```
hydai: Why do we need so many empty lines here? Why do we need an extra `exit 0`?

hydai (changes requested):

.github/workflows/test-installers.yml
```diff
@@ -248,17 +260,14 @@ jobs:
...
         # Test re-installation
         printf "\n=== Test 2: Re-installation ===\n"
         bash utils/install_v2.sh -V
         verify_plugin "libwasmedgePluginWasiNN.so"
-
         # Test specific version with custom GGML build
         printf "\n=== Test 3: Custom GGML build ===\n"
         bash utils/install_v2.sh -V --version=${TEST_VERSION} --ggmlbn=b5640
         verify_plugin "libwasmedgePluginWasiNN.so"
-
```
hydai: You are removing something not related to your PR.

.github/workflows/test-installers.yml
```diff
@@ -310,6 +319,16 @@ jobs:
...
         . ./.github/scripts/installer-test-helpers.sh
 
+        # Backup contents of existing .bash_profile (if any)
+        [ -e ~/.bash_profile ] && cp ~/.bash_profile ~/.dummy_bash_source
+
+        # Remove existing .bash_profile (file or symlink)
+        rm -f ~/.bash_profile
+
+        # Create symlink pointing to dummy source
+        ln -s ~/.dummy_bash_source ~/.bash_profile
```
hydai: And you are not removing the spaces.

0x-avi: Im sorry for not checking them properly , i have fixed it now

0x-avi:
@hydai @SAtacker please review if you get time! sorry for the ping

0x-avi:
@hydai @PhantomInTheWire please check

utils/uninstall.sh
```diff
@@ -52,28 +52,28 @@ usage() {
...
 
-    - wasmedgec-tensorflow is the AOT compiler that compiles WebAssembly 
-      bytecode programs (wasm programs) into native code (so program) on your 
+    - wasmedgec-tensorflow is the AOT compiler that compiles WebAssembly
+      bytecode programs (wasm programs) into native code (so program) on your
       deployment machine. It is aware of WamsEdge's Tensorflow extension API.
-    
-    - wasmedge-tensorflow-lite is the runtime that executes the wasm program or 
+
+    - wasmedge-tensorflow-lite is the runtime that executes the wasm program or
```
PhantomInTheWire: you should probably avoid touching whitespaces irrelevant to what you are doing

0x-avi: I figured it out after so long , zed kept auto formatting everything 😭  , ill push it in a few mins

hydai:
Please squash your commits.

0x-avi:
hi i had to find a workaround because i wanted to keep using zed , unrelated to issue but if someone knows how to disable linting or formatting on save please let me know, modifying settings.json is just not working out for me. Anyways all the requested changes have been made and the commits have been squashed and everything has been cleaned up, please let me know if there are any other changes, @hydai @SAtacker .

hydai (approved):

hydai:
Please squash your commits and follow the commit lint guidelines.

0x-avi:
@hydai do i need to squash the pull from master as well or does the current structure work?

hydai:
You should use `rebase` instead of `merge`, as it will contain changes not related to your PR.

0x-avi:
@hydai all done , anything else?

0x-avi:
@hydai upon closer inspection after rebasing, i might have performed it incorrectly so the current merged branch has the test written incorrectly (uses normal file instead of symlink for testing). i have fixed the tests on [here](https://github.com/0x-74/WasmEdge/tree/uninstaller-fix) , please let me know how to proceed or if you can merge it again to the latest code , genuinely sorry for the inconvenience :(

hydai:
Ah, I didn't notice that, please create a new PR for it. Thanks.

Issue #3866, opened 2024-11-09
Advisory: Release Alpine Linux packages for 3.22 instead of 3.16


evangineer:
### Summary

Alpine Linux 3.16 series is no longer being supported with security fixes and there have been several CVEs since the release of 3.16.9 in January 2024.

The latest stable series is v3.20.

https://alpinelinux.org/posts/Alpine-3.17.10-3.18.9-3.19.4-3.20.3-released.html

hydai:
Hi @evangineer 
Thanks for raising this, we are going to upgrade the alpine from 3.16 to 3.20. However, since this should be an easy one, I will mark it as help wanted / good first issue for those who want to contribute.

evangineer:
Thanks @hydai.

If I have the time to get a build environment together, I might even take a stab at it myself!

inosmeet:
Hey @hydai! I would like to try this one.
I just need to change the `ALPINE_VERSION` to 3.20, right?

hydai:
It may not be that easy. It seems like some packages are not provided by the latest Alpine. If so, you will need to build from source for these packages.

https://github.com/WasmEdge/WasmEdge/blob/master/utils/docker/Dockerfile.alpine-static#L4-L6

c0wb0yp0tat0:
Hi @hydai and team,

I noticed the discussion about upgrading Alpine from 3.16 to 3.20. I have a clarification question: is the current approach to manually update the existing Dockerfile for Alpine 3.16 to support Alpine 3.20?

If so, would it be more straightforward and less labor-intensive to utilize the current Alpine 3.20 Docker image and follow the workflow guidance in the "hands-on example" section of [this WasmLabs article](https://wasmlabs.dev/articles/docker-without-containers/?form=MG0AV3#native-vs-wasm)? This way, we could convert/port the existing Alpine 3.20 Docker container to a new WasmEdge container. 

Does this approach make sense? I welcome any questions or feedback.

Thanks,
c0wb0yp0tat0

hydai:
Hi @c0wb0yp0tat0 
The Alpine workflow is for building WasmEdge on Alpine for those who need to execute WasmEdge on their host environment, which is unrelated to the container. Of course, you can run a normal Wasm workload with the Wasm container directly, but not in this case.

c0wb0yp0tat0:
```markdown
### Summary of Findings and Process

#### Docker Environment Validation

1. **Setup and Run Alpine 3.20 in Docker:**
[...]

PR #4267, opened 2025-07-16
WIP: build(docker): upgrade Alpine base image from 3.16 to 3.22 for security

ShigrafS:
Closes: #3866 

This PR updates the base image in `utils/docker/Dockerfile.alpine-static` from **Alpine 3.16** to **Alpine 3.22**. 

**Why this change is necessary:**

* Alpine 3.16 is no longer maintained and has known CVEs.
* Alpine 3.22 is the current latest stable release with ongoing security support.
* Ensures future static builds are based on an up-to-date, secure foundation.

**Changes made:**

* Changed `ARG ALPINE_VERSION=3.16` to `ARG ALPINE_VERSION=3.22`.
* Removed outdated comment about lld-static availability for Alpine 3.17/3.18.sk.

**Impact:**

* Maintains current build functionality.
* Aligns the Dockerfile with current security practices.
* Unblocks use of this image in production pipelines requiring supported base images.

### DIFF
--- a/utils/docker/Dockerfile.alpine-static
+++ b/utils/docker/Dockerfile.alpine-static
@@ -58,7 +58,6 @@ RUN cmake -B build \
     -DLLVM_TOOL_BUGPOINT_BUILD=ON \
     -DLLVM_TOOL_LLVM_BCANALYZER_BUILD=ON \
     -DLLVM_TOOL_LLI_BUILD=ON \
-    -DLLVM_TOOL_LLI_BUILD=ON \
     llvm
 
 RUN cmake --build build --target install
@@ -70,7 +69,7 @@ RUN find build/bin -type f -executable -exec cp -v {} /lld-install/bin/ \;
 RUN test -f /lld-install/bin/lli || (echo "Error: lli not installed!" && false)
 
 ########################################
-# Stage 3: Alpine Static Image with LLD
+# Stage 3: Alpine Static Image with LLD and a Robust Toolchain
 ########################################
 FROM --platform=$BUILDPLATFORM alpine:${ALPINE_VERSION} AS alpine-static
 
@@ -79,10 +78,10 @@ RUN apk add --no-cache bash
 
 SHELL ["bash", "-c"]
 
+# CRITICAL FIX: Do not install g++, as it creates a standard library conflict.
 RUN apk add --no-cache \
     cmake \
     samurai \
-    g++ \
     clang \
     llvm-static \
     lld-dev \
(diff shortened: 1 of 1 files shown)


juntao:

------

## [utils/docker/Dockerfile.alpine-static](https://github.com/WasmEdge/WasmEdge/blob/ad9f5bc34a73962274fdfdf83d965508e9a25056/utils%2Fdocker%2FDockerfile.alpine-static)

#### Potential issues

3 issues:
1. Stage 4: Wasmedge-build stage should use a different base image since it's only building for Alpine and not for any other platforms.
2. Stage 5: Wasmedge-build stage should install the necessary tools, including `libstdc++-static`, which is required for building static libraries with CMake on Alpine.
3. Stage 6: The toolchain generation script `/usr/bin/xx-toolchain` is not being used anywhere in the Dockerfile.

#### Summary of changes

-1. Upgrade Alpine base image from 3.16 to 3.22.
-2. Install host dependencies like cmake, g++, clang.
-3. Create a cross-compilation toolchain file.
-4. Build LLD from LLVM source and install it.
-5. Copy all built tools, even if not installed via 'install' target.
-6. Create a comprehensive, robust toolchain file generator.
-7. Generate CMake toolchain file for the target platform.
-8. Install a clean libc++ toolchain into the sysroot.
-9. Use native version of llvm-config.
-10. Copy the read-only mounted source to a writable location, then patch and build from there.
-11. Build with CMake using the robust toolchain.
-12. Create a tar archive of the built WasmEdge package.

ShigrafS:
Hi @hydai 

Reference: https://github.com/ShigrafS/WasmEdge/actions/runs/16397889367

I wanted to follow up on the suggestion from the issue discussion about splitting the build into two separate Dockerfiles (`Dockerfile.lld-static` and `Dockerfile.alpine-static`) for building lld statically and then consuming it in the WasmEdge build.

Unfortunately, this approach fails in GitHub Actions because Docker does not support referencing build stages across different Dockerfiles unless the source stage is pushed as an image to a container registry. 

This problem is apparently specific to GitHub Actions which would lead to failed tests at later points as well. 

In our case, when `Dockerfile.alpine-static` tries to do:

```dockerfile
COPY --from=lld-static /lld-install/lib/liblld*.a ...
```

…it results in the following error during the CI build:

```
pull access denied, repository does not exist or may require authorization
```

This happens because `lld-static` is treated as an external image, and GitHub Actions runners don’t share local image state between jobs or across separate Dockerfiles unless the image is explicitly pushed to a registry.

### Possible Solutions:

1. **Merge the two Dockerfiles into one**
   Use a multi-stage Dockerfile so that the `lld-static` stage can be reliably referenced within the same build context. This is stable and does not require pushing or pulling from a registry.

2. **Push `lld-static` as a public Docker image**
   This allows us to reference it as a remote image from `Dockerfile.alpine-static`, but it adds overhead—building, tagging, pushing the image, and managing version compatibility. 
  
Please let me know your thoughts. Let me know if your have any other suggestions. 
Thanks!

ShigrafS:
As of now, 
The `utils/docker/Dockerfile.lld-static` file is working fine. 
And `.github/workflows/docker-build-and-test.yml` is meant for testing and would be removed before the PR is merged.

hydai:
I am considering creating a new repository to manage these dependencies instead of creating circular dependencies within this repository. Please refer to #4277.  
Feel free to share your thoughts with us about this issue.

ShigrafS:
> I am considering creating a new repository to manage these dependencies instead of creating circular dependencies within this repository. Please refer to #4277. Feel free to share your thoughts with us about this issue.

I personally support this.  
And I've been thinking maybe we should let `Dockerfile.lld-static` be a separate PR in itself. 
This would be in accordance to: https://github.com/WasmEdge/WasmEdge/issues/3866#issuecomment-2565229524

ShigrafS:
@hydai 
Alpine is missing other LLVM components that is required to build it, apart from `lld-static`, which we have successfully built. 
From the error logs: https://github.com/ShigrafS/WasmEdge/actions/runs/16453776888, the only way this can be built safely is by not only building `lld-static` but the whole `LLVM` for a stable build. The other way, `manually removing the broken reference`, might be fragile and is not future-proof, but can possibly fix this. 

Given the recent decision to split dockerfiles into a separate repo, how should we proceed with this PR?

hydai:
> Given the recent decision to split dockerfiles into a separate repo, how should we proceed with this PR?

Let's hold the PR until the new repo is created.

hydai:
Since we still need some time to migrate all Docker-related workflows to the new repo, let's review and merge this PR in the current repo.  
I will review it; could you explain the reason for adding a new workflow file instead of modifying the original one?  
Please remove the new workflow file and integrate its contents into the original one. After that, I will approve to trigger the workflow execution.

ShigrafS:
Hi @hydai 
I believe the alpine dockerfile should be changed and lld-static and alpine should be built in a single file. 
I'll make the necessary changes ASAP.

hydai:
Please ping me once you are ready for the review.

ShigrafS:
Hi @hydai 
This PR is ready. 
Please review it.

utils/docker/Dockerfile.alpine-static
```diff
@@ -1,96 +1,107 @@
...
-ARG XX_VERSION=1.2.1
-ARG ALPINE_VERSION=3.16
-# alpine 3.16 ships with llvm 13.
-# alpine 3.17 and 3.18 do not ship lld-static.
+ARG ALPINE_VERSION=3.22
+
+########################################
+# Stage 1: xx (cross-compilation tool)
+########################################
+FROM --platform=$BUILDPLATFORM tonistiigi/xx:1.2.1 AS xx
```
hydai, line 8: You should keep the xx version as an argument so that we can set it through the ARG instead of hardcoding.

utils/docker/Dockerfile.alpine-static
```diff
@@ -1,96 +1,107 @@
...
+
+########################################
+# Stage 2: Build LLD from LLVM source
```
hydai, line 11: You should also explain why we need to build from source.

utils/docker/Dockerfile.alpine-static
```diff
@@ -1,96 +1,107 @@
...
 
-# Make a cmake toolchain file
-RUN cat <<'EOT' > /usr/bin/xx-toolchain && chmod a+x /usr/bin/xx-toolchain
+# Set bash for all future RUN commands
+SHELL ["bash", "-c"]
+
+# Install static toolchain and build dependencies
+RUN apk add --no-cache \
+    cmake \
+    samurai \
```
hydai, line 84: What is this package for?

ShigrafS: `Samurai` is a lightweight replacement for `ninja`. 
I'm using this since it is available in Alpine's official repo.

utils/docker/Dockerfile.alpine-static
ShigrafS: Apart from missing `lld-static`, Alpine's repo is missing some files. 
Additionally, Alpine often lags behind on newer LLVM releases, and sometimes doesn’t include the static variants. 
Plus, it helps keep everything consistent and working smoothly with our cross-compilation setup for WasmEdge.

hydai: I mean, you should put this explanation into the comments, just like we did before to discuss why we stay on a pretty old version.

ShigrafS:
Hi @hydai 
I'm still actively working on this PR. 
I'm stuck with a missing header file issue in WasmEdge build process, namely `283.3 /usr/lib/gcc/x86_64-alpine-linux-musl/14.2.0/../../../../include/c++/14.2.0/cstdlib:79:15: fatal error: 'stdlib.h' file not found` 

I'll let you know when this is ready.

hydai:
Hi @ShigrafS,  
Just wanted to know if you are still working on this PR.

ShigrafS:
Yes. 
I'll be pushing the changes shortly.

ShigrafS:
@hydai The errors following this part are too tricky. 
Any suggestions on how to handle it?

hydai:
Will this help?

https://stackoverflow.com/questions/42366739/gcc-cant-find-stdio-h-in-alpine-linux

ShigrafS:
I'll let you know.

ShigrafS:
I’ve actually already included musl-dev and libc-dev using xx-apk, and confirmed that headers like stdio.h and stdlib.h are present in the sysroot.

However, Clang still can't find them during the build. I suspect the issue might be with the include paths in the cross-compilation context, or perhaps something misconfigured in the toolchain file.

hydai:
Sounds weird. We did use cross-compilation in 3.16; could you check the differences between 3.16 and 3.22? Especially those configurations related to the include path for cross-compilation.

ShigrafS:
https://github.com/ShigrafS/WasmEdge/blob/alpine5/utils/docker/Dockerfile.alpine-static 

https://github.com/ShigrafS/WasmEdge/actions/runs/18192662013/job/51791105883

hydai:
I am closing this due to inactivity. Thank you again for your contribution.

PR #4269, opened 2025-07-17
fix: add missing output for funcref result

### LINKED ISSUE #4260: bug: Unlike other runtimes, WasmEdge does not output "funcref"
### Summary

I executed the following test case with different runtimes like Wasmtime, Wasmer, WAMR, and WasmEdge. Wasmtime outputs "<null funcref>", Wasmer outputs "funcref", and WAMR outputs "func:ref.null". But, WasmEdge's compiled and interpreter modes do not produce an output similar to "funcref".

Test case:

```
(module
  (type (;0;) (func))
  (type (;1;) (func (result funcref)))
  (export "main" (func 0))
  (export "to_test" (func 0))
  (func (;0;) (type 1) (result funcref)
    ref.null func
    ref.null func
    i32.const 213
    select (result funcref)
  )
)
```

### Current State

Does not output "funcref" or "<null funcref>".

### Expected State

Should output "null funcref".

### Reproduction steps

1. Compiled: `wasmedge compile test.wasm test.so`; `wasmedge --enable-all run test.so to_test`
2. Interpreted: `wasmedge --enable-all run test.so to_test`

### Screenshots

_No response_

### Components

CLI

0.14.1-271-ga89c176f

Ubuntu 22.04

### Hardware Architecture

x86_64

_No response_

intojhanurag:
Fixes #4260

- Adds handling for `TypeCode::FuncRef` in the CLI output.
- Prints `<null funcref>` or `<funcref>` based on runtime value.
- Matches behavior of runtimes like Wasmtime and WAMR.


juntao:

------

## [lib/driver/runtimeTool.cpp](https://github.com/WasmEdge/WasmEdge/blob/99a625bc6958f5c4c16c753b17c6400305c97c8f/lib%2Fdriver%2FruntimeTool.cpp)

#### Potential issues

1. **Type Handling for `FuncRef` and `ExternRef` Missing**: The code contains a `TODO: FuncRef and ExternRef` comment indicating that handling for these types is not implemented, which could lead to incomplete or incorrect behavior when encountering these types.

2. **Unchecked Optional Access in ToolOnModule and ToolOnComponent Functions**: The code assumes that `Opt.Args.value()` will always return a valid value without checking if it contains any elements, which could cause undefined behavior if `Opt.Args` is empty.

3. **Potential Inconsistent Exit Codes**: The code uses multiple exit codes (e.g., `EXIT_SUCCESS`, `128 + SIGABRT`) in different conditions but does not define the significance of these codes uniformly, leading to potential confusion about the program's exit status.

#### Summary of changes

- **Added handling for `FuncRef` type**: The patch adds a new case to handle `TypeCode::FuncRef`, which was previously missing.
- **Null check for FuncRef**: Implemented a check to print `<null funcref>` if the function reference is null, otherwise prints `<funcref>`.
- **Comment update**: Removed the TODO comment about `FuncRef` and `ExternRef` since `FuncRef` has been implemented.

lib/driver/runtimeTool.cpp
```diff
@@ -205,6 +205,16 @@ ToolOnComponent(WasmEdge::VM::VM &VM, const std::string &FuncName,
             "{}\n"sv,
             uint128(std::get<ValVariant>((*Result)[I].first).get<uint128_t>()));
         break;
+      case TypeCode::FuncRef: {
+        auto ref = std::get<ValVariant>((*Result)[I].first).get<RefVariant>();
```
hydai: Please follow the coding style.

lib/driver/runtimeTool.cpp
```diff
@@ -205,6 +205,16 @@ ToolOnComponent(WasmEdge::VM::VM &VM, const std::string &FuncName,
...
+        if (ref.isNull()) {
+          fmt::print("<null funcref>\n"sv);
+        } else {
+          fmt::print("<funcref>\n"sv);
```
hydai, line 213: I think we should also print the result of `funcref` here.

lib/driver/runtimeTool.cpp
```diff
@@ -205,6 +205,16 @@ ToolOnComponent(WasmEdge::VM::VM &VM, const std::string &FuncName,
...
+        }
+        break;
+      }
+
       /// TODO: FuncRef and ExternRef
```
hydai: Please update the TODO.

hydai (changes requested):

lib/driver/runtimeTool.cpp
intojhanurag: Hey , you are talking about that curly brackets ?

hydai: `ref` -> `Ref`

lib/driver/runtimeTool.cpp
intojhanurag: I couldn't find a method like getFuncIdx() in RefVariant. Should I implement such a method to extract the function index, or is there an existing way to fetch it? Let me know how you'd prefer it to be handled.

lib/driver/runtimeTool.cpp
intojhanurag: Hey @hydai  I did changes accordingly, can you check ?

lib/driver/runtimeTool.cpp
hydai: Let's keep this for now.

lib/driver/runtimeTool.cpp
hydai: Do the FuncRef and ExternRef handle them in the same way? It seems like you are removing the `ExternRef` at the same time.

lib/driver/runtimeTool.cpp
```diff
@@ -205,7 +205,15 @@ ToolOnComponent(WasmEdge::VM::VM &VM, const std::string &FuncName,
             "{}\n"sv,
             uint128(std::get<ValVariant>((*Result)[I].first).get<uint128_t>()));
         break;
-      /// TODO: FuncRef and ExternRef
```
hydai, line 208: Hi @intojhanurag 
You still missed this one; only one type is handled by your PR, but you removed both.

hydai:
Hi @intojhanurag 
Are you still working on this?

intojhanurag:
> Hi @intojhanurag Are you still working on this?


> Hi @hydai , I’ve reviewed the feedback, but I’m not fully confident navigating the codebase to make the changes correctly. I’ll step back on this one so it’s not blocked. Thanks for the review and insights.

lib/driver/runtimeTool.cpp
hydai: This one. You implement the FuncRef, however, you remove the ExternRef comments.

hydai:
Hi @intojhanurag,  
I think you may have missed the above review comments.  
You are deleting comments containing `funcref` and `externref`, but you only implemented `funcref`. That's why I am requesting the changes.

q82419:
Replaced by #4324

PR #4274, opened 2025-07-20
fix(executor): add checks before dumpStackTrace

### LINKED ISSUE #4273: question: The output contains the "[2025-07-19 09:20:23.355] [error] calling stack:"
### Summary

Hello, when I use `wasmedge` to execute any wasm file in the terminal, after normal execution, it will output an extra line of information, as shown in the following image:

### Appendix

Image

Image

Image

I used three versions of the `wasmedge` test separately, and the first two had this output. All three `wasmedges` are built from source. The version of `wasmedge` is shown in the image.

- version 0.15.0-alpha.3
- version 0.15.0-alpha.4-5-g7491f8c7
- version 0.14.1

I tried to find a solution on my own, but unfortunately I didn't find the cause. I wonder if it's because I'm missing something, or if there's a bug in `wasmedge`.

Looking forward to your answer! Thank you!

PhantomInTheWire:
## Summary

Fixes #4273

This PR prevents **extra `[error]` output after normal execution** of wasm files built with `emcc` on Linux.

---

### Background

* Issue only occurs with **`emcc`-generated wasm**, not `rustc`.
* Reproducible on **Linux**, not on macOS.
* Present since `0.15.0-alpha.1`, not in `0.14`.

* likely introduced with pr https://github.com/WasmEdge/WasmEdge/pull/3967
The cause was `dumpStackTrace` being called even on `ErrCode::Terminated`, resulting in unnecessary `[error]` logs despite normal termination.

---

### Changes

* Add guard before `dumpStackTrace`:
  * Check that error is **not `ErrCode::Terminated`**.
  * Check that **`StackTraceSize > 0`**.

---

### Testing(on linux)

* `emcc` wasm files (previously showed extra error)
* `rustc` wasm files (never showed extra error)

Confirmed:

* Normal executions no longer output `[error]`.
* Real errors continue to display stack traces as intended.

---

#### Notes

* Thanks @gaaraw for the clear report and reproduction files.
* Root cause identified as introduced in #3967.

### DIFF
--- a/lib/executor/executor.cpp
+++ b/lib/executor/executor.cpp
@@ -144,7 +144,9 @@ Executor::invoke(const Runtime::Instance::FunctionInstance *FuncInst,
 
   // Call runFunction.
   EXPECTED_TRY(runFunction(StackMgr, *FuncInst, Params).map_error([](auto E) {
-    dumpStackTrace(Span<const uint32_t>{StackTrace}.first(StackTraceSize));
+    if (E != ErrCode::Value::Terminated) {
+      dumpStackTrace(Span<const uint32_t>{StackTrace}.first(StackTraceSize));
+    }
     return E;
   }));
 


juntao:

------

## [lib/executor/executor.cpp](https://github.com/WasmEdge/WasmEdge/blob/ba1690b5c8a286cc7c7ec808cdf58e5a487c97ba/lib%2Fexecutor%2Fexecutor.cpp)

#### Potential issues

1. **Error Code Handling**: The code always logs error codes but doesn't ensure that the logged errors are meaningful or helpful for debugging, which could hinder troubleshooting.
2. **Reference Type Validation**: The validation for reference types in `invoke` function does not check all necessary conditions, potentially allowing invalid references to pass through.
3. **Stack Trace Dumping Condition**: The condition for dumping stack trace in the `map_error` lambda of `runFunction` could be improved to include more detailed or specific error handling scenarios.

#### Summary of changes

- **Added a condition to check if the error is not `ErrCode::Value::Terminated` before dumping stack trace.**
- **Prevents unnecessary stack trace dump when the function execution is terminated normally.**
- **Improves performance by avoiding redundant operations during expected terminations.**

lib/executor/executor.cpp
```diff
@@ -139,7 +139,9 @@ Executor::invoke(const Runtime::Instance::FunctionInstance *FuncInst,
 
   // Call runFunction.
   EXPECTED_TRY(runFunction(StackMgr, *FuncInst, Params).map_error([](auto E) {
-    dumpStackTrace(Span<const uint32_t>{StackTrace}.first(StackTraceSize));
+    if (E != ErrCode::Value::Terminated && StackTraceSize > 0) {
```
ibmibmibm: Is `StackTraceSize > 0` checking necessary?

PhantomInTheWire: Although nothing unusual appears when I remove it, I think we should still keep the `StackTraceSize > 0` check to avoid printing empty stack traces. Empty traces would just clutter the logs and could confuse developers. LMK  if you see it differently or if this could cause some issues that I don't realize yet.

hydai: I think the original issue was caused by `E != ErrCode::Value::Terminated`.  
We should not check whether `StackTraceSize > 0`, as this may indicate a potential issue: the error occurs, but the size of the stack trace is empty.

PhantomInTheWire: sure I guess that makes more sense

hydai (approved):

hydai:
Thanks.

PR #4295, opened 2025-08-04
fix(validator): correct core module handling

LFsWang:
This fix the first test of the Component.

### DIFF
--- a/lib/validator/component_validator.cpp
+++ b/lib/validator/component_validator.cpp
@@ -73,6 +73,7 @@ Validator::validate(const AST::Component::CoreModuleSection &ModSec) noexcept {
     return E;
   }));
   CompCtx.incCoreSortIndexSize(AST::Component::Sort::CoreSortType::Module);
+  CompCtx.addCoreModule(ModSec.getContent());
   return {};
 }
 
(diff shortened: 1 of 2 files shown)


juntao:

------

## [lib/validator/component_validator.cpp](https://github.com/WasmEdge/WasmEdge/blob/fa0326d58b8d0105968f17e2450ef955c0958c52/lib%2Fvalidator%2Fcomponent_validator.cpp)

#### Potential issues

-1. The `validate` function should be renamed to something more descriptive, such as `processComponent`.
2. The use of `compCtx.enterComponent(Comp)` and `compCtx.exitComponent()` is unnecessary if the context management is handled within the `validateCoreModuleSection` and other relevant functions.
3. The variable name `ReportError` is not very helpful. Consider renaming it to something more descriptive, such as `emitValidationWarning`.
4. In the `validate(const AST::Component::ImportSection &ImpSec)` function, the loop should check if the import descriptor type matches the sort type before checking the index.
5. The variable name `OutLinkCompCnt` is not very helpful. Consider

#### Summary of changes

- **Most Important Changes:**
1. Added a method `addCoreModule` to the `CompCtx` class to add core modules.
2. Updated the condition in `validate(const AST::Component::Alias &Alias)` method to ensure that an alias is not mapped to a core type reference if it is not a core module.
3. Improved error handling to provide more detailed information about the issue when encountering an invalid core export.

- **Other Changes:**
1. Updated the sorting index size for core modules in `validate(const AST::Component::CoreModuleSection &ModSec)` method.
2. Added comments to explain the purpose and functionality of the new method `addCoreModule`.

## [test/component/spectest.cpp](https://github.com/WasmEdge/WasmEdge/blob/fa0326d58b8d0105968f17e2450ef955c0958c52/test%2Fcomponent%2Fspectest.cpp)

#### Potential issues

N/A

#### Summary of changes

- 1. `VM.validate()` is commented out.
- 2. The test case for loading a binary component is removed.

PR #4317, opened 2025-08-12
ci(runner): Use windows-2025 instead of 2022 due to the EOL of win10

hydai:

PR #4359, opened 2025-09-08
fix(wasi-nn,cmake): add missing option for the MSVC target

hydai:


juntao:

------

## [cmake/WASINNDeps.cmake](https://github.com/WasmEdge/WasmEdge/blob/2e19b352288196806610f6330ebb4b556910e49d/cmake%2FWASINNDeps.cmake)

#### Potential issues

1. The `wasmedge_setup_llama_target` function has a missing argument for `target`. It should include a reference to the target name in the `find_package` command.
2. There is a syntax error in the `target_compile_options` call within the `wasmedge_setup_willisper_target` function, where the flag `-Wno-error=unused-parameter` is not properly escaped.
3. The `wasmedge_setup_piper_target` function has an incorrect variable name for the MLX library path. It should use `${MLX_INCLUDE_DIRS}` instead of `${CMAKE_SOURCE_DIR}/plugins/wasi_nn/MLX`.

#### Summary of changes

-### Key Changes

1. **Helper Function for Removing Compile Flags**: Added a `remove_flag_from_target` function that modifies the compile options of a target to remove a specified flag.

2. **MSVC Target Options**: For Microsoft Visual Studio (MSVC) targets, removed warnings `/wd4305` and `/wd4244`.

### Most Important Changes

- Added a helper function to manage compile flags for different compilers.
- Removed specific warnings from MSVC targets to avoid errors during compilation.

ibmibmibm:
Fixed in #4317

PR #4363, opened 2025-09-12
ci(macos): don't install cmake due to the conflict inside macos runner image

hydai:


juntao:

------

## [.github/workflows/IWYU_scan.yml](https://github.com/WasmEdge/WasmEdge/blob/e3aafc2df5e4b76b17af836400205ee2f8bb2723/.github%2Fworkflows%2FIWYU_scan.yml)

#### Potential issues

-1. The code is not properly formatted or documented, which can make it difficult to understand and maintain.
- 2. There are several missing `#include` directives, such as `<string>`, `<vector>`, and `<iostream>`.
- 3. The code relies on a local package manager (`brew`) which could be removed without notice in the future.

#### Summary of changes

- - Add `brew install llvm@18 ninja` to the job configuration to ensure LLVM and Ninja are installed.
- - Remove cmake installation from the job configuration as it is already installed inside the local/pinned tap inside the macos runner image.

## [.github/workflows/wasi-testsuite.yml](https://github.com/WasmEdge/WasmEdge/blob/e3aafc2df5e4b76b17af836400205ee2f8bb2723/.github%2Fworkflows%2Fwasi-testsuite.yml)

#### Potential issues

-1. The `permissions: contents: read` directive is set for the `test` job, which allows reading the repository's contents but not writing to it. This might cause issues if a user tries to commit changes.

- 2. The `strategy.fail-fast: false` directive is used in the `test` job, which means that even if one test fails, all tests will be run and no failure is reported as an error. This might lead to misleading results.

- 3. The `steps[4]` script installs `llvm-windows` from a specific release URL, but it does not specify that the version should be downloaded. This could cause issues if a different version of `llvm-windows` is available at the time of the run.

#### Summary of changes

-1. Install llvm and ninja using brew
-2. Use a local/pinned tap to install cmake inside the macos runner image
-3. Configure LLVM_DIR and set CFLAGS and CXXFLAGS to use clang

PR #4366, opened 2025-09-12
ci: pinned the version of the used packages.

hydai:


juntao:

------

## [.github/workflows/commitlint.yml](https://github.com/WasmEdge/WasmEdge/blob/569c38fb3b66abec30ac25bb8f8dcedd865d1feb/.github%2Fworkflows%2Fcommitlint.yml)

#### Potential issues

-1. Incorrect permission set in `.github/workflows/commitlint.yml`, should be `contents: write` to allow checkout of pull requests.
- 2. Missing `--from` and `--to` flags in the commitlint command, causing it to validate all commits in the branch instead of just the current pull request.
- 3. Incorrect package names used for commitlint and its config, should use the correct versions (`conventional-changelog-conventionalcommits@19.1.0`, `commitlint@19.8.1`, and `@commitlint/config-conventional@19.8.1`).

#### Summary of changes

-```plaintext
1. Pinned the version of `conventional-changelog-conventionalcommits` to `9.1.0`.
2. Pinned the version of `commitlint` and `@commitlint/config-conventional` to `latest`.
3. Added a step to validate PR commits with commitlint using `latest` versions.
```

## [.github/workflows/misc-linters.yml](https://github.com/WasmEdge/WasmEdge/blob/569c38fb3b66abec30ac25bb8f8dcedd865d1feb/.github%2Fworkflows%2Fmisc-linters.yml)

#### Potential issues

1. The `misc` job is missing a step to install necessary dependencies using pip.
2. The `lineguard` job's checkout depth setting should be set to 0 for complete code history retrieval.
3. Ensure that the lineguard installation command uses `cargo install` instead of `python3 -m venv ...`.

#### Summary of changes

-1. **Pinned the version of `codespell`**: The patch updates the version of the `codespell` package to `2.4.1`. This is done by specifying `pip install codespell==2.4.1` in the job definition. This ensures that the code formatting tool used consistently across all development environments, maintaining consistency and reducing errors.

- 2. **Changed `git ls-files` command**: The old command `git ls-files` was used to list files in a repository, but it did not exclude files located in specific directories (`/thirdparty`, `/thirdparty/`, and `/dist/`). This change ensures that only relevant files are processed by the `codespell` tool.

- 3. **Removed output redirection**: The old command included `2>/dev/null` to discard any error messages, but this is no longer necessary since the patch specifies that only successful runs should be reported.

## [.github/workflows/test-installers.yml](https://github.com/WasmEdge/WasmEdge/blob/569c38fb3b66abec30ac25bb8f8dcedd865d1feb/.github%2Fworkflows%2Ftest-installers.yml)

#### Potential issues

-1. The `get-latest-version` job is configured to output the latest WasmEdge release version, but it does not use this output anywhere in the workflow.
- 2. The `test-install-sh-linux` and `test-install-sh-macos` jobs are missing a `verify_version` function that checks if the installed version matches the expected version. This can lead to incorrect test results.
- 3. The `test-install-v2` and `test-install-v2-sh-macos` jobs assume that the environment variable `.wasmedge/env` is set, but this is not guaranteed by default. It should be checked if the variable exists before using it.

#### Summary of changes

-### Summary of Changes
1. **Python Version**: Pinned the version of Python to v6.0.0.
2. **Black Formatter Installation**: Changed `pip install black` to `pip install black==25.1.0` to specify a specific version.
3. **Checkout Code**: Updated the checkout step to version 5.0.0.

### Most Important Changes
- **Python Version**: Ensuring consistent Python environment across jobs.
- **Black Formatter**: Updating Black formatter to version 25.1.0 for consistency and reliability.

These changes are crucial as they address a dependency issue related to Python versions and ensure that the toolchain is maintained and compatible with a specific version of `black`.

PR #4370, opened 2025-09-15
refactor(WASI-NN/GGML): parse metadata

grorge123:
- Add macro to parse metadata
- Refactor parsing code

### DIFF
--- a/plugins/wasi_nn/wasinn_ggml.cpp
+++ b/plugins/wasi_nn/wasinn_ggml.cpp
@@ -57,6 +57,42 @@ namespace {
   spdlog::error("[WASI-NN] GGML backend: "sv __VA_ARGS__);                     \
   return Error;
 
+// Basic parsing macro
+#define PARSE_Json(Type, Doc, Key, Var)                                        \
+  if ((Doc).at_key(Key).error() == simdjson::SUCCESS) {                        \
+    auto Err = (Doc)[Key].get<Type>().get(Var);                                \
+    if (Err) {                                                                 \
+      std::string Msg = fmt::format("Unable to retrieve the {} option.", Key); \
+      spdlog::error("[WASI-NN] GGML backend: {}", Msg);                        \
+      return ErrNo::InvalidArgument;                                           \
+    }                                                                          \
+  }
+
+// Macro for parsing with type casting.
+#define PARSE_Json_WithCast(FromType, ToType, Doc, Key, Var)                   \
+  if ((Doc).at_key(Key).error() == simdjson::SUCCESS) {                        \
+    FromType TempValue = FromType();                                           \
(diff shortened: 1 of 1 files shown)


juntao:

------

## [plugins/wasi_nn/wasinn_ggml.cpp](https://github.com/WasmEdge/WasmEdge/blob/0bca70426e2f3ba5267f6a82fc283487c7185272/plugins%2Fwasi_nn%2Fwasinn_ggml.cpp)

#### Potential issues

**Major Bug:** The `parseMetadata` function in the `ggml_backend_wasm.cc` file contains a major coding issue. Specifically, the code does not correctly parse the JSON metadata for the context parameters. The function uses `parseJsonWithCastAuto<int64_t>` to parse the `ctx-size`, but it does not properly handle the parsing of other context-related parameters such as `batch-size`, `ubatch-size`, and `n-keep`. This can lead to incorrect settings for these parameters when loading a model.

#### Summary of changes

N/A

plugins/wasi_nn/wasinn_ggml.cpp
```diff
@@ -56,6 +56,42 @@ namespace {
   spdlog::error("[WASI-NN] GGML backend: "sv __VA_ARGS__);                     \
   return Error;
 
+// Basic parsing macro
+#define PARSE_Json(Type, Doc, Key, Var)                                        \
```
hydai: Let's use template and expected format instead of the macro way.

plugins/wasi_nn/wasinn_ggml.cpp
```diff
@@ -56,6 +56,64 @@ namespace {
...
+                                          std::string_view Key) noexcept {
+  if (Doc.at_key(Key).error() == simdjson::SUCCESS) {
+
+    T Value{};
+    auto Err = Doc[Key].get<T>().get(Value);
+    if (Err) {
+      std::string Msg = fmt::format("Unable to retrieve the {} option.", Key);
+      spdlog::error("[WASI-NN] GGML backend: {}", Msg);
+      return WasmEdge::Unexpected<ErrNo>(ErrNo::InvalidArgument);
+    }
```
hydai, line 70: We should return `Value` here if the `get` succeeds.

plugins/wasi_nn/wasinn_ggml.cpp
```diff
@@ -56,6 +56,65 @@ namespace {
   spdlog::error("[WASI-NN] GGML backend: "sv __VA_ARGS__);                     \
   return Error;
 
+// JSON parsing helper template
+template <typename T>
+WasmEdge::Expected<T, ErrNo> getJsonValue(const simdjson::dom::element &Doc,
```
hydai: Since you are using `try-catch`, there is reason to use `WasmEdge::Expected`; you can simply return the normal value or throw the error directly.

PR #4380, opened 2025-09-24
ci(WASI-NN/MLX): support cmake version

grorge123:


juntao:

------

## [.github/workflows/matrix-extensions.json](https://github.com/WasmEdge/WasmEdge/blob/eef36c41220825a95613ebbbd46df1b0d14e3181/.github%2Fworkflows%2Fmatrix-extensions.json)

#### Potential issues

-1. **CMake Version Requirement**: The `cmake_policy_version_minimum` option is not correctly set to 3.5 as required by the build system. This might cause issues when trying to use new features or policies that are only available in newer versions of CMake.
   
- 2. **Missing OpenSSL Header File**: The code snippet includes a direct include of an OpenSSL header file without checking if it exists. If the OpenSSL library is not installed, this will result in compilation errors.

- 3. **Incorrect Platform Support for `wasi_nn-mlx`**: The platform support for `wasi_nn-mlx` is only for macOS ARM64, while other platforms are listed as having `-DCMAKE_POLICY_VERSION_MINIMUM=3.5`. This might cause build failures on some platforms that do not support the specified policies.

These issues should be addressed to ensure compatibility and correct functionality of the build system when using CMake versions 3.5 or later.

#### Summary of changes

- **Added option to specify policy version in CMake configuration**. The patch adds the `-DCMAKE_POLICY_VERSION_MINIMUM=3.5` option to ensure compatibility with newer CMake policies.

## [plugins/wasi_nn/MLX/model/utils.cpp](https://github.com/WasmEdge/WasmEdge/blob/eef36c41220825a95613ebbbd46df1b0d14e3181/plugins%2Fwasi_nn%2FMLX%2Fmodel%2Futils.cpp)

#### Potential issues

-2 major issues:
1. The `saveWeights` function does not handle the case when the file path does not end with `.safetensors` or `.npz`.
2. There is a missing include for the `AssumingUnreachable` function, which is used to report unreached code paths.

#### Summary of changes

-Added `Wasi::Environ` to the `loadBytesFromFile` function signature.
-Updated function comments for clarity.

## [plugins/wasi_nn/MLX/model/utils.h](https://github.com/WasmEdge/WasmEdge/blob/eef36c41220825a95613ebbbd46df1b0d14e3181/plugins%2Fwasi_nn%2FMLX%2Fmodel%2Futils.h)

#### Potential issues

-1. Missing #include statement for `<iomanip>` to use `std::setw()` for formatting strings.
- 2. Incorrect function names in `saveWeights` functions, they should be `saveWeights(const std::unordered_map<std::string, mx::array>& Weights, const std::string& Path)` and `saveWeights(const mx::array& Weights, const std::string& Path)`.
- 3. Improper use of `fillPlaceholders` function with template arguments, it should be called with `Oss << toString(Value);` inside the loop.

#### Summary of changes

-**Changes:**  
1. Added a new include for `host/wasi/vfs_io.h` which likely provides file system operations specific to Wasi.
2. Changed the signature of `loadBytesFromFile` to accept an additional parameter `const Host::WASI::Environ *Env`, indicating that it may need access to Wasi-specific environment information.
  
**Most Important Changes:**  
- Added support for Wasi-specific file system operations in the `host/wasi/vfs_io.h` include.
- Changed function signature to handle Wasi-specific environment information.

hydai (approved):

PR #4430, opened 2025-12-07
feat(validator): add ComponentName class and parsing logic

LFsWang:
* add Component Name validator for import
* fix a bug about access data


hydai:
@LFsWang, please rebase to the latest master branch. It should restore the Fedora CI functionality through your previous patch.

@q82419, this PR is ready for review. Please process it.

q82419:
> @LFsWang, please rebase to the latest master branch. It should restore the Fedora CI functionality through your previous patch.
> 
> @q82419, this PR is ready for review. Please process it.

I'll review in these days.

include/validator/component_context.h
```diff
@@ -52,6 +56,91 @@ class ComponentContext {
...
+          return false;
+        }
+        // By rule, a constructor [constructor]X and X are strongly-unique.
+        // if [constructor]X and its lower-case [constructor]x form both exist,
+        // it meaning [constructor]x is coming from [constructor]X.
+      }
+      ImportedNames.insert(Normal);
+      ImportedNames.insert(UniForm);
+      return true;
+    }
```
hydai: I think we should move these long implementations into .cpp files instead of the headers.

lib/validator/component_name.cpp
```diff
@@ -0,0 +1,290 @@
...
+      return;
+    }
+    if (!isKebabString(Resource) || !isKebabString(Next)) {
+      return;
+    }
+    Detail.Static.Resource = Resource;
+    Detail.Static.Method = Next;
+    Kind = ComponentNameKind::Static;
+    return;
+  }
```
hydai: I found this part is almost the same as the `[method]` one. Should we extract these into a helper function instead?

PR #4451, opened 2025-12-17
fix(runtime): refine the overflow detection when wrapping memory instance

q82419:


include/runtime/instance/memory.h
```diff
@@ -241,14 +241,28 @@ class MemoryInstance {
...
+      return Span<T>();
+    }
+    Size = static_cast<uint32_t>(Num);
+#else
+    if (unlikely(__builtin_mul_overflow(static_cast<uint32_t>(sizeof(T)), Count,
+                                        &Size))) {
+      return Span<T>();
+    }
+#endif
+    if (unlikely(!checkAccessBound(Offset, Size))) {
```
dannypsnl, line 262: In non-msvc block, `Size` seems didn't get initialized?

q82419: The function `__builtin_mul_overflow` will fill the variable.

PR #4456, opened 2025-12-19
fix(runtime): add instantiation-only mode for modules without exports

### LINKED ISSUE #4378: bug: Issue with running wasm file in interpreter mode
### Summary

I am facing an issue when trying to run a wasm file with the WasmEdge runtime. I am including a simple test case. The wasmedge interpreter mode is creating a kind of locked situation. The test case needs to be run with `--enable-all` but there is no exported function. If we run `wasmedge <testcase>`, we get the `integer too large error` likely because `--enable-all` feature was not included. `wasmedge compile` works on this test case with the `--enable-all` feature. When running `wasmedge --enable-all <testcase>`, I keep getting the error that `a function name is required when running in the reactor mode`.

### Current State

Cannot run the wasm file in interpreter mode.

### Expected State

Should be able to run the wasm file in interpreter mode without any error.

### Reproduction steps

Test:
```
(module
  (memory (;0;) 2 1024 shared)
)
```

1. Convert wat to wasm.
2. Run `wasmedge test.wasm`
3. Run `wasmedge --enable-all test.wasm`

### Screenshots

DESCRIPTION

Could this be a bug or is there another way to run this wasm file with WasmEdge? Your guidance will be highly appreciated.

### Components

CLI

0.15.0-alpha.4

Ubuntu 22.04

### Hardware Architecture

x86_64

_No response_

Harsh-6291:
Fixes #4378

This PR fixes the issue where WasmEdge couldn't run WASM modules with shared memory but no exported functions.

## Problem

Running `wasmedge --enable-all test.wasm` on a module with only memory would fail with error "function name is required"

## Solution

I modified `lib/driver/runtimeTool.cpp` to check if the module has any exported functions. If it has no exports, it now successfully instantiates instead of throwing an error.

## Testing

- Built and tested with a shared memory module
- **Before fix:** Error message
- **After fix:** Success message "Module instantiated successfully"

This allows testing shared memory modules without needing dummy functions.


Harsh-6291:
Hi @hydai Please review this PR

Harsh-6291:
@hydai , I have run clang-format on lib/driver/runtimeTool.cpp and pushed the updated commit. Please let me know if any further changes are required.

hydai:
CleanShot 2025-12-24 at 11 24 16

Your PR contains commits from others; please ensure you only commit your own code.

Harsh-6291:
hey @hydai ,I’ve rebased onto master so the PR now only contains my fix commit with DCO and passing commitlint. Please approve and run the pending workflows when convenient.

hydai:
I don't think so. Please check again.
CleanShot 2025-12-24 at 18 39 31

Harsh-6291:
Thanks for pointing that out @hydai . I’ve now rebased the branch onto the latest master and rebuilt it.
The PR currently contains only a single commit

Harsh-6291:
Hi @hydai ,
In the latest CI run, the Debian-based Docker build failed with Error reading from server - read (104: Connection reset by peer) while apt-get install was downloading libbrotli1. This looks like a transient network/mirror issue rather than something caused by the code changes.
Could you please re-run the failed job when you have a moment? If you’d prefer, I can also open a follow-up PR to add a small retry/--fix-missing option to the apt-get step to make this build more robust.

hydai:
Since we are going to have 0.16.0 before the New Year, the PR will be postponed until the release process is complete.

Harsh-6291:
Thank you for letting me know. I understand the PR will be reviewed after the 0.16.0 release is complete. Please take your time, and feel free to ping me if you need any changes or clarifications after the release.

lib/driver/runtimeTool.cpp
```diff
@@ -450,11 +450,28 @@ int Tool(struct DriverToolOptions &Opt) noexcept {
...
+        return EXIT_SUCCESS;
+      }
+
+      // Original error for modules with exports but no function name provided
       fmt::print(
           stderr,
           "A function name is required when reactor mode is enabled.\n"sv);
       return EXIT_FAILURE;
     }
+
```
hydai, line 495: May I ask why we need to add a newline here?

Harsh-6291: that line was not required , i  fixed it now

lib/driver/runtimeTool.cpp
```diff
@@ -450,6 +450,22 @@ int Tool(struct DriverToolOptions &Opt) noexcept {
...
+      if (!HasExportedFunctions) {
+        // Instantiation-only mode: Module has no _start and no exported
+        // functions This is valid for modules that only declare resources
+        // (memory, tables, etc.)
+        spdlog::info("Module instantiated successfully (no _start or exported "
+                     "functions).");
+        return EXIT_SUCCESS;
+      }
+
+      // Original error for modules with exports but no function name provided
```
PhantomInTheWire, line 489: do we need this kind of comment? if you are using ai please add to your ai agent rules to not add irrelevant comments

lib/driver/runtimeTool.cpp
```diff
@@ -450,6 +450,16 @@ int Tool(struct DriverToolOptions &Opt) noexcept {
 
     // Get the function name to invoke.
     if (Opt.Args.value().empty()) {
+      // Check if module has any exported functions
+      auto Functions = VM.getFunctionList();
+      bool HasExportedFunctions = !Functions.empty();
+
+      if (!HasExportedFunctions) {
```
PhantomInTheWire, line 480:
```suggestion
      if (Functions.empty()) {
```
we should probably avoid the needless boolean flip

hydai:
I am closing this because #4513 will be the solution.  
Thank you for your previous contribution and interest.

PR #4440, opened 2025-12-12
Upgrade Alpine base to 3.23 and restore static LLD support

blackdragoon26:
This PR updates the `utils/docker/Dockerfile.alpine-static` base image from Alpine 3.16 (EOL) to Alpine 3.23 (Latest Stable) as of Dec'2025.

This upgrade addresses the long-standing issue, i.e. #3866  where Alpine 3.17+ removed the lld-static package, which previously blocked attempts to upgrade the base image ( #4267).

### The Problem
- Alpine 3.16 is EOL: It has unpatched security vulnerabilities.
- Missing Dependencies: Newer Alpine versions dropped `lld-static` from their repositories.
- Build Failures: Previous attempts to compile LLVM/LLD from source using the existing cross-compilation toolchain (`xx-clang`) failed due to sysroot header issues (`stdlib.h not found`).

### The Solution: Native Builder Stage
This PR introduces a Multi-Stage Build strategy to solve the dependency hell:

- Stage 1 (`lld-builder`):
   - Instead of fighting with cross-compilation headers, I compiled LLD using native Alpine compilers in a separate stage.
   - built only the necessary static linker libraries (`lldELF`, `lldCOFF`, `lldWasm`, `lldMinGW`, etc.) to keep build times reasonable.
   - explicitly disable tests, examples, and benchmarks to minimize overhead.

- Stage 2 (Main Build):
   - copy the compiled static libraries from Stage 1 into the sysroot.
   - patch the broken Alpine LLVM package configuration (which incorrectly references missing `LLVMTesting` libraries) by creating dummy archives. This satisfies CMake without requiring to compile unused testing code.
   - added `zstd-static`, which is now a required dependency for LLVM on newer Alpine versions.

### detailed Changes
- Dockerfile:

   - Bumped `ALPINE_VERSION` to `3.23`
   - Added `lld-builder` stage to compile LLVM/LLD static libs from source.
   - Added `zstd-static` to `xx-apk` dependencies.
   - Added a shell loop to mock missing `libLLVMTesting*.a` files (fixing CMake configuration crashes).
   - Added logic to copy the built `liblld*.a` files into the build context.
   - Ensured `-DWASMEDGE_LINK_TOOLS_STATIC=ON` remains enabled to support full AOT functionality in the static binary.

### Verification
I have successfully built the image locally. The resulting static library includes the linker code, enabling AOT support.

Build Time: ~60-90 minutes (due to LLVM compilation).

[...]

PR #4466, opened 2025-12-24
feat(docker): add standalone alpine-base image build pipeline

blackdragoon26:
fixes the long 6hr+ build issue at #4440

### DIFF
--- a/utils/docker/Dockerfile.alpine-base
+++ b/utils/docker/Dockerfile.alpine-base
@@ -1,0 +1,62 @@
+# syntax=docker/dockerfile:1.5-labs
+
+ARG ALPINE_VERSION=3.23
+
+# ----------------------------------------------------
+# STAGE 1: Builder
+# Compiles LLD from source, linking against system LLVM
+# ----------------------------------------------------
+FROM alpine:${ALPINE_VERSION} AS builder
+
+# Use a specific tag for reproducibility
+# Pinned version to match Alpine 3.23's system LLVM (v19)
+ARG LLVM_VERSION=llvmorg-19.1.5
+
+# Build Dependencies:
+# - build-base: Compiler toolchain (gcc, g++, make, libc-dev) required for compilation.
+# - cmake: The build system used by LLVM.
+# - ninja: The build generator (faster than make).
+# - python3: Required by LLVM build scripts and configuration utilities.
+# - git: Required to clone the LLVM repository.
+# - linux-headers: Ensures system header compatibility for C++ standard library builds.
+# - llvm19-dev: Provides LLVM development headers for version 19.
+# - llvm19-static: Provides static LLVM libraries for version 19.
+# - zstd-static: Provides static zstd compression libraries.
+# - zlib-static: Provides static zlib compression libraries.
+RUN apk add --no-cache build-base cmake ninja python3 git linux-headers \
+    llvm19-dev llvm19-static zstd-static zlib-static
+
+
+# The system CMake files require testing libraries that are missing in the Alpine package.
+# We create empty dummy archives to satisfy the dependency check.
+RUN LLVM_LIB_DIR=/usr/lib/llvm19/lib && \
+    for f in libLLVMTestingAnnotations.a libLLVMTestingSupport.a libllvm_gtest.a libllvm_gtest_main.a; do \
+        ar rc "$LLVM_LIB_DIR/$f"; \
+    done
+
+# Clone LLVM Project
+WORKDIR /src
+# Shallow clone specific tag
+RUN git clone --depth 1 -b ${LLVM_VERSION} https://github.com/llvm/llvm-project.git
+
+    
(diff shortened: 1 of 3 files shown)


utils/docker/Dockerfile.alpine-base
```diff
@@ -0,0 +1,40 @@
...
+    -DBUILD_SHARED_LIBS=OFF \
+    -DLLVM_BUILD_STATIC=ON \
+    -DLLVM_INCLUDE_TESTS=OFF \
+    -DLLVM_INCLUDE_EXAMPLES=OFF \
+    -DLLVM_INCLUDE_BENCHMARKS=OFF \
+    -DLLVM_ENABLE_ZLIB=ON \
+    -DLLVM_ENABLE_ZSTD=ON
+
+# Compile only the specific linker libraries needed
+RUN cmake --build build --target lldELF lldCommon lldMachO lldCOFF lldWasm lldMinGW
```
hydai: I found that we will only use a very limited set of built assets in this Docker image. See: https://github.com/WasmEdge/WasmEdge/pull/4440/changes#diff-199c7f46bf8974f073fd0072da24c710f7798310892ce217a63e40e7114ae84fR105-R106

Should we only keep these files and remove the temporary files generated during the build process to reduce the Docker image size? This is very important for minimizing overhead when building another Docker image based on this. If we keep everything here, Docker will try to download the entire image but only utilize a few items from it.

utils/docker/docker-bake.alpine-base.hcl
```diff
@@ -0,0 +1,12 @@
...
+}
+
+target "alpine-base" {
+  dockerfile = "Dockerfile.alpine-base"
+  context    = "utils/docker"
+  # Build for both major architectures at once
+  platforms  = ["linux/amd64", "linux/arm64"]
+  # Tag it clearly so we can use it later
+  tags       = ["wasmedge/wasmedge:alpine-base-3.23"]
+}
```
hydai: Please check the end of the line; we should not have this `-` note.

utils/docker/Dockerfile.alpine-base
```diff
@@ -0,0 +1,40 @@
+# syntax=docker/dockerfile:1.5-labs
+
+ARG ALPINE_VERSION=3.23
+
+FROM alpine:${ALPINE_VERSION}
+
+# We default to release/19.x to match modern Alpine expectations, 
+# but this can be overridden if Alpine ships a newer version (e.g., 21).
+ARG LLVM_VERSION=release/19.x
```
hydai: I prefer to use a tag instead of a branch, since it will be easier to identify which version we are using in this image.

utils/docker/docker-bake.alpine-base.hcl
blackdragoon26: okay, will fix that

utils/docker/Dockerfile.alpine-base
blackdragoon26: That is actually good and much needed here, let me optimise it.

utils/docker/Dockerfile.alpine-base
blackdragoon26: alrighty

hydai:
Let's wait and see how much time it will take if this still times out. Maybe we can just use the llvm-static package and build only lld instead?

blackdragoon26:
> Let's wait and see how much time it will take if this still times out. Maybe we can just use the llvm-static package and build only lld instead?

Alrighty,
though fingers crossed :)

0yi0 (dismissed):

blackdragoon26:
> Let's wait and see how much time it will take if this still times out. Maybe we can just use the llvm-static package and build only lld instead?

It looks like, based on the current progress of just 678/2088 CXX build for 3.5 hours, I suspect the same 10 hours+ expected time build, which is just too long again

Your suggestion to use the system llvm-static package sounds like the perfect solution.

So if you agree, then
I will refactor the Dockerfile to:
- Install llvm-static (and llvm-dev) via apk to get the core LLVM artifacts instantly.
- Clone the llvm-project but configure CMake to build only the LLD sub-project.
- Point CMake to the system LLVM (-DLLVM_DIR=/usr/lib/cmake/llvm) so it links against the pre-installed package instead of compiling LLVM from scratch.

Sounds good?

hydai:
Sounds good. Could you please move forward?

hydai (changes requested):
Let me block this PR in case someone merges it accidentally.

blackdragoon26:
@hydai , the Alpine Base (3.23) action passed successfully.
I don't think so that we need to wait for rest of actions to pass since they are irrelevant and were always passing in previous attempts.
If looks good, can we proceed ahead?

hydai:
I think it will take some time to publish to our Docker Hub. Please wait a moment.

PR #4470, opened 2025-12-27
fix(wasi): resolve 100% CPU usage on macOS due to clock mismatch

### LINKED ISSUE #3851: bug: non-blocking io cause CPU 100%
### Summary

I use example here [WasmEdge/wasmedge_reqwest_demo](https://github.com/WasmEdge/wasmedge_reqwest_demo)
and hyper_demo to request `https://httpbin.org/delay/10`, CPU keep on 100%

related issue: https://github.com/second-state/wasmedge-quickjs/issues/145

### Current State

_No response_

### Expected State

I wonder if it is a bug or expected behavior?

### Reproduction steps

Just built according to README, change url to https://httpbin.org/delay/10

_No response_

### Components

CLI

0.14.1

Darwin 22.1.0 arm64 arm

### Hardware Architecture

arm64

_No response_

blackdragoon26:
Previously, the macOS implementation of  used the
 flag with . This expected a wall-clock timestamp
(Realtime), but the WASI guest often provides a Monotonic timestamp.

Because the monotonic time (e.g., uptime) is significantly smaller than the current wall-clock time (since epoch),  interpreted the timer as having already expired. This caused  to return immediately, creating a busy loop that consumed 100% CPU.

This commit fixes the issue by:
1. Identifying the correct clock source (Realtime vs Monotonic).
2. Manually calculating the relative timeout duration (Target - Current).
3. Using a relative timer in  instead of an absolute one.

This results in correct blocking behavior and near 0% CPU usage during sleeps.

Fixes #3851

Also attaching the Full Reproduction Logs
1. https://gist.github.com/blackdragoon26/1901207b895530afecae00f88ce13940
2. https://gist.github.com/blackdragoon26/54208ee0b8f85e38c63d12325ef260cc

### DIFF
--- a/.github/ISSUE_TEMPLATE/gsoc_idea.yml
+++ b/.github/ISSUE_TEMPLATE/gsoc_idea.yml
@@ -60,6 +60,11 @@ body:
       placeholder: According to the GSoC policy, it **requires** at least two mentors with one being a primary mentor.
     validations:
       required: true
+  - type: markdown
+    attributes:
+      value: |
+        ### Important Note
+        Before you apply, please make sure you have read and understood the [WasmEdge Mentorship Program Policy](https://github.com/WasmEdge/WasmEdge/blob/master/docs/mentorship-policy.md).
   - type: textarea
     id: apply-link
     attributes:
--- a/.github/ISSUE_TEMPLATE/lfx_mentorship_idea.yml
+++ b/.github/ISSUE_TEMPLATE/lfx_mentorship_idea.yml
@@ -52,6 +52,11 @@ body:
       placeholder: According to the CNCF policy, at least two mentors(one for major, another for backup) are required for each project.
     validations:
       required: true
+  - type: markdown
+    attributes:
+      value: |
+        ### Important Note
+        Before you apply, please make sure you have read and understood the [WasmEdge Mentorship Program Policy](https://github.com/WasmEdge/WasmEdge/blob/master/docs/mentorship-policy.md).
   - type: textarea
     id: apply-link
     attributes:
--- a/docs/mentorship-policy.md
+++ b/docs/mentorship-policy.md
@@ -1,0 +1,29 @@
+# WasmEdge Mentorship Program Policy
+
+Welcome to the WasmEdge community! This document outlines the basic policies and guidelines for applicants and participants of mentorship programs such as LFX Mentorship and Google Summer of Code (GSoC).
+
+## Application Phase
+
+To ensure fairness, the following actions are **strictly prohibited** before the application period concludes:
+
+1. **No Private Contact:** Applicants must not privately contact mentors via email, Discord DM, or CNCF Slack DM during the application period to seek assistance or reviews beyond what is available in public channels.
+2. **No AI Spam:** Applicants must not submit AI-generated spam ("AI slop") to the repository. This includes, but is not limited to, meaningless or fabricated Issues, Pull Requests, and Discussions.
+3. **No Plagiarism:** Applicants must not plagiarize others' work or public proposals and implementations.
+
+### Communication and Discussion
+
+If communication or discussion regarding relevant topics is needed, please observe the following:
+
+1. **General Topics:** Broad questions should be asked in GitHub Discussions, the Discord mentoring channel, or the Slack CNCF `#wasmedge` channel.
+2. **Project Ideas:** Questions related to specific project ideas should be discussed directly under the relevant GitHub Issue.
+3. **Program/Administrative Inquiries:** Issues related to specific programs (e.g., LFX Mentorship or GSoC) or financial matters should be directed to the corresponding organizing body.
+
+## During the Mentorship Program
+
+Once accepted, mentees have the following obligations:
+
+1. **Initialization:** At the start of the program, create an issue using the appropriate LFX/GSoC issue template. This issue must detail the task's timeline and milestones.
+2. **Weekly Updates:** Provide progress updates in the designated issue every week.
+3. **Timely Communication:** If you encounter difficulties or require assistance, communicate with your mentor immediately within the issue.
+4. **Schedule Management:** If you anticipate events that will significantly impact your progress (such as midterm or final exams), you must inform your mentor in the issue **in advance** and update/communicate the new timeline.
+5. **Evaluations:** Please be mindful of the midterm and final evaluation schedules. Current programs have strict time limits. If your progress or Pull Requests do not meet the standards by the deadline, you will receive a failing ("reject") evaluation. No extensions will be granted.


blackdragoon26:
Fixed clang formatting issue, which failed rest 5 lints and skipped rest of actions.
Can you review and approve workflow again @ibmibmibm

lib/host/wasi/inode-macos.cpp
```diff
@@ -1406,7 +1406,7 @@ WasiExpect<void> Poller::prepare(Span<__wasi_event_t> E) noexcept {
   return {};
 }
 
-void Poller::clock(__wasi_clockid_t, __wasi_timestamp_t Timeout,
+void Poller::clock(__wasi_clockid_t ClockId, __wasi_timestamp_t Timeout,
```
hydai, line 1409: Since you are introducing a new variable here, what is its default value? What is the default behavior? Will it behave the same as the previous status?

lib/host/wasi/inode-macos.cpp
```diff
@@ -1419,13 +1419,28 @@ void Poller::clock(__wasi_clockid_t, __wasi_timestamp_t Timeout,
 
   uint32_t FFlags = NOTE_NSECONDS;
   if (Flags & __WASI_SUBCLOCKFLAGS_SUBSCRIPTION_CLOCK_ABSTIME) {
-#ifdef NOTE_ABSOLUTE
```
hydai, line 1422: Another thing is that you can find the `ABSTIME` appended in the `Flags`. If we don't use `NOTE_ABSOLUTE`, will it still be an `ABSTIME`? If not, this is not a correct way to handle it.

lib/host/wasi/inode-macos.cpp
```diff
@@ -1419,13 +1419,28 @@ void Poller::clock(__wasi_clockid_t, __wasi_timestamp_t Timeout,
...
-    FFlags |= NOTE_ABSOLUTE;
-#else
-    Event.Valid = true;
-    Event.error = __WASI_ERRNO_NOSYS;
-    return;
-#endif
+    timespec NowTS;
+    // WASI: 0 = Realtime, 1 = Monotonic. Map to macOS system clocks.
+    clockid_t SysClockId = (ClockId == 0) ? CLOCK_REALTIME : CLOCK_MONOTONIC;
```
PhantomInTheWire:
```suggestion
    clockid_t SysClockId = toClockId(ClockId);
```

PhantomInTheWire: this should make it a little less brittle here

blackdragoon26: Thanks alot, for this

blackdragoon26:
@hydai  Thank you for the detailed review!

Regarding `toClockId`: I agree, hardcoding `0` was brittle. I will update the code to use the helper `toClockId(ClockId)` as suggested by @PhantomInTheWire 

Regarding `ABSTIME`: You are absolutely correct that the flag implies an absolute wait. Ideally, we would just pass `NOTE_ABSOLUTE `to the kernel.

However, during testing, I discovered a limitation in macOS `kevent`: `NOTE_ABSOLUTE` seems to always check against the system's Realtime (Wall) clock. When a WASI guest uses a Monotonic clock (common for timers), the timestamp passed is essentially "uptime," which is much smaller than the Wall Clock time. The kernel interprets this as "time already passed" and returns immediately, causing the 100% CPU busy loop.

To strictly honor the behavior of `ABSTIME` (waiting until a specific moment) without triggering this kernel mismatch, I am manually calculating the delta: `Relative = Target_Absolute - Current_Time`.

This ensures the thread wakes up at the exact intended timestamp while keeping the CPU usage at 0%.

I have done corrections, let me retest it and verify in local, before pushing it in here.

lib/host/wasi/inode-macos.cpp
blackdragoon26: yes, valid point. 
I will update this to use the helper `toClockId(ClockId)` as suggested by the other reviewer. This ensures the default behavior and error handling match the rest of the WasmEdge codebase safely.

lib/host/wasi/inode-macos.cpp
blackdragoon26: Yep, You right that `ABSTIME` implies an absolute wait. To ensure we strictly honor that behavior, while avoiding the macOS kernel bug with Monotonic clocks
I will update the logic to manually calculate the relative duration (`Target - Current`).

blackdragoon26:
Also, I am facing another issue about slight inconsistency in testing the 100% cpu usage
Could you guide me on what would be best approach to log and compare the cpu usage.
Thanks!

hydai:
Could you describe the issue you are encountering? Also, what is the current method you are using for benchmarking it?

lib/host/wasi/inode-macos.cpp
hydai: I found that, in this case, some of the behavior is missing. It seems like the clockid is not being honored. Could you please check the Wasmtime implementation of the same WASI function and figure out how to check it correctly?

Ref:
- https://github.com/bytecodealliance/wasmtime/blob/main/crates/wasi-preview1-component-adapter/src/lib.rs#L2147-L2193
- https://github.com/bytecodealliance/wasmtime/blob/ecd1cb3d96f046cbf1a8cc2e8ea82eb7c0205fde/crates/wasi/src/p1.rs#L2349-L2370

blackdragoon26:
> Could you describe the issue you are encountering? Also, what is the current method you are using for benchmarking it?

hi @hydai 
tldr, could you actually verify whether this issue exist anymore?

Here is my detailed method of what I am trying to do over here

I changed file format a bit of https.rs
```
use tokio::time::{sleep, Duration};

#[tokio::main(flavor = "current_thread")]
async fn main() -> Result<(), reqwest::Error> {

    println!("Test started. Sleeping for 10 seconds to trigger busy loop...");
    
    // If the bug is present, CPU should hit 100% here.
    sleep(Duration::from_secs(10)).await;

    println!("Sleep done. Making request...");

    // Some simple CLI args requirements...
    let url = "https://httpbin.org/delay/10";

    eprintln!("Fetching {:?}...", url);

    let res = reqwest::get(url).await?;

    eprintln!("Response: {:?} {}", res.version(), res.status());
    eprintln!("Headers: {:#?}\n", res.headers());

    let body = res.text().await?;
    println!("GET: {}", body);

    let client = reqwest::Client::new();

    let res = client
        .post("https://eu.httpbin.org/post")
        .body("msg=WasmEdge")
        .send()
        .await?;
    let body = res.text().await?;

    println!("POST: {}", body);

    let res = client
        .put("https://eu.httpbin.org/put")
        .body("msg=WasmEdge")
        .send()
        .await?;
    let body = res.text().await?;

    println!("PUT: {}", body);

    Ok(())
}
```

and then compiled it via
(had to set proper path for homebrew issue)
```
AR=/opt/homebrew/opt/llvm/bin/llvm-ar \
CC=/opt/homebrew/opt/llvm/bin/clang \
cargo build --target wasm32-wasip1 --release
```


I also ensure i delete the build and then only perform changes and then only do testing with updated build
i.e. 
```
cd build
rm -rf *
export LLVM_PATH="$(brew --prefix llvm@18)"                                 
unset CC CXX LDFLAGS CPPFLAGS

# 3. Configure with ALL required paths
# Added: -DLLVM_DIR (Critical fix)
cmake -DCMAKE_BUILD_TYPE=Release -DWASMEDGE_BUILD_TESTS=OFF \
      -DLLVM_DIR="${LLVM_PATH}/lib/cmake/llvm" \
      -DLLD_DIR="${LLVM_PATH}/lib/cmake/lld" \
      -DCMAKE_CXX_FLAGS="-isystem ${LLVM_PATH}/include -Wno-error -Wno-unused-parameter -Wno-unknown-warning-option" \
      -DCMAKE_EXE_LINKER_FLAGS="-L${LLVM_PATH}/lib -L${LLVM_PATH}/lib/c++" \
      -DCMAKE_SHARED_LINKER_FLAGS="-L${LLVM_PATH}/lib -L${LLVM_PATH}/lib/c++" \
      ..

# 4. Build
make -j$(sysctl -n hw.ncpu)
```
and doing testing via
these commands respectively for old(original) and new(pr fixed) `inode-macos.cpp`

below is one of the example of 
```
WASMEDGE_OLD="../WasmEdge/build/tools/wasmedge/wasmedge"
$WASMEDGE_OLD target/wasm32-wasip1/release/https.wasm &
PID=$!
sleep 3
echo "--- NEW BUILD STATS ---"
ps -p $PID -o %cpu
sample $PID 1 -f sample_old.txt
kill $PID
```

surprisingly its output is same as old and new , i.e. 0.0%
```
[1] 23821
Test started. Sleeping for 10 seconds to trigger busy loop...
--- NEW BUILD STATS ---
 %CPU
  0.0
Sampling process 23821 for 1 second with 1 millisecond of run time between samples
Sampling completed, processing symbols...
Sample analysis of process 23821 written to file sample_old.txt
[1]  + 23821 terminated  $WASMEDGE_OLD target/wasm32-wasip1/release/https.wasm  
```

here is gist of 
sample_old.txt : https://gist.github.com/blackdragoon26/5cc165a48406d79ee8f38a828657943b
sample_new.txt: https://gist.github.com/blackdragoon26/118453fe01bcc6299c9eef90cce13513


Also I had some other testing methods in past too which i attempted
https://gist.github.com/blackdragoon26/82e78e90682c9eecd8406d71e3023f4c
https://gist.github.com/blackdragoon26/01a7c168827523923f63a3f5e91ef346

lib/host/wasi/inode-macos.cpp
blackdragoon26: Sure, I will have look in there!

blackdragoon26:
https://github.com/user-attachments/assets/ce644437-5e2e-461f-8c5a-447b19b26f32

Could you assess the top 2 processes happening, and I just wanted to know your thoughts @hydai of pid 27562 and pid 5999657
In this video recording, I am testing poller::clock fn inside with `std::chrono::steady_clock` which is unpushed as of now. 

I am just trying to figure out the underlying cause as of now
and also I have overcome testing inconsistency issues,
I just use htop now and don't follow any particular script fashion.

hydai:
Hi @blackdragoon26,  
Could you do me a favor and package these scripts into a simple repository so I can clone it and reproduce the entire process in your environment, including the Rust application you used?
Ref: https://github.com/WasmEdge/WasmEdge/pull/4470#issuecomment-3713441846

blackdragoon26:
> Hi @blackdragoon26, Could you do me a favor and package these scripts into a simple repository so I can clone it and reproduce the entire process in your environment, including the Rust application you used? Ref: [#4470 (comment)](https://github.com/WasmEdge/WasmEdge/pull/4470#issuecomment-3713441846)

Hihi @hydai ,
I could n't be more happier than I am right now, with the fix in CPU usage!
Okay so those two links you gave about not honouring the clock were huge deal, thanks for that, 
Although even after that cpu was stuck at 100, but later after digging deeper with debug logs, I discovered a second issue regarding "zombie" event filters.

So as of now, two things fixed:
1. `Poller::wait` (The "Zombie" Fix)
What it fixed: The 100% CPU Loop.
The socket was stuck in "Write Mode" even after `kqueue` filter even after the application had finished writing 
`kevent` kept waking up instantly bcoz the buffer was empty (and thus writable), causing an infinite tight loop. I added logic to explicitly remove these stale events.

2. `Poller::clock` (The "Logic" Fix)
What it fixed: Correctness 
The original code completely ignored the `ClockId` argument (Realtime vs Monotonic) and relied on a macOS flag (`NOTE_ABSOLUTE`) that is known to behave inconsistently with Monotonic clocks. The fix now honors the `ClockId` and manually calculates the timeout for absolute subscriptions.

With much better satisfaction of current state of PR now,
Kindly review the changes.

Also do let me know, if you still want me to create separate package for testing related.

lib/host/wasi/inode-macos.cpp
```diff
@@ -1417,15 +1417,34 @@ void Poller::clock(__wasi_clockid_t, __wasi_timestamp_t Timeout,
...
+        static_cast<__wasi_timestamp_t>(NowTS.tv_nsec);
+
+    // Calculate the Relative Timeout (Target - Now)
+    if (Timeout > CurrentTimeNS) {
+      Timeout = Timeout - CurrentTimeNS;
+    } else {
+      Timeout = 0; // The time has already passed, return immediately.
+    }
+    // IMPORTANT: We do NOT add NOTE_ABSOLUTE here.
+    // We have already converted it to a relative duration.
```
hydai: The behavior should vary based on the clock ID. Do you handle them?

blackdragoon26: > The behavior should vary based on the clock ID. Do you handle them?

Right now I'm just handling them uniformly.
I assumed the "manual calculation" approach (target - now) was safer because `NOTE_ABSOLUTE` behaves badly with Monotonic clocks on macOS (which causes the 100% CPU loop).

But you're right, by converting everything to relative, `CLOCK_REALTIME` won't respect system time jumps like NTP updates during the sleep.

So, should I split the logic? I can use the native `NOTE_ABSOLUTE` for Realtime so it handles time jumps correctly and keep this manual calculation for Monotonic to prevent the CPU bug.

hydai: It sounds incorrect; you should split the logic.

lib/host/wasi/inode-macos.cpp
```diff
@@ -1417,15 +1417,40 @@ void Poller::clock(__wasi_clockid_t, __wasi_timestamp_t Timeout,
...
+    bool HandledNative = false;
+
+    // 1. If Realtime, use native NOTE_ABSOLUTE if available.
+    if (SysClockId == CLOCK_REALTIME) {
 #ifdef NOTE_ABSOLUTE
-    FFlags |= NOTE_ABSOLUTE;
-#else
-    Event.Valid = true;
-    Event.error = __WASI_ERRNO_NOSYS;
-    return;
```
hydai, line 1427: This part is returning an error to indicate that if users do want REALTIME but NOTE_ABSOLUTE is not available, it will show __WASI_ERRNO_NOSYS.
I think we should keep it.

hydai (changes requested):

lib/host/wasi/inode-macos.cpp
blackdragoon26: Alrighty. 
I have updated the logic to strictly use `NOTE_ABSOLUTE` for `CLOCK_REALTIME` and return `NOSYS` if it's unavailable.
For `CLOCK_MONOTONIC`, I kept the manual relative calculation to avoid the macOS kernel bug.

blackdragoon26:
hihi @hydai , all checks passed yesternight, could your review the PR?

blackdragoon26:
@ibmibmibm , Kindly review my PR.

> LGTM. @ibmibmibm, would you like to take a look?

ibmibmibm (approved):

hydai:
Thanks.

PR #4473, opened 2025-12-29
feat(WASI-NN,ggml): bump llama.cpp b7569

LFsWang:


q82419:
Bumped into a newer version

PR #4483, opened 2026-01-04
feat(docker): add Dockerfile for PyTorch 2.4 build (Fixes #4482)

### LINKED ISSUE #4482: feat: [Feature] Add Dockerfile for reproducible WASI-NN PyTorch 2.4 build
### Summary

Currently, building the WASI-NN plugin with the PyTorch backend is challenging for new contributors. The documentation often references older PyTorch versions (1.8.2), but the master branch and newer LLVM versions (Clang 14+) work better with newer PyTorch ABIs.

Attempting to build manually often leads to dependency errors regarding `Torch_DIR` paths and `llvm` version mismatches.

### Details

I propose adding a dedicated `Dockerfile` that automates the build environment on Ubuntu 22.04. This Dockerfile will:
1. Install Clang-14 and LLVM-14.
2. Download and configure LibTorch 2.4.0 (CPU).
3. Build WasmEdge with `-DWASMEDGE_PLUGIN_WASI_NN_BACKEND="PyTorch"` enabled.

**Additional Context**
I have verified this build process locally and confirmed that the generated plugin successfully runs the MobileNet image classification example.

### Appendix

_No response_

yasw175:
### Description
This PR addresses issue #4482.

The current documentation and build scripts reference older PyTorch versions (1.8.2), which can cause ABI compatibility issues with newer LLVM/Clang versions.

I have added `Dockerfile.pytorch24` which provides a reproducible build environment on Ubuntu 22.04. It automates:
- Installation of Clang-14 and LLVM-14.
- Setup of LibTorch 2.4.0 (CPU).
- Compilation of WasmEdge with the WASI-NN PyTorch backend enabled.

### Verification
I have verified this locally by building the image and running the `wasmedge --version` command to confirm the plugin is loaded.


utils/docker/Dockerfile.ubuntu-plugins-deps
```diff
@@ -52,7 +52,7 @@ ENV OPENCV_VERSION="4.8.0"
 RUN [ "/bin/bash", "install-opencvmini.sh" ]
 
 COPY wasi-nn/install-pytorch.sh .
-ENV PYTORCH_VERSION="2.5.1"
+ENV PYTORCH_VERSION="2.4.0"
```
hydai, line 55: This is so weird. We are using 2.5.1 now; what is the reason for downgrading the version?

yasw175: @hydai The reason for the change is that the current setup with PyTorch 2.5.1 was failing to link correctly in my local build environment (Ubuntu 22.04, Clang 14), causing ABI compatibility issues with the WASI-NN plugin.

I found that downgrading to 2.4.0 (cxx11-abi) resolved those linker errors and allowed the plugin to build and verify successfully (wasmedge --version shows the plugin loaded).

If you prefer to keep 2.5.1, I am happy to revert the version change, but I might need guidance on the correct build flags for 2.5.1 to avoid the ABI mismatch. Should I revert this file and only keep the install-pytorch.sh fixes?

hydai: It's strange; I don't think we should downgrade the version, as it doesn't make much sense. Could you specify which type of version you're using for PyTorch 2.5.1: the cxx11-abi version or the non-cxx11-abi version?

yasw175: I was using the cxx11-abi version. The install-pytorch.sh script defaults to libtorch-cxx11-abi, so that is exactly what was used.

Even with the correct ABI version of 2.5.1, I encountered persistent linker errors on Ubuntu 22.04. Switching to 2.4.0 (also cxx11-abi) immediately resolved the build issues.

If you prefer to keep 2.5.1, I am happy to revert this specific change and share the failure logs so we can debug why the 2.5.1 build is breaking in this environment. Let me know how you'd like to proceed.

hydai: We should keep 2.5.1 for the new features.

yasw175: Understood. I will revert the version back to 2.5.1 since new features depend on it.

However, the build is currently failing on Ubuntu 22.04 with 2.5.1 (linker errors). Since you likely have a working setup, could you point me to the correct dependencies or CMake flags needed for 2.5.1?

I am happy to implement those fixes in this PR so that 2.5.1 builds reliably for everyone.

yasw175: Understood. I wasn't aware 2.5.1 was a hard requirement for new features.

Since downgrading isn't an option, I will close this PR for now to avoid clutter. Per our discussion on Discord, I'm going to shift focus to studying the WASI-HTTP plugin architecture first to better understand the system.

Once I've leveled up on the plugin system, I'll circle back to debug the PyTorch 2.5.1 linker issues properly. Thanks for the guidance @hydai!

PR #4501, opened 2026-01-08
feat(docker): Add Dockerfile for verifiable WASI-NN PyTorch 2.5.1 build

yasw175:
This adds a dedicated Dockerfile to build WasmEdge with the WASI-NN PyTorch backend enabled.

As discussed with @hydai, this avoids downgrading to PyTorch 2.4 and instead verifies that the build works correctly with the `cxx11-abi` version of PyTorch 2.5.1 on master.

Changes:
- Adds `utils/docker/Dockerfile.wasi-nn-pytorch`.
- Uses the existing `utils/wasi-nn/install-pytorch.sh` script to correctly fetch the `cxx11-abi` version.
- Verifies the build of `libwasmedgePluginWasiNN.so`.

This ensures we have a reproducible build environment for the current master branch.

### DIFF
--- a/utils/docker/Dockerfile.wasi-nn-pytorch
+++ b/utils/docker/Dockerfile.wasi-nn-pytorch
@@ -1,0 +1,39 @@
+# Use the standard WasmEdge development base image
+FROM wasmedge/wasmedge:latest
+
+# Install necessary build dependencies
+RUN apt-get update && apt-get install -y \
+    build-essential \
+    curl \
+    wget \
+    unzip \
+    cmake \
+    libopencv-dev \
+    libxml2-dev \
+    libz-dev \
+    && rm -rf /var/lib/apt/lists/*
+
+# Setup PyTorch 2.5.1 using the project's install script
+WORKDIR /usr/local
+COPY utils/wasi-nn/install-pytorch.sh .
+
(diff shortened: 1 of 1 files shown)


yasw175:
Hi @hydai, thanks for the review!

You are completely right that the existing ubuntu-plugins-deps Dockerfile correctly installs PyTorch 2.5.1.

However, that file seems to be a dependency environment only—it installs the libraries but does not verify the build by running cmake and make.

The point of this PR is to provide a Dockerfile that performs the full end-to-end build of the wasmedgePluginWasiNN target using those 2.5.1 dependencies. This validates that the cxx11-abi linking actually works on master (which was the original concern in issue #4482) and gives users a ready-to-use image with the plugin compiled.

If you prefer, I can modify the existing Dockerfile to include the build steps instead of adding a new file?

hydai:
This Docker image is produced and used in the workflows that build the PyTorch plugins. If there is an issue, it will fail. Adding a redundant verification is just unnecessary.

yasw175:
Understood. I missed that the existing CI workflows already utilize the plugins-deps image to verify the build and linking.

Thanks for clarifying that the coverage is already there. I'll close this PR to avoid the noise and switch my focus to the WASI-HTTP plugin architecture analysis as planned.

PR #4506, opened 2026-01-09
fix(zlib): standardize missing memory instance handling

vishal2005025:
### Summary

Standardize missing memory instance handling in the zlib host plugin.

The zlib implementation previously used ad-hoc null checks and logged internal messages such as `Frame.getMemoryByIndex(0) returned nullptr`. This PR replaces those checks with a small local macro that follows the same pattern already used by other WasmEdge plugins.

### Details

- Use a single `MEMINST_CHECK` macro to retrieve and validate the memory instance
- Emit a consistent, user-facing error message:
  `[WasmEdge-Zlib] Memory instance not found.`
- Keep the existing error code (`HostFuncError`) and control flow unchanged

### Impact

This is an internal cleanup only. Runtime behavior, APIs, and return values remain the same; the change improves consistency and log clarity within the zlib plugin.

### DIFF
--- a/utils/ffmpeg/download-ffmpeg-sample-video.sh
+++ b/utils/ffmpeg/download-ffmpeg-sample-video.sh
@@ -12,9 +12,6 @@ fi
 if [ ! -d $TODIR ]; then
     mkdir $TODIR
 fi
-if [ ! -d $TODIR ]; then
-    mkdir $TODIR
-fi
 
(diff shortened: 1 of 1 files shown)


vishal2005025:
@hydai @LFsWang @q82419 Please take a look and let me know if any changes are required.

q82419:
LGTM

PR #4518, opened 2026-01-14
feat: Add Amazon Linux 2 CI build for WASI-NN ggml plugin with CUDA support

### LINKED ISSUE #3034: feat: enable ggml plugin support on Amazon Linux 2
### Summary

Some users work on AWS Deep Learning AMI based on Amazon Linux 2. When they tried our wasmedge+ggml examples, they failed due to several incompatible issues.
To avoid these kinds of issues happening again, we would like to have a new release and building assets for the Amazon Linux 2.

### Details

Major:

- [ ] Build WasmEdge on Amazon Linux 2
- [ ] Build ggml plugin with CUDA enabled option on Amazon Linux 2

Minor:

- [ ] Support the remaining plugins

### Appendix

- [Amazon Linux docker images](https://hub.docker.com/_/amazonlinux/)

MUFFANUJ:
closes #3034 

This PR adds CI support to build WasmEdge and the WASI-NN ggml plugin with CUDA on Amazon Linux 2, so users on AWS Deep Learning AMIs (which use glibc 2.26) can run the plugin without compatibility issues. It introduces a new reusable GitHub Actions workflow that builds inside an amazonlinux:2 container, checks that the resulting plugin does not depend on newer glibc versions, and produces clearly named Amazon Linux 2 artifacts. The workflow is wired into the existing CI and release pipelines so future changes automatically stay compatible with Amazon Linux 2.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -107,6 +107,15 @@ jobs:
       matrix: "[{'name':'manylinux_2_28 x86_64','runner':'ubuntu-latest','docker_tag':'manylinux_2_28_x86_64','asset_tag':'manylinux_2_28_x86_64'},
                 {'name':'manylinux_2_28 aarch64','runner':'ubuntu-24.04-arm','docker_tag':'manylinux_2_28_aarch64','asset_tag':'manylinux_2_28_aarch64'}]"
 
+  build_on_amazonlinux2:
+    permissions:
+      contents: write
+    needs: [get_version, lint]
+    name: AmazonLinux2
+    uses: ./.github/workflows/reusable-build-on-amazonlinux2.yml
+    with:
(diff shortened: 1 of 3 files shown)


hydai:
Hi @MUFFANUJ,  
Thanks for this PR. Amazon Linux 2 will end its security support in five months. I don't think we will still want to maintain it after it reaches EOL soon.  

I also found that Amazon Linux 2023 uses glibc 2.34, which is already supported by manylinux2_28 (glibc 2.28). So, I think it's perfectly fine to do nothing about the Amazon Linux distribution. Thanks. Ref: https://docs.aws.amazon.com/linux/al2023/ug/core-glibc.html

MUFFANUJ:
hi @hydai thanks, it’s reasonable to avoid adding new maintenance burden for it. Closing the PR.

PR #4521, opened 2026-01-15
chore(AI): provide AGENTS.md for the AI co-working toolchains

hydai:
Since more and more people are using AI tools to help build better applications, it's time for us to provide policies and helper files so these toolchains can follow them and ensure they do not generate more AI slop within this project.

This is a draft PR, and we welcome any contributors to discuss. Suggestions and comments are also encouraged.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,19 @@
+## Description
+<!-- Describe your changes here -->
+
+## Checklist
+
+Before submitting this PR, please ensure the following:
+
+- [ ] **DCO Signed-off**: All commits are signed-off (`git commit -s`). CI workflows will only be approved if DCO check is passed.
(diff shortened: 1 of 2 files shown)


PhantomInTheWire (commented):
how about adding a rule about logging conventions in the project? for example using sv (string view) for all logs until it has a good reason not to

PhantomInTheWire (commented):
Many times, the AI agent ends up wasting tokens trying to discover the correct paths for binaries, libraries, and plugins in order to manually verify that things work. Providing a well-defined, documented default path for the WasmEdge binary, for example <build_folder>/tools/wasmedge/wasmedge, would help avoid this.

In addition, expanding the documentation around how to set the library and plugin paths and env vars, along with a clear explanation of how the various plugin related stuff works, would make verification and debugging significantly easier for both humans and ai agents.

hydai:
Sounds not bad; I will try to include these in a section.

hydai:
This is ready for review. cc all maintainers

/vote

PR #4522, opened 2026-01-15
fix(AOT): align the NaN propagation order in {f32x4,f64x2}.max/min with the interpreter mode

### LINKED ISSUE #4257: bug: WasmEdge compiled version gives different output compared to other runtimes
### Summary

I executed the following test case with WAMR and WasmEdge runtimes. The WAMR runtime and WasmEdge's JIT and Interpreter modes output the same number. But, WasmEdge's compiled mode outputs a different number.

Test case:

```
(module
  (type (;0;) (func))
  (type (;1;) (func (result v128)))
  (export "main" (func 0))
  (export "to_test" (func 0))
  (func (;0;) (type 1) (result v128)
    v128.const i32x4 0xffffffff 0xffffffff 0xffffffff 0xfff7ffff
    v128.const i32x4 0xffffadff 0xffffff0f 0xffffffff 0xffffffff
    f32x4.max
  )
)
```

### Current State

Outputs: `340240828546070184842346363461134450687`

### Expected State

Should output: `340282366920938463463374606400976039423`

### Reproduction steps

Compile: `wasmedge compile test.wasm test.so`
Run: `wasmedge --enable-all run test.so to_test`

### Screenshots

_No response_

### Components

CLI

0.14.1-271-ga89c176f

Ubuntu 22.04

### Hardware Architecture

x86_64

_No response_

asmit27rai:
The AOT compiler's implementation of SIMD floating-point min/max
operations had the NaN handling select operations in the wrong order,
causing different results compared to the interpreter when both
operands contain NaN values.

The interpreter processes NaN checks as:
1. If LHS is NaN, use LHS
2. If RHS is NaN, use RHS (wins if both are NaN)

The AOT compiler was incorrectly doing:
1. If RHS is NaN, use RHS
2. If LHS is NaN, use LHS (incorrectly wins if both are NaN)

This fix swaps the order of NaN handling select operations in
compileVectorVectorFMin and compileVectorVectorFMax to match the
interpreter's behavior.

Fixes #4257

### DIFF
--- a/.github/workflows/IWYU_scan.yml
+++ b/.github/workflows/IWYU_scan.yml
@@ -59,7 +59,6 @@ jobs:
       run: |
         # Retrieve annotated tags. Details: https://github.com/actions/checkout/issues/290
         git fetch --tags --force
-        git config --global --add safe.directory $(pwd)
         echo "version=$(git describe --match '[0-9].[0-9]*' --tag)" >> $GITHUB_OUTPUT
 
   build_fedora:
--- a/.github/workflows/build-extensions.yml
+++ b/.github/workflows/build-extensions.yml
@@ -162,10 +162,6 @@ jobs:
       - name: Ensure git safe directory
         run: |
           git config --global --add safe.directory $(pwd)
-      - name: Install dependency
-        uses: crazy-max/ghaction-chocolatey@2526f467ccbd337d307fe179959cabbeca0bc8c0 # v3.4.0
-        with:
-          args: install cmake ninja vswhere
       - uses: GuillaumeFalourd/setup-windows10-sdk-action@9979bedbf6ae74f038cc338622ba015942ed2408 # v2.4
         with:
           sdk-version: 26100
--- a/.github/workflows/build_for_openwrt.yml
+++ b/.github/workflows/build_for_openwrt.yml
@@ -67,7 +67,7 @@ jobs:
       uses: actions/cache/restore@8b402f58fbc84540c8b491a91e594a4576fec3d7 # v5.0.2
       with:
         path: ./openwrt-x86-64-generic-rootfs.tar.gz
-        key: openwrt
+        key: openwrt-${{ hashFiles('.github/workflows/build_for_openwrt.yml') }}
     - name: Install dependency
       if: steps.restore-openwrt.outputs.cache-hit != 'true'
       run: |
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -26,7 +26,6 @@ jobs:
       contents: write
     outputs:
       version: ${{ steps.prep.outputs.version }}
-      upload_url: ${{ steps.create_release.outputs.upload_url }}
     steps:
       - name: Harden the runner (Audit all outbound calls)
         uses: step-security/harden-runner@e3f713f2d8f53843e71c69a996d56f51aa9adfb9 # v2.14.1
--- a/.github/workflows/reusable-build-extensions-on-macos.yml
+++ b/.github/workflows/reusable-build-extensions-on-macos.yml
@@ -125,10 +125,6 @@ jobs:
           name: ${{ steps.var.outputs.artifact }}
           path: ${{ steps.var.outputs.artifacts }}
       - if: ${{ inputs.release }}
-        name: Install gh for release
-        run: |
-          brew install gh
-      - if: ${{ inputs.release }}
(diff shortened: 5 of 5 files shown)


hydai:
Hi @asmit27rai,  
Have you tested and verified the results, or is it possible to add a test case to check this behavior?

asmit27rai:
@hydai I added one.
Please have a look.

test/llvm/CMakeLists.txt
```diff
@@ -12,3 +12,17 @@ target_link_libraries(wasmedgeLLVMCoreTests
...
+)
+
+add_test(wasmedgeSIMDNaNTest wasmedgeSIMDNaNTest)
+
+target_link_libraries(wasmedgeSIMDNaNTest
+  PRIVATE
+  ${GTEST_BOTH_LIBRARIES}
+  wasmedgeLLVM
+  wasmedgeVM
+)
```
hydai: The test case should be inside one of the previous suites instead of creating a new one.

test/llvm/SIMDNaNTest.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+std::array<WasmEdge::Byte, 88> SIMDNaNTestWasm{
+    0x00, 0x61, 0x73, 0x6d, 0x01, 0x00, 0x00, 0x00, 0x01, 0x05, 0x01, 0x60,
+    0x00, 0x01, 0x7b, 0x03, 0x02, 0x01, 0x00, 0x07, 0x16, 0x01, 0x12, 0x74,
+    0x65, 0x73, 0x74, 0x5f, 0x66, 0x33, 0x32, 0x78, 0x34, 0x5f, 0x6d, 0x61,
+    0x78, 0x5f, 0x6e, 0x61, 0x6e, 0x00, 0x00, 0x0a, 0x2b, 0x01, 0x29, 0x00,
+    0xfd, 0x0c, 0x01, 0x00, 0xc0, 0x7f, 0x01, 0x00, 0xc0, 0x7f, 0x01, 0x00,
+    0xc0, 0x7f, 0x01, 0x00, 0xc0, 0x7f, 0xfd, 0x0c, 0x00, 0x00, 0xc0, 0x7f,
+    0x00, 0x00, 0xc0, 0x7f, 0x00, 0x00, 0xc0, 0x7f, 0x00, 0x00, 0xc0, 0x7f,
+    0xfd, 0xe9, 0x01, 0x0b};
```
hydai: You must provide the WAT as a comment to explain this test asset.

asmit27rai:
@hydai Addressed your comments.

PhantomInTheWire:
I don't think you addressed this comment https://github.com/WasmEdge/WasmEdge/pull/4522#discussion_r2704957259
also DCO is failing and your commits are not in format, please read CONTRIBUTING.md

asmit27rai:
@PhantomInTheWire 

> I don't think you addressed this comment [#4522 (comment)](https://github.com/WasmEdge/WasmEdge/pull/4522#discussion_r2704957259) also DCO is failing and your commits are not in format, please read CONTRIBUTING.md

Please check i missed that in commit.
Now i squashed all the commit and make it DCO signoff and commitlint.
Thank You.

test/llvm/LLVMcoreTest.cpp
```diff
@@ -591,6 +592,85 @@ TEST(Configure, ConfigureTest) {
...
+  ASSERT_TRUE(VM.instantiate());
+
+  auto Result = VM.execute("test_f32x4_max_nan");
+  ASSERT_TRUE(Result);
+  ASSERT_EQ((*Result).size(), 1U);
+
+  auto ResultVal = (*Result)[0].first.get<WasmEdge::uint128_t>();
+
+  uint32_t Lanes[4];
+  std::memcpy(Lanes, &ResultVal, 16);
```
hydai: We are no longer accepting `memcpy`; instead, we are using `std::copy_n`.

asmit27rai:
@hydai I addressed your recent comment.
Thanks.

test/llvm/LLVMcoreTest.cpp
```diff
@@ -26,6 +26,7 @@
 #include <array>
 #include <chrono>
 #include <cstdint>
+#include <cstring>
```
hydai: Why do we need `cstring`?

asmit27rai: Removed it.

Issue #3450, opened 2024-06-04
feat: Add an option to cli/sdk to change the logging level


hydai:
### Summary

Currently, the WasmEdge CLI and SDK only provide limited logging levels, and we should extend that to a complete list so that users or developers can control them.

Also, our current warning message is sent to stdout, which should be redirected to stderr to avoid making the stdout mess.

### Details

* Support Debug, Info, Warning, Error, Fatal logging level
* Send level = {Warning, Error, Fatal} to stderr
* Send level = {Debug, Info} to stdout
* Allow users to enable/disable them

vishal2005025:
@hydai I’ve completed Phase 1 for the --log-level option.

The flag works with both wasmedge and wasmedge run, supports all documented levels, defaults to info, and routes logs to stdout/stderr as expected. Invalid values are handled with a clear error.

I’ve opened PR#4525 with only the required changes. Happy to adjust anything if needed before moving on to Phase 2.

PR #4525, opened 2026-01-16
feat(cli): add CLI option to control logging level

vishal2005025:
### Summary

This PR adds a `--log-level` option to the WasmEdge runtime CLI, allowing users to control logging verbosity at runtime.

The option integrates with the existing logging infrastructure and applies the selected level during tool startup.

### What’s Added

- New `--log-level` CLI option for the runtime tool
- Runtime-level control over log verbosity
- Default behavior preserved (`info` when unspecified)
- Proper stdout/stderr routing based on log severity

## Testing

- Verified `--log-level` is visible in both `wasmedge --help` and `wasmedge run --help`
- Verified all supported levels: `off`, `trace`, `debug`, `info`, `warning`, `error`, `fatal`
- Confirmed default log level is `info`
- Confirmed stdout/stderr routing works as expected
- Verified invalid values are rejected with a clear error message
- Confirmed `compile` subcommand is unaffected

## Notes

This PR completes Phase 1 of the logging level control enhancement. 
fixes: #3450 

Screenshot 2026-01-18 235623


vishal2005025:
@hydai @LFsWang @q82419  Please take a look and let me know if any changes are required.

vishal2005025:
@hydai  @ibmibmibm 
Please take a look when you get a chance.
I’ve added terminal screenshot in the PR description for verification also all the supported levels ,defaults to info, and routes logs to stdout/stderr has been verified and is working as expected.

Thanks!

include/common/spdlog.h
```diff
@@ -35,6 +35,14 @@ void setErrorLoggingLevel();
...
 void setCriticalLoggingLevel();
 
+/// Set logging level from string (debug, info, warning, error, fatal)
+/// Returns true if level was set successfully, false if invalid level string
+bool setLoggingLevelFromString(std::string_view Level);
+
+/// Configure spdlog with dual sinks: stdout for debug/info, stderr for
+/// warn/error/critical
+void configureDualSinkLogger();
+
```
hydai: This is over-designed.

lib/common/spdlog.cpp
```diff
@@ -16,9 +16,13 @@
 #ifdef _WIN32
 #include <spdlog/sinks/wincolor_sink.h>
 using color_sink_t = spdlog::sinks::wincolor_stdout_sink_mt;
+using color_stdout_sink_t = spdlog::sinks::wincolor_stdout_sink_mt;
+using color_stderr_sink_t = spdlog::sinks::wincolor_stderr_sink_mt;
```
hydai: No need to do this.

lib/common/spdlog.cpp
```diff
@@ -40,6 +44,51 @@ void setErrorLoggingLevel() { spdlog::set_level(spdlog::level::err); }
...
+    setWarnLoggingLevel();
+    return true;
+  } else if (Level == "error"sv) {
+    setErrorLoggingLevel();
+    return true;
+  } else if (Level == "fatal"sv || Level == "critical"sv) {
+    setCriticalLoggingLevel();
+    return true;
+  }
+  return false;
```
PhantomInTheWire, line 72: a switch statement would be much better here

vishal2005025:
@hydai @PhantomInTheWire  
I’ve addressed all the review comments. Please take another look and let me know if any changes are required.
Thanks!

lib/common/spdlog.cpp
```diff
@@ -15,17 +15,44 @@
...
-using color_sink_t = spdlog::sinks::ansicolor_stdout_sink_mt;
 #endif
 
 using namespace std::literals;
 
 namespace WasmEdge {
 namespace Log {
 
+namespace {
+// Anonymous namespace: internal helper for dual-sink logger setup
```
hydai: We don't need the `dual-sink`. There is no reason to change the original behavior, since this PR is only related to "ADD CLI OPTION".

vishal2005025:
@hydai I’ve updated the PR to strictly limit the scope to adding the --log-level CLI option only.
All dual-sink / stdout–stderr routing changes have been removed to preserve the original logging behavior.

If stdout/stderr separation is still desired, I can propose it separately in a dedicated PR to keep concerns clearly isolated.
Thanks for the clarification.

lib/common/spdlog.cpp
```diff
@@ -15,10 +15,8 @@
 #endif
 #ifdef _WIN32
 #include <spdlog/sinks/wincolor_sink.h>
-using color_sink_t = spdlog::sinks::wincolor_stdout_sink_mt;
```
hydai, line 18: It's pretty nonsensical to remove these. Are you using an AI tool to generate this pull request?  
Have you checked the code you wrote?  
It's so strange to do lots of unrelated things that don't align with the original goal.

vishal2005025:
@hydai Thanks for pointing this out.
Earlier iterations of this PR briefly included some logging structure changes because I initially tried to align with the broader context discussed in the original issue. That was a mistake on my side for this PR’s scope.
I’ve now reverted all unrelated changes and kept this PR strictly focused on adding the --log-level CLI option, without altering any existing logging behavior.

Apologies for the noise earlier, and thanks for your patience.

lib/driver/runtimeTool.cpp
```diff
@@ -240,7 +240,17 @@ ToolOnComponent(WasmEdge::VM::VM &VM, const std::string &FuncName,
...
+
+  // Set log level from CLI option (default is "info")
+  const std::string &Level =
+      Opt.LogLevel.value().empty() ? "info" : Opt.LogLevel.value();
+
+  if (!Log::setLoggingLevelFromString(Level)) {
+    spdlog::error("Invalid log level: {}. Valid values are: off, trace, debug, "
+                  "info, warning, error, fatal",
+                  Opt.LogLevel.value());
+    return EXIT_FAILURE;
```
hydai: If the logging level is not found, should we fail the entire execution?  
Would it be better to show a warning message and use `info` as the default logging level, just like the previous behavior?

cc @q82419 @ibmibmibm

lib/driver/runtimeTool.cpp
```diff
@@ -240,7 +240,17 @@ ToolOnComponent(WasmEdge::VM::VM &VM, const std::string &FuncName,
 
 int Tool(struct DriverToolOptions &Opt) noexcept {
   std::ios::sync_with_stdio(false);
-  Log::setInfoLoggingLevel();
```
hydai, line 243: You only changed runtimeTool.cpp. However, I think uniTool.cpp should also apply these changes. So, we can move the whole thing into `Log::` and replace this single line.

vishal2005025:
@hydai Thanks for the feedback.

I’ve updated the PR to address both points:
- Removed the unconditional Log::setInfoLoggingLevel() so the CLI --log-level is no longer overridden.
- Changed invalid log-level handling to emit a warning and fall back to info, preserving the previous behavior instead of aborting execution.

When no --log-level is provided, the behavior remains unchanged.
Please let me know if any further adjustments are needed.

lib/driver/uniTool.cpp
```diff
@@ -16,7 +16,6 @@ int UniTool(int Argc, const char *Argv[], const ToolType ToolSelect) noexcept {
   using namespace std::literals;
 
   std::ios::sync_with_stdio(false);
-  Log::setInfoLoggingLevel();
```
hydai, line 19: You should follow the same approach as you did in `runtimeTool.cpp`.

lib/driver/runtimeTool.cpp
```diff
@@ -240,7 +240,17 @@ ToolOnComponent(WasmEdge::VM::VM &VM, const std::string &FuncName,
...
+
+  // Set log level from CLI option (default is "info")
+  const std::string &Level =
+      Opt.LogLevel.value().empty() ? "info" : Opt.LogLevel.value();
```
hydai: This will never be `empty`. The reason is that you already set the default value inside the PO. Therefore, it cannot be empty anywhere else.

lib/common/spdlog.cpp
```diff
@@ -40,6 +40,65 @@ void setErrorLoggingLevel() { spdlog::set_level(spdlog::level::err); }
...
+    setErrorLoggingLevel();
+    return true;
+  case LogLevelTag::Critical:
+    setCriticalLoggingLevel();
+    return true;
+  case LogLevelTag::Invalid:
+    return false;
+  }
+  return false;
+}
```
hydai: I believe this is already overengineered. What if we use only if blocks instead?


```suggestion
bool setLoggingLevelFromString(std::string_view Level) {
  if (Level == "off"sv) {
    setLogOff();
    return true;
  }
  if (Level == "trace"sv) {
    setTraceLoggingLevel();
    return true;
  }
  if (Level == "debug"sv) {
    setDebugLoggingLevel();
    return true;
  }
  if (Level == "info"sv) {
    setInfoLoggingLevel();
    return true;
  }
  if (Level == "warning"sv || Level == "warn"sv) {
    setWarnLoggingLevel();
    return true;
  }
  if (Level == "error"sv) {
    setErrorLoggingLevel();
    return true;
  }
  if (Level == "fatal"sv || Level == "critical"sv) {
    setCriticalLoggingLevel();
    return true;
  }
  return false;
}

```

vishal2005025:
@hydai I have addressed all the review comments. Could you please take another look and let me know if any further adjustments are needed.

q82419:
It's fine to add the options for setting logging levels in CLI options.
But it only maps the logging level from `string` to `enum` in tools, I think it's better to implement the mapping mechanism in tools, not in `lib/common`.

vishal2005025:
@q82419 @hydai Just to confirm, is runtimetool.cpp the intended place to implement the mapping mechanism?

lib/common/spdlog.cpp
```diff
@@ -40,6 +40,38 @@ void setErrorLoggingLevel() { spdlog::set_level(spdlog::level::err); }
...
+    setErrorLoggingLevel();
+    return true;
+  }
+  if (Level == "fatal"sv || Level == "critical"sv) {
+    setCriticalLoggingLevel();
+    return true;
+  }
+  return false;
+}
+
```
hydai, line 74: I believe @q82419 is referring to this part, which is used only inside `runtimeTool` and `uniTool`, both of which are under `driver`. So, the question is: where should we place this part?

IMO, this duty belongs to `Log`, so it's fine.

vishal2005025: @hydai Thanks for clarifying.
So i am keeping this log-level mapping remain in spdlog.cpp file as said, and the tools will continue to call into it. also please let me know if any further adjustments are required from my side.

Issue #4510, opened 2026-01-12
bug: `br_on_cast` Jump Error on `ref extern`


akldc:
### Test Case

```
(module
  (type $0 (func (result f32 (ref null extern))))
  (export "main" (func 0))
  (func $0 (type 0)
    (block (type 0)
      (f32.const 1.8)
      (i32.const 0)
      (ref.i31)
      (extern.convert_any)
      (br_on_cast 0 (ref extern) (ref extern))
      (drop)
      (drop)
      (f32.const 3.2)
      (ref.null extern)
    )
  )
)
```
### Description

I executed the test case in wasmtime and wasmedge, and the results are as follows
```
$ wasmtime run -W all-proposals=true --invoke main test1.wasm
1.8
<externref>

$ wasmedge run --enable-all test1.wasm main
3.2
<null externref>
```
Under expected conditions, the `br_on_cast` jump will always succeed, and the instructions following it will not be executed.

### Current State

If the top of the stack before `br_on_cast` is i31ref, i.e., after removing `extern.convert_any`, and executing `br_on_cast 0 (ref i31) (ref i31)`, no error occurs, and it correctly returns `1.8:f32`. 
Therefore, the issue likely lies with the `br_on_cast` instruction when encountering `ref extern`.

### Reproduction steps

1. Build with options '...'
2. Execute with flags '....'
3. Execute with inputs '....'
4. Get error

### Components

CLI

### WasmEdge Version or Commit you used

0.16.1

### Operating system information

Ubuntu 20.04

### Hardware Architecture

x86_64

Vikaspal8923:
@akldc  @q82419 Hi, I’m a new contributor exploring the WasmEdge codebase and looked into this issue while trying to understand the behavior. It seems the problem is in runRefConvOp, where the code externalizes non-null references without updating the heap type, causing br_on_cast to fail. I’ve tested a local fix that restores the expected br_on_cast behavior.

If this understanding is correct, I can open a PR with the fix.

q82419:
> [@akldc](https://github.com/akldc) [@q82419](https://github.com/q82419) Hi, I’m a new contributor exploring the WasmEdge codebase and looked into this issue while trying to understand the behavior. It seems the problem is in runRefConvOp, where the code externalizes non-null references without updating the heap type, causing br_on_cast to fail. I’ve tested a local fix that restores the expected br_on_cast behavior.
> 
> If this understanding is correct, I can open a PR with the fix.

[...]

PR #4536, opened 2026-01-19
fix(executor): preserve heap type when externalizing references

Vikaspal8923:
what is the Issue:
- extern.convert_any only set an internal flag but didn't update the heap type code from the original type (e.g., i31ref) to externref, causing br_on_cast type checks to fail.
Fix: 
-Added Ref.getType() = ValType(TypeCode::Ref, TypeCode::ExternRef); to actually change the heap type to externref in the externalize path of runRefConvOp().

Issue: https://github.com/WasmEdge/WasmEdge/issues/4510
output after the fix : 

Screenshot 2026-01-18 222731


hydai (changes requested):
Since a valid test case is provided in the issue, we should add it to the tests to verify the behavior.

q82419:
I have the impression that I set the external flag here but not forcibly changed the heap type because we should keep the original heap type.
Are you sure that all of the spec tests are passed?

Vikaspal8923:
@q82419 I ran the WasmEdge  test I added, and it passed successfully.

As a new contributor ,  I’m not fully sure yet how WasmEdge runs the WebAssembly spec tests .

 Could you please confirm the recommended way to run the relevant spec tests for this change?

 I’d be happy to run them  to check my fix.

hydai:
Use `cmake --build <build folder> --target test` to run all test suites.

q82419:
> @q82419 I ran the WasmEdge test I added, and it passed successfully.
> 
> As a new contributor , I’m not fully sure yet how WasmEdge runs the WebAssembly spec tests .
> 
> Could you please confirm the recommended way to run the relevant spec tests for this change?
> 
> I’d be happy to run them to check my fix.

https://wasmedge.org/docs/contribute/source/os/linux

Vikaspal8923:
@q82419 I fixed the br_on_cast issue after extern.convert_any. The fix preserves the internal GC type and only treats externalized refs as externref during cast matching. I also removed an invalid ValType mutation that caused test regressions.
Tests are passing now.

all tests passed :: 
Screenshot 2026-01-21 022323





Exact test mention in issue description :: 


Screenshot 2026-01-21 030518

lib/executor/engine/controlInstr.cpp
```diff
@@ -105,6 +105,7 @@ Expect<void> Executor::runBrTableOp(Runtime::StackManager &StackMgr,
   return branchToLabel(StackMgr, LabelTable[LabelTableSize], PC);
 }
 
+
```
hydai: What's the reason to add this empty line?

lib/executor/engine/controlInstr.cpp
```diff
@@ -121,13 +122,28 @@ Expect<void> Executor::runBrOnCastOp(Runtime::StackManager &StackMgr,
...
+  ValType NormalizedVT = VT;
+  
+  if (NormalizedVT.isExternalized()) {
+    // An externalized reference must appear as an 'externref' to the matcher.
+    // We preserve the nullability (Ref vs RefNull).
+    NormalizedVT = ValType(
+        VT.isNullableRefType() ? TypeCode::RefNull : TypeCode::Ref,
+        TypeCode::ExternRef
+    );
+
```
hydai: Ditto

test/externref/ExternrefTest.cpp
```diff
@@ -583,6 +634,7 @@ TEST(ExternRefTest, Ref__STL) {
   WasmEdge_ModuleInstanceDelete(HostMod);
 }
 
+
```
hydai: Ditto

hydai (changes requested):

lib/executor/engine/controlInstr.cpp
```diff
@@ -121,11 +121,24 @@ Expect<void> Executor::runBrOnCastOp(Runtime::StackManager &StackMgr,
...
+    // An externalized reference must appear as an 'externref' to the matcher.
+    // We preserve the nullability (Ref vs RefNull).
+    NormalizedVT = ValType(
+        VT.isNullableRefType() ? TypeCode::RefNull : TypeCode::Ref,
+        TypeCode::ExternRef
+    );
   }
 
-  if (AST::TypeMatcher::matchType(ModInst->getTypeList(),
+  bool matchResult = AST::TypeMatcher::matchType(ModInst->getTypeList(),
```
hydai: Wrong naming style.

lib/executor/engine/controlInstr.cpp
```diff
@@ -121,11 +121,24 @@ Expect<void> Executor::runBrOnCastOp(Runtime::StackManager &StackMgr,
...
                                   Instr.getBrCast().RType2, GotTypeList,
-                                  VT) != IsReverse) {
+                                  NormalizedVT);
+
```
hydai: Please keep the code tidy; adding redundant empty lines in many places is very annoying.

hydai (changes requested):

test/externref/ExternrefTest.cpp
```diff
@@ -473,6 +482,45 @@ TEST(ExternRefTest, Ref__Functions) {
   WasmEdge_ModuleInstanceDelete(HostMod);
 }
 
+TEST(ExternRefTest, ExternConvertAnyWithBrOnCast) {
+  WasmEdge_ConfigureContext *ConfCxt = WasmEdge_ConfigureCreate();
+  WasmEdge_ConfigureAddProposal(ConfCxt, WasmEdge_Proposal_ReferenceTypes);
+  WasmEdge_ConfigureAddProposal(ConfCxt, WasmEdge_Proposal_GC);
```
q82419: Currently WasmEdge use the WASM 3.0 standard as default, which includes these proposals.
Therefore the configuration is not necessary.

q82419 (changes requested):

test/externref/ExternrefTest.cpp
```diff
@@ -473,6 +482,42 @@ TEST(ExternRefTest, Ref__Functions) {
   WasmEdge_ModuleInstanceDelete(HostMod);
 }
 
+TEST(ExternRefTest, ExternConvertAnyWithBrOnCast) {
+  WasmEdge_ConfigureContext *ConfCxt = WasmEdge_ConfigureCreate();
```
q82419: The meaningless configure context is not necessary.

hydai:
Hi @Vikaspal8923  
Please rebase onto the latest master branch and squash all your commits into a single one.

PR #4540, opened 2026-01-20
refactor: simplify WasmFunction instr allocation

Sanyamjain2105:
### Summary

This change simplifies the instruction vector allocation in `WasmFunction` by removing a redundant `reserve(+1)` call during construction.

Instead of reserving extra capacity and assigning later, the vector is now initialized with the correct size directly. This makes the allocation intent clearer, removes the stale `FIXME`, and avoids unnecessary capacity management.

There are **no functional or behavioral changes**. This is a small refactor focused on improving code clarity and maintainability around instruction allocation.


q82419:
The reservation of this capacity is for preventing from the connection of the instruction sequences in different WASM functions, since the program counter jumping in interpreter mode need this to check the end of sequences.
Therefore this fixing not makes sense and the tests will obviously not pass.

PR #4541, opened 2026-01-21
fix(WASI-NN/whisper): implement resource cleanup in finalizeExecCtx

Parship12:
### Summary
While reading the code, I noticed a TODO comment regarding missing resource cleanup. So, in this PR I implemented that TODO.

### Tests
Screenshot 2026-01-21 074030

### DIFF
--- a/plugins/wasi_nn/wasinn_whisper.cpp
+++ b/plugins/wasi_nn/wasinn_whisper.cpp
@@ -1113,7 +1113,11 @@ Expect<ErrNo> finalizeExecCtx(WasiNNEnvironment &Env,
     spdlog::info(
         "[WASI-NN][Debug] Whisper backend: finalize_execution_context"sv);
   }
-  // TODO: Free resources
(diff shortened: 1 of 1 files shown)


plugins/wasi_nn/wasinn_whisper.cpp
```diff
@@ -1113,7 +1113,11 @@ Expect<ErrNo> finalizeExecCtx(WasiNNEnvironment &Env,
     spdlog::info(
         "[WASI-NN][Debug] Whisper backend: finalize_execution_context"sv);
   }
-  // TODO: Free resources
+  // Free resources
+  CxtRef.InputPCM.clear();
+  CxtRef.InputPCMs.clear();
+  CxtRef.Outputs.clear();
```
hydai, line 1119: RAII will handle it; there's no need to do this.

hydai (changes requested):

Issue #2102, opened 2022-11-21
feat: Limit the range of randomly generated fd in the WASI component


hydai:
## Motivation

Due to some applications using the old select API, our current implementation follows the standard of WASI which means we will have a map to handle the randomly generated fd number (`<= 2**31`) to the actual fd number. However, in some cases, the application only accepts the fd number which is less than `FD_SETSIZE (usually 1024)`.

It would be nice to have an API to set the range of the randomly generated fd numbers.

## Details

- [ ] Implement an internal function to control the range
- [ ] Export this function as a configuration option in C API
- [ ] Export a CLI option to control this flag
- [ ] Write a section in the document

## Appendix

* https://github.com/WebAssembly/WASI/blob/main/phases/snapshot/docs.md#-path_openfd-fd-dirflags-lookupflags-path-string-oflags-oflags-fs_rights_base-rights-fs_rights_inheriting-rights-fdflags-fdflags---resultfd-errno

uddhavphatak:
Hi @hydai, I wish to work on this issue, but have a confusion, is this still a pending feature, can I work for this for the upcoming release?

hydai:
This is still available.

uddhavphatak:
Thank you,
is there any documentation related to this issue, the appendix written in the issue description is giving error-404

uddhavphatak:
I am assuming the 404'd appendix referred to the FD_SETSIZE (1024) limitation in the select() syscall.

I plan to implement a configuration option that constrains the randomized FDs to be within the FD_SETSIZE range. 

I'll start by looking at how the WASI implementation currently maps FDs in "lib/host/wasi". Does that sound like the right direction? Please correct me if I am going in the wrong direction

uddhavphatak:
hi @hydai , I have made a PR about this. though I didn't have any documentation to look through, I figured out that the API for generating random fd's by doing a 
```
grep -ir "generaterandom"
```

and got the following output:
```
uddhav@uddhav-A15:~/Desktop/lfx/WasmEdge$ grep -ir "generaterandom"
include/host/wasi/environ.h:    return generateRandomFdToNode(Node);
include/host/wasi/environ.h:    return generateRandomFdToNode(Node);
include/host/wasi/environ.h:    return generateRandomFdToNode(NewNode);
include/host/wasi/environ.h:  WasiExpect<__wasi_fd_t> generateRandomFdToNode(std::shared_ptr<VINode> Node) {
```

[...]

PR #4549, opened 2026-01-25
feat: Limit the range of randomly generated fd in the WASI component

uddhavphatak:
## Motivation
 
Due to some applications using the old select API, our current implementation follows the standard of WASI which means we will have a map to handle the randomly generated fd number (`<= 2**31`) to the actual fd number. However, in some cases, the application only accepts the fd number which is less than `FD_SETSIZE (usually 1024)`.
 
 It would be nice to have an API to set the range of the randomly generated fd numbers.
 ## Details
 * [x]  Implement an internal function to control the range
 * [x]  Export this function as a configuration option in C API
 * [x] Export a CLI option to control this flag
 * [ ] Write a section in the document
 
 solves issue #2102

### DIFF
--- a/include/common/configure.h
+++ b/include/common/configure.h
@@ -169,7 +169,7 @@ public:
     return AllowAFUNIX.load(std::memory_order_relaxed);
   }
 
-  void setMaxWasiFd(uint32_t Fd) noexcept { MaxWasiFd = Fd; }
+  void setMaxWasiFd(uint32_t MaxFd) noexcept { MaxWasiFd = MaxFd; }
 
   uint32_t getMaxWasiFd() const noexcept { return MaxWasiFd; }
 
@@ -180,7 +180,7 @@ private:
   std::atomic<bool> CoredumpWasmgdb = false;
   std::atomic<bool> ForceInterpreter = false;
   std::atomic<bool> AllowAFUNIX = false;
-  uint32_t MaxWasiFd = 0x7FFFFFFF;
+  std::atomic<uint32_t> MaxWasiFd = 0x7FFFFFFF;
 };
 
 class StatisticsConfigure {
--- a/include/common/configure.h
+++ b/include/common/configure.h
@@ -169,6 +169,10 @@ public:
     return AllowAFUNIX.load(std::memory_order_relaxed);
   }
 
+  void setMaxWasiFd(uint32_t Fd) noexcept { MaxWasiFd = Fd; }
+
+  uint32_t getMaxWasiFd() const noexcept { return MaxWasiFd; }
+
 private:
   std::atomic<uint32_t> MaxMemPage = 65536;
   std::atomic<bool> EnableJIT = false;
@@ -176,6 +180,7 @@ private:
   std::atomic<bool> CoredumpWasmgdb = false;
   std::atomic<bool> ForceInterpreter = false;
   std::atomic<bool> AllowAFUNIX = false;
+  uint32_t MaxWasiFd = 0x7FFFFFFF;
 };
 
 class StatisticsConfigure {
--- a/include/driver/tool.h
+++ b/include/driver/tool.h
@@ -117,13 +117,13 @@ struct DriverToolOptions {
                 "Limitation of maximum time(in milliseconds) for execution, "
                 "default value is 0 for no limitations"sv),
             PO::MetaVar("TIMEOUT"sv), PO::DefaultValue<uint64_t>(0)),
-        MaxWasiFd(PO::Description(
-                      "Limitation of max range for randomly generated file "
-                      "descriptors. The value will be clamped to the range [3, "
-                      "2**31-1]. "
-                      "Default value is 2**31-1."sv),
-                  PO::MetaVar("MAX_FD"sv),
-                  PO::DefaultValue<uint32_t>(0x7FFFFFFF)),
+        MaxWasiFd(
+            PO::Description(
+                "Limitation of max range for randomly generated file "
+                "descriptors. The value will be clamped to the range [1024, "
+                "2**31-1]. "
+                "Default value is 2**31-1."sv),
+            PO::MetaVar("MAX_FD"sv), PO::DefaultValue<uint32_t>(0x7FFFFFFF)),
         GasLim(
             PO::Description(
                 "Limitation of execution gas. Upper bound can be specified as "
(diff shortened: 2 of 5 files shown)


uddhavphatak:
This is my attempt at trying to solve this issue, please review and tell if I can improve anything.

If everything is fine, I will write this into the documentation

include/driver/tool.h
```diff
@@ -227,6 +233,7 @@ struct DriverToolOptions {
         .add_option("enable-component"sv, PropComponent)
         .add_option("enable-all"sv, PropAll)
         .add_option("time-limit"sv, TimeLim)
+	.add_option("max-fd"sv, MaxWasiFd)
```
LFsWang: lint problem

uddhavphatak: @LFsWang Fixed here, also spotted this issue in other file declarations, mainly due to my habitual use of tab instead of spaces. Thus fixed at those places as well

hydai (changes requested):
This PR doesn't contain any basic verifications and exhibits some unusual behaviors, including:
1. What happens if users set the maximum WASI FD to 0?
2. Hard-coding 0x7FFFFFF in the C API.
3. I don't think we need to change the `init` series functions. The `generateRandomFdToNode` is not used in these functions. Instead, we should set the MaxFD in another phase.

hydai:
Converted to a draft PR since the contributor is not running CI on their own forked repo. There are some very basic CI checks that must fail. It's not necessary to approve the CI to avoid wasting our resources on this PR.

uddhavphatak:
@hydai, It was mistake from my end, I had understood the issue wrong, thus thought that I had to make changes in the init API. 

I also have setup the CI on my fork, and done the CI testing before reopening the issue. should I undraft this PR?

include/common/configure.h
```diff
@@ -169,13 +169,18 @@ class RuntimeConfigure {
...
+  uint32_t getMaxWasiFd() const noexcept { return MaxWasiFd; }
+
 private:
   std::atomic<uint32_t> MaxMemPage = 65536;
   std::atomic<bool> EnableJIT = false;
   std::atomic<bool> EnableCoredump = false;
   std::atomic<bool> CoredumpWasmgdb = false;
   std::atomic<bool> ForceInterpreter = false;
   std::atomic<bool> AllowAFUNIX = false;
+  uint32_t MaxWasiFd = 0x7FFFFFFF;
```
hydai: This should be atomic.

include/host/wasi/environ.h
```diff
@@ -1200,6 +1200,10 @@ class Environ : public PollerContext {
     return std::string(Buffer.data(), Buffer.size());
   }
 
+  void setMaxWasiFd(uint32_t Limit) noexcept {
```
hydai: Your naming is inconsistent. Some of them use `FD`, while others use `Limit`.

include/host/wasi/environ.h
```diff
@@ -1200,6 +1200,10 @@ class Environ : public PollerContext {
...
+    MaxFd = std::clamp(Limit, 3U, 0x7FFFFFFFU);
```
hydai: You are silently modifying the user's given limit; however, this is not mentioned in the option description.

hydai (changes requested):

include/host/wasi/environ.h
```diff
@@ -1221,7 +1226,8 @@ class Environ : public PollerContext {
   }
 
   WasiExpect<__wasi_fd_t> generateRandomFdToNode(std::shared_ptr<VINode> Node) {
-    std::uniform_int_distribution<__wasi_fd_t> Distribution(0, 0x7FFFFFFF);
+    std::uniform_int_distribution<__wasi_fd_t> Distribution(
+        0, static_cast<__wasi_fd_t>(MaxFd));
```
hydai, line 1237: There are many possible cases that are not covered.  
For example, what if `MaxFd = 3`, but there are more than one preopened folder?  
What if `MaxFd` is fully used, but we are trying to open a new one? The original limitation is large enough to prevent such a case. However, your new approach may not work.

hydai (changes requested):

include/host/wasi/environ.h
uddhavphatak: @hydai I see the issue, but I have a few doubts.
can the user set the MaxFd as 3?, as it is a CLI based command (sets max limit when option "max-fd" is used), but by default, the limitation is large enough.

and in the case if  MaxFd is fully used, but we are trying to open a new one. should we return an error? or should I create an implementation that will create a queue to track the files which are open/closed? WDYT

hydai: That's why I am considering it.  
The reason we want this feature is that the issue report uses select, which accepts fewer than 1024 file descriptors. I prefer to set the minimum to 1024; if the given MaxFd is less than this, then we should clamp it to 1024. If MaxFd is fully used, we should return an error directly. I don't think having a queue for this is a good idea, since execution may rely on this new FD to perform certain tasks.

uddhavphatak:
@hydai I have clamped the lower limit to be 1024. added a check for check for FdMap.size() >= MaxFd. 
and even added a for loop of 50 times, which will check for collisions. I think most collisions can happen in the lower limit, and thought 50 should be a good enough number, is that alright, or should I raise it?

include/host/wasi/environ.h
```diff
@@ -1221,15 +1226,25 @@ class Environ : public PollerContext {
...
+        return WasiUnexpect(__WASI_ERRNO_MFILE);
+      }
+    }
+
+    std::uniform_int_distribution<__wasi_fd_t> Distribution(
+        0, static_cast<__wasi_fd_t>(MaxFd));
     __wasi_fd_t NewFd;
-    while (!Success) {
+
+    for (int i = 0; i < 50; i++) {
```
hydai, line 1240: I don't understand the reason for modifying the original while loop to this.

hydai (changes requested):
You should also add tests for the configure.

include/host/wasi/environ.h
uddhavphatak: I modified the loop because as the limit goes lower, the number of collisions would also increase, and at some point, it might just wait for an fd to get allocated, the program might get into an infinite loop there. That's why I modified this loop

itsme-ranger:
Dear @uddhavphatak and @hydai 

is this PR still in development or postponed? I'm willing to continue this effort till finished. Thank you

hydai:
Thanks @uddhavphatak for your contribution.  
After careful consideration of this feature, I believe we will no longer need to provide it unless there is a solid use case that genuinely relies on this limited range of the FD number. The reason is that mapping the real FD into a larger range of randomly generated FDs is intended to ensure safety and reduce collisions between the random number generator. However, limiting it to a small range significantly decreases this safeguard.

I still want to thank you again for all your time discussing and contributing to this. Unfortunately, I have to close it for the reasons mentioned above.

PR #4562, opened 2026-01-27
feat(statistics): conditionally show gas limit and usage percentage

### LINKED ISSUE #4561: feat: conditionally expose gas limit and usage in --enable-all-statistics
### Summary

The WasmEdge runtime already tracks gas costs and gas limits internally, but the CLI statistics output currently only reports total gas consumption. Since the default gas limit is unlimited, it does not makes sense to display this information by default. However, when a user explicitly configures a gas limit, exposing the configured limit and gas usage percentage would provide clearer insight for benchmarking and resource control, while keeping the default behavior unchanged.

### Details

While exploring the existing statistics implementation, I noticed that the gas limit and related information already exist in the runtime but are not shown in the CLI output.

Since the default gas limit is unlimited, it makes sense not to display it by default. However, when a user explicitly sets `--gas-limit`, exposing the configured limit and usage percentage could be useful for `benchmarking` and `resource control`.

A possible approach would be to conditionally display the gas limit and gas usage only when a non-default limit is explicitly provided by the user (for example, via `wasmedge --gas-limit <VALUE> program.wasm`).

### Appendix

_No response_

vishal2005025:
### Summary

This PR enhances the existing statistics output by conditionally exposing the configured gas limit and gas usage percentage when a non-default gas limit is explicitly provided by the user.

### Testing

- Built WasmEdge from a clean build directory
- Ran the WasmEdge CLI locally with `--enable-all-statistics`
- Verified that existing statistics remain unchanged
- Verified that when `--gas-limit` is explicitly set:
  - Gas limit is reported
  - Gas usage percentage is reported
- Verified that no additional output appears when the default (unlimited) gas
  limit is used

All statistics are displayed correctly and runtime behavior remains unchanged.

Screenshot 2026-01-27 155223


vishal2005025:
@hydai @q82419 @ibmibmibm  Please take a look when you get a chance and let me know if any changes are needed. Thanks!

q82419:
LGTM.

include/common/statistics.h
```diff
@@ -173,6 +173,12 @@ class Statistics {
     }
     if (StatConf.isCostMeasuring()) {
       spdlog::info(" Gas costs: {}"sv, getTotalCost());
+      if (getCostLimit() != UINT64_MAX) {
+        spdlog::info(" Gas limit: {}"sv, getCostLimit());
+        spdlog::info(" Gas usage: {:.2f}%"sv,
+                     (static_cast<double>(getTotalCost()) / getCostLimit()) *
```
hydai, line 179: What will happen if `getCostLimit()` returns `0`?

hydai (changes requested):

include/common/statistics.h
vishal2005025: Good point - if a user sets --gas-limit 0, this would result in a division by zero.
I have added a guard to ensure the limit is non-zero before calculating usage.

ravindra-RKB:
Hi @hydai 
I noticed that PR #4562 already addresses this issue and includes the requested guard for `--gas-limit 0`. Is this PR still under consideration or would you prefer a refreshed PR rebased on the current master branch?

I'm happy to help update tests or rebase the implementation if needed.......

PR #4564, opened 2026-01-27
Feat: non-static string for ErrorCode and spectest

### LINKED ISSUE #4444: feat: Component Model: non-static string for ErrorCode and spectest
### Summary

Allow print the error with parameter and pass the spectest

### Details

Our spec test use static string to compare the error message, but there are some spec will have dynamic part, like

`type index {} is not an instance type`

Otherwise, we need to modify the testcases.

A possible solution is that let `Res.error()` carries the additional error string. But this way needs a significant refactor of the Error Code system. Therefore I suggest anyone interested in tackling this should provide a comprehensive plan before implementation.

https://github.com/WasmEdge/WasmEdge/blob/master/test/spec/spectest.cpp#L743
```
  // Helper function to check trap on validation.
  auto TrapValidate = [&](const std::string &FileName,
                          const std::string &Text) {
    if (auto Res = onValidate(FileName); Res) {
      EXPECT_TRUE(false);
    } else {
      EXPECT_TRUE(Res.error().getErrCodePhase() ==
                  WasmEdge::WasmPhase::Validation);
      EXPECT_TRUE(
          stringContains(Text, WasmEdge::ErrCodeStr[Res.error().getEnum()]));
    }
  };
```

### Appendix

_No response_

xonx4l:
Fixes #4444

Implements support for dynamic error messages in the Component Model validator to satisfy `spectests` requirements.

I modified the Validator to store a formatted error string internally (using fmt) when a specific error occurs. The CLI tool then checks this string first if it's set it prints the dynamic message instead of the static one.

Validated that the build passes.

### DIFF
--- a/include/validator/validator.h
+++ b/include/validator/validator.h
@@ -20,6 +20,8 @@
 #include "validator/formchecker.h"
 
 #include <cstdint>
+#include <string>
+#include <fmt/core.h>
 
 namespace WasmEdge {
 namespace Validator {
@@ -34,6 +36,8 @@ public:
   Expect<void> validate(const AST::Module &Mod);
   /// Validate AST::Component.
   Expect<void> validate(const AST::Component::Component &Comp) noexcept;
+
+  std::string_view getErrorStr() const noexcept { return ErrorMsg; }
 
 private:
   /// \name Validate WASM AST nodes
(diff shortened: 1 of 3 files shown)


LFsWang:
@xonx4l Hello, I don’t think this PR addresses the core issue. The main goal is to ensure that the section above works as intended.

https://github.com/WasmEdge/WasmEdge/blob/master/test/spec/spectest.cpp#L751

LFsWang:
I’m closing this PR because it does not resolve the issue.

PR #4566, opened 2026-01-27
fix(loader): validate memory op alignment during loading

### LINKED ISSUE #4292: bug: WasmEdge interpreter missed alignment validation check
### Summary

I have generated a test case and I executed it with different runtimes. Other runtimes like wasmtime, wamr, wasmer, and wasmedge's compiled version output an invalid alignment error. But, wasmedge's interpreter does not check this alignment error and instead outputs `unreachable`.

### Current State

```
[2025-07-31 05:40:12.411] [error] execution failed: unreachable, Code: 0x40a
[2025-07-31 05:40:12.411] [error]     In instruction: unreachable (0x00) , Bytecode offset: 0x000000f4
[2025-07-31 05:40:12.412] [error] calling stack:0
[2025-07-31 05:40:12.412] [error]     When executing function name: "main"
```

### Expected State

```
[2025-07-31 05:40:38.020] [error] loading failed: malformed memop flags, Code: 0x11c
[2025-07-31 05:40:38.020] [error]     Bytecode offset: 0x000000f9
[2025-07-31 05:40:38.020] [error]     At AST node: instruction
[2025-07-31 05:40:38.020] [error]     At AST node: expression
[2025-07-31 05:40:38.020] [error]     At AST node: code segment
[2025-07-31 05:40:38.020] [error]     At AST node: code section
[2025-07-31 05:40:38.020] [error]     At AST node: module
[2025-07-31 05:40:38.020] [error] Parse Module failed. Error code: 284
```

Or,

```
invalid memop flags: alignment must not be larger than natural
```

### Reproduction steps

**Test case:**
[test.txt](https://github.com/user-attachments/files/21527449/test.txt)

**Steps to reproduce:**
1. Download test case and rename from .txt to .wasm.
2. Run: `wasmedge run --enable-all test.wasm main`

### Screenshots

_No response_

_No response_

### Components

CLI

0.15.0-alpha.4

Ubuntu 22.04

### Hardware Architecture

x86_64

- LLVM Version: 18
- cmake options: `cmake -DCMAKE_BUILD_TYPE=Debug ..`

Meetjain1:
The WasmEdge interpreter was missing alignment validation for memory instructions, allowing invalid `memarg.align` values to pass load time and fail later at runtime with an `unreachable` error.

### What changed

This PR adds instruction-level alignment validation in the interpreter loader.
If the specified alignment is larger than the natural alignment of the instruction, the module is now rejected during load/validation time with a proper error.

**Fixes #4292**

### Result

* Interpreter behavior is now consistent with validator/AOT paths
* Invalid alignment is caught early instead of crashing at runtime

### Reproduction / Proof

Command:

```
wasmedge run /tmp/test_invalid_align.wasm
```

Observed result:

```
validation failed: alignment must not be larger than natural
```

image

### DIFF
--- a/.github/workflows/build-extensions.yml
+++ b/.github/workflows/build-extensions.yml
@@ -179,7 +179,7 @@ jobs:
           $Env:CXX = "$pwd\LLVM-21.1.3-win64-MultiThreadedDLL\LLVM-21.1.3-win64\bin\clang-cl.exe"
           $llvm_mt = "mt"
           $cmake_sys_version = "10.0.26100.0"
-          cmake -Bbuild -GNinja -DCMAKE_SYSTEM_VERSION="$cmake_sys_version" -DCMAKE_MSVC_RUNTIME_LIBRARY=MultiThreadedDLL -DCMAKE_MT="$llvm_mt" -DLLVM_DIR="$llvm_dir" -DWASMEDGE_BUILD_TESTS=ON -DWASMEDGE_USE_LLVM=OFF -DWASMEDGE_BUILD_TOOLS=OFF .
+          cmake -Bbuild -GNinja -DCMAKE_BUILD_TYPE=Release -DCMAKE_SYSTEM_VERSION="$cmake_sys_version" -DCMAKE_MSVC_RUNTIME_LIBRARY=MultiThreadedDLL -DCMAKE_MT="$llvm_mt" -DLLVM_DIR="$llvm_dir" -DWASMEDGE_BUILD_TESTS=ON -DWASMEDGE_USE_LLVM=OFF -DWASMEDGE_BUILD_TOOLS=OFF .
 
           Write-Output "Building $Env:tar_names backend:"
           cmake -Bbuild -GNinja "$Env:build_options"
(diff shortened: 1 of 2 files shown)


Meetjain1:
@hydai PTAL on this PR, Thanks!

q82419:
This (error code `invalid alignment`) should be in the validation phase, I don't think it can be processed in the loader, which only check for the syntax.
Furthermore, I don't think it's a good idea to extend and list the instruction OpCodes into a such large if-condition you wrote.

Meetjain1:
> This (error code `invalid alignment`) should be in the validation phase, I don't think it can be processed in the loader, which only check for the syntax. Furthermore, I don't think it's a good idea to extend and list the instruction OpCodes into a such large if-condition you wrote.

Thanks for the review @q82419,

I understand your point alignment validation makes more sense in the validation phase rather than the loader, 
and I see the concern around listing many opcodes in the loader logic.

I’ll revisit this and look into how this can be handled properly in the validation path instead.

q82419:
Since the root cause is not compatible with your description and solution, close this.

PR #4568, opened 2026-01-28
fix(examples): avoid unbounded strcpy in unix domain socket examples

vishal2005025:
### Summary

This PR fixes unsafe usage of `strcpy` in the Unix domain socket C API examples.

The affected examples copy user-provided socket paths into `sockaddr_un::sun_path` without any bounds checking. Since `sun_path` has a fixed size (typically 108 bytes on Linux), this can lead to buffer overflow or  undefined behavior if a longer path is provided. 
This PR replaces strcpy with a bounded copy to ensure the buffer is not overrun and is always NUL-terminated.

Although these are example files, they are often used as reference code, so it’s important that they demonstrate safe patterns.

### Real-world references

- Python / mod_wsgi hardening against long sun_path
https://github.com/GrahamDumpleton/mod_wsgi/issues/35

- Linux man-pages explicitly document the size limit of sun_path
https://man7.org/linux/man-pages/man7/unix.7.html

- Discussion on Unix socket path length limits
https://unix.stackexchange.com/questions/367008/why-is-socket-path-length-limited-to-a-hundred-chars

### Testing

- Build verified locally
- No functional changes

### DIFF
--- a/examples/capi/unix_domain_socket/client.cpp
+++ b/examples/capi/unix_domain_socket/client.cpp
@@ -22,7 +22,13 @@ int main() {
   }
 
   ClientAddr.sun_family = AF_UNIX;
-  strcpy(ClientAddr.sun_path, ClientSockPath);
+  if (std::strlen(ClientSockPath) >= sizeof(ClientAddr.sun_path)) {
+    std::fprintf(stderr, "Socket path is too long\n");
+    return -1;
+  }
+  std::strncpy(ClientAddr.sun_path, ClientSockPath,
+               sizeof(ClientAddr.sun_path) - 1);
+  ClientAddr.sun_path[sizeof(ClientAddr.sun_path) - 1] = '\0';
   Size = offsetof(sockaddr_un, sun_path) + strlen(ClientAddr.sun_path);
 
   // unlink(ClientSockPath);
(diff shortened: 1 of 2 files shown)


vishal2005025:
@hydai @q82419  Please review this PR and let me know if any changes are required. Thanks!

examples/capi/unix_domain_socket/client.cpp
```diff
@@ -22,7 +22,8 @@ int main() {
   }
 
   ClientAddr.sun_family = AF_UNIX;
-  strcpy(ClientAddr.sun_path, ClientSockPath);
+  std::snprintf(ClientAddr.sun_path, sizeof(ClientAddr.sun_path), "%s",
```
hydai: This will silently truncate. If the purpose of this PR is to ensure the behavior will not result in a buffer overflow and to make it work correctly, perhaps we should also add a basic size check before calling this.

hydai (changes requested):

vishal2005025:
@hydai i have addressed the review changes please take a look and let me know if any changes are required .
Thanks!

vishal2005025:
@hydai Thanks for checking this more closely and pointing it out.
You’re right - the socket paths in these examples are fixed literals defined in the code, so `strcpy` will not overflow in the current implementation. I initially assumed they were user-provided. 
Given that, I agree this is not fixing a real bug today. I’m happy to close this PR unless you feel it’s still worth keeping as a defensive or best-practice improvement for example code.

hydai:
Thank you for your contribution. I am closing this because the example is just fine.

Issue #4574, opened 2026-01-28
bug: WASI-NN GGML-backend on Windows CI takes significantly longer than before.


hydai:
### Summary

The current WASI-NN GGML backend on Windows CI takes approximately 3 hours to execute. However, it was only less than 20 minutes before. We need to investigate the root cause to reduce the duration.

The workflow is here: https://github.com/WasmEdge/WasmEdge/blob/master/.github/workflows/build-extensions.yml#L140-L202

### Current State

See: https://github.com/WasmEdge/WasmEdge/actions/runs/21398835844/job/61605008983

<img width="1851" height="952" alt="Image" src="https://github.com/user-attachments/assets/1dab25b9-df7a-4b70-b42b-d89d4b130162" />

### Expected State

See: https://github.com/WasmEdge/WasmEdge/actions/runs/20924454888/job/60118891102
<img width="1860" height="959" alt="Image" src="https://github.com/user-attachments/assets/ab9cf728-7d36-42cf-af68-54819616101a" />

### Reproduction steps

1. Fork this repo
2. Create a PR which will trigger the WASI-NN GGML on Windows CI
3. Get the very long running jobs

### Screenshots

Check above sections.

### Any logs you want to share for showing the specific issue

The expected output should be:

```
Testing wasi_nn-ggml backend:
Running main() from D:\a\WasmEdge\WasmEdge\build\_deps\gtest-src\googletest\src\gtest_main.cc
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from WasiNNTest
[ RUN      ] WasiNNTest.GGMLBackend
common_init_from_params: failed to load model 'ggml-model.bin', try reducing --n-gpu-layers if you're running out of VRAM
Error: 1-12 15:29:55.350] [error] [WASI-NN] GGML backend: load: unable to init model.
Error: 1-12 15:29:55.351] [error] [WASI-NN] Failed when accessing the return GraphID memory.
Error: 1-12 15:29:55.351] [error] [WASI-NN] Failed when accessing the GraphBuilder memory.
Error: 1-12 15:29:55.351] [error] [WASI-NN] Failed when accessing the Builder[0] memory.
Error: 1-12 15:29:56.087] [error] [WASI-NN] GGML backend: parse metadata error.
Error: 1-12 15:29:56.087] [error] [WASI-NN] GGML backend: load: Failed to parse metadata.
common_init_from_params: added </s> logit bias = -inf
[...]

PR #4579, opened 2026-01-28
fix(ci): cache LLVM to speed up Windows WASI-NN GGML workflow

Meetjain1:
## Summary

This PR improves Windows WASI-NN GGML CI performance by caching the prebuilt LLVM binary used during the build.

## Context

While working on **issue #4574**, we noticed that Windows CI runs were still slower than expected.  
A large part of the remaining time was spent repeatedly downloading and extracting the LLVM toolchain on every run.

This happens even when the build and tests themselves are unchanged.

## What this PR does

- Adds GitHub Actions caching for the prebuilt LLVM (win64 MultiThreadedDLL)
- Skips the LLVM download and extraction step when the cache is available
- Leaves all existing build and test logic exactly the same

## Why this helps

- Avoids repeated downloads of a large LLVM binary
- Reduces unnecessary network and I/O overhead
- Makes repeated Windows CI runs faster and more stable

## Scope and Safety

- CI-only change
- No changes to build flags, test behavior, or runtime logic
- Falls back to downloading LLVM if the cache is missing

## Related

- Addresses CI performance observations discussed in **#4574**


.github/workflows/build-extensions.yml
```diff
@@ -165,15 +165,29 @@ jobs:
       - uses: GuillaumeFalourd/setup-windows10-sdk-action@9979bedbf6ae74f038cc338622ba015942ed2408 # v2.4
         with:
           sdk-version: 26100
+      - name: Cache LLVM
+        uses: actions/cache@v4
```
hydai: Use `@hash` not `tag`.

.github/workflows/build-extensions.yml
```diff
@@ -165,15 +165,29 @@ jobs:
...
-          Expand-Archive -Path $llvm
+          $llvm_extracted_dir = "LLVM-21.1.3-win64-MultiThreadedDLL"
+          if (-Not (Test-Path $llvm_extracted_dir)) {
+            Write-Output "Downloading LLVM..."
+            Invoke-WebRequest -Uri $uri -HttpVersion 2.0 -OutFile $llvm
+            Write-Output "Extracting LLVM..."
+            Expand-Archive -Path $llvm
+          } else {
+            Write-Output "LLVM cache found, skipping download..."
+          }
```
hydai: You have already used `actions/cache`, which allows you to use `cache-hit` to check the status.
https://github.com/actions/cache?tab=readme-ov-file#skipping-steps-based-on-cache-hit

Meetjain1:
@hydai made the changes, PTAL.

.github/workflows/build-extensions.yml
Meetjain1: Done

.github/workflows/build-extensions.yml
Meetjain1: Yes, makes sense Updated.

hydai:
I don't think you've run it even once on your own. Your commit messages will never pass our commitlint. You're just submitting something you haven't verified. I will be closing it unless you can prove that you've already performed the basic checks on your code and commit history.

Meetjain1:
> I don't think you've run it even once on your own. Your commit messages will never pass our commitlint. You're just submitting something you haven't verified. I will be closing it unless you can prove that you've already performed the basic checks on your code and commit history.

@hydai I noticed the failure was due to the cache action version. 
Please have a look I have updated - It takes 13m 30sec

https://github.com/Meetjain1/WasmEdge/actions/runs/21470716826/job/61842486159

image


Regarding commitlint, I’ll fix once the approach is approved.

Updated and fixed here Takes  12m 47s, Thankyou!

https://github.com/Meetjain1/WasmEdge/actions/runs/21472569322/job/61848643583?pr=1

image

Issue #3159, opened 2024-01-16
bug: Unexpected exception triggered by the instruction `memory.init`


erxiaozhou:
### Summary

The test case contains a `memory.init` instruction and it is supposed to be valid. However, an unexpected exception indicating "memory OOB" is triggered.
The instructions in the test case.
```
    i32.const 32768
    i32.const 4
    i32.const 1
    memory.init 0
```

### Current State

There is an exception indicating "memory OOB". 
Furthermore, the second line is confusing because 0x00008000 is smaller than 0x0000ffff.
```
[2024-01-16 17:39:10.332] [error] execution failed: out of bounds memory access, Code: 0x408
[2024-01-16 17:39:10.332] [error]     Accessing offset from: 0x00008000 to: 0x00008000 , Out of boundary: 0x0000ffff
[2024-01-16 17:39:10.332] [error]     In instruction: memory.init (0xfc 0x08) , Bytecode offset: 0x00000141
[2024-01-16 17:39:10.332] [error]     When executing function name: "to_test"
```

### Expected State

No exception

### Reproduction steps

1. Build: ' cmake  -DCMAKE_BUILD_TYPE=Release -DWASMEDGE_BUILD_AOT_RUNTIME=OFF .. '
2. Execute: 'wasmedge --reactor <case_name> to_test'
Case:

[all_wamr_memory.init_no_exception.zip](https://github.com/WasmEdge/WasmEdge/files/13948149/all_wamr_memory.init_no_exception.zip)

3. Get error

### Screenshots

![DESCRIPTION](LINK.png)
![image](https://github.com/WasmEdge/WasmEdge/assets/32102519/f092056d-a199-49b2-97dd-570f25d19dd8)

### Components

CLI

### WasmEdge Version or Commit you used

930abbedb7738db582e4996671827bbe0550d8e8

### Operating system information

Ubuntu 20.04

### Hardware Architecture

x86_64

### Compiler flags and options

CMake flags: -DCMAKE_BUILD_TYPE=Release -DWASMEDGE_BUILD_AOT_RUNTIME=OFF

TerryXhx:
Hi @erxiaozhou , the exception is caused by the fact the input data segment is invalid with size 0.

[...]

PR #4582, opened 2026-01-29
fix: correct memory.init error log logic

souravk29:
The memory.init instruction was incorrectly logging the Memory limit when the Data Segment (Slice) was out of bounds.
This change updates the log to print Slice.size() to clarify the error (addresses #3159).

Also adds static_cast<uint32_t> to prevent MSVC warning C4267 on Windows (Since WasmEdge uses /WX (Treat Warnings as Errors)).
In linux it worked even without static_cast<uint32_t> but had to add this for warning C4267.


souravk29:
@q82419 please review

PR #4585, opened 2026-01-30
build: add CMake option for maximum WASM memory pages

Meetjain1:
This PR adds a build-time option to configure the maximum WASM memory pages.

The memory page limit was previously hard-coded, which made it difficult to use WasmEdge in embedded or memory-constrained environments.

### What changed
- Added `WASMEDGE_MAX_MEMORY_PAGES` CMake option
- Default value remains unchanged (65536 pages)
- Replaced hard-coded memory limit with the configurable value

### Testing
- Verified the default configuration uses the original memory page limit (65536)
- Confirmed the value is generated via the CMake configuration (`config.h`)
- Build-time override is supported via the new WASMEDGE_MAX_MEMORY_PAGES option

image

### DIFF
--- a/.github/workflows/build-extensions.yml
+++ b/.github/workflows/build-extensions.yml
@@ -145,7 +145,6 @@ jobs:
     env:
       output_dir: build/plugins/wasi_nn
       test_dir: build/test/plugins/wasi_nn
-      build_options: -DWASMEDGE_PLUGIN_WASI_NN_BACKEND=GGML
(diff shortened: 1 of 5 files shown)


hydai:
Closing this because it is invalid. The wasm32 specification grants a 4 GB memory space as the maximum possible size, which can be limited and configured during runtime through options. I also do not see any strong reason to have this feature.

PR #4592, opened 2026-01-31
feat(statistics): add VM lifecycle timing for module load and instantiation

vishal2005025:
### Summary

This PR extends the existing statistics infrastructure to expose VM lifecycle timing for module load and module instantiation when `--enable-all-statistics` is enabled. These metrics provide useful insight into startup and cold-start behavior for benchmarking WasmEdge-based workloads, while reusing the existing statistics mechanism and keeping the default runtime behavior unchanged.

### What changed

- Added new internal timing tags for module load and instantiation
- Instrumented VM load and instantiation paths to record timing
- Ensured timers are correctly stopped on both success and error paths
- Extended statistics output to include the new timing metrics

## Testing

- Built WasmEdge from a clean build directory
- Ran the WasmEdge CLI locally with `--enable-all-statistics`
- Verified that the following new statistics are reported with non-zero values:
  - Module load time
  - Module instantiation time

All statistics are displayed correctly and the runtime behavior remains unchanged.

Screenshot 2026-01-19 105516


vishal2005025:
@hydai @q82419 @dannypsnl @ibmibmibm Please take a look when you get a chance and let me know if any changes are needed. Thanks!

PR #4586, opened 2026-01-30
feat: component model validator for type checking

Vikaspal8923:
Implemented validator logic for component type definitions, with full support for variant case refinement, duplicate case detection, invalid self-refinement, and out-of-bounds or undefined type references, as specified by definedtypes.wast.

 How I currently check my validator logic: I create a separate .wat file for each assert_invalid test case, then run the .wasm version of each .wat file against the updated wasmedge locally to check whether I get the expected results for each test case. I run each test and get the expected result. I have attached the screenshots as well.
 
 Note : I split the changes into two parts. One set of changes is related to validation, and the other is related to extending ast for variant refine . The validation logic changes depend on the loader changes. I have created individual PRS for both changes. 
 So this is the validation logic PR, which depends on the loader logic PR https://github.com/WasmEdge/WasmEdge/pull/4593.
 this Pr need to be merged after the loader PR https://github.com/WasmEdge/WasmEdge/pull/4593
 
 Issue: https://github.com/WasmEdge/WasmEdge/issues/4529
 
Screenshot 2026-01-30 015919
Screenshot 2026-01-30 015940
Screenshot 2026-01-30 020006

Issue #4529, opened 2026-01-18
feat: Component Model: validator for type check


LFsWang:
### Summary

Support type check and pass the test

https://github.com/WebAssembly/component-model/blob/main/test/wasm-tools/definedtypes.wast

### Details

Add more tests

Vikaspal8923:
@LFsWang , I would like to work on it

LFsWang:
@Vikaspal8923 sure

Vikaspal8923:
@q82419 @LFsWang I was looking at the variant-related tests that use (case ... (refines ...)).
While checking the current AST, I see that AST::Component::VariantTy only stores the case name and optional value type, and I don’t see a place where refines (or case identifiers like $x) are represented.  Before I move forward, just wanted to confirm: should support for refines be added to the AST/parser, or is there already an existing representation that I may be missing?

q82419:
Because of re-constructing the implementation status of component-model proposal, close this.
Please refer to the main issue #4236 and validator issue #4268 .

PR #4593, opened 2026-01-31
feat(loader): support component variant refines

Vikaspal8923:
This PR extends the component AST and loader to support variant case refinement
as defined by the component model.   PR https://github.com/WasmEdge/WasmEdge/pull/4586  is depending on this pr so this pr need to be merged before pr https://github.com/WasmEdge/WasmEdge/pull/4586

Related to the issue number : https://github.com/WasmEdge/WasmEdge/issues/4529

### DIFF
--- a/lib/loader/ast/component/component_type.cpp
+++ b/lib/loader/ast/component/component_type.cpp
@@ -381,8 +381,7 @@ Expect<void> Loader::loadType(AST::Component::RecordTy &Ty) {
 
 Expect<void> Loader::loadType(AST::Component::VariantTy &Ty) {
   // variant ::= case*:vec(<case>) => (variant case+) (if |case*| > 0)
-  // case    ::= l:<label'> t?:<valtype>? r?:<u32>? 0x00
-  //   where r? is the optional refines index
+  // case    ::= l:<label'> t?:<valtype>? r?:<u32>?
 
   auto LoadCase =
       [this](AST::Component::VariantTy::Case &Case) -> Expect<void> {


dannypsnl:
Which part of https://github.com/WebAssembly/component-model/blob/main/design/mvp/Binary.md mention this?

Vikaspal8923:
@dannypsnl The loader change here is intentionally forward-compatible: the AST and loader are extended to carry an optional refines index because variant case refinement is required by the component model validation rules and exercised by [definedtypes.wast ](https://github.com/WebAssembly/component-model/blob/main/test/wasm-tools/definedtypes.wast)

dannypsnl:
ok, can you compile `definedtypes.wast` and add it into loading test?

Vikaspal8923:
Done . i have added tests

test/component/CMakeLists.txt
```diff
@@ -1,21 +1,14 @@
...
-# SPDX-FileCopyrightText: 2019-2024 Second State INC
+#SPDX - License - Identifier : Apache - 2.0
+#SPDX - FileCopyrightText : 2019 - 2024 Second State INC
 
-wasmedge_add_executable(componentTests
-  spectest.cpp
-  componet_nameparser.cpp
-  componentvalidatortest.cpp
-)
+wasmedge_add_executable(
```
dannypsnl: Fix format here.

test/component/componentloadertest.cpp
```diff
@@ -0,0 +1,46 @@
...
+  // Component with simple variant without refines (backward compatibility)
+  // (type (variant (case "x")))
+  std::vector<uint8_t> Vec = {0x00, 0x61, 0x73, 0x6d, 0x0d, 0x00,
+                              0x01, 0x00, 0x07, 0x07, 0x01, 0x71,
+                              0x01, 0x01, 0x78, 0x00, 0x00};
+
+  ASSERT_TRUE(VM.loadWasm(Vec));
+}
+
+} // namespace
```
dannypsnl: eof

Vikaspal8923: Done

test/component/CMakeLists.txt
Vikaspal8923: Done

dannypsnl (approved):

q82419:
Sorry, I still consider that the structure should be added into the validation context, not in the loader structure.
Aspecially, the spec said:
```
case          ::= l:<label'> t?:<valtype>? 0x00
```
Where you commented in your code changes may confuse.

Vikaspal8923:
@q82419 I'm confused about how to implement variant refines validation based on your feedback.
The issue:

The binary format spec shows: `case ::= l:<label'> t?:<valtype>? 0x00 `(no refines field)
But the test cases I need to validate contain: (case "x" $t (refines 1))

My confusion:
If I don't store the refines index in the AST , how will the validator know which case is refining which at validation time?
For example, to validate this test case:
```
wat(type $v (variant 
  (case "x" $t) 
  (case "y" u64 (refines 2))))
```
The validator needs to know that case index 1 claims to refine case index 2 (so it can error because 2 > 1).

 did I misunderstand something you were trying to point out? Could you help me clarify this part further?

q82419:
> @q82419 I'm confused about how to implement variant refines validation based on your feedback. The issue:
> 
> The binary format spec shows: `case ::= l:<label'> t?:<valtype>? 0x00 `(no refines field) But the test cases I need to validate contain: (case "x" $t (refines 1))
> 
> My confusion: If I don't store the refines index in the AST , how will the validator know which case is refining which at validation time? For example, to validate this test case:
> 
> ```
> wat(type $v (variant 
>   (case "x" $t) 
>   (case "y" u64 (refines 2))))
> ```
> 
> The validator needs to know that case index 1 claims to refine case index 2 (so it can error because 2 > 1).
> 
> did I misunderstand something you were trying to point out? Could you help me clarify this part further?

According to the example you listed in the test case:

```
(component $C
  (type (variant
    (case "x")
    (case "y" string (refines 0))
    (case "z" string (refines 1))))
)
```

binary format by the `wasm-tool`:

```
0x00, 0x61, 0x73, 0x6d, 0x0d, 0x00, 0x01, 0x00, // WASM header
0x07, 0x13, // Type section, size = 13
0x01, // Vector, size = 1
0x71, 0x03, // type[0]: Variant, case vector size = 3
0x01, 0x78, 0x00, 0x00, // case[0]: label = "x", no valtype
0x01, 0x79, 0x01, 0x73, 0x01, 0x00, // case[1]: label = "y", valtype = string, refines = 0
0x01, 0x7a, 0x01, 0x73, 0x01, 0x01 // case[2]: label = "z", valtype = string, refines = 1
```

The `refines` seems like `<u32>?` in the binary format, which would be single `0x00` if there's no refines.
That is, the `refines` may be the replacement / extension for the original `0x00` flag at the end of `case` type.

Therefore, as the original rule in the spec:

```
case    ::= l:<label'> t?:<valtype>? 0x00
```

The comment you modified:

```
case    ::= l:<label'> t?:<valtype>? r?:<u32>? 0x00
//   where r? is the optional refines index
```

Must be wrong.

The correct one may be:

```
case    ::= l:<label'> t?:<valtype>? r?:<u32>?
```

But the spec doesn't mentioned about this.
I recognized that the current spec may not totally be correct, but I suggest you can create an issue in the [component model repo](https://github.com/WebAssembly/component-model) for asking a question about this first.
Because this is the issue about the binary format, and definitely not related to the validation phase.
When implementing the WASM spec, we should clearly separate the binary format in loader (for syntax checking) and the validator (for semantic checking).

Vikaspal8923:
I made the  issue regarding the clarification https://github.com/WebAssembly/component-model/issues/603   and get this clarification : 
 

>  Thanks for filing! For a while there was a refines clause along with function subtyping in the spec, but we removed them in https://github.com/WebAssembly/component-model/pull/255 since they hadn't been implemented yet and there were some subtle open questions and we wanted to release a Preview 2. The trailing 0x00 in the binary format is left over from that (to avoid breaking existing binaries that didn't use refines). So I believe that particular test should just be removed. cc @alexcrichton

So, what should I do? Should I revert loader changes, then remove the refine-related variant test case, and do the same in the validation pr?

dannypsnl:
Well, then you can open an issue to track it, and refers to this PR as reference of implementation in the future. But I don't think we will merge a feature that still not finalized by spec.

q82419:
> I made the issue regarding the clarification [WebAssembly/component-model#603](https://github.com/WebAssembly/component-model/issues/603) and get this clarification :
> 
> > Thanks for filing! For a while there was a refines clause along with function subtyping in the spec, but we removed them in [WebAssembly/component-model#255](https://github.com/WebAssembly/component-model/pull/255) since they hadn't been implemented yet and there were some subtle open questions and we wanted to release a Preview 2. The trailing 0x00 in the binary format is left over from that (to avoid breaking existing binaries that didn't use refines). So I believe that particular test should just be removed. cc @alexcrichton
> 
> So, what should I do? Should I revert loader changes, then remove the refine-related variant test case, and do the same in the validation pr?

Thanks for your issue with asking about the spec.
I think we can close this PR due to the feature is not in the spec now.
Let's look into the validation PR then.

PR #4596, opened 2026-02-01
perf(jit): improve performance of memory.size in tight loops

### LINKED ISSUE #4595: bug: Micro-benchmark: The JIT mode of wasmedge performs poorly when executing the memory.size instruction.
### Summary

Hello, I noticed that in differential micro-benchmarking, WasmEdge's JIT mode performs poorly when executing the `memory.size` instruction.

### Current State

The specific timing data is as follows:

| runtime           	|   time   	|
|-------------------	|:--------:	|
| wasmedge_jit      	|  8.7981  	|
| wasmer_llvm       	|  9.74033 	|
| wasmtime          	|  1.9795  	|
| wamr_llvm_jit     	|  0.01739 	|
| wamr_fast_jit     	|  1.32305 	|

The data is in seconds, and each data is the result of ten executions and averages.

I also submitted the relevant issue to Wasmer.

### Expected State

When changing the number of different memory pages, the execution time remains basically the same. 

In the test cases, I placed the `memory.size` instruction in a loop to amplify performance differences. 

For comparison: the timing data for an empty loop without the `memory.size` instruction is as follows:

|                 	|     time    	|
|-----------------	|:-----------:	|
| wasmedge_jit    	|   0.017236  	|
| wasmer_llvm 		|   0.01226 	|
| wasmtime         	|  0.91258  |
| wamr_llvm_jit   	|   0.01792   	|
| wamr_fast_jit   	|   0.88375   	|

### Reproduction steps

The minimal reproducible code is as follows:

<details>
<summary> test_case.wat </summary>

```
(module
  (type (func (param i32)))
  (type (func))

  (import "wasi_snapshot_preview1" "proc_exit" (func (type 0)))

  (func (type 1)
    (local $i i64)
    (local.set $i (i64.const 4294967296))
    (block $exit
      (loop $body
        memory.size
        i32.const 2147483647
        i32.gt_s
        br_if $exit
        (local.set $i (i64.sub (local.get $i) (i64.const 1)))
        (br_if $body (i64.ne (local.get $i) (i64.const 0)))
      )
    )
    (call 0 (i32.const 0))
  )

  (memory $m0 12768)

  (export "_start" (func 1))
  (export "memory" (memory 0))

)
```

</details>

```
wat2wasm test_case.wat -o test_case.wasm

# Execute the wasm file and collect data
perf stat -r 10 -e 'task-clock' /path/to/wasmedge --enable-jit test_case.wasm
perf stat -r 10 -e 'task-clock' /path/to/wasmer run -l test_case.wasm
perf stat -r 10 -e 'task-clock' /path/to/wasmtime test_case.wasm
perf stat -r 10 -e 'task-clock' /path/to/build_llvm_jit/iwasm test_case.wasm
perf stat -r 10 -e 'task-clock' /path/to/build_fast_jit/iwasm test_case.wasm
```

**Environment**

The runtime tools are all built on release and use JIT mode.

- wasmedge: 0.16.1-18-gc457fe30
- wasmer: 6.1.0
- WAMR: iwasm 2.4.3
- wabt: 1.0.27
- llvm: 21.1.5

_No response_

### Components

CLI

0.16.1-18-gc457fe30

Host OS: Ubuntu 22.04.5 LTS

### Hardware Architecture

x86_64

- g++: 11.4.0
- cmake: 3.22.1
- cmake flags: `-DCMAKE_BUILD_TYPE=Release -DWASMEDGE_BUILD_TESTS=ON`

vishal2005025:
### Summary

In WasmEdge JIT mode, `memory.size` is lowered to an indirect runtime call and executed on every iteration when used inside tight loops. This causes severe performance degradation compared to empty loops, even though the memory size is loop-invariant in many cases.

This PR improves the performance of `memory.size` in JIT execution by caching its value once per function **when semantics guarantee safety**, avoiding repeated runtime calls in hot loops.

### Changes

- Analyze each function to identify `memory.size` instructions that are safe to cache
- Cache `memory.size` values in the function entry block to ensure SSA dominance
- Reuse the cached value for all subsequent `memory.size` instructions in the function
- Disable caching when:
  - `memory.grow` targets the same memory in the function
  - the Threads proposal is enabled and the memory is shared
- Preserve the original behavior for all non-cacheable cases

This change is scoped to JIT code generation and does not affect the interpreter or AOT execution.

### Testing 

- The same test cases described in the issue were used to reproduce and verify the performance improvement.

<details>
<summary><code>test.wat</code></summary>

```wat
(module
  (type (func (param i32)))
  (type (func))

  (import "wasi_snapshot_preview1" "proc_exit" (func (type 0)))

  (func (type 1)
    (local $i i64)
    (local.set $i (i64.const 4294967296))
    (block $exit
      (loop $body
        memory.size
        i32.const 2147483647
        i32.gt_s
        br_if $exit
        (local.set $i (i64.sub (local.get $i) (i64.const 1)))
        (br_if $body (i64.ne (local.get $i) (i64.const 0)))
      )
    )
    (call 0 (i32.const 0))
  )

  (memory $m0 12768)

  (export "_start" (func 1))
  (export "memory" (memory 0))
)
```
</details>

<details>
<summary><code>emptyloop.wat</code></summary>

```wat
(module
  (type (func (param i32)))
  (type (func))

  (import "wasi_snapshot_preview1" "proc_exit" (func (type 0)))

  (func (type 1)
    (local $i i64)
    (local.set $i (i64.const 4294967296))
    (block $exit
      (loop $body
        ;; EMPTY LOOP BODY
        (local.set $i (i64.sub (local.get $i) (i64.const 1)))
        (br_if $body (i64.ne (local.get $i) (i64.const 0)))
      )
    )
    (call 0 (i32.const 0))
  )

  (memory $m0 12768)

  (export "_start" (func 1))
)

```
</details>

Screenshot: 

Screenshot 2026-02-01 181100

### Performance

| Test case | Before | After |
|----------|--------|-------|
| `memory.size` in tight loop (JIT) | ~7.0s | ~0.22s |
| Empty loop (JIT) | ~0.06s | ~0.06s |

### DIFF
--- a/lib/llvm/compiler.cpp
+++ b/lib/llvm/compiler.cpp
@@ -128,6 +128,8 @@ static inline LLVMCodeGenOptLevel toLLVMCodeGenLevel(
 struct LLVM::Compiler::CompileContext {
   LLVM::Context LLContext;
   LLVM::Module &LLModule;
+  const Configure &Conf;
+  const AST::Module &Module;
   LLVM::Attribute Cold;
   LLVM::Attribute NoAlias;
   LLVM::Attribute NoInline;
@@ -204,9 +206,9 @@ struct LLVM::Compiler::CompileContext {
   std::vector<LLVM::Type> Globals;
   LLVM::Value IntrinsicsTable;
   LLVM::FunctionCallee Trap;
-  CompileContext(LLVM::Context C, LLVM::Module &M,
-                 bool IsGenericBinary) noexcept
-      : LLContext(C), LLModule(M),
+  CompileContext(LLVM::Context C, LLVM::Module &M, const Configure &Conf,
+                 const AST::Module &Module, bool IsGenericBinary) noexcept
+      : LLContext(C), LLModule(M), Conf(Conf), Module(Module),
         Cold(LLVM::Attribute::createEnum(C, LLVM::Core::Cold, 0)),
         NoAlias(LLVM::Attribute::createEnum(C, LLVM::Core::NoAlias, 0)),
         NoInline(LLVM::Attribute::createEnum(C, LLVM::Core::NoInline, 0)),
@@ -519,7 +521,8 @@ public:
       : Context(Context), LLContext(Context.LLContext),
         Interruptible(Interruptible), F(F), Builder(LLContext) {
     if (F.Fn) {
-      Builder.positionAtEnd(LLVM::BasicBlock::create(LLContext, F.Fn, "entry"));
+      EntryBB = LLVM::BasicBlock::create(LLContext, F.Fn, "entry");
+      Builder.positionAtEnd(EntryBB);
       ExecCtx = Builder.createLoad(Context.ExecCtxTy, F.Fn.getFirstParam());
 
       if (InstructionCounting) {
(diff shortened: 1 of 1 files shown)


vishal2005025:
@hydai @q82419 @ibmibmibm Please take a look and let me know if any changes are required.
Thanks!

lib/llvm/compiler.cpp
```diff
@@ -558,12 +561,64 @@ class FunctionCompiler {
...
+    }
+  }
+
   Expect<void>
   compile(const AST::CodeSegment &Code,
           std::pair<std::vector<ValType>, std::vector<ValType>> Type) noexcept {
     auto RetBB = LLVM::BasicBlock::create(LLContext, F.Fn, "ret");
     Type.first.clear();
     enterBlock(RetBB, {}, {}, {}, std::move(Type));
+    computeCacheableMemories(Code.getExpr().getInstrs());
```
hydai, line 617: What will happen if the following code executes? Does your cache work in this case?

```wat
  (module
    (memory 1)
    (func $grow_helper (drop (memory.grow (i32.const 1))))
    (func $main (export "main")
      (drop (memory.size))
      (call $grow_helper)
      (drop (memory.size))
    )
  )
```

hydai (changes requested):
You are changing the behavior without covering any test cases.

vishal2005025:
@hydai Thanks for the example, 
Previously, memory.size could be cached even when a subsequent call indirectly executed memory.grow, which could lead to a stale value being reused.
I’ve tightened the cacheability logic to conservatively disable memory.size caching when the function contains any call or call_indirect, and added a test that traps if a stale value is observed. 

testgrow.wat 
```(module
  (memory 1)

  (func $grow_helper
    i32.const 1
    memory.grow
    drop
  )

  (func (export "_start")
    ;; First memory.size must be 1
    memory.size
    i32.const 1
    i32.ne
    if
      unreachable
    end

    ;; Grow memory via call
    call $grow_helper

    ;; Second memory.size must be 2
    memory.size
    i32.const 2
    i32.ne
    if
      unreachable
    end
  )
)
```

Before :
Screenshot 2026-02-02 122822

After :
Screenshot 2026-02-02 124607

Please take an another look. Thanks!

lib/llvm/compiler.cpp
```diff
@@ -558,12 +561,73 @@ class FunctionCompiler {
     return BB;
   }
 
+  void computeCacheableMemories(AST::InstrView Instrs) noexcept {
+    // Find memory indices that are grown in this function
+    std::unordered_set<uint32_t> GrowingMemories;
+    // Check for any call instructions (may execute memory.grow transitively)
+    bool HasCall = false;
+    for (const auto &Instr : Instrs) {
+      if (Instr.getOpCode() == OpCode::Memory__grow) {
```
PhantomInTheWire, line 570: this incorrectly caches memory.size results by only checking if the current function contains Memory__grow instructions. However, it fails to account for function calls (Call, Call_indirect, Call_ref, Return_call, Return_call_indirect, Return_call_ref) that can invoke other functions which may grow memory.

PhantomInTheWire (commented):
the solution you propose is incorrect, I would recommend looking at how others(node/v8, wasmtime) handle this and coming up with good test cases to verify correctness

vishal2005025:
Thanks for the review  - after looking more closely at how Wasmtime (Cranelift) and V8 handle `memory.size`, and I think the key difference is in the architecture rather than a specific optimization.

In Wasmtime, `memory.size` is not special-cased or cached. It lowers to normal load / atomic_load instructions that read memory metadata and are intentionally not marked `readonly` or `can_move`. Because of that, LICM and other loop
optimizations only apply when the effect system can prove safety. The performance there comes from cheaper IR lowering (direct loads) and a generic effect-aware optimizer, not from opcode-level caching.

In contrast, WasmEdge’s JIT currently lowers `memory.size` via an indirect runtime call, which explains the performance gap in tight loops. Trying to close that gap purely at the frontend level (e.g. caching or hoisting based on local inspection) does not match the correctness model used by Wasmtime/V8, especially in the presence of transitive calls, indirect calls, or imports.

I also put together adversarial cases (transitive `memory.grow`, indirect calls) to reason about correctness. These cases illustrate why opcode-local or function-local caching is fragile without an effect system.

Based on this, I think the fixes that align best with Wasmtime/V8 design principles would be along the lines of:
  * introducing effect-aware infrastructure (so optimizations are driven by `readonly` / `can_move`–style properties rather than opcode checks), or
  * lowering `memory.size` to direct IR loads instead of a runtime call, so that existing compiler optimizations can apply naturally.

I’d like to confirm whether this direction aligns with the intended approach for WasmEdge, before going further in that direction.
@PhantomInTheWire @hydai please take an another look.
Thanks!

q82419:
> Thanks for the review - after looking more closely at how Wasmtime (Cranelift) and V8 handle `memory.size`, and I think the key difference is in the architecture rather than a specific optimization.
> 
> In Wasmtime, `memory.size` is not special-cased or cached. It lowers to normal load / atomic_load instructions that read memory metadata and are intentionally not marked `readonly` or `can_move`. Because of that, LICM and other loop optimizations only apply when the effect system can prove safety. The performance there comes from cheaper IR lowering (direct loads) and a generic effect-aware optimizer, not from opcode-level caching.
> 
> In contrast, WasmEdge’s JIT currently lowers `memory.size` via an indirect runtime call, which explains the performance gap in tight loops. Trying to close that gap purely at the frontend level (e.g. caching or hoisting based on local inspection) does not match the correctness model used by Wasmtime/V8, especially in the presence of transitive calls, indirect calls, or imports.
> 
> I also put together adversarial cases (transitive `memory.grow`, indirect calls) to reason about correctness. These cases illustrate why opcode-local or function-local caching is fragile without an effect system.
> 
> Based on this, I think the fixes that align best with Wasmtime/V8 design principles would be along the lines of:
> 
> * introducing effect-aware infrastructure (so optimizations are driven by `readonly` / `can_move`–style properties rather than opcode checks), or
> * lowering `memory.size` to direct IR loads instead of a runtime call, so that existing compiler optimizations can apply naturally.
> 
> I’d like to confirm whether this direction aligns with the intended approach for WasmEdge, before going further in that direction. @PhantomInTheWire @hydai please take an another look. Thanks!

In my opinion, the second one - lowering `memory.size` to direct IR loads instead of a runtime call is the better choice.
For the AOT/JIT related mechanism cc @ibmibmibm

q82419:
replaced by #5088

PR #4599, opened 2026-02-01
feat(plugin): add hot-reloading support for plugins

asmit27rai:
This PR adds support for hot-reloading (live reloading) of plugins in WasmEdge. This feature allows users to dynamically load, unload, or update plugins at runtime without restarting the WasmEdge host process.

As discussed in #4581, while WASI-NN plugins can dynamically change models, updating the plugin's inference logic, preprocessing code, or swapping backends still requires restarting the host process. This PR addresses that limitation by enabling full plugin hot-reloading.


asmit27rai:
After the discussion, I agree this doesn't address a real use case. The scenarios I proposed are better handled by existing mechanisms. Closing this PR.

PR #4600, opened 2026-02-01
ci: cache homebrew deps

PhantomInTheWire:
**Before (recent CI run)**
[https://github.com/WasmEdge/WasmEdge/actions/runs/21559718073/job/62130271902?pr=4594](https://github.com/WasmEdge/WasmEdge/actions/runs/21559718073/job/62130271902?pr=4594)

**After (CI run for this PR on fork)**
[https://github.com/PhantomInTheWire/WasmEdge/actions/runs/21569964185/job/62147471372?pr=4](https://github.com/PhantomInTheWire/WasmEdge/actions/runs/21569964185/job/62147471372?pr=4)

On Intel macOS 15 runners, the core workflow consistently took the longest, around 30 minutes, making it the primary bottleneck. The main reason was that several Homebrew dependencies were being installed and `lld` was built from source on every run.

By caching `lld` and other heavy dependencies like LLVM, Ninja, WABT, and gRPC, the Intel macOS 15 job time was reduced by approximately 5 to 7 minutes. Since this job is the bottleneck, the overall core workflow time is also reduced by roughly the same amount.

At this point, the second slowest job is the Windows MSVC build, which completes in about 17 to 18 minutes. This implies that further optimizations in non macOS workflows will not meaningfully reduce the total core CI duration unless the Intel macOS 15 runtime can be brought down by an additional 5 to 7 minutes.(which I think is non trivial since most of the remaining time is build and test, so unless we can change the process itself it would be difficult, although I think if we build wasmedge in release mode the tests should complete much faster but I am not sure if we want that since the workflow has a special non release section for normal ci runs, would love some thoughts on this part)

cc: @0yi0

### DIFF
--- a/.github/workflows/reusable-build-on-macos.yml
+++ b/.github/workflows/reusable-build-on-macos.yml
@@ -43,17 +43,75 @@ jobs:
       - name: Setup build environment - non-release
         if: ${{ !inputs.release }}
         run: |
+          echo "BREW_PREFIX=$(brew --prefix)" >> $GITHUB_ENV
           brew uninstall cmake # conflict with the one used by WasmEdge/llvm/lld
           brew tap WasmEdge/llvm
+      - name: Install Homebrew tools - non-release
+        if: ${{ !inputs.release }}
+        run: |
           brew install llvm ninja wabt grpc
-          brew install --build-from-source WasmEdge/llvm/lld
+      - name: Get lld version
+        if: ${{ !inputs.release }}
+        id: lld-version-non-release
+        run: |
+          VERSION=$(brew info --json WasmEdge/llvm/lld | jq -r '.[0].versions.stable')
+          echo "version=${VERSION}" >> $GITHUB_OUTPUT
+      - name: Cache lld and cmake - non-release
+        if: ${{ !inputs.release }}
+        id: cache-lld-non-release
+        uses: actions/cache@cdf6c1fa76f9f475f3d7449005a359c84ca0f306 # v5.0.3
+        with:
+          path: |
+            ${{ env.BREW_PREFIX }}/Cellar/lld
+            ${{ env.BREW_PREFIX }}/Cellar/cmake
+          key: homebrew-cellar-lld-${{ runner.os }}-${{ matrix.arch }}-${{ steps.lld-version-non-release.outputs.version }}-${{ hashFiles('.github/workflows/reusable-build-on-macos.yml') }}
+          restore-keys: |
+            homebrew-cellar-lld-${{ runner.os }}-${{ matrix.arch }}-${{ steps.lld-version-non-release.outputs.version }}-
+            homebrew-cellar-lld-${{ runner.os }}-${{ matrix.arch }}-
+      - name: Install or link lld and cmake - non-release
+        if: ${{ !inputs.release }}
+        run: |
+          if [ "${{ steps.cache-lld-non-release.outputs.cache-hit }}" != "true" ]; then
+            brew install --build-from-source WasmEdge/llvm/lld
+          fi
+          brew link lld || true
+          brew link cmake || true
       - name: Setup build environment - release
         if: ${{ inputs.release }}
         run: |
+          echo "BREW_PREFIX=$(brew --prefix)" >> $GITHUB_ENV
           brew uninstall cmake # conflict with the one used by WasmEdge/llvm/lld
           brew tap WasmEdge/llvm
+      - name: Install Homebrew tools - release
+        if: ${{ inputs.release }}
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/reusable-build-on-macos.yml
```diff
@@ -43,17 +43,71 @@ jobs:
...
+          else
+            echo "BREW_PREFIX=/usr/local" >> $GITHUB_ENV
+          fi
           brew uninstall cmake # conflict with the one used by WasmEdge/llvm/lld
           brew tap WasmEdge/llvm
-          brew install llvm ninja wabt grpc
-          brew install --build-from-source WasmEdge/llvm/lld
+      - name: Install Homebrew tools - non-release
+        if: ${{ !inputs.release }}
+        uses: tecolicom/actions-use-homebrew-tools@b9c066b79607fa3d71e0be05d7003bb75fd9ff34 # v1
```
hydai: Why do you want to introduce a rather old and seemingly unmaintained action here instead of simply using the `brew install` command?

.github/workflows/reusable-build-on-macos.yml
```diff
@@ -43,17 +43,71 @@ jobs:
...
+        with:
+          tools: llvm ninja wabt grpc
+      - name: Cache lld and cmake - non-release
+        if: ${{ !inputs.release }}
+        id: cache-lld-non-release
+        uses: actions/cache@1bd1e32a3bdc45362d1e726936510720a7c30a57 # v4.2.0
```
hydai: Why do you use v4 instead of the latest v5?

.github/workflows/reusable-build-on-macos.yml
```diff
@@ -43,17 +43,71 @@ jobs:
...
+        with:
+          path: |
+            ${{ env.BREW_PREFIX }}/Cellar/lld
+            ${{ env.BREW_PREFIX }}/Cellar/cmake
+          key: homebrew-cellar-lld-${{ runner.os }}-${{ matrix.arch }}-${{ hashFiles('.github/workflows/reusable-build-on-macos.yml') }}
```
hydai: Since your key uses `hashFiles`, what will happen if `lld` has a newer version? Will it still hit the cache?

.github/workflows/reusable-build-on-macos.yml
PhantomInTheWire: I also thought that brew install should be the same as caching since the network i/o component should dominate to the point where there isn't a significant difference in time consumed. turns out that isnt the case, raw brew install takes 2min20secs compared to 40 secs with cache hit.
https://github.com/PhantomInTheWire/WasmEdge/actions/runs/21592486686/job/62216033723?pr=4
that said I can replace the action by pulling all of its code, inside our codebase.

that said this can wait for now and be a follow up pr(since it would be a big change, around 200 loc), this would also help us a lot more in extensions workflow, where this kind of caching was saving a lot more time.(i couldnt find the run but if i recall correctly it was saving 3-4 minutes.)

.github/workflows/reusable-build-on-macos.yml
PhantomInTheWire: my bad, didn't realize v5 was out. thanks.

.github/workflows/reusable-build-on-macos.yml
PhantomInTheWire: added lld version to cache key

.github/workflows/reusable-build-on-macos.yml
```diff
@@ -43,17 +43,83 @@ jobs:
       - name: Setup build environment - non-release
         if: ${{ !inputs.release }}
         run: |
+          if [ "${{ matrix.arch }}" == "arm64" ]; then
+            echo "BREW_PREFIX=/opt/homebrew" >> $GITHUB_ENV
+          else
+            echo "BREW_PREFIX=/usr/local" >> $GITHUB_ENV
+          fi
```
hydai: The prefix can be retrieved using `brew --prefix`. What is the reason for exporting it with a hardcoded string?

hydai (changes requested):

.github/workflows/reusable-build-on-macos.yml
PhantomInTheWire: gh actions can't evaluate bash commands in "path", see line 69,70

hydai: Why not `echo "BREW_PREFIX=$(brew --prefix)" >> $GITHUB_ENV`? Is it also not available in this block?

PhantomInTheWire: no i think that should be fine

PR #4602, opened 2026-02-02
fix(cli): handle invalid numeric arguments gracefully

### LINKED ISSUE #4588: bug: CLI crashes on invalid numeric arguments due to unhandled std::sto* exceptions
### Summary

When executing a WASM function via the WasmEdge CLI, passing invalid numeric arguments (e.g. non-numeric strings or out-of-range values) causes the runtime to crash with an unhandled exception instead of showing a user-friendly error.

### Current State

The CLI directly uses `std::stol`, `std::stoll`, `std::stof`, and `std::stod` for argument parsing in `runtimeTool.cpp` without exception handling.
If a user passes invalid input (like `"abc"` for an integer), the process terminates with `SIGABRT` due to an uncaught exception.

### **Expected State**

The CLI should catch conversion errors and return a clear, user-friendly error message, such as:

```
Error: Invalid argument 'abc' for parameter 1: expected i32
```

The process should exit gracefully instead of crashing.

### Reproduction steps

1. Install WasmEdge (tested with released version):

   ```bash
   wasmedge --version
   ```
2. Create a simple WASM module with an integer parameter (e.g. `add(i32, i32)`).
3. Execute with an invalid argument:

   ```bash
   wasmedge --reactor add.wasm add abc 5
   ```
4. Observe the crash with an unhandled exception.

### Screenshots

Not applicable (CLI output shown in logs below).

```
libc++abi: terminating due to uncaught exception of type std::invalid_argument: stol: no conversion
zsh: abort      wasmedge --reactor add.wasm add abc 5
Exit code: 134 (SIGABRT)
```

### Components

CLI

0.14.1

macOS

### Hardware Architecture

arm64

_No response_

Meetjain1:
This PR fixes a crash in the CLI when invalid numeric arguments are passed.

When non-numeric or out-of-range values (for example `"abc"` or very large integers) are provided, the CLI previously threw uncaught exceptions and aborted. This change adds exception handling around numeric argument parsing so the CLI fails gracefully with a clear error message and a proper exit code.

Closes #4588

CI checks - https://github.com/Meetjain1/WasmEdge/pull/2

image

### DIFF
--- a/.github/scripts/run-riscv64-quick-tests.sh
+++ b/.github/scripts/run-riscv64-quick-tests.sh
@@ -63,7 +63,7 @@ QUICK_TESTS="
   wasmedgeLoaderASTTests
   wasmedgeLoaderSerializerTests
   wasmedgeValidatorSubtypeTests
-  wasmedgeMemLimitTests
+  wasmedgeMemInstanceTests
   wasmedgeExternrefTests
   wasmedgeHostMockTests
   wasmedgeAPIUnitTests
--- a/.github/workflows/build_for_riscv.yml
+++ b/.github/workflows/build_for_riscv.yml
@@ -78,13 +78,13 @@ jobs:
           gcc-riscv64-linux-gnu g++-riscv64-linux-gnu \
           zlib1g-dev:riscv64
 
-        # llvm-18-dev:riscv64 cannot be installed via apt-get because its
-        # dependency chain pulls in python3.12-minimal:riscv64, whose postinst
-        # script tries to execute /usr/bin/python3.12 (a riscv64 binary) on
-        # the x86_64 host, causing "Exec format error".
+        # llvm-20-dev:riscv64 cannot be installed via apt-get because its
+        # dependency chain pulls in python3-minimal:riscv64, whose postinst
+        # script tries to execute a riscv64 binary on the x86_64 host,
+        # causing "Exec format error".
         # Workaround: download the packages and extract without running postinst.
         apt-get install -y -d --no-install-recommends \
-          llvm-18-dev:riscv64 liblld-18-dev:riscv64
+          llvm-20-dev:riscv64 liblld-20-dev:riscv64
         for deb in /var/cache/apt/archives/*_riscv64.deb; do
           dpkg-deb -x "$deb" /
         done
@@ -98,7 +98,7 @@ jobs:
           -DWASMEDGE_BUILD_TOOLS=ON \
           -DWASMEDGE_BUILD_SHARED_LIB=ON \
           -DWASMEDGE_USE_LLVM=ON \
-          -DLLVM_DIR=/usr/lib/riscv64-linux-gnu/cmake/llvm-18 \
+          -DLLVM_DIR=/usr/lib/riscv64-linux-gnu/cmake/llvm-20 \
           .
         cmake --build build -j $(nproc)
 
@@ -164,8 +164,8 @@ jobs:
         # Install riscv64 LLVM/LLD runtime libraries (these do NOT depend on
         # python3:riscv64, so apt-get install works without postinst issues)
         apt-get install -y --no-install-recommends \
-          libllvm18:riscv64 \
-          liblld-18:riscv64 \
+          libllvm20:riscv64 \
+          liblld-20:riscv64 \
           zlib1g:riscv64
 
         # Create symlinks so QEMU sysroot can find multi-arch libraries.
--- a/include/ast/component/canonical.h
+++ b/include/ast/component/canonical.h
@@ -13,6 +13,7 @@
 //===----------------------------------------------------------------------===//
 #pragma once
 
+#include "ast/component/valtype.h"
 #include "common/span.h"
 
 #include <cstdint>
@@ -44,6 +45,7 @@ public:
     PostReturn = 0x05,
     Async = 0x06,
     Callback = 0x07,
+    AlwaysTaskReturn = 0x08,
   };
 
   CanonOpt() noexcept : Code(OptCode::Encode_UTF8), Idx(0) {}
@@ -193,12 +195,22 @@ public:
     Opts = std::move(List);
   }
 
+  Span<const LabelValType> getResultList() const noexcept { return ResultList; }
+  void setResultList(std::vector<LabelValType> &&R) noexcept {
+    ResultList = std::move(R);
+  }
+  void setResultList(const ComponentValType &VT) noexcept {
+    ResultList.clear();
+    ResultList.emplace_back(VT);
+  }
+
 private:
   OpCode Code;
   bool IsAsync;
   uint32_t Idx, TargetIdx;
   uint32_t I32;
   std::vector<CanonOpt> Opts;
+  std::vector<LabelValType> ResultList;
 };
 
 } // namespace Component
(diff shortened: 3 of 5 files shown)


lib/driver/runtimeTool.cpp
```diff
@@ -35,45 +35,67 @@ ToolOnModule(WasmEdge::VM::VM &VM, const std::string &FuncName,
...
-    case TypeCode::F64: {
-      const double Value = std::stod(Opt.Args.value()[I + 1]);
-      FuncArgs.emplace_back(Value);
-      FuncArgTypes.emplace_back(TCode);
-      break;
-    }
-    // TODO: FuncRef and ExternRef
-    default:
-      break;
+    try {
```
PhantomInTheWire: how about using a template here to avoid duplication?
```cpp
template <typename T, typename Func>
bool parseNumericArg(const std::string &Value, size_t ParamIndex, T &Out,
                     Func Converter) {
  try {
    Out = Converter(Value);
    return true;
  } catch (const std::invalid_argument &) {
    spdlog::error(
        "Invalid argument '{}' for parameter {}: expected numeric value"sv,
        Value, ParamIndex + 1);
    return false;
  } catch (const std::out_of_range &) {
    spdlog::error("Argument '{}' for parameter {}: value out of range"sv, Value,
                  ParamIndex + 1);
    return false;
  }
}
```
something like this

Meetjain1: Sure will add

Meetjain1:
@hydai PTAL on this PR.

lib/driver/runtimeTool.cpp
```diff
@@ -23,6 +23,25 @@ using namespace std::literals;
...
+// Helper template to parse numeric arguments and catch conversion exceptions
+template <typename T, typename Converter>
+inline bool parseNumericArg(const std::string &Value, size_t ParamIndex, T &Out,
+                            Converter Conv) {
+  try {
+    Out = Conv(Value);
+    return true;
+  } catch (const std::invalid_argument &) {
+    spdlog::error(
+        "Invalid argument '{}' for parameter {}: expected numeric value"sv,
```
hydai: Since you already know the type, you should display the typename instead of "numeric value."

lib/driver/runtimeTool.cpp
```diff
@@ -23,6 +23,25 @@ using namespace std::literals;
...
+        Value, ParamIndex);
+    return false;
+  } catch (const std::out_of_range &) {
+    spdlog::error("Argument '{}' for parameter {}: value out of range"sv, Value,
```
hydai: Ditto, we should provide enough information to let users know the expected typename, so it's clearly out of range.

lib/driver/runtimeTool.cpp
```diff
@@ -23,6 +23,25 @@ using namespace std::literals;
 namespace WasmEdge {
 namespace Driver {
 
+// Helper template to parse numeric arguments and catch conversion exceptions
+template <typename T, typename Converter>
+inline bool parseNumericArg(const std::string &Value, size_t ParamIndex, T &Out,
```
hydai: I don't think we need `inline` keyword here. Templates are implicitly inline.

lib/driver/runtimeTool.cpp
```diff
@@ -23,6 +23,25 @@ using namespace std::literals;
...
+    spdlog::error(
+        "Invalid argument '{}' for parameter {}: expected numeric value"sv,
+        Value, ParamIndex);
+    return false;
+  } catch (const std::out_of_range &) {
+    spdlog::error("Argument '{}' for parameter {}: value out of range"sv, Value,
+                  ParamIndex);
+    return false;
+  }
+}
```
hydai, line 45:
```suggestion
template <typename Converter, typename ValVec, typename TypeVec, typename TC>
bool parseNumericArg(const std::string &Value, size_t ParamIndex,
                     Converter Conv, ValVec &FuncArgs, TypeVec &FuncArgTypes,
                     TC TCode) {
  try {
    auto Out = Conv(Value);
    FuncArgs.emplace_back(Out);
    FuncArgTypes.emplace_back(TCode);
    return true;
  } catch (const std::invalid_argument &) {
    spdlog::error(
        "Invalid argument '{}' for parameter {}: expected numeric value"sv,
        Value, ParamIndex + 1);
    return false;
  } catch (const std::out_of_range &) {
    spdlog::error("Argument '{}' for parameter {}: value out of range"sv,
                  Value, ParamIndex + 1);
    return false;
  }
}
```

lib/driver/runtimeTool.cpp
```diff
@@ -35,29 +54,47 @@ ToolOnModule(WasmEdge::VM::VM &VM, const std::string &FuncName,
        I < FuncType.getParamTypes().size() && I + 1 < Opt.Args.value().size();
        ++I) {
     const auto TCode = FuncType.getParamTypes()[I].getCode();
+    const std::string &ArgValue = Opt.Args.value()[I + 1];
```
hydai:
```suggestion
    const auto &ArgValue = Opt.Args.value()[I + 1];
```

lib/driver/runtimeTool.cpp
```diff
@@ -35,29 +54,47 @@ ToolOnModule(WasmEdge::VM::VM &VM, const std::string &FuncName,
...
+      int32_t Value = 0;
+      if (!parseNumericArg(ArgValue, I, Value, [](const std::string &S) {
+            return static_cast<int32_t>(std::stol(S));
+          })) {
+        return EXIT_FAILURE;
+      }
       FuncArgs.emplace_back(Value);
       FuncArgTypes.emplace_back(TCode);
       break;
     }
```
hydai, line 72:
```cpp
    case TypeCode::I32: {
      if (!parseNumericArg(
              ArgValue, I,
              [](const std::string &S) {
                return static_cast<int32_t>(std::stol(S));
              },
              FuncArgs, FuncArgTypes, TCode)) {
        return EXIT_FAILURE;
      }
    }
```

lib/driver/runtimeTool.cpp
Meetjain1: Done.

lib/driver/runtimeTool.cpp
Meetjain1: Done.

lib/driver/runtimeTool.cpp
Meetjain1: Removed.

lib/driver/runtimeTool.cpp
Meetjain1: Fixed.

lib/driver/runtimeTool.cpp
Meetjain1: Fixed.

lib/driver/runtimeTool.cpp
Meetjain1: fixed.

Meetjain1:
@hydai Addressed all comments with latest CI checks https://github.com/Meetjain1/WasmEdge/pull/2.

Meetjain1:
@hydai anything else needed in this PR?

q82419:
@Meetjain1 

Please rebase to the latest master branch and resolve the conflicts, and then we would merge this PR. Thanks!

Meetjain1:
> @Meetjain1
> 
> Please rebase to the latest master branch and resolve the conflicts, and then we would merge this PR. Thanks!

@q82419 done.

hydai (approved):

PR #4590, opened 2026-01-31
feat(component): add extended interface version canonicalization supp…

sagar-h007:
### Summary
This PR adds support for extended interface version canonicalization to WasmEdge’s Component Model loader, following the Component Model spec (PR #536).

In practical terms, this lets WasmEdge correctly handle versioned component interfaces, including canonical version matching, while remaining fully compatible with existing components.

### What’s included

Versioned interfaces
- Import and export names now support canonical interface versions
- Both formats are handled:
- Legacy 0x00 encoding
- Extended 0x01 encoding with version suffixes

Version handling
- Introduces a small utility for:
- Parsing semantic versions
- Canonicalizing them according to the spec
- Checking version compatibility
#### Examples:
- 1.2.3 → 1
- 0.2.6-rc.1 → 0.2
- 0.0.1-alpha → 0.0.1

### Loader changes
The binary loader now understands extended import/export names
Version suffixes are stored separately so the full version is preserved for:
Better error messages
Future validation and tooling

### Tests
Adds unit tests covering:
Valid and invalid semantic versions
Canonicalization rules
Version compatibility
Real WASI Preview 2 cases
Updates interface name parsing tests to include canonical versions

### Why this matters
The Component Model relies on canonical version matching to allow compatible components to link even when their patch versions differ.

Without this support, versioned interfaces either fail to match or rely on fragile string comparisons. This PR fixes that and brings WasmEdge in line with the spec.

comprehensive unit tests for semver parsing, canonicalization, and WASI Preview 2 cases

### Testing
- New unit tests added
- Existing test suite passes

Fixes: #4276.

Issue #4268, opened 2025-07-17
feat: Component Model: Validator completion after refactoring component AST structures


q82419:
### Summary

This feature is after the merging of PR #4173 .

1. Complete the empty validation functions in `lib/validator/component_validator.cpp`.
2. Add the necessary error codes in `include/common/enum.inc`.
3. Pass the validation tests in `test/component/`.

### Details

1. The code architectures are ready in the validator part. For the adjustment of some AST nodes, you can refer to the code in `loader`.
2. The component context for the validation should be refined.
3. The current codes may not be correct. You should check the current codes again according to the spec.

### Appendix

1. Validation rules are mainly in the [explainer](https://github.com/WebAssembly/component-model/blob/main/design/mvp/Explainer.md). The proposal repo has no validation rules page.
2. For some info, can check in the previous LFX issue: #4044

LFsWang:
I can do it

q82419:
@LFsWang 

I think the following information may help if you want some test or details of validation.
Take the [component module in this test file](https://github.com/WebAssembly/component-model/blob/main/test/resources/multiple-resources.wast#L4) for example.
You can use the [wasm-tools](https://github.com/bytecodealliance/wasm-tools) to translate the component text format into binary format: `wasm-tools parse test.wat -o test.wasm`.
WasmEdge loader can load the binary format into the AST structure. And also please feel free to point out the wrong implementation in loader if found.
Then for the validation context, you can use the `wasm-tool` to translate the binary format back into text format: `wasm-tools print test.wasm`.

The original component text format may be like:

```
(component
  (component $C
    (core module $Indirect
      (table (export "ftbl") 2 funcref)
      (type $FT (func (param i32)))
      (func (export "R1-dtor") (param i32)
        (call_indirect (type $FT) (local.get 0) (i32.const 0))
      )
      (func (export "R2-dtor") (param i32)
        (call_indirect (type $FT) (local.get 0) (i32.const 1))
      )
    )
    (core instance $indirect (instantiate $Indirect))
    (type $R1' (resource (rep i32) (dtor (func $indirect "R1-dtor"))))
    (type $R2' (resource (rep i32) (dtor (func $indirect "R2-dtor"))))
    (export $R1 "R1" (type $R1'))
    (export $R2 "R2" (type $R2'))
    (canon resource.new $R1' (core func $R1.resource.new))
    (canon resource.new $R2' (core func $R2.resource.new))

    (core module $CM
[...]

Issue #4236, opened 2025-07-07
feat: Component Model Proposal and WASI Preview 2 in WasmEdge


q82419:
### Summary

This issue is for tracking the implmentation status of [Component Model proposal](https://github.com/WebAssembly/component-model/tree/main) and [WASI preview 2](https://github.com/orgs/bytecodealliance/projects/10/views/1).

### Details

TBD

### Appendix

1. Previous component model issue: #2943
2. Previous WASI preview2 issue: #2939

Ankitdotraider:
Hi @q82419 , I'm interested in contributing to the Component Model validator. I noticed several spec tests still show ❌ for validation (e.g., alias.wast, func.wast, import.wast). Are any of these available to work on, or are they already being picked up by someone? Happy to start with a small one if you can point me in the right direction.

q82419:
Currently, we are in a large scale of refactoring and completion of component model proposal (see PR #5170)
The completion needs the full picture of architecture instead of just minor fixing for passing single test suites.
Therefore we would CLOSE the un-reviewed or conflicted existing component-model PRs and not accept new component-model PRs before the completion.

cc reviewer @LFsWang  @dannypsnl

PR #4604, opened 2026-02-02
Validator support for versioned interface names

sagar-h007:
This PR follows up on the #4590 that added loader support for parsing versioned component interface names.

It was suggested that validator logic should be split out and properly wired into the validator pipeline. This PR does exactly that.

### What’s included
- Validate version suffixes on component import and export interface names.
- Ensure version strings follow a semver-compatible major.minor.patch format.
- Emit clear validator errors for invalid version formats.
- Keep existing behavior unchanged for legacy interfaces without version suffixes.

### Tests

- Added validator tests covering valid and invalid versioned imports and exports, as well as legacy interfaces without versions.

This keeps the validator changes small, focused, and directly connected to the earlier loader work.

### DIFF
--- a/include/validator/component_name.h
+++ b/include/validator/component_name.h
@@ -25,9 +25,7 @@ bool tryRead(std::string_view prefix, std::string_view &name);
 bool tryReadKebab(std::string_view &input, std::string_view &output);
 
 bool isValidSemVer(std::string_view version);
-
 }; // namespace ComponentNameParser
-
 
 enum class ComponentNameKind {
   Invalid,
--- a/lib/validator/component_name.cpp
+++ b/lib/validator/component_name.cpp
@@ -141,7 +141,47 @@ bool isValidSemVer(std::string_view Version) {
     return false;
   }
 
-  return true;
+  if (!readNumber()) {
+    return false;
+  }
+
+  if (Pos >= Version.size()) {
+    return true;
+  }
+
+  // Pre-release
+  if (Version[Pos] == '-') {
+    Pos++;
+    if (Pos >= Version.size()) {
+      return false;
+    }
+    while (Pos < Version.size()) {
+      if (!isalnum(Version[Pos]) && Version[Pos] != '-' && Version[Pos] != '.') {
+        break;
+      }
+      Pos++;
+    }
+  }
+
+  if (Pos >= Version.size()) {
+    return true;
+  }
+
+  // Build metadata
+  if (Version[Pos] == '+') {
+    Pos++;
+    if (Pos >= Version.size()) {
+      return false;
+    }
+    while (Pos < Version.size()) {
+      if (!isalnum(Version[Pos]) && Version[Pos] != '-' && Version[Pos] != '.') {
+        break;
+      }
+      Pos++;
+    }
+  }
+
+  return Pos == Version.size();
 }
 
 } // namespace ComponentNameParser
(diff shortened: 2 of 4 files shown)


lib/validator/component_name.cpp
```diff
@@ -105,8 +105,48 @@ bool tryReadKebab(std::string_view &Input, std::string_view &Output) {
   return isKebabString(Output);
 }
 
+bool isValidSemVer(std::string_view Version) {
```
PhantomInTheWire, line 108: The returns true after successfully parsing MAJOR.MINOR.PATCH without verifying that the entire version string has been consumed. There could be a suffix at the end that it doesn't consider. like 1.0.0-alpth, is this intentional?

lib/validator/component_name.cpp
```diff
@@ -105,8 +105,48 @@ bool tryReadKebab(std::string_view &Input, std::string_view &Output) {
...
+  if (!readNumber()) {
+    return false;
+  }
+
+  return true;
+}
+
 } // namespace ComponentNameParser
 
+
```
PhantomInTheWire, line 189: please avoid such whitespace changes

include/validator/component_name.h
```diff
@@ -24,8 +24,11 @@ bool readUntil(std::string_view &input, char delim, std::string_view &output);
 bool tryRead(std::string_view prefix, std::string_view &name);
 bool tryReadKebab(std::string_view &input, std::string_view &output);
 
+bool isValidSemVer(std::string_view version);
+
 }; // namespace ComponentNameParser
 
+
```
PhantomInTheWire: ditto

lib/validator/component_validator.cpp
```diff
@@ -762,6 +762,21 @@ Expect<void> Validator::validate(const AST::Component::Import &Im) noexcept {
...
+    if (!InterfaceDetails.Version.empty()) {
+      if (!ComponentNameParser::isValidSemVer(InterfaceDetails.Version)) {
+        spdlog::error(ErrCode::Value::ComponentInvalidVersion);
+        spdlog::error("    Import: Invalid semver format in version '{}'"sv,
+                      InterfaceDetails.Version);
+        spdlog::error(ErrInfo::InfoAST(ASTNodeAttr::Comp_Import));
+        return Unexpect(ErrCode::Value::ComponentInvalidVersion);
+      }
+    }
+  }
```
PhantomInTheWire: The logic for validating the version suffix is identical in both Validator::validate(const AST::Component::Import &Im) and Validator::validate(const AST::Component::Export &Ex).
can you please extract it into a private helper method in the Validator class.

lib/validator/component_name.cpp
sagar-h007: @PhantomInTheWire Thanks for pointing that out.

That wasn’t intentional. The validator should ensure the entire version string is consumed, including any optional prerelease or build metadata. I’ll update the implementation to verify that parsing consumes the full input and properly rejects trailing invalid suffixes like 1.0.0-alpth.

I’ll push a fix shortly.

lib/validator/component_validator.cpp
sagar-h007: Yes, it’s duplicated. I kept it inline initially to make the validation flow explicit while iterating, but I agree it shouldn’t stay that way.

I’ll extract the shared logic into a private helper on the validator and update both import and export paths to use it.

I’ll push the refactor shortly.

q82419:
@sagar-h007 

Because of re-constructing the implementation status of component-model proposal, close this.
Please refer to the main issue #4236 and validator issue #4268 .
Your can re-summarize your code and create anothor PR with following the guide in #4268.
Thanks!

PR #4615, opened 2026-02-04
fix(compiler): enforce 32-bit integer wrapping for memory offsets

### LINKED ISSUE #3063: bug: Different output when running a same binary
### Summary

I used the AOT mode of different runtimes to run randomly generated wasm binaries, and the output from WASMEDGE was different from the others. 
[filea29176_1.zip](https://github.com/WasmEdge/WasmEdge/files/13486786/filea29176_1.zip)

### Current State

```bash
root@4252f5ec38df:/a# wasmedge_0.13.5 compile  filea29176_1.wasm filea29176_1.so ; wasmedge_0.13.5 --reactor filea29176_1.so main
[2023-11-28 02:27:04.278] [info] compile start
[2023-11-28 02:27:04.281] [info] verify start
[2023-11-28 02:27:04.282] [info] optimize start
[2023-11-28 02:27:04.313] [info] codegen start
[2023-11-28 02:27:04.345] [info] output start
[2023-11-28 02:27:04.350] [info] compile done
[2023-11-28 02:27:04.368] [error] execution failed: out of bounds memory access, Code: 0x88
[2023-11-28 02:27:04.368] [error]     When executing function name: "main"
```

### Expected State

```bash
-----------------wasmtime-----------------
warning: using `--invoke` with a function that returns values is experimental and may break in the future
340282366920938463444927863362353627135
340282366920938463444927863362353627135
340282366920938463444927863362353627135
623261522
340282366920938463444927863362353627135

-----------------wasmer-----------------
Compiler: cranelift
Target: x86_64-unknown-linux-gnu
✔ File compiled successfully to `/a/filea29176_1.wasmu`.
340282366920938463444927863362353627135 340282366920938463444927863362353627135 340282366920938463444927863362353627135 623261522 340282366920938463444927863362353627135

-----------------wamr-----------------
Create AoT compiler with:
  target:        x86_64
  target cpu:    znver3
  target triple: x86_64-unknown-linux-gnu
  cpu features:  
  opt level:     3
  size level:    3
  output format: AoT file
Compile success, file /a/filea29176_1.wasm.aot was generated.
<0x00000000ffffffff 0xffffffffffffffff>:v128,<0x00000000ffffffff 0xffffffffffffffff>:v128,<0x00000000ffffffff 0xffffffffffffffff>:v128,0x25263752:i32,<0x00000000ffffffff 0xffffffffffffffff>:v128

```

### Reproduction steps

```bash
wasmedge_0.13.5 compile  filea29176_1.wasm filea29176_1.so ; wasmedge_0.13.5 --reactor filea29176_1.so main
```

```python
f"{wasmtime} compile {filename} -o {filename_cwasm}\
            ; {wasmtime} --allow-precompiled --invoke main {filename_cwasm}"
f"{wasmer} compile {filename} -o {filename}u\
            ; {wasmer} run -i main  {filename}u"
f"{wamrc} --bounds-checks=1 -o {filename}.aot {filename}\
            ; {iwasm} --heap-size=0 -f main {filename}.aot"
```

### Screenshots

_No response_

_No response_

### Components

CLI

wasmedge version 0.13.5

Docker version 24.0.5, build ced0996

### Hardware Architecture

Ubuntu 22.04.3 LTS (GNU/Linux 5.15.0-79-generic x86_64)

_No response_

blackdragoon26:
## Summary
This PR fixes a logic error in `compileLoadOp` and `compileStoreOp` where 32-bit address offsets were being calculated in 64-bit space. This caused valid Wasm32 wrap-around logic (e.g., `0xFFFFFFFF + 1`) to overflow into the Guard Page instead of wrapping to `0`, resulting in false segmentation faults/traps on valid Wasm code.

## The Bug
Previously, the compiler logic was:
1. Zero-extend the 32-bit base address to 64-bit.
2. Add the offset in 64-bit space.
3. **Result:** `(uint64)0xFFFFFFFF + 1` = `4GB` (Hits Guard Page -> Crash/Trap).

## The Fix
The logic has been reordered to enforce Wasm32 semantics:
1. Add the offset to the base address in **32-bit space** (allowing natural overflow/wrapping).
2. Zero-extend the wrapped result to 64-bit.
3. **Result:** `(uint32)0xFFFFFFFF + 1` = `0` -> extends to `0` (Valid Memory Access).

## Verification
I have verified this fix using the official `ubuntu:22.04` environment (emulating x86_64).
* **Before:** The reported test case crashed with `Segmentation fault`.
* **After:** The test case executes successfully and produces the expected output.

Fixes #3063

### DIFF
--- a/docs/ROADMAP.md
+++ b/docs/ROADMAP.md
@@ -29,22 +29,18 @@ WasmEdge roadmap will be updated quarterly with the following steps.
 
 ## Current Roadmap
 
-Last Updated: Q4 / 2025
+Last Updated: Q1 / 2026
 
 | Theme    | Description | Timeline | Assignee |
 | ---      | ----------- | -------- | -------- |
-| Feature | WasmEdge 0.16.0 release | Q4 / 2025 | [@hydai](https://github.com/hydai) |
-| Proposal | [WASM 3.0 supporting](https://github.com/WasmEdge/WasmEdge/issues/4382) | Q4 / 2025 | [@q82419](https://github.com/q82419) |
-| Proposal | [WASM memory64 proposal](https://github.com/WasmEdge/WasmEdge/pull/2964) | Q4 / 2025 | [@q82419](https://github.com/q82419) |
-| Proposal | [Component Model completion](https://github.com/WasmEdge/WasmEdge/issues/4236) | Q4 / 2025 | [@q82419](https://github.com/q82419) |
-| Proposal | [Component Model value and value type refactoring](https://github.com/WasmEdge/WasmEdge/issues/4333) | Q4 / 2025 | [@q82419](https://github.com/q82419) |
-| Proposal | [Component Model canonical section refactoring](https://github.com/WasmEdge/WasmEdge/issues/4334) | Q4 / 2025 | [@q82419](https://github.com/q82419) |
-| Proposal | [WASI preview2 as plugins](https://github.com/WasmEdge/WasmEdge/issues/4236) | Q4 / 2025 | [@q82419](https://github.com/q82419) |
-| Proposal | [Runtime GC support in WasmEdge](https://github.com/WasmEdge/WasmEdge/pull/4156) | Q4 / 2025 | [@ibmibmibm](https://github.com/ibmibmibm) |
-| Proposal | Update the Android NDK build examples | Q4 / 2025 | [@dm4](https://github.com/dm4) |
-| Proposal | WASI-NN GGML plugin with latest llama.cpp integration for Q4/2025 | Q4 / 2025 | [@dm4](https://github.com/dm4) |
+
 | Document | WasmEdge documentation refactoring | Q1 / 2026 | [@q82419](https://github.com/q82419) |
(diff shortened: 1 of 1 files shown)


blackdragoon26:
@hydai kindly review it

q82419:
Close this due to the following reasons:

1. This will conflict with the memory64 proposal, which will be merged soon.
2. The offset SHOULD be overflow from 32-bit if it's out-of-bound to trigger the OOB trap.

blackdragoon26:
> Close this due to the following reasons:
> 
> 1. This will conflict with the memory64 proposal, which will be merged soon.
> 2. The offset SHOULD be overflow from 32-bit if it's out-of-bound to trigger the OOB trap.

Thanks for the context regarding the Memory64 proposal. @q82419 
I agree that we cannot simply force 32-bit arithmetic globally if it breaks future 64-bit memory features.
However, for standard Wasm32, the current behavior is technically incorrect according to the spec. 0xFFFFFFFF + 1 is required to wrap to 0 (modulo $2^{32}$), but WasmEdge currently promotes it to 64-bit (4GB), causing a false OOB trap on valid code.

Instead of a global change, we should apply this wrapping logic conditionally.
1. If the module uses Wasm32 memory: Enforce 32-bit wrapping (fixing this bug).
2. If the module uses Memory64: Allow the 64-bit overflow behaviour you described.

Does this align with your vision for the Memory64 implementation? If so, I can create the PR to include this check.

PR #4613, opened 2026-02-04
perf(jit): optimize call_indirect with fast-path type check

Harsh-6291:
## What does this PR do?
* This PR resolves a significant performance regression in the JIT engine where call_indirect operations were order(s) of magnitude slower than expected.
## What are the changes?
* Modified File: lib/executor/engine/proxy.cpp
### Conclusion
* The performance overhead of call_indirect has been completely eliminated. Indirect calls now have identical performance characteristics to direct calls for intra-module execution.
#### Closing Issue
* ~~Fixes # 4369~~

PR #4621, opened 2026-02-05
perf: optimise call_indirect insn

### LINKED ISSUE #4369: bug: The call_indirect instruction causes a performance decline in WasmEdge JIT mode
### Summary

Hello, I noticed that the `wasmedge JIT` mode shows a significant performance drop when executing the `call_indirect` instruction, while the JIT modes of other runtime tools perform well.

### Current State

The specific execution time data is as follows:

|                  	|      time     	|
|------------------	|:-------------:	|
| wasmer_llvm      	|    4.667317   	|
| **wasmedge_jit** 	| **88.334651** 	|
| wamr_llvm_jit    	|    8.001169   	|
| wasmer_cranelift 	|    5.713164   	|
| wasmtime         	|    5.872757   	|
| wamr_fast_jit    	|   12.705185   	|

The data is in seconds, and each data is the result of ten executions and averages.
I placed the `call_indirect` instruction in a loop to amplify the performance difference.

### Expected State

Perhaps `wasmedge` has differences in the implementation or optimization of the `call_indirect` instruction compared to other runtime tools, leading to this performance anomaly.

### Reproduction steps

The minimal reproducible code is as follows:

<details>
<summary> test_case.wat </summary>

```
(module
  (type (;0;) (func (param i32)))
  (type (;1;) (func))
  (type (;2;) (func (result i32)))
  (import "wasi_snapshot_preview1" "proc_exit" (func (;0;) (type 0)))
  (func (;1;) (type 1))
  (func (;2;) (type 2) (result i32)
    (i32.const 0))
  (func (;3;) (type 0) (param i32))
  (func (;4;) (type 1))
  (func (;5;) (type 0) (param i32))
  (func (;6;) (type 1)
    (local i32 i32 i32 i32)
    (local.set 3
      (i32.const 0))
    (loop

      (call_indirect (type 1)
        (i32.const 7))
        
      (local.set 3
        (i32.add
          (local.get 3)
          (i32.const 1)))
      (br_if 0
        (i32.ne
          (local.get 3)
          (i32.const 0))))
    (call 0
      (i32.const 0))
    (unreachable))
  (func (;7;) (type 0) (param i32)
    (global.set 0
      (local.get 0)))
  (func (;8;) (type 2) (result i32)
    (global.get 0))
  (table (;0;) 14 14 funcref)
  (memory (;0;) 258 43198)
  (global (;0;) (mut i32) (i32.const 66688))
  (export "memory" (memory 0))
  (export "__indirect_function_table" (table 0))
  (export "_start" (func 6))
  (export "_emscripten_stack_restore" (func 7))
  (export "emscripten_stack_get_current" (func 8))
  (elem (;0;) (i32.const 1) func 2 2 2 2 2 2 4 4 4 4 4 4 1)
  (data (;0;) (i32.const 1024) "\01\00\00\00\02\00\00\00\03\00\00\00\04\00\00\00\05\00\00\00\06")
  (data (;1;) (i32.const 1056) "\07\00\00\00\08\00\00\00\09\00\00\00\0a\00\00\00\0b\00\00\00\0c"))
```

</details>

```
wat2wasm test_case.wat -o test_case.wasm

# Execute the wasm file and collect data
perf stat -r 10 -e 'task-clock' /path/to/wasmer run test_case.wasm --llvm
perf stat -r 10 -e 'task-clock' /path/to/wasmedge --enable-jit test_case.wasm
perf stat -r 10 -e 'task-clock' /path/to/build_llvm_jit/iwasm test_case.wasm
perf stat -r 10 -e 'task-clock' /path/to/wasmer run test_case.wasm
perf stat -r 10 -e 'task-clock' /path/to/wasmtime test_case.wasm
perf stat -r 10 -e 'task-clock' /path/to/build_fast_jit/iwasm test_case.wasm
```

**Environment**

The runtime tools are all built on release and use JIT mode.

- wasmer: 6.0.1
- wasmtime: 35.0.0 (9c2e6f17c 2025-06-17)
- wasmedge: 0.15.0
- WAMR: iwasm 2.4.0
- wabt: 1.0.27

_No response_

### Components

CLI

0.15.0

Ubuntu 22.04.5 LTS 

### Hardware Architecture

x86_64

- g++: 11.4.0
- cmake: 3.22.1
- cmake flags: `-DCMAKE_BUILD_TYPE=Release -DWASMEDGE_BUILD_TESTS=ON`

sleepingeight:
Fixes #4369.

Results on running 
` $ sudo perf stat -r 10 -e 'task-clock' /path/to/wasmedge --enable-jit ~/wasmedge/test_case.wasm` 

on Ubuntu 24.04, Intel(R) Core(TM) i9-14900HX, x86_64, 32 logical CPU(s).

Old code - 
```
 Performance counter stats for '/home/surya/wasmedge/build/tools/wasmedge/wasmedge --enable-jit /home/surya/wasmedge/test_case.wasm' (10 runs):

         90,054.12 msec task-clock                       #    1.000 CPUs utilized               ( +-  0.01% )

          90.07134 +- 0.00860 seconds time elapsed  ( +-  0.01% )
```

New code -
```
 Performance counter stats for '/home/surya/wasmedge/build/tools/wasmedge/wasmedge --enable-jit /home/surya/wasmedge/test_case.wasm' (10 runs):

         29,796.19 msec task-clock                       #    1.000 CPUs utilized               ( +-  0.04% )

           29.8036 +- 0.0121 seconds time elapsed  ( +-  0.04% )
```

The optimisation was based on the following flamegraph for the old code:
trace_old
Most of the time was being spent while locking and unlocking the shared lock. The new flamegraph after updating the code to use the unsafe methods is:
trace_new

In my analysis, I found that removing locks is safe and does not cause any race condition because during runtime the shared variable being protected is immutable and is mutable only during the module instantiation phase but not during the execution phase. The lock seemed to protect against the rare case where a host application might try to dynamically add new functions/types to a module while it is running. And my optimisation was based on considering this as a rare case for webassembly. But I'd love to hear your opinion on this. 

_As a side note, if my above reasoning is correct we can make sure that all the intrinsics use the unsafe methods to improve performance._

Also there are other optimisation which can be performed from the second flamegraph, one of which is implemented in https://github.com/WasmEdge/WasmEdge/pull/4613, however these do not improve performance greatly since 66% of performance is being lost at the locks.

### DIFF
--- a/include/runtime/instance/module.h
+++ b/include/runtime/instance/module.h
@@ -74,7 +74,11 @@ class ModuleInstance {
 public:
   ModuleInstance(std::string_view Name, void *Data = nullptr,
                  std::function<void(void *)> Finalizer = nullptr)
-      : ModName(Name), HostData(Data), HostDataFinalizer(Finalizer) {}
+      : ModName(Name), HostData(Data), HostDataFinalizer(Finalizer) {
+    auto InitialTypes = std::make_unique<std::vector<const AST::SubType *>>();
+    AtomicTypes.store(InitialTypes.get(), std::memory_order_relaxed);
+    TypeSnapshots.push_back(std::move(InitialTypes));
+  }
   virtual ~ModuleInstance() noexcept {
     // When destroying this module instance, call the callbacks to unlink to the
     // store managers.
@@ -126,14 +130,21 @@ public:
     unsafeAddHostInstance(
         Name, OwnedFuncInsts, FuncInsts, ExpFuncs,
         std::make_unique<FunctionInstance>(
-            this, static_cast<uint32_t>(Types.size()) - 1, std::move(Func)));
+            this,
+            static_cast<uint32_t>(
+                AtomicTypes.load(std::memory_order_relaxed)->size()) -
+                1,
+            std::move(Func)));
   }
   void addHostFunc(std::string_view Name,
                    std::unique_ptr<FunctionInstance> &&Func) {
     std::unique_lock Lock(Mutex);
     assuming(Func->isHostFunction());
     unsafeImportDefinedType(Func->getHostFunc().getDefinedType());
-    Func->linkDefinedType(this, static_cast<uint32_t>(Types.size()) - 1);
+    Func->linkDefinedType(
+        this, static_cast<uint32_t>(
+                  AtomicTypes.load(std::memory_order_relaxed)->size()) -
+                  1);
     unsafeAddHostInstance(Name, OwnedFuncInsts, FuncInsts, ExpFuncs,
                           std::move(Func));
   }
@@ -237,7 +248,13 @@ protected:
   void addDefinedType(const AST::SubType &SType) {
     std::unique_lock Lock(Mutex);
     OwnedTypes.push_back(std::make_unique<AST::SubType>(SType));
-    Types.push_back(OwnedTypes.back().get());
+
+    auto *CurrentTypes = AtomicTypes.load(std::memory_order_relaxed);
+    auto NewTypes =
+        std::make_unique<std::vector<const AST::SubType *>>(*CurrentTypes);
+    NewTypes->push_back(OwnedTypes.back().get());
+    AtomicTypes.store(NewTypes.get(), std::memory_order_release);
+    TypeSnapshots.push_back(std::move(NewTypes));
   }
 
   /// Create and add instances into this module instance.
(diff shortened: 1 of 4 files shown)


q82419:
There may be a **data race issue** with removing the lock:

**The Problem:**
While your reasoning about type immutability during normal execution is correct, `std::vector::operator[]` is **not thread-safe** when concurrent modifications are possible. Even though:
- Pointers in `Types[Idx]` remain stable
- Elements are never deleted

The vector container itself can be in an inconsistent state during `push_back()`/resize() operations:
- Internal data pointer may be temporarily invalid during reallocation
- Size field may be updated non-atomically
- Other internal metadata (capacity, etc.) can be in intermediate states

**Specifically:** If a host application calls `addHostFunc()` during execution (which modifies the `Types` vector), a concurrent `unsafeGetType()` call can:
1. Read a stale/invalid internal pointer
2. Access memory that's being deallocated during vector resize
3. Read inconsistent size vs. data pointer state
4. Result in undefined behavior (crashes, corrupted data, etc.)

However, to remove the lock here, I think we can discuss for the other approches, such as the policy of forbidding the insertion of new host function in execution, or the `std::atomic` for the management of `Types` vector.

sleepingeight:
Hi @q82419, thanks for the feedback. From my understanding, we have a possibility of race condition when the `Types` vector is modified by some other procedure. So, I've modified the code to use atomics implementing Read-Copy-Update, along with snapshots.

This has resulted in the following performance - 
```
 Performance counter stats for '/path/to/wasmedge --enable-jit /home/surya/wasmedge/test_case.wasm' (10 runs):

         35,462.70 msec task-clock                       #    1.000 CPUs utilized               ( +-  0.02% )

          35.47496 +- 0.00876 seconds time elapsed  ( +-  0.02% )
```
trace_new2

This approach is also performing ~3x better than the original implementation.

This approach as a memory overhead due to storing snapshots. But i don't think this matters a lot in our case. Along with it I've also added a test verifying that there exists no race condtion in the above situation.

sleepingeight:
Hi @q82419 , I'm taking a look at the CI failures. Could you please provide any other feedback on the PR.

q82419:
@sleepingeight 
Please refer to the #4633, Thanks!

q82419:
Replaced by #5052

PR #4624, opened 2026-02-07
Build completed and logs added

Daiwik-1901:
This PR fixes the build issues and ensures the project builds successfully.
The setup was verified locally using CMake and Ninja.
No functional changes were introduced.


Daiwik-1901:
@hydai Please review this and suggest changes/correction if needed.

PhantomInTheWire (commented):
I don't understand what you're trying to fix here. please explain the issue this is trying to fix. from what I understand you probably encountered some build errors during setup and have now submitted the logs of the build failure and some "fixes" that look very confusing without more context. please explain yourself a little more.

q82419:
Your PR mixed lots of things and we cannot understand your goal for fixing.
Please split your codes and describe the reason for each entries.

Daiwik-1901:
I ran into build failures during local setup and worked through them to get a successful build. While debugging, I mixed logs and setup-related changes in one PR, which caused confusion. My intent was only to fix local build issues, and I’ll clean this up into smaller, clearer changes.

PR #4625, opened 2026-02-07
feat: support parameterized error messages

### LINKED ISSUE #4444: feat: Component Model: non-static string for ErrorCode and spectest

Sujanian1304:
Implements support for dynamic error strings in ErrCode (e.g., "type index 5 is not...") to fix spec test failures, while maintaining backward compatibility with existing static constexpr strings.

## Key Changes

  -  Extended ErrCode: Added optional std::string member and getErrString() accessor to return dynamic messages when present.

   - New Helper: Added UnexpectMsg for one-line construction of formatted errors.

   - Test Infrastructure: Updated spectest.cpp trap helpers to verify dynamic error strings.

   - Validator: Migrated 5 priority call sites (Import, Function, Start, Tag sections) to use UnexpectMsg.
   

   * closes #4444 *

### DIFF
--- a/include/common/enum_errcode.hpp
+++ b/include/common/enum_errcode.hpp
@@ -19,6 +19,7 @@
 #include "common/spdlog.h"
 
 #include <cstdint>
+#include <string>
 #include <string_view>
 
 namespace WasmEdge {
@@ -81,13 +82,26 @@ public:
                : static_cast<WasmPhase>((getCode() >> 8) & 0x0FU);
   }
 
-  constexpr ErrCode() noexcept : Inner(0) {}
-  constexpr ErrCode(const ErrCode &E) noexcept : Inner(E.Inner.Num) {}
-  constexpr ErrCode(const ErrCode::Value E) noexcept : Inner(E) {}
-  constexpr ErrCode(const uint32_t N) noexcept
+  const std::string &getMessage() const noexcept { return Message; }
+
+  bool hasMessage() const noexcept { return !Message.empty(); }
+
+  void setMessage(std::string Msg) noexcept { Message = std::move(Msg); }
+
+  /// Note: returns string_view; lifetime is either Message or static storage.
+  std::string_view getErrString() const noexcept;
+
+  ErrCode() noexcept : Inner(0) {}
+  ErrCode(const ErrCode &E) noexcept : Inner(E.Inner.Num), Message(E.Message) {}
+  ErrCode(ErrCode &&E) noexcept
+      : Inner(E.Inner.Num), Message(std::move(E.Message)) {}
+  ErrCode(const ErrCode::Value E) noexcept : Inner(E) {}
+  ErrCode(const ErrCode::Value E, std::string Msg) noexcept
+      : Inner(E), Message(std::move(Msg)) {}
+  ErrCode(const uint32_t N) noexcept
       : Inner((static_cast<uint32_t>(ErrCategory::UserLevelError) << 24) +
               (N & 0x00FFFFFFU)) {}
-  constexpr ErrCode(const ErrCategory C, const uint32_t N) noexcept
+  ErrCode(const ErrCategory C, const uint32_t N) noexcept
       : Inner((static_cast<uint32_t>(C) << 24) + (N & 0x00FFFFFFU)) {}
 
   friend constexpr bool operator==(const ErrCode &LHS,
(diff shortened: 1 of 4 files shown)


q82419:
Please make sure your PR satisfy the requirements in the [contribution guide](https://wasmedge.org/docs/contribute/), or the review won't be started.

Sujanian1304:
@q82419 ,I have updated the PR  .Ready for review now

q82419:
> @q82419 ,I have updated the PR .Ready for review now

No, you didn't fix the commit lint issue.

Sujanian1304:
@q82419 ,fixed the commit lint as well .

Sujanian1304:
Hi @q82419 ,whenever you have some free time, could you please take a look at this PR? I’d really appreciate your review. Thank you!

q82419:
> Hi @q82419 ,whenever you have some free time, could you please take a look at this PR? I’d really appreciate your review. Thank you!

Sorry, I'm considering if this change would affect the performance because the `ErrCode` class in `Expected<T, ErrCode>` may cause implicit copy in every returning. Originally the member of this class is only the `uint32_t` value, but now it contains the `std::string` object.
Please help to perf the time of execution (take the spec tests in interpreter for example) before and after applying this change to simply compare the performance, both in DEBUG and RELEASE build. Thanks!

Sujanian1304:
@q82419 
Screenshot 2026-02-12 164746
Screenshot 2026-02-12 170144
Screenshot 2026-02-12 173700
Screenshot 2026-02-12 174855

Sujanian1304:
Release Build (Optimized):

    Before: ~1.15s

    After: ~1.27s


Debug Build (Unoptimized):

    Before: ~16.97s

    After: ~19.72s

hydai:
So, the performance has a significant downgrade:

```
Release: -0.1043478261
Debug: -0.1620506777
```

@q82419 I think we should not introduce this kind of error message format into the ErrCode system. Maybe we should modify the test suites as an alternative solution.

q82419:
> So, the performance has a significant downgrade:
> 
> ```
> Release: -0.1043478261
> Debug: -0.1620506777
> ```
> 
> @q82419 I think we should not introduce this kind of error message format into the ErrCode system. Maybe we should modify the test suites as an alternative solution.

Agree with this.
I think the error messages in this case such as function name or index can be in the debug info, not in the error code system.

q82419:
In the WasmEdge spec test suite, we can remove the parameterized messages and list all modified tests for description.

cc @LFsWang

q82419:
@Sujanian1304 
Thanks for your perf and contribution.
We would close this and related issues as completed soon and record this in the component model parent issue.

q82419:
@Sujanian1304 

I would close the PR and issue about the parameterized error messages in spec tests.
We welcome you to add the comments [here](https://github.com/WasmEdge/WasmEdge/blob/master/include/common/enum_errcode.hpp#L56) to explain why we don't implement the parameterized error messages in `ErrCode`, and noticed to refer to this PR, for your contribution.
Thanks!

Sujanian1304:
@q82419 ,Thank you for the detailed feedback and benchmarks! I understand the performance
concern — adding `std::string` to ErrCode introduces heap allocations on every
`Expected<T, ErrCode>` return, which is clearly unacceptable for a hot path.

I've added a Doxygen comment above the `ErrCode `class in
`include/common/enum_errcode.hpp` documenting this design decision 
PR: #4666

Thanks again for the guidance!

PR #4626, opened 2026-02-08
fix: add conditional checks for empty matrix in reusable workflows

Vikaspal8923:


.github/workflows/reusable-build-extensions.yml
```diff
@@ -163,6 +164,7 @@ jobs:
 
   build_on_macos:
     needs: prepare
+    if: ${{ needs.prepare.outputs.macos_x86_64 != '[]' || needs.prepare.outputs.macos_arm64 != '[]' }}
```
0yi0, line 167: This might not work for `wasi_nn-mlx` since it's only built on `macos_arm64`.

Vikaspal8923:
@0yi0 thanks for the clarification. I see the issue now, using or at the job level still allows matrix entries with an empty plugin list.

What about building the matrix explicitly in prepare and including only platform variants with non-empty plugin lists,
and then skipping the job when the matrix is empty?

0yi0:
> What about building the matrix explicitly in prepare and including only platform variants with non-empty plugin lists, and then skipping the job when the matrix is empty?

I commented here because it was specifically about the implementation in this PR.

Please keep related discussions in the issue. You may either raise your ideas there, or make another PR to the issue directly.

I'll lock conversation here. Thank you for your contributions again!

PR #4632, opened 2026-02-09
Fix: Strict validation for memory instruction flags

### LINKED ISSUE #4631: bug: Loader fails to reject reserved bits in memory instruction flags (memarg)
### Description
The current implementation of `readMemImmediate` in the Loader is too lenient with memory instruction flags. It does not strictly validate reserved bits as required by the WebAssembly specification.

Currently, if a memory instruction (like `i32.load`) has reserved bits set (e.g., Bit 7 `0x80`), the Loader ignores them. This leads to one of two incorrect behaviors:
1. The **Validator** catches it later as an Alignment Error (`0x201`) or Type Mismatch (`0x202`), instead of the Loader catching it as a Malformed Error (`0x1xx`).
2. If the bits form a valid alignment value but are structurally illegal (e.g., `0x80`), the parser may desynchronize.

### Current Behavior
The Loader accepts the flags. The execution eventually fails in the Validator or Executor with an unrelated error code:

### Expected State

According to the WebAssembly Spec, any reserved bit set in `memarg` must result in a **Malformed** error during the Loading phase.

### Reproduction Steps
1. Create a WASM binary with a manually constructed `i32.load` instruction.
2. Set the flags byte to `0x80 0x01` (LEB128 for 128), which has the reserved Bit 7 set. ( eg. 

[final_check.wasm.txt](https://github.com/user-attachments/files/25190014/final_check.wasm.txt)

3. Run WasmEdge: `wasmedge reserved.wasm`

### Screenshots

Image

_No response_

### Components

CLI

0.16.1

windows

### Hardware Architecture

amd64

x64

souravk29:
### Summary
This PR implements strict validation for memory argument (memarg) flags in the Loader, ensuring compliance with the WebAssembly specification.

### Fixes
Fixes #4631

### Details
Modified lib/loader/ast/instruction.cpp:

Added Reserved Bit Check: The loader now validates that Flags & ~0x7F is zero. Any set bit above Bit 6 triggers ErrCode::Value::MalformedMemoryOpFlags (0x11c).

Conditional MultiMemory Check: Bit 6 (0x40) is now strictly checked against the MultiMemories proposal configuration.

### Verification
I verified this fix using a custom-generated WASM binary with an illegal reserved bit.

Screenshot 2026-02-10 193345

 Reserved Bit Test (0x80)

### Before Fix:
 Passed Loader, failed in Validator (0x201).

### After Fix: 
Fails in Loader (0x11c MalformedMemoryOpFlags).

### Reproduction Script (Python): Run this to generate the test file:

```bash
# Creates 'fail_reserved.wasm' with i32.load using flags 0x80 (1000 0000)
# 0x80 is a reserved bit and MUST fail validation in the Loader.
python -c "open('fail_reserved.wasm', 'wb').write(bytearray([0x00, 0x61, 0x73, 0x6D, 0x01, 0x00, 0x00, 0x00, 0x01, 0x05, 0x01, 0x60, 0x00, 0x01, 0x7F, 0x03, 0x02, 0x01, 0x00, 0x05, 0x03, 0x01, 0x00, 0x01, 0x07, 0x05, 0x01, 0x01, 0x6D, 0x00, 0x00, 0x0A, 0x0A, 0x01, 0x08, 0x00, 0x41, 0x00, 0x28, 0x80, 0x01, 0x00, 0x0B]))"
```

### DIFF
--- a/lib/loader/ast/instruction.cpp
+++ b/lib/loader/ast/instruction.cpp
@@ -253,17 +253,31 @@ Expect<void> Loader::loadInstruction(AST::Instruction &Instr) {
 
   auto readMemImmediate = [this, readU32, readU64, &Instr]() -> Expect<void> {
     Instr.getTargetIndex() = 0;
-    EXPECTED_TRY(readU32(Instr.getMemoryAlign()));
-    if (Conf.hasProposal(Proposal::MultiMemories) &&
-        Instr.getMemoryAlign() >= 64) {
-      Instr.getMemoryAlign() -= 64;
-      EXPECTED_TRY(readU32(Instr.getTargetIndex()));
+    uint32_t Flags = 0;
+    EXPECTED_TRY(readU32(Flags));
+
+    static constexpr uint32_t kWasmMemFlagsAllowedMask = 0x7FU;
+    if (unlikely(Flags & ~kWasmMemFlagsAllowedMask)) {
+      return logLoadError(ErrCode::Value::MalformedMemoryOpFlags,
+                          FMgr.getLastOffset(), ASTNodeAttr::Instruction);
     }
+
+    Instr.getMemoryAlign() = Flags & 0x3FU;
+
+    bool HasMultiMemories = Conf.hasProposal(Proposal::MultiMemories);
+    if (HasMultiMemories && (Flags & 0x40U)) {
+      EXPECTED_TRY(readU32(Instr.getTargetIndex()));
+    } else if (!HasMultiMemories && (Flags & 0x40U)) {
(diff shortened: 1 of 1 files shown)


souravk29:
@q82419 please do review

q82419:
> @q82419 please do review

Before reviewing this, you should provide the detailed description in the issue.

souravk29:
> > @q82419 please do review
> 
> Before reviewing this, you should provide the detailed description in the issue.

i have corrected the description and changed the test

q82419:
Close this because this is not the situation described by the author.
Refer to the issue #4631

PR #4640, opened 2026-02-10
ci: fix BitNet FetchContent checkout failure

vishal2005025:
### Summary

This PR Fixes CI failure in the WASI-NN BitNet backend by updating the pinned BitNet.cpp commit. The previously used commit (`404980e`) no longer exists upstream, causing FetchContent checkout to fail. This change updates the `GIT_TAG` to a valid commit from the BitNet `main` branch, unblocking CI.

### Changes
- Updated BitNet.cpp `GIT_TAG` in `cmake/WASINNDeps.cmake` to a valid upstream commit

Reference: https://github.com/microsoft/BitNet/commits/main/

### DIFF
--- a/cmake/WASINNDeps.cmake
+++ b/cmake/WASINNDeps.cmake
@@ -639,7 +639,7 @@ function(wasmedge_setup_bitnet_target target)
     FetchContent_Declare(
       bitnet
       GIT_REPOSITORY https://github.com/microsoft/BitNet.git
-      GIT_TAG        404980e
+      GIT_TAG        8fd3412fbc9319f0764f82da6e9406c0323140f3
       GIT_SHALLOW    TRUE
       ${BITNET_PATCH_ARGS}
     )


vishal2005025:
@hydai @q82419 Please take a look
Thanks!

hydai:
Have you verified that the build is working after bumping the version?

vishal2005025:
@hydai I verified this locally by deleting the build directory, reconfiguring, and rebuilding, and the build completes successfully with the updated BitNet commit. 
I also checked the CI logs and the previous error:
"fatal: invalid reference: 404980e"
is no longer present
CI Jobs: https://github.com/vishal2005025/WasmEdge/pull/3/checks

Divyansh200102:
> @hydai I verified this locally by deleting the build directory, reconfiguring, and rebuilding, and the build completes successfully with the updated BitNet commit. I also checked the CI logs and the previous error: "fatal: invalid reference: 404980e" is no longer present CI Jobs: https://github.com/vishal2005025/WasmEdge/pull/3/checks

Hi, CI checks are still failing on ubuntu https://github.com/vishal2005025/WasmEdge/actions/runs/21863594641/job/63099345148?pr=3

vishal2005025:
@Divyansh200102 The original CI failure was caused by BitNet failing to fetch due to an invalid pinned commit. With this change, BitNet is fetched successfully and CI progresses further.
The current failure in `wasi_nn-bitnet-x86-tl2` is a compilation error inside BitNet upstream code (`ggml-bitnet-mad.cpp`) with Clang 18, which appears unrelated to the commit pin update.

Divyansh200102:
> @Divyansh200102 The original CI failure was caused by BitNet failing to fetch due to an invalid pinned commit. With this change, BitNet is fetched successfully and CI progresses further. The current failure in `wasi_nn-bitnet-x86-tl2` is a compilation error inside BitNet upstream code (`ggml-bitnet-mad.cpp`) with Clang 18, which appears unrelated to the commit pin update.

I know but currently the bitnet workflow  seems to be working fine after disabling the shadow clone so bumping the version for this specific problem doesn't make sense to me and it isn't solving the issue.
https://github.com/WasmEdge/WasmEdge/actions/runs/21870696747/job/63126374247?pr=4443

> 404980e) no longer exists upstream, causing FetchContent

also you said here that this commit doesn't exist but I do see it in the upstream https://github.com/microsoft/BitNet/commit/404980eecae38affa4871c3e419eae3f44536a95 . So this was likely a shadow clone problem and not something that needs a version bump.

image

vishal2005025:
@Divyansh200102 Thanks, I just checked and it looks like the BitNet workflow is now working after disabling the shallow clone and allowing a full history fetch.
In this PR, I tried an alternative approach by updating the pinned BitNet commit while keeping shallow clone enabled, since the previously pinned commit appeared unreachable in CI and I wanted to ensure we were using a valid and up-to-date upstream commit.
Given that the shallow-clone fix already resolves the issue, bumping the BitNet commit may not be necessary. If maintainers prefer to keep the existing pinned version, I’m happy to close this PR.

PR #4641, opened 2026-02-11
fix(executor): check Expect<> before dereference in runCallIndirectOp

sumleo:
## Summary

- `ModInst->getType()` in `runCallIndirectOp` (`controlInstr.cpp:196`) returns `Expect<const AST::SubType *>` which can be in error state when the type index is out of bounds
- The previous code dereferenced it unconditionally with `**`, causing undefined behavior
- Use `EXPECTED_TRY` to propagate the error instead, matching the pattern already used elsewhere in the same function

## Test plan

- [x] Added `test/executor/executorNullDerefTest.cpp` with 3 tests:
  - Valid `call_indirect` executes correctly
  - Validator rejects module with out-of-bounds type index
  - Bypassing validation returns error gracefully instead of UB
- [x] All tests pass locally

### DIFF
--- a/lib/executor/engine/controlInstr.cpp
+++ b/lib/executor/engine/controlInstr.cpp
@@ -193,7 +193,9 @@ Expect<void> Executor::runCallIndirectOp(Runtime::StackManager &StackMgr,
 
   // Get function type at index x.
   const auto *ModInst = StackMgr.getModule();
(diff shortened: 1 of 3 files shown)


q82419:
The WASM execution here is validated and your mechanism is not necessary.

PR #4642, opened 2026-02-11
fix(executor): check Expect<> before dereference in proxyCallIndirect

sumleo:
## Summary

- `ModInst->getType()` in `proxyCallIndirect` (`proxy.cpp:167,603`) returns `Expect<const AST::SubType *>` which can be in error state when the type index is out of bounds
- The previous code dereferenced it unconditionally with `**`, causing undefined behavior
- Use `EXPECTED_TRY` to propagate the error instead, matching the pattern already used elsewhere in the same file

## Test plan

- [x] Added `test/executor/executorNullDerefTest.cpp` verifying the interpreter `call_indirect` path still works (proxy functions are AOT-only and not directly testable without LLVM)
- [x] All tests pass locally


q82419:
The WASM execution here is validated and your mechanism is not necessary.

PR #4643, opened 2026-02-11
fix(executor): add explicit assuming for getType in toBottomType

sumleo:
## Summary

- `StackMgr.getModule()->getType()` in `toBottomType` (`helper.cpp:425`) returns `Expect<>` which was dereferenced without checking, causing silent undefined behavior on error
- Since `toBottomType` returns `TypeCode` (not `Expect<>`), it cannot propagate errors via `EXPECTED_TRY`
- Add an explicit `assuming()` to make the postcondition contract clear and provide a debug assertion instead of silent UB
- The loader already provides defense-in-depth by rejecting modules with invalid type references before this code path is reached

## Test plan

- [x] Added `test/executor/executorNullDerefTest.cpp` with 2 tests:
  - Valid `struct.new` with GC types works correctly
  - Loader rejects module with struct field referencing non-existent type
- [x] All tests pass locally

### DIFF
--- a/lib/executor/helper.cpp
+++ b/lib/executor/helper.cpp
@@ -421,9 +421,9 @@ TypeCode Executor::toBottomType(Runtime::StackManager &StackMgr,
         assumingUnreachable();
       }
     } else {
-      const auto &CompType =
-          (*StackMgr.getModule()->getType(Type.getTypeIndex()))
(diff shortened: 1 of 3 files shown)


q82419:
The WASM execution here is validated here and your mechanism is not necessary.

PR #4645, opened 2026-02-11
ci: optimize s390x workflow with apt cache and Ninja

vishal2005025:
### Summary

This PR optimizes the s390x CI workflow by caching APT package downloads and switching the build system from GNU Make to Ninja to reduce build overhead time.

### Changes

- Cache `/var/cache/apt` to reuse downloaded APT packages across runs
- Install `ninja-build` and use the Ninja generator in CMake (`-G Ninja`)
- Replace `make` / `make install` with `ninja` / `ninja install`

CI jobs: https://github.com/vishal2005025/WasmEdge/actions/runs/21919080922/job/63294198242?pr=4


vishal2005025:
@hydai @q82419 Could you please take a look and let me know if any changes are required
Thanks!

.github/workflows/build_for_s390x.yml
```diff
@@ -58,6 +58,13 @@ jobs:
     - name: Ensure git safe directory
       run: |
         git config --global --add safe.directory $(pwd)
+    - name: Cache s390x apt packages
+      uses: actions/cache@cdf6c1fa76f9f475f3d7449005a359c84ca0f306 # v5.0.3
+      with:
+        path: .s390x-apt-cache
+        key: s390x-apt-${{ runner.os }}-jammy-llvm18
```
hydai, line 65: What's the purpose of the `jammy`?

vishal2005025: @hydai The workflow runs the s390x build inside an ubuntu22.04 container, so the `jammy` segment in the cache key is just to scope the cache to that distro and avoid potential collisions if the distro changes in the future. If you prefer, I can also drop it and keep the key simpler.

PR #4649, opened 2026-02-14
fix(jit): trap compiled stack overflow before native crash (#4422)

### LINKED ISSUE #4422: bug: AddressSanitizer: SEGV on unknown address 0x000000000000
### Summary

Run the following command:

```bash
wasmedge run --enable-jit test.wasm main
```

Observed output:
```
/work/harnesses/wasmedge/bin/wasmedge run --enable-jit /work/Data/sample/wasmedge/test.wasm main
[2025-11-24 06:54:52.025] [info] compile start
[2025-11-24 06:54:52.026] [info] verify start
[2025-11-24 06:54:52.026] [info] optimize start
[2025-11-24 06:54:52.029] [info] optimize done
Segmentation fault (core dumped)
```
with asan:
```
AddressSanitizer:DEADLYSIGNAL
=================================================================
[1m[31m==2283740==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000 (pc 0x000000000000 bp 0x000000000000 sp 0x7fffffffdca8 T0)
[1m[0m==2283740==Hint: pc points to the zero page.
==2283740==The signal is caused by a READ memory access.
==2283740==Hint: address points to the zero page.
    #0 0x0  (<unknown module>)

AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV (<unknown module>) 
==2283740==ABORTING
```

### Current State

_No response_

### Expected State

_No response_

### Reproduction steps

wasmedge/bin/wasmedge run --enable-jit test.wasm main

[test.wasm.txt](https://github.com/user-attachments/files/23707155/test.wasm.txt)

### Screenshots

DESCRIPTION

_No response_

### Components

CLI

0.15.0

ubuntu20.04

### Hardware Architecture

amd64

_No response_

karimali03:
**Description:**
Fixes #4422

This PR fixes a segmentation fault (SEGV) where deep recursion in JIT/AOT compiled modules hits the OS guard page. It implements a soft stack-limit guard at the function entry, ensuring the runtime traps deterministically with `ErrCode::StackOverflow` instead of crashing the process.

**Changes:**

* `include/common/enum.inc`: Added `ErrCode::Value::StackOverflow` (`0x041A`).
* `lib/executor/helper.cpp`: Implemented `computeThreadStackLimit` to calculate the thread's safe boundary with a **64KB safety margin** (supports Linux, macOS, Windows).
* `lib/llvm/compiler.cpp`: Updated the JIT/AOT prologue to read `StackLimit` from `ExecutionContext`, compare it with `llvm.stacksave` (SP), and trap if exceeded.
* `test/llvm/LLVMcoreTest.cpp`: Added `JITStackOverflow` regression test to verify the fix.

**Verification:**
Verified locally that `wasmedge run --enable-jit` now correctly traps with `stack overflow (0x41a)` on the reproduction case.
Ran `lineguard` and `ctest` to ensure no regressions.


hydai:
Hi @karimali03 
Thanks for the PR. Since most of our maintainers are out of office due to the Lunar New Year holiday, your PR will be reviewed after we return.

karimali03:
Hi @ibmibmibm and @q82419, just checking in on this PR. Let me know if you've had a chance to review it or if there's anything I should update. Thanks!

include/executor/executor.h
```diff
@@ -1088,6 +1088,7 @@ class Executor {
     std::atomic_uint64_t *Gas;
     uint64_t GasLimit;
     std::atomic_uint32_t *StopToken;
+    uint8_t *StackLimit;
```
hydai: May I ask the reason for choosing `uint8_t` here?

lib/executor/helper.cpp
```diff
@@ -11,8 +11,77 @@
...
+#include <pthread.h>
+#elif WASMEDGE_OS_WINDOWS
+#include "system/winapi.h"
+#endif
+
 namespace WasmEdge {
 namespace Executor {
+namespace {
+
+constexpr size_t StackGuardSize = 64 * 1024;
```
hydai, line 24: I would like to know the reason for choosing `64*1024` as the `StackGuardSize`. It seems to be a fixed magic number here.

lib/executor/helper.cpp
```diff
@@ -11,8 +11,77 @@
...
+    if (pthread_attr_getstack(&Attr, &StackBase, &StackSize) == 0) {
+      if (pthread_attr_getguardsize(&Attr, &GuardSize) != 0) {
+        GuardSize = 0;
+      }
+      pthread_attr_destroy(&Attr);
+      auto *StackLow = static_cast<uint8_t *>(StackBase);
+      const size_t GuardMargin =
+          GuardSize > StackGuardSize ? GuardSize : StackGuardSize;
+      const size_t Margin =
+          StackSize < GuardMargin ? StackSize : GuardMargin;
```
hydai: I found that your margin calculation differs between Linux and Windows/MacOS. Could you please explain why?

lib/llvm/compiler.cpp
```diff
@@ -250,6 +250,8 @@ struct LLVM::Compiler::CompileContext {
                 Int64Ty,
                 // StopToken
                 Int32PtrTy,
+                // StackLimit
+                Int8PtrTy,
```
hydai: Since this will change the binary format, we should also update `kBinaryVersion`.

lib/executor/helper.cpp
```diff
@@ -11,8 +11,77 @@
...
 namespace WasmEdge {
 namespace Executor {
+namespace {
+
+constexpr size_t StackGuardSize = 64 * 1024;
+
+uint8_t *computeThreadStackLimit() noexcept {
+  uint8_t Marker;
+  auto *Fallback = const_cast<uint8_t *>(
+      reinterpret_cast<const uint8_t *>(&Marker) - StackGuardSize);
```
hydai, line 29: I think this fallback is sketchy. Could you provide an example to demonstrate that it is useful?

hydai (changes requested):

include/executor/executor.h
karimali03: I originally used ```uint8_t``` to simplify byte-level pointer arithmetic during the stack calculation. However, I will update it to ```uintptr_t```, as that is more appropriate for storing the address limit.

lib/executor/helper.cpp
karimali03: Exactly ```64KB``` was chosen as a safe buffer. Standard OS guard pages are typically much smaller (```4KB``` or ```16KB```). By reserving ```64KB```, we guarantee there is plenty of usable stack space left to safely execute trap logic and throw the C++ exception before the actual OS page boundary is hit.

lib/executor/helper.cpp
karimali03: I decided to keep GuardSize for safety. On Linux, it's not always guaranteed that ```pthread_attr_getstack``` excludes the guard page from the reported range. By using ```GuardSize + StackGuardSize```, we ensure the Wasm limit is always in writable memory. Windows and macOS already return usable boundaries, so they are unaffected.

lib/executor/helper.cpp
karimali03: The fallback is intended as a failsafe for restricted environments where OS-specific APIs (like ```pthread_getattr_np```) might fail or be blocked by security sandboxes. Without it, a failure in OS-level queries would result in a null/zero limit, leaving Wasm execution completely unprotected. While conservative—guaranteeing 64KB of headroom from the current frame—this 'best-effort' boundary is strictly safer than no protection.

karimali03:
Hi @hydai, I’ve updated the PR to address your feedback. Could you please take another look when you have a moment? Thanks!

lib/executor/helper.cpp
```diff
@@ -11,8 +11,71 @@
...
+#include "system/winapi.h"
+#endif
+
 namespace WasmEdge {
 namespace Executor {
+namespace {
+
+constexpr size_t StackGuardSize = 64 * 1024;
+
+uintptr_t computeThreadStackLimit() noexcept {
```
ibmibmibm, line 26: These platform-dependent functions should be moved to `lib/system` and a new header should be created for them.

ibmibmibm (changes requested):

q82419:
replace by #5040

PR #4656, opened 2026-02-16
fix: catch std::bad_alloc in WasmEdge_StringCreateByBuffer

### LINKED ISSUE #4651: bug: Missing Exception Handlers in C API Boundary
### Summary

The C API implementation uses C++ constructs that can throw exceptions, specifically `new` (`std::bad_alloc`) and `std::vector` allocations. Functions like`WasmEdge_StringCreateByBuffer` perform raw `new` operations without `try-catch` blocks. If a C++ exception propagates out of an `extern "C"` function, the behavior is undefined (often causing an immediate crash/abort). A C library must never let C++ exceptions escape to the C caller.

### Current State

Location: `lib/api/wasmedge.cpp:737`
```C++
// lib/api/wasmedge.cpp
WASMEDGE_CAPI_EXPORT WasmEdge_String
WasmEdge_StringCreateByBuffer(const char *Buf, const uint32_t Len) {
  if (Buf && Len) {
    char *Str = new char[Len]; // Throws std::bad_alloc on failure
    std::copy_n(Buf, Len, Str);
    return WasmEdge_String{/* Length */ Len, /* Buf */ Str};
  }
  return WasmEdge_String{/* Length */ 0, /* Buf */ nullptr};
}

```

### Expected State

Wrap C++ implementations in try-catch blocks and return appropriate error codes or handle failure gracefully.

```C++
try {
  char *Str = new char[Len];
  // ...
} catch (const std::bad_alloc&) {
  return WasmEdge_String{0, nullptr};
}
```

### Reproduction steps

1)Locate:`WasmEdge_StringCreateByBuffer`
2)Create a C++ test program that links against `libwasmedge`.
3)In the test, configure the memory limit (or mock `operator new`) to fail.
4)Call `WasmEdge_StringCreateByBuffer` with a large size.
5)Observe the program crashing/aborting instead of handling the error cleanly, or use a tool like `valgrind` or `ASan` to catch the unwinding across C boundary.

### Screenshots

DESCRIPTION

_No response_

### Components

C SDK

v0.16.1

macOS m4 pro

### Hardware Architecture

arm64

_No response_

eslamsamy226:
Description:
Wraps the raw `new char[Len]` allocation in `WasmEdge_StringCreateByBuffer` with a `try-catch` block. This prevents `std::bad_alloc` from escaping across the FFI boundary, which previously caused undefined behavior/abort.

Returns an empty `WasmEdge_String (Length 0, Buf nullptr)` on failure.

Related Issue

Fixes #4651 

Verification

I verified this locally using fault injection (manually throwing `std::bad_alloc` inside the function) and confirmed that the C API returns an empty `WasmEdge_String` instead of crashing.

Why no test case is included:

I attempted to add a unit test using `UINT32_MAX` to force an allocation failure. However, on high-RAM machines (for me it's Apple M4 with 16GB RAM), the 4GB allocation actually succeeds instead of throwing `std::bad_alloc`. This causes the subsequent `std::copy_n` to segfault due to buffer over-read, rather than testing the OOM handling logic.

Since reliable OOM testing is highly dependent on available hardware and OS limits, I have omitted the test case.

### DIFF
--- a/lib/api/wasmedge.cpp
+++ b/lib/api/wasmedge.cpp
@@ -734,9 +734,13 @@ WasmEdge_StringCreateByCString(const char *Str) {
 WASMEDGE_CAPI_EXPORT WasmEdge_String
 WasmEdge_StringCreateByBuffer(const char *Buf, const uint32_t Len) {
   if (Buf && Len) {
-    char *Str = new char[Len];
-    std::copy_n(Buf, Len, Str);
-    return WasmEdge_String{/* Length */ Len, /* Buf */ Str};
+    try {
+      char *Str = new char[Len];
+      std::copy_n(Buf, Len, Str);
+      return WasmEdge_String{/* Length */ Len, /* Buf */ Str};
+    } catch (const std::bad_alloc &) {
+      return WasmEdge_String{/* Length */ 0, /* Buf */ nullptr};
+    }
   }
   return WasmEdge_String{/* Length */ 0, /* Buf */ nullptr};
 }


mrhapile:
@eslamsamy226 The fix correctly handles std::bad_alloc in WasmEdge_StringCreateByBuffer.
However, similar allocation patterns exist in WasmEdge_BytesCreate, which currently does not guard against allocation failure.

Since this is part of the C API surface, it might be safer to apply the same try/catch pattern there to avoid any potential exception propagation across the C boundary.

Screenshot 2026-02-17 at 8 27 43 AM

q82419:
How do you think about the exception handler of bad alloc (not only in `WasmEdge_String`)? @hydai  @ibmibmibm

hydai:
IMO, `std::nothrow` is better. Ref: https://en.cppreference.com/w/cpp/memory/new/nothrow.html

hydai:
I changed my mind; let's unify all the `new` patterns with `try-catch` blocks. I think we only need a single PR for all the similar issues.

q82419:
> I changed my mind; let's unify all the `new` patterns with `try-catch` blocks. I think we only need a single PR for all the similar issues.

Agree with this.

hydai:
Replaced by #4703 

Thanks for your contribution.

Issue #4513, opened 2026-01-13
LFX mentorship (2026/term1): Extend sub-command of WasmEdge CLI tool


q82419:
### Project Title

Extend sub-command of WasmEdge CLI tool

### Description

WasmEdge command line tool currently supports only the sub-command about run and compile the WASM files. As a WebAssembly runtime tool, it's welcome to have more functions about verifying the WASM files. In this mentorship, we expect the mentee to plan and implement the functionality of WasmEdge command line tool about parsing, validating, and instantiating WASM files, which developers can do with WasmEdge C API.

### Expected Outcome

1. Extend the sub-commands for WasmEdge CLI tool contains not only `parse`, `validate`, or `instantiate`, etc.
2. Provide the test cases to verify your implementation.
3. Use WasmEdge C API to trigger the WasmEdge CLI tools for further test cases.

### Recommend skills

1. C/C++
2. WebAssembly
3. GitHub workflows

### Pre-tests

Announce after the starting of program application.

### Mentor(s)

YiYing He (@q82419 , yiying@secondstate.io )

### Apply Link

TBD

### Appendix

TBD

meldegwi:
Hi @q82419,

I find this idea very interesting and would love to be part of WasmEdge through the LFX mentorship with it.

I have started exploring the architecture to understand how the new subcommands would fit. I analyzed the patterns in `run` and `compile` and built a PoC for `validate`. It successfully validates WASM files by leveraging `configure`, `loader`, and `validator` functionality.

You can check on the PoC here: https://github.com/MohamedElDegwi/WasmEdge/tree/playground

Note: This is currently a rough prototype to test my grasp of the architecture and used design patterns.

Before I go further, does this implementation align with what you had in mind for the new subcommands? I'm happy to refine it, look into other documentation, or tackle related bugs while waiting for the pre-tests. :)

Thanks!

q82419:
For the fairness of the process, we won't review any individual proposals from candidates before the announce of mentorship on the LFX website.
And also, in order to prevent from cheating, we won't give the feedback for any candidates' proposals, neither under this issue, nor by e-mail.
We'll select the mentee by their quality of the resume, plans, and pre-tests.
The only thing I can advise is that please be sure that you have the qualification and complete the process of the LFX mentorship program before you submit your report of pre-test.

Vikaspal8923:
Hello @q82419,

[...]

PR #4657, opened 2026-02-17
test: add component model fuzz testing

LFsWang:
Some TODOs :
* 100 times testing is enough ? (~= 1s CI time)
* Should we add a version for `wasm-tools` to prevent breaking change from upstream ?
* If #4513 Finished, we can remove `fuzz_component.cpp` and use the wasmedge binary to test.

---
Add fuzz testing for the component model parse-to-validate pipeline. A new test binary (componentFuzzTest) accepts a wasm file path and runs Loader::parseWasmUnit followed by Validator::validate, treating any failure as an error since wasm-tools wit-smith guarantees valid output.

A shell script (run_fuzz.sh) drives the binary in a configurable loop (FUZZ_ITERATION_COUNT, default 100), generating random component wasm via wasm-tools wit-smith. On failure, the WAT is dumped for debugging.

CI integration is added to the Ubuntu reusable workflow.

### DIFF
--- a/.github/workflows/reusable-build-on-ubuntu.yml
+++ b/.github/workflows/reusable-build-on-ubuntu.yml
@@ -81,7 +81,7 @@ jobs:
         run: |
           curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
           source "$HOME/.cargo/env"
-          cargo install wasm-tools
+          cargo install wasm-tools@1.245.1
       - name: Component Fuzz Test
         if: ${{ !matrix.coverage && matrix.tests }}
         env:


q82419 (dismissed):

.github/workflows/reusable-build-on-ubuntu.yml
```diff
@@ -76,6 +76,20 @@ jobs:
           ./tools/wasmedge/wasmedge -v
           ctest
           cd -
+      - name: Install wasm-tools
+        if: ${{ !matrix.coverage && matrix.tests }}
+        run: |
+          curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+          source "$HOME/.cargo/env"
+          cargo install wasm-tools
```
hydai: I prefer to add a pinned version here to prevent supply chain attacks and ensure the standard is fixed.

Issue #4558, opened 2026-01-26
GSoC (2026): Refine the WASM instruction structure in WasmEdge


q82419:
### Project Title

GSoC (2026): Refine the WASM instruction structure in WasmEdge

### Description

According to the definition of WASM instructions, there are variety of immediates in every instructions. And in WasmEdge data structures currently, the immediates data of each instructions is packaged in the instruction class. This causes the situation that the instruction class should have the enough size to fit both the instructions with large and small size of immediates, and the runtime memory usage is wasted. In this program, we expect the mentee to refactor the data structures to split the instruction OpCode vector and the immediates data buffer vector to refine the memory usage of WasmEdge.

### Expected Outcome

* Refactor the instruction class to split the OpCode vector and immediates data.
* Update all tests which contain the instruction class and pass all tests.
* Perf the memory usage to prove your implementation.

### Recommend skills

* C/C++
* WebAssembly
* Github workflow

### Pre-tests

https://hyd.ai/lfx-wasmedge-pretests/2026-gsoc-4558.html

### Expected Project Size

Large.

### Mentors

YiYing He (@q82419 , yiying@secondstate.io)

### Apply Link

TBD

### Appendix

- https://github.com/cncf/mentoring/blob/main/programs/summerofcode/2026.md#wasmedge
- https://webassembly.github.io/spec/core/index.html

EricMurray-e-m-dev:
Hi @q82419 and @hydai,

I'm a student developer interested in this project for GSoC 2026. I’ve been looking into the WasmEdge codebase (specifically the AST::Instruction class) and have been playing around with some small local prototypes to better understand the problem.

I’d love to contribute to this! Before Is there a preferred design pattern or a specific method the team has in mind for this refactor? Looking forward to any advice you can share! Thanks!

tushar1977:
Hi @hydai , i was half way there in my proposal for this problem statement. Kindly clarify whether this would be there or not

hydai:
The pretest will be updated this Friday (3/6).

tushar1977:
hello @hydai 
I have prepared a draft proposal for this issue
Can you just go through and take a look?
My proposal involves using the similar instruction structure as implemented by WAVM(https://github.com/WAVM/WAVM/blob/master/Include/WAVM/IR/Operators.h)

[...]

PR #4658, opened 2026-02-17
docs: improve union Inner comments in instruction.h

P-Suraj:
## Summary
While studying the Instruction class as part of my GSoC 2026 
preparation for Issue #4558, I noticed the union Inner comments 
were quite minimal and didn't explain which WASM instructions 
use each member or why certain design decisions were made.

## Changes
- Document which WASM instructions use each union member
- Explain why BrTable and SelectT use heap-allocated pointers
- Clarify why uint128_t is needed for SIMD v128.const instructions
- Note how Type 8 (Num) being 16 bytes causes every instruction 
  to occupy at least 16 bytes regardless of actual data needs

No functional changes.


hydai:
This is 100% spam PR. Closing it.

PR #4659, opened 2026-02-18
fix broken Uninstaller and add support for fish

tavallaie:


hydai:
Since you are introducing support for a new shell, could you please add some tests for the Fish shell to verify its behavior?  
None of our maintainers are using Fish at this moment, so the test cases will help us ensure that this will not break anyone's environment.

tavallaie:
do you have any example of test for bash? 
maybe I can do that with github actions and light docker image.

hydai:
Here you go: https://github.com/WasmEdge/WasmEdge/blob/master/.github/workflows/test-installers.yml

tavallaie:
I'm on it.

hydai:
Closing due to inactivity. Please feel free to re-open it based on the latest master branch if you still want to work on this. Thanks.

PR #4669, opened 2026-02-26
fix(validator): handle unreachable type in ref.test/ref.cast validation

sumleo:
## Description

Fix logic error in `ref.test`/`ref.cast` validation that incorrectly rejects valid modules containing these instructions after `unreachable`.

After `unreachable`, `popType()` returns `std::nullopt` representing the bottom/unreachable type. The condition at `formchecker.cpp:950` used `!Type.has_value() || !Type->isRefType()`, which incorrectly rejects `nullopt`. Per the Wasm spec, unreachable code is polymorphic and the bottom type should be accepted by any type check.

Changed `!Type.has_value() ||` to `Type.has_value() &&`, matching the existing correct pattern used by `ref.is_null` at line 698.

## Checklist

## Test Evidence

Added two tests in `ValidatorSubtypeTest.cpp`:
- `RefTestAfterUnreachable` - validates `unreachable` followed by `ref.test (ref func)` passes validation
- `RefCastAfterUnreachable` - validates `unreachable` followed by `ref.cast_null (ref func)` passes validation

### DIFF
--- a/.github/workflows/IWYU_scan.yml
+++ b/.github/workflows/IWYU_scan.yml
@@ -44,7 +44,7 @@ jobs:
       version: ${{ steps.prep.outputs.version }}
     steps:
     - name: Harden the runner (Audit all outbound calls)
-      uses: step-security/harden-runner@5ef0c079ce82195b2a36a210272d6b661572d83e # v2.14.2
+      uses: step-security/harden-runner@a90bcbc6539c36a85cdfeb73f7e2f433735f215b # v2.15.0
       with:
         egress-policy: audit
 
@@ -72,7 +72,7 @@ jobs:
 
     steps:
     - name: Harden the runner (Audit all outbound calls)
-      uses: step-security/harden-runner@5ef0c079ce82195b2a36a210272d6b661572d83e # v2.14.2
+      uses: step-security/harden-runner@a90bcbc6539c36a85cdfeb73f7e2f433735f215b # v2.15.0
       with:
         egress-policy: audit
 
@@ -126,7 +126,7 @@ jobs:
       if: always()
       run: cat iwyu_fedora.log
 
-    - uses: actions/upload-artifact@b7c566a772e6b6bfb58ed0dc250532a479d7789f # v6.0.0
+    - uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f # v7.0.0
       with:
         name: iwyu_fedora.log
         path: iwyu_fedora.log
@@ -141,7 +141,7 @@ jobs:
 
     steps:
     - name: Harden the runner (Audit all outbound calls)
-      uses: step-security/harden-runner@5ef0c079ce82195b2a36a210272d6b661572d83e # v2.14.2
+      uses: step-security/harden-runner@a90bcbc6539c36a85cdfeb73f7e2f433735f215b # v2.15.0
       with:
         egress-policy: audit
 
@@ -194,7 +194,7 @@ jobs:
       if: always()
       run: cat iwyu_macOS.log
 
-    - uses: actions/upload-artifact@b7c566a772e6b6bfb58ed0dc250532a479d7789f # v6.0.0
+    - uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f # v7.0.0
       with:
         name: iwyu_macOS.log
         path: iwyu_macOS.log
--- a/.github/workflows/build-extensions.yml
+++ b/.github/workflows/build-extensions.yml
@@ -54,7 +54,7 @@ jobs:
       version: ${{ steps.prep.outputs.version }}
     steps:
       - name: Harden the runner (Audit all outbound calls)
-        uses: step-security/harden-runner@5ef0c079ce82195b2a36a210272d6b661572d83e # v2.14.2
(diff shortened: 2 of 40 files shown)


dannypsnl:
Can you add a test that contains this instruction for the validation?

sumleo:
Hi @dannypsnl, tests have been added — see the second commit (`test(validator): add tests for ref.test/ref.cast after unreachable`) which adds `RefTestAfterUnreachable` and `RefCastAfterUnreachable` tests in `ValidatorSubtypeTest.cpp`. Both construct a module with `unreachable` followed by `ref.test`/`ref.cast` and verify validation passes.

lib/validator/formchecker.cpp
```diff
@@ -947,7 +947,7 @@ Expect<void> FormChecker::checkInstr(const AST::Instruction &Instr) {
   case OpCode::Ref__cast_null: {
     EXPECTED_TRY(validate(Instr.getValType()));
     EXPECTED_TRY(auto Type, popType());
-    if (!Type.has_value() || !Type->isRefType()) {
+    if (Type.has_value() && !Type->isRefType()) {
```
q82419, line 950: You allow unreachable case here.
According to the [validation algorithm](https://webassembly.github.io/spec/core/appendix/algorithm.html#validation-of-opcode-sequences), the type should be reference type and should not be unreachable.
Please explain why you want to modify this.

sumleo: The spec validation algorithm uses `pop_val()` which returns `Bot` (bottom type) when the stack is polymorphic after `unreachable`. Per the spec appendix:

> "if the stack is polymorphic, then any type is valid."
> "the Bot type is returned, because that is a principal choice trivially satisfying all use constraints."

In WasmEdge, `popType()` returns `std::nullopt` to represent this bottom type. The original condition `!Type.has_value() || !Type->isRefType()` incorrectly rejects `nullopt`, which the spec says should be accepted.

This is consistent with how WasmEdge already handles unreachable stack in similar instructions:

- `ref.is_null` (line 698): `if (Type.has_value() && !Type->isRefType())`
- `br_on_null` (line 565): `if (ResT.has_value() && !ResT->isRefType())`
- `ref.as_non_null` (line 721): explicitly checks `Type == unreachableVType()` and accepts it

All of these already let `nullopt` (bottom type) pass through. The fix here aligns `ref.test`/`ref.cast` with the same pattern.

sumleo:
@q82419 Just checking in — does the explanation above address your concern about the unreachable case? CI is all green and the approach is consistent with how `ref.is_null`, `br_on_null`, and `ref.as_non_null` already handle `nullopt` from `popType()`. Happy to discuss further if needed.

q82419:
> @q82419 Just checking in — does the explanation above address your concern about the unreachable case? CI is all green and the approach is consistent with how `ref.is_null`, `br_on_null`, and `ref.as_non_null` already handle `nullopt` from `popType()`. Happy to discuss further if needed.

Sure, I tested with wasm-tools and your test binary can also be validated successfully.
Therefore I think your explain fits the spec, but please fix with the requested changes, thanks.

lib/validator/formchecker.cpp
```diff
@@ -947,14 +947,17 @@ Expect<void> FormChecker::checkInstr(const AST::Instruction &Instr) {
...
       // For getting bottom valtype here, matching must fail.
```
q82419: Because you modified the if-condition here, please fix the comment.
The trap occurs when the got type is not a reference type.

test/validator/ValidatorRegressionTest.cpp
```diff
@@ -257,6 +257,58 @@ TEST_F(ValidatorSubtypeTest, ExactMaxSubtypeDepth) {
   EXPECT_TRUE(ValidationResult);
 }
 
+// Test that ref.test after unreachable passes validation.
```
q82419, line 260: Just as the tests in executor, please help to modify the file name and cmake target to `ValidatorRegressionTest`.

q82419 (changes requested):

test/validator/CMakeLists.txt
```diff
@@ -1,13 +1,13 @@
...
+  ValidatorRegressionTest.cpp
 )
 
-add_test(wasmedgeValidatorSubtypeTests wasmedgeValidatorSubtypeTests)
+add_test(wasmedgeValidatorRegressionTests wasmedgeValidatorRegressionTests)
 
-target_link_libraries(wasmedgeValidatorSubtypeTests
+target_link_libraries(wasmedgeValidatorRegressionTests
   PRIVATE
   wasmedgeTestSpec
```
q82419: This target would not link to the spec test target.
Please help to remove this.

test/validator/CMakeLists.txt
```diff
@@ -1,14 +1,13 @@
...
 )
 
-add_test(wasmedgeValidatorSubtypeTests wasmedgeValidatorSubtypeTests)
+add_test(wasmedgeValidatorRegressionTests wasmedgeValidatorRegressionTests)
 
-target_link_libraries(wasmedgeValidatorSubtypeTests
+target_link_libraries(wasmedgeValidatorRegressionTests
   PRIVATE
-  wasmedgeTestSpec
   wasmedgeVM
```
q82419, line 13: You should link the gtest, such as other tests.

q82419:
This would be merged after the fixing of llvm-22 and lld-22 for CI #4696

PR #4670, opened 2026-02-26
fix(wasi): apply F_SETFL to accepted socket instead of listening socket

sumleo:
## Description

Fix `fcntl(F_SETFL)` in `sockAccept()` being applied to the listening socket (`Fd`) instead of the accepted socket (`NewFd`) on both Linux and macOS.

When `FDFLAGS_NONBLOCK` is requested:
- Line reads flags from `NewFd` (correct): `fcntl(NewFd, F_GETFL, 0)`
- Line sets flags on `Fd` (wrong): `fcntl(Fd, F_SETFL, SysFlag)`

This causes accepted sockets to remain blocking when they should be non-blocking, and the listening socket to incorrectly become non-blocking causing spurious `EAGAIN` on subsequent `accept()` calls.

Changed `Fd` to `NewFd` in the `F_SETFL` call in both `inode-linux.cpp` and `inode-macos.cpp`.

## Checklist

## Test Evidence

Added `SockAcceptNonblockFlag` test in `wasi.cpp` that creates a TCP server/client pair, calls `sock_accept` with `NONBLOCK` flag, and verifies the accepted socket has `NONBLOCK` set via `fd_fdstat_get`.


hydai (changes requested):
The windows related CI are all failed.

hydai (approved):

PR #4671, opened 2026-02-26
fix(executor): use memmove semantics for same-memory memory.copy

sumleo:
## Description

Fix data corruption in `memory.copy` when source and destination regions overlap within the same memory instance.

`runMemoryCopyOp` uses `getBytes()` returning a live `Span` into the memory buffer, then `setBytes()` which calls `std::copy` (forward iteration only). When `dst > src` within the same memory, source bytes are overwritten before being read.

The Wasm spec requires memmove semantics for `memory.copy`. This was already correctly implemented for `table.copy` in `TableInstance::setRefs`, but the fix was not propagated to memory.

When source and destination are the same memory instance, use `std::memmove` which correctly handles overlapping regions. The existing `getBytes`/`setBytes` path is preserved for cross-memory copies where overlap is not possible.

## Checklist

## Test Evidence

Added 8 test cases in `memCopyOverlapTest.cpp`:
- `ForwardOverlap` (primary regression case: dst > src with overlap)
- `BackwardOverlap`, `ExactOverlap`, `NonOverlapping`
- `ZeroLength`, `LargeForwardOverlap`, `AdjacentRegions`, `SingleByteOverlap`

All 8 tests pass locally.

### DIFF
--- a/.github/workflows/IWYU_scan.yml
+++ b/.github/workflows/IWYU_scan.yml
@@ -44,7 +44,7 @@ jobs:
       version: ${{ steps.prep.outputs.version }}
     steps:
     - name: Harden the runner (Audit all outbound calls)
-      uses: step-security/harden-runner@5ef0c079ce82195b2a36a210272d6b661572d83e # v2.14.2
+      uses: step-security/harden-runner@a90bcbc6539c36a85cdfeb73f7e2f433735f215b # v2.15.0
       with:
         egress-policy: audit
 
@@ -72,7 +72,7 @@ jobs:
 
     steps:
     - name: Harden the runner (Audit all outbound calls)
-      uses: step-security/harden-runner@5ef0c079ce82195b2a36a210272d6b661572d83e # v2.14.2
+      uses: step-security/harden-runner@a90bcbc6539c36a85cdfeb73f7e2f433735f215b # v2.15.0
       with:
         egress-policy: audit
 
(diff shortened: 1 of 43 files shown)


lib/executor/engine/memoryInstr.cpp
```diff
@@ -72,6 +74,27 @@ Executor::runMemoryCopyOp(Runtime::StackManager &StackMgr,
...
+      spdlog::error(
+          ErrInfo::InfoInstruction(Instr.getOpCode(), Instr.getOffset()));
+      return E;
+    }));
+    if (likely(Len > 0)) {
+      std::memmove(MemInstDst.getDataPtr() + Dst, MemInstSrc.getDataPtr() + Src,
+                   Len);
+    }
+    return {};
+  }
```
q82419, line 109: I think use `if-else` is better.

test/executor/CMakeLists.txt
```diff
@@ -12,3 +12,15 @@ target_link_libraries(wasmedgeExecutorCoreTests
   wasmedgeTestSpec
   wasmedgeVM
 )
+
```
q82419: I prefer not to put the test here.
You can rename the `memlimit` folder to `meminstance` and put the test file there.
Take care about the file naming style.

test/meminstance/CMakeLists.txt
```diff
@@ -12,3 +12,15 @@ target_link_libraries(wasmedgeMemLimitTests
   ${GTEST_BOTH_LIBRARIES}
   wasmedgeVM
 )
+
+wasmedge_add_executable(wasmedgeCrossMemoryCopyTests
+  CrossMemoryCopyTest.cpp
```
q82419: Sorry, you may misunderstand.

You can modify the original `MemLimitTest` target to `MemInstanceTest`, and put your tests into the same test file.

q82419 (approved):

q82419:
What's your latest commit?

q82419:
@sumleo DO NOT push a commit to trigger CI, or your workflow will not be approved to run.
Use such that commit is no use.
The workflow should be approved by maintainer to run.
We'll approve the CI workflow once the PR is ready and merge when passing.
Please squash your commits and push again.

PR #4672, opened 2026-02-26
fix(executor): correct address type assignment in copy operations

sumleo:
## Description

Fix swapped source/destination address types in `memory.copy` and `table.copy` for cross-instance operations with mixed address widths.

In `runMemoryCopyOp` and `runTableCopyOp`, `AddrType1` is derived from the source instance and `AddrType2` from the destination instance, but `Src` was extracted using `AddrType2` (wrong) and `Dst` using `AddrType1` (wrong). This causes incorrect address extraction when copying between instances with different address types (e.g., memory32 and memory64).

Swap the address type arguments so `Src` uses `AddrType1` (source's type) and `Dst` uses `AddrType2` (destination's type).

## Checklist

## Test Evidence

Added 3 test cases in `crossMemoryCopyTest.cpp` using a module with memory32 and memory64:
- `CopyFromMemory64ToMemory32`
- `CopyFromMemory32ToMemory64`
- `BidirectionalCopy` (round-trip verification)

All tests pass locally.

### DIFF
--- a/test/executor/crossMemoryCopyTest.cpp
+++ b/test/executor/crossMemoryCopyTest.cpp
@@ -17,8 +17,9 @@
 ///   uint64_t Dst = extractAddr(StackMgr.pop(), AddrType1); // wrong: src type
 ///
 /// The fix corrects the assignment:
-///   uint64_t Src = extractAddr(StackMgr.pop(), AddrType1); // correct: src type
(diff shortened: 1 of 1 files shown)


q82419:
This is not reasonable becasue the original order fits the spec.

Issue #4410, opened 2025-11-07
bug: Segmentation fault on recursive function


Q1IQ:
## Crash Command

```bash
wasmedge/bin/wasmedge run --enable-jit 0_0_call_crash.wasm main -959
```

[0_0_call_crash.wasm.txt](https://github.com/user-attachments/files/23413852/0_0_call_crash.wasm.txt)

### Standard Output
```
[2025-11-07 09:08:24.541] [info] compile start
[2025-11-07 09:08:24.731] [info] verify start
[2025-11-07 09:08:24.757] [info] optimize start
[2025-11-07 09:08:24.992] [info] optimize done

```

### Standard Error
```
Segmentation fault (core dumped)

```

### Reproduction steps

1. Build with options '...'
2. Execute with flags '....'
3. Execute with inputs '....'
4. Get error

### Screenshots

![DESCRIPTION](LINK.png)

### Components

CLI

### WasmEdge Version or Commit you used

0.15.0

### Operating system information

ubuntu 20.04

### Hardware Architecture

amd64

q82419:
This is the stack-overflow issue. If implement the `assert_exhaustion` tests in spec test, this issue can be fixed.

q82419:
Issue is at #2079

Issue #2079, opened 2022-11-13
Unexpected behavior


erxiaozhou:
### Current State
It seems the program is in an infinite loop and it doesn't output results

### Expected

Raise an exception indicating stack overflow.

## Environment

-  x86\_64 
- Operating system: Ubuntu 20.04

The following information is optional. Please provide them only if you have built from source.
- C++ Compiler version: clang-14
- CMake version: 3.21.6
- CMake flags:  `-DCMAKE_BUILD_TYPE=Release -DWASMEDGE_BUILD_AOT_RUNTIME=ON`
- Boost version: libboost1.71-dev:amd64

- build commands
  ```
  export CC=/usr/bin/clang-14
  export CXX=/usr/bin/clang++-14
  rm -rf build ; mkdir build ; 
  cd build ; cmake -DCMAKE_BUILD_TYPE=Release -DWASMEDGE_BUILD_AOT_RUNTIME=ON .. ; make -j ; cd ..
  ```

## Steps to Reproduce
[wasmedge_timeout.zip](https://github.com/WasmEdge/WasmEdge/files/9997777/wasmedge_timeout.zip)

1. build the wasmedge
2. run the test case by `build/tools/wasmedge/wasmedge --reactor <test_case_name> to_test `

q82419:
The `assert_exhaustion` error is not implemented yet.
We'll support it in the future.

Harsh-6291:
hii @hydai , @q82419 , I am a beginner looking to contribute to WasmEdge.
From the description it looks like this issue is about the runtime going into an infinite loop instead of throwing a stack‑overflow / assert_exhaustion error when running the wasmedge_timeout test case.
Could you please confirm if this issue is still reproducible on the latest master and if it is open for new contributors?
If yes, I would appreciate any hints on where in the codebase to start looking ,
Thank you!

PR #4673, opened 2026-02-26
fix(executor): enforce call stack depth limit to prevent unbounded recursion

sumleo:
## Description

Add a call stack depth limit (default 1000) to prevent unbounded heap growth from recursive Wasm modules.

`StackManager::pushFrame()` unconditionally grows `FrameStack` with no depth check. A recursive Wasm module causes unbounded heap growth until host process OOM. Every other major Wasm runtime enforces a call depth limit (Wasmtime: 512, Wasm3: 2000, V8: ~1000).

Changes:
- Add `CallStackOverflow` error code (`0x041A`) to `enum.inc`
- Add `MaxCallDepth = 1000` constant and depth check in `pushFrame()`
- Tail calls bypass the check since they reuse the current frame
- All `pushFrame` callers updated to propagate the error via `EXPECTED_TRY`

## Checklist

## Test Evidence

Added `RecursionLimit` test in `callStackDepthTest.cpp` that constructs a recursive Wasm module (`(func $f (call $f))`) and verifies it traps with `CallStackOverflow` instead of OOM. Test passes in ~1ms.

### DIFF
--- a/include/common/enum.inc
+++ b/include/common/enum.inc
@@ -1159,6 +1159,8 @@ E(AccessNullException, 0x0417, "null exception reference")
 E(CastFailed, 0x0418, "cast failure")
 // Uncaught Exception
 E(UncaughtException, 0x0419, "uncaught exception")
+// Call stack depth exceeded
+E(CallStackOverflow, 0x041A, "call stack exhausted")
 // @}
 
 #undef E
--- a/include/runtime/stackmgr.h
+++ b/include/runtime/stackmgr.h
@@ -100,11 +100,17 @@ public:
     return Vec;
   }
 
+  static constexpr uint32_t MaxCallDepth = 1000;
+
   /// Push a new frame entry to stack.
-  void pushFrame(const Instance::ModuleInstance *Module,
-                 AST::InstrView::iterator From, uint32_t LocalNum = 0,
-                 uint32_t Arity = 0, bool IsTailCall = false) noexcept {
+  Expect<void> pushFrame(const Instance::ModuleInstance *Module,
+                         AST::InstrView::iterator From, uint32_t LocalNum = 0,
+                         uint32_t Arity = 0, bool IsTailCall = false) noexcept {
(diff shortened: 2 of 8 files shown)


dannypsnl (commented):
It seems reasonable, can you add a test?

sumleo:
Hi @dannypsnl, a regression test is already included — the second commit (`test(executor): add call stack depth limit regression test`) adds `callStackDepthTest.cpp` with a `RecursionLimit` test that constructs a recursive Wasm module `(func $f (call $f))` and verifies it traps with `CallStackOverflow` instead of running forever.

q82419:
Same solution at #4559.
Issues related to #2079 and #4410.

The same question: is the value of max call depth (1000 you used here) reasonable?

sumleo:
Thanks for pointing out #4559, #2079, and #4410.

Regarding the max call depth value: PR #4559 uses 10000 with a configurable API. I used 1000 here as a conservative default — V8 uses 1000, SpiderMonkey ~512-1024, Wasmtime 512, and WAMR 100 by default. 1000 handles typical recursive patterns while catching runaway recursion before it crashes the host process.

That said, if the plan is to land #4559's configurable approach (which adds `WasmEdge_ConfigureSetMaxCallDepth`/`WasmEdge_ConfigureGetMaxCallDepth`), I'm happy to close this PR in favor of that and help review #4559 instead. Or I can align the default value here with whatever the project prefers.

What would you like to do?

sumleo:
Thanks for pointing out #4559 and the related issues.

Regarding the max call depth value: I chose 1000 to match V8 default (~1000). For reference, other runtimes use Wasmtime 512, Wasm3 2000.

I see PR #4559 uses 10000 and makes it configurable via RuntimeConfigure. If the preference is to make it configurable rather than hardcoded, I am happy to adjust this PR to align with that approach. Let me know what value and approach you would prefer.

q82419:
replace by #5040

PR #4674, opened 2026-02-26
fix(loader): add nesting depth limit for component model parsing

sumleo:
## Description

Add a nesting depth limit (100) for component model loading to prevent native stack overflow from deeply nested component sections.

`loadSection(ComponentSection)` calls `loadComponent()` recursively with no depth counter. A crafted binary with ~2000 nested component sections (~16KB) overflows the C++ call stack, causing a host process segfault during parsing.

Changes:
- Add `ComponentNestingDepth` counter and `MaxComponentNestingDepth = 100` constant to the Loader class
- Check depth at entry to `loadComponent()` and return `MalformedSection` error if exceeded
- Use RAII guard for exception-safe depth decrement on all return paths

## Checklist

## Test Evidence

Added 4 tests in `componentNestingTest.cpp`:
- `WithinLimit` (depth 5, succeeds)
- `ExceedsLimit` (depth 101, fails with MalformedSection)
- `ExactlyAtLimit` (depth 99 nested sections = 100 calls, succeeds)
- `OneOverLimit` (depth 100 nested sections = 101 calls, fails)

All 15 component tests pass (11 existing + 4 new).

### DIFF
--- a/include/loader/loader.h
+++ b/include/loader/loader.h
@@ -294,7 +294,8 @@ private:
 
   // Load component.
   Expect<void> loadComponent(AST::Component::Component &Comp,
-                             std::optional<uint64_t> Bound = std::nullopt);
+                             std::optional<uint64_t> Bound = std::nullopt,
+                             uint32_t Depth = 0);
   /// @}
 
   /// \name Load AST section node helper functions
@@ -419,7 +420,8 @@ private:
   Expect<void> loadSection(AST::Component::CoreModuleSection &Sec);
   Expect<void> loadSection(AST::Component::CoreInstanceSection &Sec);
   Expect<void> loadSection(AST::Component::CoreTypeSection &Sec);
-  Expect<void> loadSection(AST::Component::ComponentSection &Sec);
+  Expect<void> loadSection(AST::Component::ComponentSection &Sec,
+                           uint32_t Depth = 0);
   Expect<void> loadSection(AST::Component::InstanceSection &Sec);
   Expect<void> loadSection(AST::Component::AliasSection &Sec);
   Expect<void> loadSection(AST::Component::TypeSection &Sec);
@@ -514,6 +516,11 @@ private:
   const Executable::IntrinsicsTable *IntrinsicsTable;
   std::recursive_mutex Mutex;
   bool HasDataSection;
+  // Limit chosen to prevent native stack overflow during recursive parsing.
+  // Each nested loadComponent call adds ~2KB to the C++ call stack; 100 levels
+  // uses ~200KB which is well within default thread stack sizes (1-8MB) while
+  // still being far beyond any realistic component nesting in practice.
+  static constexpr uint32_t MaxComponentNestingDepth = 100;
 
   /// Input data type enumeration.
   enum class InputType : uint8_t { WASM, UniversalWASM, SharedLibrary };
--- a/lib/loader/ast/component/component.cpp
+++ b/lib/loader/ast/component/component.cpp
@@ -41,7 +41,15 @@ Expect<std::pair<std::vector<Byte>, std::vector<Byte>>> Loader::loadPreamble() {
 }
 
 Expect<void> Loader::loadComponent(AST::Component::Component &Comp,
-                                   std::optional<uint64_t> Bound) {
+                                   std::optional<uint64_t> Bound,
+                                   uint32_t Depth) {
+  if (unlikely(Depth >= MaxComponentNestingDepth)) {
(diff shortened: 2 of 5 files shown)


include/loader/loader.h
```diff
@@ -514,6 +514,8 @@ class Loader {
   const Executable::IntrinsicsTable *IntrinsicsTable;
   std::recursive_mutex Mutex;
   bool HasDataSection;
+  uint32_t ComponentNestingDepth = 0;
+  static constexpr uint32_t MaxComponentNestingDepth = 100;
```
dannypsnl, line 523: @hydai Do you think we need to make this be configurable?

lib/loader/ast/component/component.cpp
```diff
@@ -42,6 +42,18 @@ Expect<std::pair<std::vector<Byte>, std::vector<Byte>>> Loader::loadPreamble() {
 
 Expect<void> Loader::loadComponent(AST::Component::Component &Comp,
                                    std::optional<uint64_t> Bound) {
+  if (unlikely(ComponentNestingDepth >= MaxComponentNestingDepth)) {
+    spdlog::error("component nesting depth exceeded limit of {}"sv,
+                  MaxComponentNestingDepth);
+    return logLoadError(ErrCode::Value::MalformedSection,
+                        FMgr.getLastOffset(), ASTNodeAttr::Component);
+  }
+  ++ComponentNestingDepth;
```
dannypsnl: Does this mean even I have a component that contains 100 components (hence invokes `loadComponent` over 100 times) will be rejected?

If so, I suggest you to use immutable design.

dannypsnl (changes requested):

test/component/componentNestingTest.cpp
```diff
@@ -0,0 +1,140 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2019-2024 Second State INC
```
dannypsnl: It's 2026.

include/loader/loader.h
hydai: I am unsure if this kind of limitation is specified in the component model. However, if it is similar to the wasm spec, which allows runtimes to decide the behavior, it should be fine to have a fixed value here. But I think we should have a strong reason for choosing 100 instead of other values.

lib/loader/ast/component/component.cpp
sumleo: No — a component with 100 sibling child components will load fine. The counter tracks recursive *nesting* depth, not total invocations. Each sibling is loaded at depth 2 (outer + child), and the RAII guard decrements the counter when each child's `loadComponent` returns, so the depth stays at 2 throughout.

I've added a `ManySiblingComponents` test to demonstrate this: it loads a component with 150 sibling children successfully.

test/component/componentNestingTest.cpp
sumleo: Fixed, thanks.

include/loader/loader.h
sumleo: The component model spec doesn't prescribe a specific limit — this is a runtime safety measure similar to the core Wasm spec allowing implementation-defined limits.

Rationale for 100: each recursive `loadComponent` call adds ~2KB to the native C++ call stack. At 100 levels that's ~200KB, well within default thread stack sizes (1-8MB). 100 is also far beyond realistic nesting in practice — actual component binaries rarely nest beyond single-digit levels.

Added a comment at the constant definition explaining this. Happy to adjust if you have a preference for a different value.

sumleo: I think a fixed value is sufficient here since this is purely a stack overflow safety guard, not a semantic constraint. If there's a future need for deeply nested components, it could be made configurable via `Configure`, but in practice even 50 levels of nesting would be unusual.

lib/loader/ast/component/component.cpp
dannypsnl: Good, but I still consider RAII guard as a latent bug.

sumleo: Agreed. Refactored to pass depth as an immutable parameter through `loadComponent` and `loadSection(ComponentSection)` instead of using mutable member state + RAII guard. The depth counter is now threaded explicitly through the call chain.

dannypsnl (approved):

q82419:
Close this due to out-of-date from current latest branch.
Component-model now is planned to base on the active `dev/component` branch.
Please update and re-create PR to that branch if this is still in active development.

PR #4675, opened 2026-02-26
fix(wasi): add rights checks to sockBind, sockListen, and sockConnect

sumleo:
## Description

Add `can()` rights checks to `sockBind`, `sockListen`, and `sockConnect` in VINode, matching the pattern used by file operations.

Socket operations `sockBind`, `sockListen`, and `sockConnect` call directly into INode methods without `can()` rights checks. File operations like `fdAdvise`, `fdAllocate`, `fdRead`, etc. all correctly check rights before proceeding. If an application restricts socket rights via `fdFdstatSetRights`, the restriction has no effect because VINode's socket methods bypass the capability check.

Added `can(__WASI_RIGHTS_SOCK_BIND)` checks at the start of each method, returning `__WASI_ERRNO_NOTCAPABLE` when the right is not present.

## Checklist

## Test Evidence

Added `SocketOpsRightsChecks` test in `wasi.cpp` that:
1. Creates a TCP socket with `SOCK_BIND` right
2. Strips `SOCK_BIND` via `fdFdstatSetRights`
3. Verifies `sockBind`, `sockListen`, `sockConnect` all return `NOTCAPABLE`

### DIFF
--- a/include/host/wasi/vinode.h
+++ b/include/host/wasi/vinode.h
@@ -568,10 +568,16 @@ public:
   WasiExpect<void> sockBind(__wasi_address_family_t AddressFamily,
                             Span<const uint8_t> Address,
                             uint16_t Port) noexcept {
+    if (!can(__WASI_RIGHTS_SOCK_BIND)) {
+      return WasiUnexpect(__WASI_ERRNO_NOTCAPABLE);
+    }
     return Node.sockBind(AddressFamily, Address, Port);
   }
 
   WasiExpect<void> sockListen(int32_t Backlog) noexcept {
+    if (!can(__WASI_RIGHTS_SOCK_BIND)) {
+      return WasiUnexpect(__WASI_ERRNO_NOTCAPABLE);
+    }
     return Node.sockListen(Backlog);
   }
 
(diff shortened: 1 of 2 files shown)


.github/scripts/run-riscv64-quick-tests.sh
```diff
@@ -62,7 +62,7 @@ QUICK_TESTS="
   wasmedgeLoaderFileMgrTests
   wasmedgeLoaderASTTests
   wasmedgeLoaderSerializerTests
-  wasmedgeValidatorSubtypeTests
+  wasmedgeValidatorRegressionTests
```
dannypsnl: Why need to rename this test?

hydai:
Hi @sumleo 
Since there is a conflict, I am unable to continue deal with this PR, would you like to resolve it and force-push this branch?

sumleo:
Hi @hydai, thanks for the heads-up. I've rebased the branch onto the latest `master` and force-pushed (`82ff9c9`).

The conflict was in `test/host/wasi/wasi.cpp`: `master` added the `SockAcceptNonblockFlag` test at the same spot where this PR appends `SocketOpsRightsChecks`. Both are independent tests, so I kept both — this PR's change stays purely additive (+94/-0, `vinode.h` rights checks + the new test). GitHub now shows the PR as mergeable. PTAL 🙏

PR #4676, opened 2026-02-26
fix(loader): validate try_table catch flag byte range

sumleo:
## Description

Add range validation for the catch flag byte in `try_table` instruction loading.

The catch flag byte was not validated to be in the range 0x00-0x03. Values 0x04-0xFF were silently accepted with upper bits ignored via masking. For example, flag=0xFF would behave identically to 0x03 (`catch_all_ref`).

Added `if (Flag > 0x03) return logLoadError(...)` after reading the flag byte, before the bit-masking operations.

## Checklist

## Test Evidence

Added 2 tests in `instructionTest.cpp`:
- `LoadTryTableValidCatchFlags` - try_table with flags 0x00-0x03 loads successfully
- `LoadTryTableInvalidCatchFlag` - try_table with flags 0x04 and 0xFF fails to load

All 15 loader instruction tests pass (13 existing + 2 new).

### DIFF
--- a/.github/workflows/reusable-build-extensions-on-linux.yml
+++ b/.github/workflows/reusable-build-extensions-on-linux.yml
@@ -28,6 +28,7 @@ permissions:
 
 jobs:
   build:
+    if: ${{ inputs.plugins != '[]' }}
     permissions:
       contents: write
     strategy:
--- a/.github/workflows/reusable-build-extensions-on-macos.yml
+++ b/.github/workflows/reusable-build-extensions-on-macos.yml
@@ -25,6 +25,7 @@ permissions:
 
 jobs:
   build:
+    if: ${{ inputs.plugins != '[]' }}
     permissions:
       contents: write
     strategy:
--- a/.github/workflows/reusable-build-extensions.yml
+++ b/.github/workflows/reusable-build-extensions.yml
@@ -104,6 +104,10 @@ jobs:
 
   build_on_ubuntu:
     needs: prepare
+    if: >-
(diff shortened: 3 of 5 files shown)


test/loader/instructionTest.cpp
```diff
@@ -1167,4 +1167,148 @@ TEST(InstructionTest, LoadSIMDInstruction) {
   };
   EXPECT_FALSE(Ldr.parseModule(prefixedVec(Vec)));
 }
+
+TEST(InstructionTest, LoadTryTableValidCatchFlags) {
```
q82419: Merge the tests into one (`LoadTryTable`).
Please re-construct the cases as the similar style as above.

lib/loader/ast/instruction.cpp
```diff
@@ -339,6 +339,10 @@ Expect<void> Loader::loadInstruction(AST::Instruction &Instr) {
       auto &Desc = Instr.getTryCatch().Catch[I];
       // Read the catch flag.
       EXPECTED_TRY(uint8_t Flag, FMgr.readByte().map_error(ReportError));
+      if (unlikely(Flag > 0x03U)) {
+        return logLoadError(ErrCode::Value::IllegalOpCode, FMgr.getLastOffset(),
```
q82419: Since there's no similar test cases on the spec repo, I think we can add a new enum in `enum.inc`:

```
// Malformed catch flag (Exception-handling proposal).
E(MalformedCatchFlags, 0x011E, "malformed catch flags")
```

q82419 (approved):

PR #4677, opened 2026-02-26
fix(wasi): only skip literal dot component in canonicalGuest path

sumleo:
## Description

Fix `canonicalGuest` incorrectly dropping dot-prefixed path components like `.hidden` and `.config`.

The condition `Part.front() != '.' || Parts.size() != 1` was intended to skip the literal `.` (current directory) component, but it matches any component starting with `.` when `Parts.size() == 1`. This causes `canonicalGuest(".hidden")` to return an empty string instead of `.hidden`.

Changed the condition to check both `Part.size() == 1` and `Part.front() == '.'`, so only the exact `.` component is skipped while `.hidden`, `.config`, etc. are preserved.

## Checklist

## Test Evidence

Added `CanonicalGuest` test in `wasi.cpp` with 17 assertions covering:
- Single dot (`.`) correctly stripped
- Dot-prefixed names (`.hidden`, `.config`, `.gitignore`) preserved
- Multi-component paths with dot-prefixed directories
- Double-dot (`..`) parent resolution
- Leading slash stripping, empty paths, consecutive slashes

### DIFF
--- a/lib/host/wasi/vinode.cpp
+++ b/lib/host/wasi/vinode.cpp
@@ -68,7 +68,8 @@ std::string VINode::canonicalGuest(std::string_view Path) {
       if (!Parts.empty()) {
         Parts.pop_back();
       }
-    } else if (Part.front() != '.' || Parts.size() != 1) {
+    } else if (!(Part.size() == 1 && Part.front() == '.') ||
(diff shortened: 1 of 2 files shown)


q82419:
Out of date and inactivity

PR #4679, opened 2026-02-26
fix(executor): preserve nullability for externalized refs in ref.test/ref.cast

sumleo:
## Description

Fix `ref.test`/`ref.cast` losing nullability information when handling externalized references.

In `runRefTestOp` (refInstr.cpp:330), when an externalized reference is encountered, the value type is always created as non-nullable:
```cpp
VT = ValType(TypeCode::Ref, TypeCode::ExternRef);
```

This discards the original nullability. The fix preserves nullability using the same pattern already used by `runBrOnCastOp` in `controlInstr.cpp`:
```cpp
VT = ValType(VT.isNullableRefType() ? TypeCode::RefNull : TypeCode::Ref,
             TypeCode::ExternRef);
```

## Checklist

## Test Evidence

Added 5 tests in `refTestNullabilityTest.cpp`:
- `ExternalizedNonNullAnyRef` - ref.test on non-null externalized reference
- `ExternalizedNullAnyRef` - ref.test on null externalized reference
- `ExternalizedRoundTripExternRef` - internalize+externalize round-trip
- `ExternalizedNullExternRefRoundTrip` - null round-trip
- `ValTypeNullabilityPreservation` - direct unit test of the ValType fix logic


hydai:
Hi @sumleo,  
I assume that you may be using some tools to help you identify issues and create PRs because your PR template does not use ours. That's fine, but most of your PRs do not include corresponding tests. Could you please add the relevant tests to cover your fixes?

sumleo:
Hi @hydai, thanks for the feedback\! I've added tests to all PRs and updated the descriptions to follow the project's PR template. Each fix now includes corresponding regression tests. Let me know if anything else needs to be addressed.

test/executor/CMakeLists.txt
```diff
@@ -12,3 +12,17 @@ target_link_libraries(wasmedgeExecutorCoreTests
   wasmedgeTestSpec
   wasmedgeVM
 )
+
+wasmedge_add_executable(wasmedgeExecutorRefTestNullabilityTests
+  refTestNullabilityTest.cpp
+)
```
hydai: What is the reason for adding a standalone test binary without integrating it into the previous one?

sumleo: The existing `wasmedgeExecutorCoreTests` is built against `wasmedgeTestSpec` for running Wasm spec conformance suites with a parameterized test pattern. This regression test uses a different setup — direct `wasmedgeVM` + gtest with an embedded Wasm binary, so it can't easily be integrated into the spec test runner.

Happy to consolidate all executor regression tests into a single shared binary (e.g., `wasmedgeExecutorRegressionTests`) if that's preferred.

lib/executor/engine/refInstr.cpp
```diff
@@ -327,7 +327,8 @@ Executor::runRefTestOp(const Runtime::Instance::ModuleInstance *ModInst,
   // Copy the value type here due to handling the externalized case.
   auto VT = Val.get<RefVariant>().getType();
   if (VT.isExternalized()) {
-    VT = ValType(TypeCode::Ref, TypeCode::ExternRef);
+    VT = ValType(VT.isNullableRefType() ? TypeCode::RefNull : TypeCode::Ref,
```
q82419, line 330: Still consider about the correctness.
Run CI for spec test first.

test/executor/CMakeLists.txt
sumleo: The existing executor tests use \`wasmedgeTestSpec\` for spec conformance suites with parameterized WAT loading. This regression test constructs a Wasm module in-memory via the VM API to test a specific edge case (externalized ref nullability in ref.test/ref.cast), which doesn't fit the spec runner pattern.

If you'd prefer, I can consolidate all such executor regression tests into a shared \`wasmedgeExecutorRegressionTests\` binary, or integrate it into an existing test target. What approach works best for the project?

lib/executor/engine/refInstr.cpp
sumleo: CI spec tests have run and all pass — the core test suite including the reference type and GC proposal spec tests are green on all platforms (Ubuntu, macOS 14, macOS 15, Fedora, etc.).

The fix adds a check before dereferencing the externalized reference: if the original reference is null, it preserves the null result rather than attempting to strip the extern wrapper. This prevents the ref.test/ref.cast instruction from incorrectly reporting a non-null type for an externalized null reference.

The regression test specifically validates:
1. \`(ref.test (ref extern))\` on a null externref returns 0 (not 1)
2. \`(ref.cast (ref null extern))\` on a null externref succeeds without trapping

Both align with the Wasm GC spec's semantics for extern conversions.

test/executor/CMakeLists.txt
q82419: I think the new test binary target `wasmedgeExecutorRegressionTests` sounds great.
Please set the file name as `ExecutorRegressionTest.cpp`, and collect all your tests in this PR into a test unit `TEST(ExecutorRegression, RefTestNullAbility)`.

test/executor/ExecutorRegressionTest.cpp
```diff
@@ -0,0 +1,318 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2019-2026 Second State INC
+
+//===-- wasmedge/test/executor/ExecutorRegressionTest.cpp
+//------------------===//
```
q82419: Fix the comment here.

test/executor/ExecutorRegressionTest.cpp
```diff
@@ -0,0 +1,318 @@
...
+//
+// Part of the WasmEdge Project.
+//
+//===----------------------------------------------------------------------===//
+///
+/// \file
+/// Regression test for ref.test on externalized nullable references.
```
q82419: This file is for the regression or misc tests for executor.
Please fix the file description here and move the `ref.test` related description to the TEST unit below.

q82419 (changes requested):

q82419:
Seems like the LLVM on macOS and Fedora rawhide updated causes error.

cc @hydai  @ibmibmibm

q82419:
@sumleo DO NOT push a commit to trigger CI, or your workflow will not be approved to run.
Use such that commit is no use.
The workflow should be approved by maintainer to run.
We'll approve the CI workflow once the PR is ready and merge when passing.
Please squash your commits and push again.

PR #4683, opened 2026-02-26
test: add C-API boundary test for empty strings

Sriram-B-Srivatsa:
### Description
This PR introduces a boundary case unit test for `WasmEdge_StringCreateByCString` within the `APICoreTest` suite. 

### Technical Details
While the C-API handles standard strings, boundary conditions like empty C-strings (`""`) must be explicitly verified to ensure:
1. The `WasmEdge_String.Length` is correctly initialized to `0`.
2. The internal buffer mapping remains consistent with `WasmEdge_StringWrap`.
3. Memory deallocation via `WasmEdge_StringDelete` handles these descriptors without regression.

This improves the reliability of the string-processing layer for host-function integrations.

### DIFF
--- a/.github/workflows/build_for_nix.yml
+++ b/.github/workflows/build_for_nix.yml
@@ -42,7 +42,7 @@ jobs:
         egress-policy: audit
 
     - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
-    - uses: cachix/install-nix-action@19effe9fe722874e6d46dd7182e4b8b7a43c4a99 # v31.10.0
+    - uses: cachix/install-nix-action@1ca7d21a94afc7c957383a2d217460d980de4934 # v31.10.1
       with:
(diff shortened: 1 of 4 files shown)


test/api/APIUnitTest.cpp
```diff
@@ -3136,6 +3136,17 @@ TEST(APICoreTest, Async) {
   WasmEdge_ExecutorDelete(Exec);
 }
 
+TEST(APICoreTest, StringEmptyTest) {
```
q82419: Please merge this test into `TEST(APICoreTest, String)`. Thanks.

Sriram-B-Srivatsa: Done! I've moved the empty string boundary checks inside the main TEST(APICoreTest, String) block and rebased against the latest master. Thank you for the review!

Sriram-B-Srivatsa:
PR #4683: Noticed a Debian build failure with exit code 100. I think its an upstream apt-update network error.

PR #4686, opened 2026-02-27
fix(api): guard WasmEdge_StringCopy against null Buf

Sanchit2662:
## Summary

Found a null pointer dereference in `WasmEdge_StringCopy` where `memset` was guarded against null `Buf` but `std::copy_n` wasn't. When callers pass `Buf = nullptr` with `Len > 0` and `Str.Length > 0`, we crash on the copy.

The existing test only caught `Len = 0` cases. Realistic trigger: someone trying the "query size first" pattern with a non-zero length.

## Fix

Moved `std::copy_n` inside the existing `if (Buf)` guard and hoisted `RealLength` so the return value stays correct for null buffer calls.

**Before:**
```cpp
  if (Buf) {
    std::memset(Buf, 0, Len);
  }
  uint32_t RealLength = std::min(Len, Str.Length);
  if (RealLength > 0) {
    std::copy_n(Str.Buf, RealLength, Buf);  // crash: null Buf
  }
```

**After:**
```cpp
  uint32_t RealLength = std::min(Len, Str.Length);
  if (Buf) {
    std::memset(Buf, 0, Len);
    if (RealLength > 0) {
      std::copy_n(Str.Buf, RealLength, Buf);
    }
  }
```


Sanchit2662:
Hi @q82419 , please review the changes when you get a chance. Happy to make any adjustments if needed. Thanks!

Sanchit2662:
Hi @q82419, is there any update on this PR, or would you like me to change anything?

hydai (approved):
The basic idea is correct. Since the return value is not handled, I will create a followup PR to fix it later.

PR #4704, opened 2026-03-10
fix(driver): update ToolOnComponent to use ValVariant for argument types

LFsWang:
## Description

hotfix: allow CLI arguments to be passed for running the component model.

I am not sure about the details of the above variant, but this fix makes the CLI work now. However, I think it means we are missing a converter somewhere.

```
using ComponentValVariant = std::variant<
    // constant types in component types
    uint8_t, uint16_t, uint32_t, uint64_t, int8_t, int16_t, int32_t, int64_t,
    float, double, bool, std::string,
    // composition type like List, Record, Variant
    //
    // we need to copy them at many place, so we just use shared_ptr
    // std::shared_ptr<ValComp>,
    // wasm values
    ValVariant>;
```

```
/// ValVariant for all value types.
using ValVariant =
    Variant<uint32_t, int32_t, uint64_t, int64_t, float, double, uint128_t,
            int128_t, uint64x2_t, int64x2_t, uint32x4_t, int32x4_t, uint16x8_t,
            int16x8_t, uint8x16_t, int8x16_t, floatx4_t, doublex2_t,
            RefVariant>;
```
## Checklist

## Test Evidence

---

image


hydai:
I think we just missed the conversion between the arguments list and the vector.
It's fine to merge this as a workaround.

PR #4705, opened 2026-03-10
si, fix(wasi-nn): prevent host memory overrun in openvino getOutput backend

jay7-tech:
## Description
[wasinn_openvino.cpp](cci:7://file:///c:/Users/JAYADEEP%20GOWDA%20K%20B/Desktop/New%20folder/targets/WasmEdge/plugins/wasi_nn/wasinn_openvino.cpp:0:0-0:0) blindly copies the model output into the guest buffer using `std::copy_n` without checking if `OutBuffer` is actually large enough to hold `BytesWritten`. Added a quick bounds check to prevent host heap overruns. 

## Checklist
- [x] **DCO Signed-off**
- [x] **Commit Messages**
- [x] **Local Tests Passed**

## Test Evidence
Change is just a 5-line bounds check in [wasinn_openvino.cpp](cci:7://file:///c:/Users/JAYADEEP%20GOWDA%20K%20B/Desktop/New%20folder/targets/WasmEdge/plugins/wasi_nn/wasinn_openvino.cpp:0:0-0:0). Relying on CI for the full C++ backend regression tests.

### DIFF
--- a/plugins/wasi_nn/wasinn_openvino.cpp
+++ b/plugins/wasi_nn/wasinn_openvino.cpp
@@ -160,6 +160,15 @@ Expect<WASINN::ErrNo> getOutput(WASINN::WasiNNEnvironment &Env,
     const ov::Tensor &OutputTensor =
         CxtRef.OpenVINOInferRequest.get_output_tensor(Index);
     BytesWritten = OutputTensor.get_byte_size();
+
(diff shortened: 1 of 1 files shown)


hydai:
Thanks for this PR. This will be replaced by #5091, which unified all wasi-nn backends in the same way to handle this kind of issues.

PR #4713, opened 2026-03-14
fix(validator): refactor component instance export handling and impro…

LFsWang:
…ve type mapping

## Description

This PR resolves inconsistencies between export and inline export in instance handling across two different paths:

* export includes sort and optionally externdesc.
* inline export includes externdesc but does not contain sort.

fix the RegressionTest.
fix some the spec test (export) with core sort

## Checklist

## Test Evidence

---

### DIFF
--- a/lib/validator/component_validator.cpp
+++ b/lib/validator/component_validator.cpp
@@ -837,7 +837,22 @@ Expect<void> Validator::validate(const AST::Component::Export &Ex) noexcept {
     }));
   }
   const auto &Sort = Ex.getSortIndex().getSort();
-  if (!Sort.isCore()) {
+  uint32_t Idx = Ex.getSortIndex().getIdx();
+  if (Sort.isCore()) {
+    if (Idx >= CompCtx.getCoreSortIndexSize(Sort.getCoreSortType())) {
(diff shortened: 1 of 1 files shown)


lib/validator/component_validator.cpp
```diff
@@ -13,6 +13,29 @@ namespace Validator {
...
+    return AST::Component::Sort::SortType::Type;
+  case AST::Component::ExternDesc::DescType::ComponentType:
+    return AST::Component::Sort::SortType::Component;
+  case AST::Component::ExternDesc::DescType::InstanceType:
+    return AST::Component::Sort::SortType::Instance;
+  default:
+    assumingUnreachable();
+  }
+}
+
```
dannypsnl, line 38: Can we use Typecast Operator for this purpose?

q82419:
@LFsWang 

I think we have the structural problems in validation context due to lack of in-sight of the component model proposal before.
How do you think if we fully refactor the validator of component model?

LFsWang:
> @LFsWang
> 
> I think we have the structural problems in validation context due to lack of in-sight of the component model proposal before. How do you think if we fully refactor the validator of component model?

I think it is good, because I also found some issue should resolve from loader. (parts whose specifications have not been changed.)

PR #4726, opened 2026-03-18
feat: implement module instance dependency tree in WASM store

### LINKED ISSUE #4692: LFX Workspace: Implement module instance dependency tree in WASM store
### Project Title

Implement module instance dependency tree in WASM store

### Motivation

In the current WasmEdge implementation, multiple WebAssembly modules can be loaded, linked, and registered into the `Store`. However, there is a lack of a safety mechanism to manage the lifecycle of these module instances. If a module is deleted while others still reference its functions or memory, it leads to dangling pointers and system crashes.

### Expected Outcome

- Implement the module instance dependency tree and the maintainance methodology.
- Design the multiple module instances test cases to verify the implementation.

### Details

The goal of this project is to implement a Module Instance Dependency Tree and a corresponding maintenance methodology. The implementation will focus on the following aspects:
- Static Dependency Tracking: During the module linking phase, the system will identify "Import" relationships. When Module A imports an entity from Module B, a directed edge will be recorded .
- In-Degree Management: Each module instance will maintain an "In-degree" counter. The In-degree represents the number of other instances depending on it.
- Safe Deletion Logic: We will modify the unregistration process to check the dependency state.
- Testing & Validation: A comprehensive test suite will be developed to simulate complex scenarios, including multi-level dependencies and circular references.

### Week 1-2: Source Code Analysis & Design
- [x] Analyze `WasmEdge::Runtime::Store` implementation for instance management.
- [x] Study the `WasmEdge::Runtime::Instance` classes and their export/import mechanisms.
- [x] Trace the module linking process to find the injection point for dependency tracking.
- [x] Design the C++ data structures (e.g., adjacency list or reference counting) for the dependency tree.

### Week 3-4: Dependency Recording Logic
- [x] Implement logic to increment "In-degree" counts during the module linking phase.
- [x] Ensure dependency links are correctly registered when a new instance is added to the Store.
- [x] Implement automated "In-degree" decrementing when a dependent module is removed.

### Week 5-6: Safe Unregistration Implementation
- [x] Modify the `Unregister` API to check a module’s In-degree before deletion.
- [ ] Implement error handling and return codes for blocked deletion attempts.
- [ ] [Milestone] Submit Mid-term PR by April 7.

### Week 7-8: Mid-term Review & Feedback Integration
- [ ] Complete the mid-term evaluation with mentors (April 14).
- [ ] Refactor the core tracking logic based on mentor and community code review.
- [ ] Ensure coding style and architectural alignment with WasmEdge standards.

### Week 9-10: Handling Complex Edge Cases
- [ ] Implement validation for deep dependency trees (cascading dependency checks).
- [ ] Design and implement protection mechanisms for circular dependencies.
- [ ] Investigate `StackManager` integration to check if a module is active on the execution stack.

### Week 11: Comprehensive Testing & Integration
- [ ] Develop multiple-module test cases to verify the dependency tree accuracy.
- [ ] Perform stress tests for frequent registration/unregistration cycles.
- [ ] Conduct memory leak checks using Valgrind or similar tools.
- [ ] [Milestone] Submit Final PR by May 19 (Ready for Merging).

### Week 12: Documentation & Final Evaluation
- [ ] Finalize technical documentation for the new dependency management feature.
- [ ] Write a developer guide on how the Store handles instance lifecycles.
- [ ] Complete the final LFX mentorship evaluation (May 26).

### Appendix

* Upstream Issue: https://github.com/WasmEdge/WasmEdge/issues/4514

Wang-Yan-Hao:
## Checklist

## Test Evidence

image

---

### DIFF
--- a/include/api/wasmedge/wasmedge_vm.h
+++ b/include/api/wasmedge/wasmedge_vm.h
@@ -604,18 +604,15 @@ WasmEdge_VMGetFunctionList(const WasmEdge_VMContext *Cxt,
                            const WasmEdge_FunctionTypeContext **FuncTypes,
                            const uint32_t Len) WASMEDGE_CAPI_NOEXCEPT;
 
-/// Forcibly delete a registered module from the VM context.
+/// Unregister and trigger the deletion of a registered module from the VM.
 ///
-/// \warning This function does not check whether other modules depend on the
-/// target. Deleting a module that is still in use may cause undefined behavior
-/// or crashes.
-///
-/// A safer deletion API may be introduced in the future once module dependency
-/// management is implemented.
+/// This function unregisters the module with the given name from the VM.
+/// The module will be safely destroyed only when there are no remaining
+/// dependencies from other modules
 ///
 /// \param Cxt the WasmEdge_VMContext to delete the module from.
 /// \param ModuleName the name of the module to delete.
-WASMEDGE_CAPI_EXPORT extern void WasmEdge_VMForceDeleteRegisteredModule(
+WASMEDGE_CAPI_EXPORT extern void WasmEdge_VMDeleteRegisteredModule(
     const WasmEdge_VMContext *Cxt,
     const WasmEdge_String ModuleName) WASMEDGE_CAPI_NOEXCEPT;
 
--- a/include/runtime/instance/module.h
+++ b/include/runtime/instance/module.h
@@ -78,16 +78,21 @@ public:
                  std::function<void(void *)> Finalizer = nullptr)
       : ModName(Name), HostData(Data), HostDataFinalizer(Finalizer) {}
   virtual ~ModuleInstance() noexcept {
-    // When destroying this module instance, call the callbacks to unlink to the
-    // store managers.
-    for (auto &&[Key, Callback] : LinkedStore) {
-      assuming(Callback);
-      Callback(Key, this);
-    }
+    // Some module instances (e.g., stack-allocated ones) may not trigger
+    // the explicit termination process. We call unlinkAllStores() as
+    // a fallback to ensure all store manager callbacks are cleared.
+    unlinkAllStores();
     if (HostDataFinalizer.operator bool()) {
       HostDataFinalizer(HostData);
     }
+    for (auto *Provider : DependencyList) {
+      if (Provider) {
+        Provider->decrementInDegree();
+      }
+    }
   }
+
+  void terminate() noexcept { resetSelfDegree(); }
 
   std::string_view getModuleName() const noexcept {
     std::shared_lock Lock(Mutex);
@@ -574,6 +579,50 @@ protected:
     LinkedStore.erase(LinkedStoreKey{Store, std::string(Name)});
   }
 
+  void unlinkAllStores() noexcept {
+    // When destroying this module instance, call the callbacks to unlink to the
+    // store managers.
+    for (auto &&[Key, Callback] : LinkedStore) {
+      assuming(Callback);
+      Callback(Key, this);
+    }
+    LinkedStore.clear();
+  }
+
+  void incrementInDegree() noexcept {
+    InDegree.fetch_add(1, std::memory_order_relaxed);
+  }
+
+  void decrementInDegree() noexcept {
+    if (InDegree.fetch_sub(1, std::memory_order_acq_rel) == 1) {
+      this->tryToDelete();
+    }
+  }
+
+  void linkDependency(ModuleInstance &Provider) {
+    std::unique_lock Lock(Mutex);
+    auto It =
+        std::find(DependencyList.begin(), DependencyList.end(), &Provider);
+    if (It != DependencyList.end()) {
+      return;
+    }
+    DependencyList.push_back(&Provider);
+    Provider.incrementInDegree();
+  }
+
+  void resetSelfDegree() noexcept {
+    unlinkAllStores();
+    SelfDegree.store(0, std::memory_order_release);
+    this->tryToDelete();
+  }
+
+  void tryToDelete() noexcept {
+    if (SelfDegree.load(std::memory_order_acquire) == 0 &&
+        InDegree.load(std::memory_order_acquire) == 0) {
+      delete this;
+    }
+  }
+
   /// Mutex.
   mutable std::shared_mutex Mutex;
 
@@ -633,6 +682,15 @@ protected:
   /// External data and its finalizer function pointer.
   void *HostData;
   std::function<void(void *)> HostDataFinalizer;
+
+  /// In-degree of the reference counting. Represents the number of
+  /// moduleimporting this instance.
+  std::atomic<uint32_t> InDegree{0};
+  /// Self-degree of the reference counting. 1 if this instance is held by
+  /// somebody, 0 otherwise.
+  std::atomic<uint32_t> SelfDegree{1};
+  /// Dependency list of the provider module instances.
+  std::vector<Instance::ModuleInstance *> DependencyList;
 };
 
 } // namespace Instance
--- a/lib/api/wasmedge.cpp
+++ b/lib/api/wasmedge.cpp
@@ -2665,7 +2665,10 @@ WASMEDGE_CAPI_EXPORT void WasmEdge_ModuleInstanceAddGlobal(
 
 WASMEDGE_CAPI_EXPORT void
 WasmEdge_ModuleInstanceDelete(WasmEdge_ModuleInstanceContext *Cxt) noexcept {
-  delete fromModCxt(Cxt);
+  auto *Mod = fromModCxt(Cxt);
+  if (Mod) {
+    Mod->terminate();
+  }
 }
 
 // <<<<<<<< WasmEdge module instance functions <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
@@ -3532,26 +3535,29 @@ WASMEDGE_CAPI_EXPORT void WasmEdge_VMCleanup(WasmEdge_VMContext *Cxt) noexcept {
   }
 }
 
-void WasmEdge_VMForceDeleteRegisteredModule(
-    const WasmEdge_VMContext *Cxt, const WasmEdge_String ModuleName) noexcept {
+WASMEDGE_CAPI_EXPORT void
+WasmEdge_VMDeleteRegisteredModule(const WasmEdge_VMContext *Cxt,
+                                  const WasmEdge_String ModuleName) noexcept {
   if (!Cxt || !ModuleName.Buf) {
     return; // Invalid input
   }
 
-  // Cast away const to match WasmEdge_VMGetStoreContext signature
-  WasmEdge_StoreContext *StoreCxt =
-      WasmEdge_VMGetStoreContext(const_cast<WasmEdge_VMContext *>(Cxt));
-  if (!StoreCxt) {
-    return; // Invalid store context
-  }
+  auto &VM = const_cast<WasmEdge_VMContext *>(Cxt)->VM;
+  auto &StoreMgr = VM.getStoreManager();
+  auto NameStr = genStrView(ModuleName);
 
-  const WasmEdge_ModuleInstanceContext *ModInst =
-      WasmEdge_StoreFindModule(StoreCxt, ModuleName);
-  if (ModInst) {
-    fromStoreCxt(StoreCxt)->unregisterModule(genStrView(ModuleName));
-    WasmEdge_ModuleInstanceDelete(
-        const_cast<WasmEdge_ModuleInstanceContext *>(ModInst));
+  // Unregister from VM
+  VM.unregisterModule(NameStr);
+
+  if (StoreMgr.findModule(NameStr)) {
+    // Unregister from StoreManager
+    StoreMgr.unregisterModule(NameStr);
   }
+}
+
+WASMEDGE_CAPI_EXPORT void WasmEdge_VMForceDeleteRegisteredModule(
+    const WasmEdge_VMContext *Cxt, const WasmEdge_String ModuleName) noexcept {
+  WasmEdge_VMDeleteRegisteredModule(Cxt, ModuleName);
 }
 
 WASMEDGE_CAPI_EXPORT uint32_t
--- a/lib/vm/vm.cpp
+++ b/lib/vm/vm.cpp
@@ -159,7 +159,7 @@ void VM::unsafeInitVM() {
 void VM::unsafeLoadBuiltInHosts() {
   // Load the built-in host modules from configuration.
   // TODO: This will be extended for versioned WASI in the future.
-  BuiltInModInsts.clear();
+  cleanupModInstContainer(BuiltInModInsts);
   if (Conf.hasHostRegistration(HostRegistration::Wasi)) {
     std::unique_ptr<Runtime::Instance::ModuleInstance> WasiMod =
         std::make_unique<Host::WasiModule>();
@@ -170,7 +170,7 @@ void VM::unsafeLoadBuiltInHosts() {
 void VM::unsafeLoadPlugInHosts() {
   // Load the plugins and mock them if not found.
   using namespace std::literals::string_view_literals;
-  PlugInModInsts.clear();
+  cleanupModInstContainer(PlugInModInsts);
 
   PlugInModInsts.push_back(
       createPluginModule<Host::WasiNNModuleMock>("wasi_nn"sv, "wasi_nn"sv));
@@ -344,13 +344,55 @@ VM::unsafeRegisterModule(std::string_view Name,
   return ExecutorEngine.registerModule(StoreRef, ModInst, Name);
 }
 
+Expect<void> VM::unsafeUnregisterModule(std::string_view Name) {
+  auto RegMapIt = RegModMap.find(std::string(Name));
+  if (RegMapIt != RegModMap.end()) {
+    size_t ModInstIdx = RegMapIt->second[1];
+
+    if (ModInstIdx < RegModInsts.size()) {
+      auto *ModInst = RegModInsts[ModInstIdx].release();
+      if (ModInst) {
+        ModInst->terminate();
+      }
+    }
+
+    RegModMap.erase(RegMapIt);
+
+    return {};
+  }
+  for (auto It = BuiltInModInsts.begin(); It != BuiltInModInsts.end(); ++It) {
+    if (It->second && It->second->getModuleName() == Name) {
+      auto *ModInst = It->second.release();
+      BuiltInModInsts.erase(It);
+
+      if (ModInst) {
+        ModInst->terminate();
+      }
+      return {};
+    }
+  }
+  for (auto It = PlugInModInsts.begin(); It != PlugInModInsts.end(); ++It) {
+    if (*It && (*It)->getModuleName() == Name) {
+      auto *ModInst = It->release();
+      PlugInModInsts.erase(It);
+
+      if (ModInst) {
+        ModInst->terminate();
+      }
+      return {};
+    }
+  }
+
+  return {};
+}
+
 Expect<std::vector<std::pair<ValVariant, ValType>>>
 VM::unsafeRunWasmFile(const std::filesystem::path &Path, std::string_view Func,
                       Span<const ValVariant> Params,
                       Span<const ValType> ParamTypes) {
   if (Stage == VMStage::Instantiated) {
-    // When running another module, the instantiated module in the store will be
-    // reset. Therefore the instantiation should restart.
+    // When running another module, the instantiated module in the store will
+    // be reset. Therefore the instantiation should restart.
     Stage = VMStage::Validated;
   }
   // Load wasm unit.
@@ -373,8 +415,8 @@ VM::unsafeRunWasmFile(Span<const Byte> Code, std::string_view Func,
                       Span<const ValVariant> Params,
                       Span<const ValType> ParamTypes) {
   if (Stage == VMStage::Instantiated) {
-    // When running another module, the instantiated module in the store will be
-    // reset. Therefore the instantiation should restart.
+    // When running another module, the instantiated module in the store will
+    // be reset. Therefore the instantiation should restart.
     Stage = VMStage::Validated;
   }
   // Load wasm unit.
@@ -397,8 +439,8 @@ VM::unsafeRunWasmFile(const AST::Component::Component &Component,
                       std::string_view, Span<const ValVariant>,
                       Span<const ValType>) {
   if (Stage == VMStage::Instantiated) {
-    // When running another module, the instantiated module in the store will be
-    // reset. Therefore the instantiation should restart.
+    // When running another module, the instantiated module in the store will
+    // be reset. Therefore the instantiation should restart.
     Stage = VMStage::Validated;
   }
   EXPECTED_TRY(ValidatorEngine.validate(Component));
@@ -411,8 +453,8 @@ VM::unsafeRunWasmFile(const AST::Module &Module, std::string_view Func,
                       Span<const ValVariant> Params,
                       Span<const ValType> ParamTypes) {
   if (Stage == VMStage::Instantiated) {
-    // When running another module, the instantiated module in the store will be
-    // reset. Therefore the instantiation should restart.
+    // When running another module, the instantiated module in the store will
+    // be reset. Therefore the instantiation should restart.
     Stage = VMStage::Validated;
   }
   EXPECTED_TRY(ValidatorEngine.validate(Module));
@@ -814,13 +856,16 @@ void VM::unsafeCleanup() {
     Comp.reset();
   }
   if (ActiveModInst) {
-    ActiveModInst.reset();
+    auto *RawMod = ActiveModInst.release();
+    if (RawMod) {
+      RawMod->terminate();
+    }
   }
   if (ActiveCompInst) {
     ActiveCompInst.reset();
   }
   StoreRef.reset();
-  RegModInsts.clear();
+  cleanupModInstContainer(RegModInsts);
   RegASTModules.clear();
   RegModMap.clear();
   Stat.clear();
@@ -955,10 +1000,10 @@ Expect<void> VM::lazyCompileFunctions(const std::string &ID, uint32_t FuncIdx) {
     return {};
   }
 
-  spdlog::debug(
-      "[lazy-jit]: Lazy compiling batch ({} local funcs) for wasm entry local "
-      "{}, module ID: {}"sv,
-      BatchLocals.size(), LocalFuncIdx, ID);
+  spdlog::debug("[lazy-jit]: Lazy compiling batch ({} local funcs) for wasm "
+                "entry local "
+                "{}, module ID: {}"sv,
+                BatchLocals.size(), LocalFuncIdx, ID);
 
   LLVM::Compiler Compiler(Conf);
   auto ConfigResult = Compiler.checkConfigure();
--- a/test/api/APIUnitTest.cpp
+++ b/test/api/APIUnitTest.cpp
@@ -223,6 +223,59 @@ char *Preopens[] = {&PreopensVec[0], &PreopensVec[12], &PreopensVec[21],
                     &PreopensVec[32], &PreopensVec[49]};
 char TPath[] = "apiTestData/test.wasm";
 
+/// Binary Wasm module (Provider 1):
+///
+/// (module
+///   (func $add (param i32 i32) (result i32)
+///     local.get 0
+///     local.get 1
+///     i32.add)
+///   (export "add_func" (func $add))
+/// )
+std::vector<uint8_t> Provider1Wasm = {
+    0x00, 0x61, 0x73, 0x6d, 0x01, 0x00, 0x00, 0x00, 0x01, 0x07, 0x01, 0x60,
+    0x02, 0x7f, 0x7f, 0x01, 0x7f, 0x03, 0x02, 0x01, 0x00, 0x07, 0x0c, 0x01,
+    0x08, 0x61, 0x64, 0x64, 0x5f, 0x66, 0x75, 0x6e, 0x63, 0x00, 0x00, 0x0a,
+    0x09, 0x01, 0x07, 0x00, 0x20, 0x00, 0x20, 0x01, 0x6a, 0x0b};
+
+/// Binary Wasm module (Provider 2):
+/// Adds 100 to the sum of two integers.
+///
+/// (module
+///   (func $add (param i32 i32) (result i32)
+///     local.get 0
+///     local.get 1
+///     i32.add
+///     i32.const 100
+///     i32.add)
+///   (export "add_func" (func $add))
+/// )
+std::vector<uint8_t> Provider2Wasm = {
+    0x00, 0x61, 0x73, 0x6d, 0x01, 0x00, 0x00, 0x00, 0x01, 0x07,
+    0x01, 0x60, 0x02, 0x7f, 0x7f, 0x01, 0x7f, 0x03, 0x02, 0x01,
+    0x00, 0x07, 0x0c, 0x01, 0x08, 0x61, 0x64, 0x64, 0x5f, 0x66,
+    0x75, 0x6e, 0x63, 0x00, 0x00, 0x0a, 0x0d, 0x01, 0x0b, 0x00,
+    0x20, 0x00, 0x20, 0x01, 0x6a, 0x41, 0xe4, 0x00, 0x6a, 0x0b};
+
+/// Binary Wasm module (Consumer):
+/// Imports the add function from the provider and calls it.
+///
+/// (module
+///   (import "provider" "add_func" (func $add (param i32 i32) (result i32)))
+///
+///   (func (export "call_add") (param i32 i32) (result i32)
+///    local.get 0
+///    local.get 1
+///    call $add)
+/// )
+std::vector<uint8_t> ConsumerWasm = {
+    0x00, 0x61, 0x73, 0x6d, 0x01, 0x00, 0x00, 0x00, 0x01, 0x07, 0x01, 0x60,
+    0x02, 0x7f, 0x7f, 0x01, 0x7f, 0x02, 0x15, 0x01, 0x08, 0x70, 0x72, 0x6f,
+    0x76, 0x69, 0x64, 0x65, 0x72, 0x08, 0x61, 0x64, 0x64, 0x5f, 0x66, 0x75,
+    0x6e, 0x63, 0x00, 0x00, 0x03, 0x02, 0x01, 0x00, 0x07, 0x0c, 0x01, 0x08,
+    0x63, 0x61, 0x6c, 0x6c, 0x5f, 0x61, 0x64, 0x64, 0x00, 0x01, 0x0a, 0x0a,
+    0x01, 0x08, 0x00, 0x20, 0x00, 0x20, 0x01, 0x10, 0x00, 0x0b};
+
 void hexToFile(cxx20::span<const uint8_t> Wasm, const char *Path) {
   std::ofstream TFile(std::filesystem::u8path(Path), std::ios_base::binary);
   TFile.write(reinterpret_cast<const char *>(Wasm.data()),
@@ -1305,14 +1358,13 @@ TEST(APICoreTest, Compiler) {
   EXPECT_TRUE(isErrMatch(WasmEdge_ErrCode_IllegalPath,
                          WasmEdge_CompilerCompile(Compiler, TPath, nullptr)));
   // "" must match nullptr behavior.
-  EXPECT_TRUE(isErrMatch(
-      WasmEdge_ErrCode_IllegalPath,
-      WasmEdge_CompilerCompile(Compiler, "", "not_exist_aot.wasm")));
+  EXPECT_TRUE(
+      isErrMatch(WasmEdge_ErrCode_IllegalPath,
+                 WasmEdge_CompilerCompile(Compiler, "", "not_exist_aot.wasm")));
   EXPECT_TRUE(isErrMatch(WasmEdge_ErrCode_IllegalPath,
                          WasmEdge_CompilerCompile(Compiler, TPath, "")));
-  EXPECT_TRUE(isErrMatch(
-      WasmEdge_ErrCode_IllegalPath,
-      WasmEdge_CompilerCompile(Compiler, nullptr, nullptr)));
+  EXPECT_TRUE(isErrMatch(WasmEdge_ErrCode_IllegalPath,
+                         WasmEdge_CompilerCompile(Compiler, nullptr, nullptr)));
   // Parse failed
   EXPECT_TRUE(isErrMatch(WasmEdge_ErrCode_UnexpectedEnd,
                          WasmEdge_CompilerCompile(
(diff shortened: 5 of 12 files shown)


q82419:
@Wang-Yan-Hao 
Reminder that the mid-term evaluation will be in next month.
Also, #4732 have changes about the module linking in store. Please rebase on it after this PR being merged.

Wang-Yan-Hao:
> @Wang-Yan-Hao Reminder that the mid-term evaluation will be in next month. Also, #4732 have changes about the module linking in store. Please rebase on it after this PR being merged.

Understood. I will keep an eye on #4732 and rebase after it's merged. Also, I'm aware of the mid-term evaluation.

include/api/wasmedge/wasmedge_vm.h
```diff
@@ -603,18 +603,19 @@ WasmEdge_VMGetFunctionList(const WasmEdge_VMContext *Cxt,
...
 ///
-/// A safer deletion API may be introduced in the future once module dependency
-/// management is implemented.
+/// This mechanism ensures memory safety and prevents undefined behavior during
+/// module lifecycle transitions.
 ///
 /// \param Cxt the WasmEdge_VMContext to delete the module from.
 /// \param ModuleName the name of the module to delete.
-WASMEDGE_CAPI_EXPORT extern void WasmEdge_VMForceDeleteRegisteredModule(
+WASMEDGE_CAPI_EXPORT extern void WasmEdge_VMDeleteRegisteredModule(
```
q82419, line 614: Don't just delete the API.
You should move the old API to the deprecated header and make sure that still works.

lib/api/wasmedge.cpp
```diff
@@ -3508,7 +3508,7 @@ WASMEDGE_CAPI_EXPORT void WasmEdge_VMCleanup(WasmEdge_VMContext *Cxt) noexcept {
   }
 }
 
-void WasmEdge_VMForceDeleteRegisteredModule(
+void WasmEdge_VMDeleteRegisteredModule(
```
q82419: As mentioned. And redirect the old API to the new one.

lib/vm/vm.cpp
```diff
@@ -235,6 +235,21 @@ VM::unsafeRegisterModule(const Runtime::Instance::ModuleInstance &ModInst,
   return ExecutorEngine.registerModule(StoreRef, ModInst, Name);
 }
 
+Expect<std::unique_ptr<Runtime::Instance::ModuleInstance>>
+VM::unsafeUnregisterModule(std::string_view Name) {
```
q82419: When unregister module, I don't think we need to return the module instance to caller.

lib/vm/vm.cpp
```diff
@@ -309,6 +324,7 @@ VM::unsafeRunWasmFile(const AST::Module &Module, std::string_view Func,
   EXPECTED_TRY(ValidatorEngine.validate(Module));
   EXPECTED_TRY(ActiveModInst,
                ExecutorEngine.instantiateModule(StoreRef, Module));
+  StoreRef.linkAnonymousModule(ActiveModInst.get());
```
q82419: Link the anonymous module to store is not necessary.

lib/vm/vm.cpp
```diff
@@ -486,6 +502,7 @@ Expect<void> VM::unsafeInstantiate() {
 
     EXPECTED_TRY(ActiveModInst,
                  ExecutorEngine.instantiateModule(StoreRef, *Mod));
+    StoreRef.linkAnonymousModule(ActiveModInst.get());
```
q82419: Ditto.

q82419 (changes requested):
First, please check the sign-off and lint of your commits.

As our discussion, I realized that you explained the implementation for degree (or ref count) mechanism is in the module instance. But the code doesn't fit your description.
According to your [issue comment](https://github.com/WasmEdge/WasmEdge/issues/4692#issuecomment-4204101626), you've recognized that the store is only for module searching and linking when instantiation, and the store only "notices" the module instances for the new importing from the other modules. Therefore it's wierd that the `DependencyTree` is put in store.
You can make the full picture more simply and easy -- maintain the in-degree and out-degree only in the module instances as you originally mentioned, but not relate to the linking system currently in store.

Linking is linking in store -- this is only notification for modules and stores to record the linking state. When a module is in deletion state, even though the physically deletion not happend due to still has dependency, the unlinking to all stores occurs. Also, when a store is deleted, the unlinking to all modules occurs.

Degree is degree -- this IS the main feature in this project. The module instances have their in-degrees when being imported by other modules, and out-degrees for their dependency modules. You can seem they have self-degree for the ownership. When the deletion API being called (i.e. the self-degree becoming 0), unlinking to stores occurs. The physically deletion occurs when in-degree + self-degree becoming 0. In this case, notice other modules in out-degree to remove their in-degree from this module.

Maybe above is the algorithm you imagined.

Wang-Yan-Hao:
Sorry for the confusion earlier. I mistakenly thought the dependency tree should be implemented in `StoreManager`. I have now refactored the logic and moved the dependency tracking to the `ModuleInstance` level as discussed. I'm currently performing additional tests to ensure the implementation is solid.

Regarding the API changes:

1. I have moved the old `WasmEdge_VMForceDeleteRegisteredModule` (the previous placeholder) to `wasmedge_deprecated.h`.
2. The new `WasmEdge_VMDeleteRegisteredModule` is now implemented in `wasmedge_vm.h` with the degree-based lifecycle management.
3. I've kept `WasmEdge_VMForceDeleteRegisteredModule` as a wrapper in `wasmedge.cpp` to maintain backward compatibility.

All commits have been signed off with `commit -s` (DCO). Thanks for your patience!

include/api/wasmedge/wasmedge_vm.h
```diff
@@ -603,18 +603,18 @@ WasmEdge_VMGetFunctionList(const WasmEdge_VMContext *Cxt,
...
+/// Unregister and trigger the deletion of a registered module from the VM
+/// context.
 ///
-/// A safer deletion API may be introduced in the future once module dependency
-/// management is implemented.
+/// This function initiates the lifecycle management for the target module.
+/// With the implemented dependency tracking, the module is not necessarily
+/// destroyed immediately. It will be unlinked from the VM, but physical
+/// deallocation only occurs when its "Self-Degree" is reset and its
+/// "In-Degree" (references from other modules) reaches zero.
```
q82419: In the explaination of header, it's not necessary to describe the degree system about the implmentation. You only need to explain that the module would be safely deleted when there's no dependency on it.

lib/vm/vm.cpp
```diff
@@ -235,6 +235,18 @@ VM::unsafeRegisterModule(const Runtime::Instance::ModuleInstance &ModInst,
   return ExecutorEngine.registerModule(StoreRef, ModInst, Name);
 }
 
+Expect<void> VM::unsafeUnregisterModule(std::string_view Name) {
+  auto It = std::find_if(
+      RegModInsts.begin(), RegModInsts.end(),
+      [&](const auto &Mod) { return Mod->getModuleName() == Name; });
+
+  if (It != RegModInsts.end()) {
+    It->release();
```
q82419: Should notify the modules to reset self-degree.
Also should consider the case about the imported modules not owned by this VM.

q82419 (changes requested):
I would review the naming convensions and code architecture after fixing for the correctness of deletion logic.

q82419:
@Wang-Yan-Hao 
It's a reminder that today is the deadline of midterm evolution.
Please help to update your progress of requested changes. Thanks!

Wang-Yan-Hao:
I have updated the PR to address your feedback. The main changes include:

1. Header Documentation: Simplified the comments in `wasmedge_vm.h` to focus on the high-level safety of the deletion API, removing internal implementation details about degrees.
2. Ownership-Aware Unregistration: Re-implemented the unregistration logic in `VM::unsafeUnregisterModule`.

Regarding the unregistration flow:

1. Store Level (`Store::unregisterModule`): This acts as a "book record." It simply unlinks the `ModuleName` from the `ModuleInstance` mapping. This happens regardless of which VM owns the module, ensuring that no new modules can import this specific instance by name.
2. VM Level (`VM::unsafeUnregisterModule`).

Wang-Yan-Hao:
@q82419, Thank you for the reminder regarding the midterm evaluation. I am fully aware of the deadline and have been working intensively on the requested changes.

Regarding the midterm evaluation, besides pushing the code updates, is there anything else I should prepare (e.g., a formal report)? Also, I noticed that the Midterm Evaluation Form is not appearing on my LFX dashboard, and I haven't received the related email yet. Do I need to fill some form?

q82419:
> @q82419, Thank you for the reminder regarding the midterm evaluation. I am fully aware of the deadline and have been working intensively on the requested changes.
> 
> Regarding the midterm evaluation, besides pushing the code updates, is there anything else I should prepare (e.g., a formal report)? Also, I noticed that the Midterm Evaluation Form is not appearing on my LFX dashboard, and I haven't received the related email yet. Do I need to fill some form?

No, the mid-term evaluation is for mentors to evaluate the progress of mentorship programs.
If fail, the program will be interrupted and closed.
In WasmEdge project, we expect the main features would be completed before the mid-term, and the tests and docs for verification for final evaluation.

q82419 (changes requested):
For your code, you should plan carefully for the deletion routine of modules, with considering the VM and store life cycle, because the module instances have its own ownership other than VMs and stores.

Think about the following questions.

1. We made the delayed deletion now for keeping the module instances with handling the dependencies. It's correct, but it's easier for causing leak or segfault in current code base. You can use the custom deleter for unique pointer template to refine the architecture.
2. Besides the VM, the deletion of owned module instances from API users should also consider the memory management.
3. You still didn't handle the module unregistration when VM being deleted. Mentioned in discussion, although VM owns the module instances, it's still possible for users to get the module instance by name and register into other stores or VMs. This causes complex dependencies.
4. When unregistration, you didn't consider the imported modules for unlinking. Not every modules are owned by this VM.
5. You can discover that in store, we mainly use the given name for linking rather than the name in the module instance. Please extend the `RegModInsts` to the map with name key.

Wang-Yan-Hao:
Thanks for the suggestions on using a custom deleter for ModuleInstance (Points 1 & 2). I have successfully implemented the SelfDegree and InDegree logic within a nested Deleter struct. Regarding the architecture refinement, I have a question about the scope: Should I propagate the custom deleter type to all `std::unique_ptr<ModuleInstance>` across the codebase (e.g., `Executor::registerModule` and other C++ APIs)?

q82419:
> Thanks for the suggestions on using a custom deleter for ModuleInstance (Points 1 & 2). I have successfully implemented the SelfDegree and InDegree logic within a nested Deleter struct. Regarding the architecture refinement, I have a question about the scope: Should I propagate the custom deleter type to all `std::unique_ptr<ModuleInstance>` across the codebase (e.g., `Executor::registerModule` and other C++ APIs)?

Sure, it's the better pattern to prevent from issues of memory management.
Noticed that the "deleter" is not the physical deletion, but the self-degree reset. The physical deletion occurs when the in-degree and self-degree become 0.

q82419:
On the other hand, I would suggest you to reconstruct your system design again and describe your algorithm entirely with the cases we discussed in the issue, because your implementation still weird and lack of the situations I mentioned.
Since you did your work in the first half term if this mentorship, that meets parts goal of the mid-term evaluation.
But I would ask you to refine and explain your algorithm/mechanism again -- whether in description, graph, whatever -- with the situations about VM, non-VM, multiple modules and stores registration, etc. Please summarize the cases in discussions we mentioned and raise your report in issue. I think this can help you to have a fully and better picture about the system design.
Please finish this before 4/28, and this will be the part of final evaluation. Thanks!

Wang-Yan-Hao:
I have pushed a partial fix commit to make sure my current thought process is on the right track.

* **About point 1:** I have introduced a custom `Deleter` for the module instances. It first resets the `Self-Degree` and then checks whether both the `In-Degree` and `Self-Degree` are actually zero. If they are, it triggers the actual physical deletion (`delete this`).
* **About point 5:** I changed `RegModInsts` to use the structure `std::unordered_map<std::string, std::unique_ptr<Runtime::Instance::ModuleInstance, Runtime::Instance::ModuleInstance::Deleter>>`. Using the name as the key aligns with WasmEdge's new design (where one module can have multiple names), and the custom `Deleter` ensures the lifecycle is properly managed.
* **About point 4:** As shown in the commit, the VM's `unsafeUnregisterModule()` function no longer just deletes modules from `RegModInsts`. It now also scans the `BuiltInModInsts` and `PlugInModInsts` members. Since these three members now all use `unique_ptr` with the custom `Deleter`, their lifecycles will be correctly monitored by our dependency tree mechanism when they are erased.
* **About point 2:** I believe `WasmEdge_ModuleInstanceDelete` is the only C API related to this manual deletion. Because that API passes a raw C pointer, I updated it to explicitly use my custom `Deleter` to maintain consistency.
* **About point 3 (Complex Dependencies):** I want to confirm my understanding of the scenario you mentioned. Suppose `VM_1` owns `Module_1` (via `RegModInsts`), and `VM_2` imports `Module_1`. Both `StoreMgr_1` and `StoreMgr_2` will have linking records of `Module_1`. If `VM_1` is deleted, its `StoreMgr_1` is destroyed, but `Module_1` remains alive because it is still used by `VM_2`. The key realization here is that when `VM_1` is deleted, `Module_1`'s `Self-Degree` becomes 0, and at this exact moment, it **must** unlink from ALL StoreManagers (including `StoreMgr_2`), even if physical deletion hasn't occurred yet.
    Therefore, I implemented the unlinking logic inside `resetSelfDegree()`. This ensures that whenever a module enters the deletion state (`Self-Degree` set to 0), it actively unlinks from all stores to prevent dangling pointers.

Hope the above implementation aligns with your expectations. If this logic is correct, I will proceed to apply this custom `Deleter` to all `std::unique_ptr`s of the module instances ASAP.

Wang-Yan-Hao:
> On the other hand, I would suggest you to reconstruct your system design again and describe your algorithm entirely with the cases we discussed in the issue, because your implementation still weird and lack of the situations I mentioned. Since you did your work in the first half term if this mentorship, that meets parts goal of the mid-term evaluation. But I would ask you to refine and explain your algorithm/mechanism again -- whether in description, graph, whatever -- with the situations about VM, non-VM, multiple modules and stores registration, etc. Please summarize the cases in discussions we mentioned and raise your report in issue. I think this can help you to have a fully and better picture about the system design. Please finish this before 4/28, and this will be the part of final evaluation. Thanks!

Thanks for the feedback! Sure, I will take a step back and reconstruct the system design as you suggested. I completely agree that writing out all the specific cases we've discussed (VM, non-VM, multiple modules/stores registration, etc.) is the best way to ensure my design and algorithm are perfectly aligned with the architecture.

q82419:
> I have pushed a partial fix commit to make sure my current thought process is on the right track.
> 
> * **About point 1:** I have introduced a custom `Deleter` for the module instances. It first resets the `Self-Degree` and then checks whether both the `In-Degree` and `Self-Degree` are actually zero. If they are, it triggers the actual physical deletion (`delete this`).
> * **About point 5:** I changed `RegModInsts` to use the structure `std::unordered_map<std::string, std::unique_ptr<Runtime::Instance::ModuleInstance, Runtime::Instance::ModuleInstance::Deleter>>`. Using the name as the key aligns with WasmEdge's new design (where one module can have multiple names), and the custom `Deleter` ensures the lifecycle is properly managed.
> * **About point 4:** As shown in the commit, the VM's `unsafeUnregisterModule()` function no longer just deletes modules from `RegModInsts`. It now also scans the `BuiltInModInsts` and `PlugInModInsts` members. Since these three members now all use `unique_ptr` with the custom `Deleter`, their lifecycles will be correctly monitored by our dependency tree mechanism when they are erased.
> * **About point 2:** I believe `WasmEdge_ModuleInstanceDelete` is the only C API related to this manual deletion. Because that API passes a raw C pointer, I updated it to explicitly use my custom `Deleter` to maintain consistency.
> * **About point 3 (Complex Dependencies):** I want to confirm my understanding of the scenario you mentioned. Suppose `VM_1` owns `Module_1` (via `RegModInsts`), and `VM_2` imports `Module_1`. Both `StoreMgr_1` and `StoreMgr_2` will have linking records of `Module_1`. If `VM_1` is deleted, its `StoreMgr_1` is destroyed, but `Module_1` remains alive because it is still used by `VM_2`. The key realization here is that when `VM_1` is deleted, `Module_1`'s `Self-Degree` becomes 0, and at this exact moment, it **must** unlink from ALL StoreManagers (including `StoreMgr_2`), even if physical deletion hasn't occurred yet.
>   Therefore, I implemented the unlinking logic inside `resetSelfDegree()`. This ensures that whenever a module enters the deletion state (`Self-Degree` set to 0), it actively unlinks from all stores to prevent dangling pointers.
> 
> Hope the above implementation aligns with your expectations. If this logic is correct, I will proceed to apply this custom `Deleter` to all `std::unique_ptr`s of the module instances ASAP.

These seems more make sense than you implemented before.
As I mentioned, instead of answering through the listed questions, please summary for the mechanism with each partitions, such as what would happen when registering/unregistering/deletion, etc. in module instances, VMs, stores.
It's too fragmented for only just answering these questions.

Wang-Yan-Hao:
> 
> These seems more make sense than you implemented before. As I mentioned, instead of answering through the listed questions, please summary for the mechanism with each partitions, such as what would happen when registering/unregistering/deletion, etc. in module instances, VMs, stores. It's too fragmented for only just answering these questions.

I have mention different scenarios and core algorithm in the issue #4692.

include/runtime/instance/module.h
```diff
@@ -76,15 +76,20 @@ class ModuleInstance {
                  std::function<void(void *)> Finalizer = nullptr)
       : ModName(Name), HostData(Data), HostDataFinalizer(Finalizer) {}
   virtual ~ModuleInstance() noexcept {
-    // When destroying this module instance, call the callbacks to unlink to the
-    // store managers.
-    for (auto &&[Key, Callback] : LinkedStore) {
-      assuming(Callback);
-      Callback(Key, this);
-    }
+    unlinkAllStores();
```
q82419, line 84: Since already moved the store linking procedure to the timing of reseting the self-degree, I don't think the unlinking is needed here.

include/runtime/instance/module.h
```diff
@@ -76,15 +76,20 @@ class ModuleInstance {
...
+    for (auto *Provider : DependencyList) {
+      if (Provider) {
+        Provider->decrementInDegree();
+      }
+    }
+  }
+
+  void terminate() noexcept {
+    this->resetSelfDegree();
+    this->tryToDelete();
```
q82419: Why don't you call `tryToDelete` in `resetSelfDegree` and `decrementInDegree`?

q82419 (changes requested):
The `APIVMCoreTest` is mainly for spec tests.
Please move the tests into a test scope of `TEST(APICoreTest, VMDeleteRegistered)` and `TEST(APICoreTest, ModuleDeletion)`.

include/runtime/instance/module.h
Wang-Yan-Hao: As you mention previously, some instances are allocated on the stack and do not go through the `resetSelfDegree` process. For these modules, the destructor is triggered automatically upon returning. To ensure that even these stack-based instances safely detach from all Stores, I think it's safer to keep the unlinking logic within the destructor as a fallback.

include/runtime/instance/module.h
Wang-Yan-Hao: I will update the implementation. Now `resetSelfDegree()` and `decrementInDegree()` are responsible for calling `tryToDelete()`. As for `terminate()`, it simply calls `resetSelfDegree()` to trigger the automated deletion flow.

Wang-Yan-Hao:
> The `APIVMCoreTest` is mainly for spec tests. Please move the tests into a test scope of `TEST(APICoreTest, VMDeleteRegistered)` and `TEST(APICoreTest, ModuleDeletion)`.

Regarding the refactoring of the test scope, I have a quick clarification:

I noticed that some existing tests like `TEST(APICoreTest, VM)` are quite consolidated and long. When you mentioned moving tests into `VMDeleteRegistered` and `ModuleDeletion`, did you mean I should:

1. Consolidate all related logic (e.g., Cross-VM, Shared Store, Host-owned) into these two specific `TEST` functions with comments to separate cases?
2. Or use them as naming prefixes (e.g., `ModuleDeletion_SharedStore`, `ModuleDeletion_CrossVM`) to keep the test cases granular?

Currently, I have multiple scenarios for dependency trees, and putting them into a single `TEST` might result in a very large function. What would be your preferred style for the WasmEdge codebase? Thanks

q82419:
> > The `APIVMCoreTest` is mainly for spec tests. Please move the tests into a test scope of `TEST(APICoreTest, VMDeleteRegistered)` and `TEST(APICoreTest, ModuleDeletion)`.
> 
> Regarding the refactoring of the test scope, I have a quick clarification:
> 
> I noticed that some existing tests like `TEST(APICoreTest, VM)` are quite consolidated and long. When you mentioned moving tests into `VMDeleteRegistered` and `ModuleDeletion`, did you mean I should:
> 
> 1. Consolidate all related logic (e.g., Cross-VM, Shared Store, Host-owned) into these two specific `TEST` functions with comments to separate cases?
> 2. Or use them as naming prefixes (e.g., `ModuleDeletion_SharedStore`, `ModuleDeletion_CrossVM`) to keep the test cases granular?
> 
> Currently, I have multiple scenarios for dependency trees, and putting them into a single `TEST` might result in a very large function. What would be your preferred style for the WasmEdge codebase? Thanks

I think the first one (consolidate all related logic and comments to separate cases) is okey, even if a single TEST would be very large. 

You can take this for example (but this is a minor scope example): https://github.com/WasmEdge/WasmEdge/blob/master/test/loader/segmentTest.cpp#L36
First list the explaination of each cases, and comments to separate cases.
I know there may be several test styles in WasmEdge, but this way is more convenient for the future refactoring after this mentorship program.

include/runtime/instance/module.h
q82419: Make sense. Please add some comments for why unlink stores here.

include/runtime/instance/module.h
```diff
@@ -76,17 +76,19 @@ class ModuleInstance {
...
       HostDataFinalizer(HostData);
     }
+    for (auto *Provider : DependencyList) {
+      if (Provider) {
+        Provider->decrementInDegree();
+      }
+    }
   }
 
+  void terminate() noexcept { this->resetSelfDegree(); }
```
q82419: Remove `this->`.

include/runtime/instance/module.h
```diff
@@ -530,6 +532,53 @@ class ModuleInstance {
     LinkedStore.erase(LinkedStoreKey{Store, std::string(Name)});
   }
 
+  void unlinkAllStores() noexcept {
+    if (LinkedStore.empty()) {
```
q82419: Is it redundant?
If the map is empty, I think the for loop below would do nothing.

q82419 (changes requested):

include/runtime/instance/module.h
Wang-Yan-Hao: Fix it.

include/runtime/instance/module.h
Wang-Yan-Hao: Fix it.

Wang-Yan-Hao:
> I think the first one (consolidate all related logic and comments to separate cases) is okey, even if a single TEST would be very large.
> 
> You can take this for example (but this is a minor scope example): https://github.com/WasmEdge/WasmEdge/blob/master/test/loader/segmentTest.cpp#L36 First list the explaination of each cases, and comments to separate cases. I know there may be several test styles in WasmEdge, but this way is more convenient for the future refactoring after this mentorship program.

I have refactored and expanded the unit tests to provide a more comprehensive validation of the Module Instance Dependency Tree logic. To improve clarity and maintainability, I’ve categorized the tests into two distinct suites based on their trigger mechanisms:

### APICoreTest.VMDeleteRegistered (API-Triggered Unlinking)

This suite focuses on the explicit use of the WasmEdge_VMDeleteRegisteredModule API. It verifies that logical unlinking from the store registry behaves correctly while maintaining the physical instance's availability (deferred deletion).

- Case 1: Verifies unlinking a provider with active registered consumer dependencies.
- Case 2: Verifies that anonymous (active) modules correctly link to the latest provider after an old one is unlinked.
- Case 3: Tests cross-VM dependency stability when a provider is unlinked from a shared store.
- Case 4: Ensures that a "borrower" VM can unlink a module from the shared store without affecting other VMs holding the instance.
- Case 5: Verifies unlinking a host-owned module from one VM's registry while others still use it.
- Case 6: Tests unlinking a host-owned module from a shared registry environment.

### APICoreTest.ModuleDeletion (Lifecycle & Ownership Management)

This suite focuses on resource destruction and ownership transitions. It ensures the dependency tree correctly manages reference counts (in-degree/self-degree) when the containers themselves (VM/Store) or the host owner are released.

- Case 1: Verifies that destroying the originating VM of a provider doesn't crash consumers in other VMs (Cross-VM persistence).
- Case 2: Ensures module instances remain executable via direct handles even after their parent StoreContext is destroyed.
- Case 3: Tests that destroying a borrower VM correctly decrements the provider's in-degree while keeping the module alive for others.
- Case 4: Verifies that calling WasmEdge_ModuleInstanceDelete (releasing self-degree) triggers deferred deletion across multiple independent VMs.
- Case 5: Tests host ownership release within a shared store environment, ensuring global registry unlinking.

These updates ensure that the implementation handles both explicit API calls and implicit resource cleanup accurately. Looking forward to your feedback!

q82419:
> > I think the first one (consolidate all related logic and comments to separate cases) is okey, even if a single TEST would be very large.
> > You can take this for example (but this is a minor scope example): https://github.com/WasmEdge/WasmEdge/blob/master/test/loader/segmentTest.cpp#L36 First list the explaination of each cases, and comments to separate cases. I know there may be several test styles in WasmEdge, but this way is more convenient for the future refactoring after this mentorship program.
> 
> I have refactored and expanded the unit tests to provide a more comprehensive validation of the Module Instance Dependency Tree logic. To improve clarity and maintainability, I’ve categorized the tests into two distinct suites based on their trigger mechanisms:
> 
> ### APICoreTest.VMDeleteRegistered (API-Triggered Unlinking)
> This suite focuses on the explicit use of the WasmEdge_VMDeleteRegisteredModule API. It verifies that logical unlinking from the store registry behaves correctly while maintaining the physical instance's availability (deferred deletion).
> 
> * Case 1: Verifies unlinking a provider with active registered consumer dependencies.
> * Case 2: Verifies that anonymous (active) modules correctly link to the latest provider after an old one is unlinked.
> * Case 3: Tests cross-VM dependency stability when a provider is unlinked from a shared store.
> * Case 4: Ensures that a "borrower" VM can unlink a module from the shared store without affecting other VMs holding the instance.
> * Case 5: Verifies unlinking a host-owned module from one VM's registry while others still use it.
> * Case 6: Tests unlinking a host-owned module from a shared registry environment.
> 
> ### APICoreTest.ModuleDeletion (Lifecycle & Ownership Management)
> This suite focuses on resource destruction and ownership transitions. It ensures the dependency tree correctly manages reference counts (in-degree/self-degree) when the containers themselves (VM/Store) or the host owner are released.
> 
> * Case 1: Verifies that destroying the originating VM of a provider doesn't crash consumers in other VMs (Cross-VM persistence).
> * Case 2: Ensures module instances remain executable via direct handles even after their parent StoreContext is destroyed.
> * Case 3: Tests that destroying a borrower VM correctly decrements the provider's in-degree while keeping the module alive for others.
> * Case 4: Verifies that calling WasmEdge_ModuleInstanceDelete (releasing self-degree) triggers deferred deletion across multiple independent VMs.
> * Case 5: Tests host ownership release within a shared store environment, ensuring global registry unlinking.
> 
> These updates ensure that the implementation handles both explicit API calls and implicit resource cleanup accurately. Looking forward to your feedback!

I'll review for your test cases later.
But first, please list your cases on the begining of the test scope, so that the reader can realize what would be the goal in this test scope.
And, please squash your commits into 2: implementation and tests, and rebase onto the current master.
Remember to fix the spell reported by the misc linter, and add the new API to the symbol exposure list.
Thanks!

test/api/APIUnitTest.cpp
```diff
@@ -223,6 +223,62 @@ char *Preopens[] = {&PreopensVec[0], &PreopensVec[12], &PreopensVec[21],
...
+/// Binary Wasm module (Provider 1):
+///
+/// (module
+///   (func $add (param i32 i32) (result i32)
+///     local.get 0
+///     local.get 1
+///     i32.add)
+///   (export "add_func" (func $add))
+/// )
+unsigned char provider_1_wasm[] = {
```
q82419: Follow the format of defining WASM binary above (like `FibonacciWasm`).

test/api/APIUnitTest.cpp
```diff
@@ -3911,6 +3967,594 @@ TEST(APICoreTest, VM) {
   WasmEdge_VMDelete(VM);
 }
 
+TEST(APICoreTest, VMDeleteRegistered) {
+  WasmEdge_ConfigureContext *Conf = WasmEdge_ConfigureCreate();
```
q82419: For the VMs/Loaders/etc. you didn't modify the configuration, the `Conf` is not needed.

test/api/APIUnitTest.cpp
```diff
@@ -3911,6 +3967,594 @@ TEST(APICoreTest, VM) {
...
+                                    WasmEdge_ValTypeGenI32()};
+  WasmEdge_ValType RetTypes[1] = {WasmEdge_ValTypeGenI32()};
+  WasmEdge_FunctionTypeContext *FType =
+      WasmEdge_FunctionTypeCreate(ParamTypes, 2, RetTypes, 1);
+
+  auto DummyAdd =
+      [](void *Data, const WasmEdge_CallingFrameContext *CallFrameCxt,
+         const WasmEdge_Value *In, WasmEdge_Value *Out) -> WasmEdge_Result {
+    (void)Data;
+    (void)CallFrameCxt;
```
q82419: Use `[](void *, const WasmEdge_CallingFrameContext *, ...`

test/api/APIUnitTest.cpp
q82419: And fix the naming style.

test/api/APIUnitTest.cpp
```diff
@@ -3911,6 +3967,594 @@ TEST(APICoreTest, VM) {
   WasmEdge_VMDelete(VM);
 }
 
+TEST(APICoreTest, VMDeleteRegistered) {
+  WasmEdge_ConfigureContext *Conf = WasmEdge_ConfigureCreate();
+  WasmEdge_ExecutorContext *Exec = WasmEdge_ExecutorCreate(Conf, nullptr);
+
+  WasmEdge_Result Res;
+  uint32_t originalCount = 0;
```
q82419: fix the naming style.

test/api/APIUnitTest.cpp
```diff
@@ -3911,6 +3967,594 @@ TEST(APICoreTest, VM) {
...
+
+  WasmEdge_String PName = WasmEdge_StringCreateByCString("provider");
+  WasmEdge_String CName1 = WasmEdge_StringCreateByCString("consumer_1");
+  WasmEdge_String CName2 = WasmEdge_StringCreateByCString("consumer_2");
+  WasmEdge_String FuncName = WasmEdge_StringCreateByCString("call_add");
+  WasmEdge_String PFunc = WasmEdge_StringCreateByCString("add_func");
```
q82419: `PFuncName` is better.

test/api/APIUnitTest.cpp
```diff
@@ -3911,6 +3967,594 @@ TEST(APICoreTest, VM) {
...
+    EXPECT_EQ(WasmEdge_VMListRegisteredModuleLength(VMCxt), originalCount - 1);
+
+    // Consumer 1 should still work (Result 30)
+    Res = WasmEdge_VMExecute(VMCxt, FuncName, Params, 2, Returns, 1);
+    EXPECT_TRUE(WasmEdge_ResultOK(Res));
+    EXPECT_EQ(WasmEdge_ValueGetI32(Returns[0]), 30);
+
+    // Register Provider 2 with the same name
+    Res = WasmEdge_VMRegisterModuleFromBuffer(VMCxt, PName, provider_2_wasm,
+
```
q82419: Remove the redundant line

test/api/APIUnitTest.cpp
```diff
@@ -3911,6 +3967,594 @@ TEST(APICoreTest, VM) {
...
+    // Unregister Provder 2
+    WasmEdge_VMDeleteRegisteredModule(VMCxt, PName);
+    EXPECT_EQ(WasmEdge_VMListRegisteredModuleLength(VMCxt), originalCount);
+
+    // Test Consumer 2 uses Provider 2 (Result 130)
+    Res = WasmEdge_VMExecuteRegistered(VMCxt, CName2, FuncName, Params, 2,
+                                       Returns, 1);
+    EXPECT_TRUE(WasmEdge_ResultOK(Res));
+    EXPECT_EQ(WasmEdge_ValueGetI32(Returns[0]), 130);
+  }
```
q82419, line 4074: You should call VMDelete before the end of this scope.

test/api/APIUnitTest.cpp
```diff
@@ -3911,6 +3967,594 @@ TEST(APICoreTest, VM) {
...
+    // Unregister Provider 2 (It becomes Zombie)
+    WasmEdge_VMDeleteRegisteredModule(VMCxt, PName);
+    EXPECT_EQ(WasmEdge_StoreFindModule(StoreCxt, PName), nullptr);
+    EXPECT_EQ(WasmEdge_VMListRegisteredModuleLength(VMCxt), originalCount - 1);
+
+    // Consumer 2 should still work (Result 130)
+    Res = WasmEdge_VMExecute(VMCxt, FuncName, Params, 2, Returns, 1);
+    EXPECT_TRUE(WasmEdge_ResultOK(Res));
+    EXPECT_EQ(WasmEdge_ValueGetI32(Returns[0]), 130);
+  }
```
q82419, line 4140: You should call VMDelete before the end of this scope.

test/api/APIUnitTest.cpp
```diff
@@ -3911,6 +3967,594 @@ TEST(APICoreTest, VM) {
...
+    // Unregister Provider 1 (It become Zombie)
+    WasmEdge_VMDeleteRegisteredModule(VM1, PName);
+    EXPECT_EQ(WasmEdge_StoreFindModule(SharedStore, PName), nullptr);
+
+    // We can still call the zombie provider from Consumer 1 (Result 30)
+    Res = WasmEdge_VMExecuteRegistered(VM2, CName1, FuncName, Params, 2,
+                                       Returns, 1);
+    EXPECT_TRUE(WasmEdge_ResultOK(Res));
+    EXPECT_EQ(WasmEdge_ValueGetI32(Returns[0]), 30);
+  }
```
q82419, line 4178: You should call VMDelete and StoreDelete before the end of this scope.

test/api/APIUnitTest.cpp
```diff
@@ -3911,6 +3967,594 @@ TEST(APICoreTest, VM) {
...
+    // VM1
+    EXPECT_EQ(WasmEdge_StoreFindModule(SharedStore, PName), nullptr);
+    EXPECT_EQ(WasmEdge_VMListRegisteredModuleLength(VM1), originalCount);
+    EXPECT_EQ(WasmEdge_VMListRegisteredModuleLength(VM2), originalCount);
+
+    // We still can execute functions from the module in VM1
+    Res = WasmEdge_ExecutorInvoke(Exec, FuncInst, Params, 2, Returns, 1);
+    EXPECT_TRUE(WasmEdge_ResultOK(Res));
+    EXPECT_EQ(WasmEdge_ValueGetI32(Returns[0]), 30);
+  }
```
q82419, line 4226: You should call VMDelete and StoreDelete before the end of this scope.

test/api/APIUnitTest.cpp
```diff
@@ -3911,6 +3967,594 @@ TEST(APICoreTest, VM) {
...
+    EXPECT_EQ(WasmEdge_StoreFindModule(Store1, PName), nullptr);
+    // We still can find module in VM2
+    EXPECT_NE(WasmEdge_StoreFindModule(Store2, PName), nullptr);
+
+    // We still can execute functions from the host provider module in VM1
+    Res = WasmEdge_VMExecuteRegistered(VM1, CName1, FuncName, Params, 2,
+                                       Returns, 1);
+    EXPECT_TRUE(WasmEdge_ResultOK(Res));
+    EXPECT_EQ(WasmEdge_ValueGetI32(Returns[0]), 30);
+  }
```
q82419, line 4281: You should call VMDelete and StoreDelete before the end of this scope.

test/api/APIUnitTest.cpp
```diff
@@ -3911,6 +3967,594 @@ TEST(APICoreTest, VM) {
...
+                                       Returns, 1);
+    EXPECT_TRUE(WasmEdge_ResultOK(Res));
+    EXPECT_EQ(WasmEdge_ValueGetI32(Returns[0]), 30);
+
+    // Execution in VM2 via consumer_2 should also still work
+    Res = WasmEdge_VMExecuteRegistered(VM2, CName2, FuncName, Params, 2,
+                                       Returns, 1);
+    EXPECT_TRUE(WasmEdge_ResultOK(Res));
+    EXPECT_EQ(WasmEdge_ValueGetI32(Returns[0]), 30);
+  }
```
q82419, line 4424: You should call VMDelete and StoreDelete before the end of this scope.

test/api/APIUnitTest.cpp
```diff
@@ -3911,6 +3967,594 @@ TEST(APICoreTest, VM) {
...
+      [](void *Data, const WasmEdge_CallingFrameContext *CallFrameCxt,
+         const WasmEdge_Value *In, WasmEdge_Value *Out) -> WasmEdge_Result {
+    (void)Data;
+    (void)CallFrameCxt;
+    Out[0] = WasmEdge_ValueGenI32(WasmEdge_ValueGetI32(In[0]) +
+                                  WasmEdge_ValueGetI32(In[1]));
+    return WasmEdge_Result_Success;
+  };
+
+  WasmEdge_FunctionInstanceContext *HostFunc =
```
q82419: Host function allocation should not be here.
Please refer to the explaination of `WasmEdge_ModuleInstanceAddFunction` API. This API will move the ownership of host function object to the callee.

test/api/APIUnitTest.cpp
```diff
@@ -3911,6 +3967,594 @@ TEST(APICoreTest, VM) {
...
```
q82419: Host function allocation should not be here.
Please refer to the explaination of WasmEdge_ModuleInstanceAddFunction API. This API will move the ownership of host function object to the callee.

test/api/APIUnitTest.cpp
```diff
@@ -3911,6 +3967,594 @@ TEST(APICoreTest, VM) {
...
+                                    WasmEdge_ValTypeGenI32()};
+  WasmEdge_ValType RetTypes[1] = {WasmEdge_ValTypeGenI32()};
+  WasmEdge_FunctionTypeContext *FType =
+      WasmEdge_FunctionTypeCreate(ParamTypes, 2, RetTypes, 1);
+
+  auto DummyAdd =
+      [](void *Data, const WasmEdge_CallingFrameContext *CallFrameCxt,
+         const WasmEdge_Value *In, WasmEdge_Value *Out) -> WasmEdge_Result {
+    (void)Data;
+    (void)CallFrameCxt;
```
q82419: As above.

test/api/APIUnitTest.cpp
```diff
@@ -3911,6 +3967,594 @@ TEST(APICoreTest, VM) {
...
+  WasmEdge_StringDelete(CName2);
+  WasmEdge_StringDelete(CName1);
+  WasmEdge_StringDelete(PName);
+  WasmEdge_FunctionInstanceDelete(HostFunc);
+  WasmEdge_FunctionTypeDelete(FType);
+  WasmEdge_ExecutorDelete(Exec);
+  WasmEdge_ConfigureDelete(Conf);
+}
+
+TEST(APICoreTest, ModuleDeletion) {
```
q82419: I think the TEST() here is for the module deletion API. Why below uses VM?

q82419 (changes requested):

test/api/APIUnitTest.cpp
Wang-Yan-Hao: My initial thought was that  the entire VM is deleted, which subsequently triggers the deletion of its registered modules is also part of `ModuleDeletion` scenario. That's why I put this test case in `ModuleDeletion` test case.

I will move it to the VM test case.

test/api/APIUnitTest.cpp
Wang-Yan-Hao: Fix it.

test/api/APIUnitTest.cpp
Wang-Yan-Hao: Fix it.

test/api/APIUnitTest.cpp
Wang-Yan-Hao: Fix it.

test/api/APIUnitTest.cpp
Wang-Yan-Hao: Fix it.

test/api/APIUnitTest.cpp
Wang-Yan-Hao: Fix it.

test/api/APIUnitTest.cpp
Wang-Yan-Hao: Fix it.

test/api/APIUnitTest.cpp
Wang-Yan-Hao: Fix it.

test/api/APIUnitTest.cpp
Wang-Yan-Hao: Fix it.

test/api/APIUnitTest.cpp
Wang-Yan-Hao: Fix it.

test/api/APIUnitTest.cpp
Wang-Yan-Hao: Fix it.

test/api/APIUnitTest.cpp
Wang-Yan-Hao: Fix it.

test/api/APIUnitTest.cpp
Wang-Yan-Hao: Fix it.

test/api/APIUnitTest.cpp
Wang-Yan-Hao: Fix it.

test/api/APIUnitTest.cpp
Wang-Yan-Hao: I don't declare `HostFunc` variable and use `WasmEdge_ModuleInstanceAddFunction( HostMod_P1, PFuncName, WasmEdge_FunctionInstanceCreate(FType, DummyAdd, nullptr, 0));` directly.

test/api/APIUnitTest.cpp
Wang-Yan-Hao: I don't declare `HostFunc` variable and use `WasmEdge_ModuleInstanceAddFunction( HostMod_P1, PFuncName, WasmEdge_FunctionInstanceCreate(FType, DummyAdd, nullptr, 0));` directly.

Wang-Yan-Hao:
Thanks for the detailed review! I have addressed the requested changes in the latest push:

* WASM Binary Definition: Updated the WASM binary format to use `std::vector<uint8_t>` (e.g., `Provider1Wasm`, `ConsumerWasm`) to stay consistent with existing tests like FibonacciWasm.
* Naming Style: Refactored variable names to follow the PascalCase convention (e.g., `OriginalCount`, `PFuncName`).
* Configuration Optimization: Removed redundant `WasmEdge_ConfigureContext` creation in test cases where specialized configuration wasn't required, passing `nullptr` directly instead.
* Lambda Refactoring: Updated host function lambdas to remove unused parameters and avoid explicit (`void`) casts.
* Ownership Management: Moved host function allocation directly into `WasmEdge_ModuleInstanceAddFunction` to ensure the ownership is correctly transferred to the module instance.
* Resource Cleanup: Added explicit `Delete` calls (VM, Store, and ModuleInstance) at the end of each test scope.
* Test Reorganization:
  * Reorganized `TEST(APICoreTest, VMDeleteRegistered)` to focus on lifecycle management triggered by VM-level APIs.
  * Refactored `TEST(APICoreTest, ModuleDeletion)` to focus strictly on direct module and store deletion APIs.
  * Added descriptive comments for each test case as suggested.
* Rebase to only two commits.
* Fix misc linter error.
* Add new API to exposure symbol.

test/api/APIUnitTest.cpp
q82419: For the `VMDeleteRegistered` test above, please put all your VM related there (includes cross VM, shared store, etc.)
Please construct your tests WITHOUT VM (for module instances and loader, executor, etc.) and mainly for module deletion behavior here.

q82419 (changes requested):

Wang-Yan-Hao:
I have organized the VM-related test cases and added 4 new low-level core API test cases to `ModuleDeletion` to thoroughly verify the dependency tree's lifecycle behavior and improve Codecov branch coverage.

Here is a quick summary of the 4 new cases:
1. Delete a host provider module and re-register a new one under the same name.
2. Delete a WASM provider module and re-register a new one under the same name.
3. Delete the store manager while leaving instantiated modules active.
4. Delete a shared provider module and destroy one of its associated stores.

Please let me know if you have any feedback or suggestions!

Wang-Yan-Hao:
Hi @q82419,

I'm currently reviewing the [Codecov](https://app.codecov.io/gh/WasmEdge/WasmEdge/pull/4726?dropdown=coverage&src=pr&el=continue&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=WasmEdge) report for the new module deletion APIs and would like to get your advice on a few uncovered lines before digging deeper into the implementation:

1. `BuiltInModInsts` & `PluginModInsts` Loops: Codecov flags the deletion loops for these two containers as uncovered. Conceptually, they process module instances similarly to RegModInsts. Do you think it is necessary to introduce dedicated unit tests (e.g., mock a plugin or enable WASI) to cover these specific paths, or is the current core module validation sufficient for this PR?
2. Component Model Lines (`Comp.reset()` & `ActiveCompInst.reset()`): There are a few red lines regarding the Component Model state resets. Since this PR focuses strictly on the Core Module dependency tree lifecycle, should I ignore these component-specific paths, or do we want to cross-test them here?
3. Guard Clauses (`Invalid store context`): The defensive guard clauses for handling invalid or `nullptr` contexts are currently uncovered. Would you prefer me to add quick negative test cases (passing nullptr) to clear these Codecov alerts?

---

While implementing Case 3 to verify cross-VM/borrower dependency survival within a shared store, I noticed that the test actually passes successfully, but it spams the console with multiple `0x300` error logs during `VM2` construction:

```cpp
WasmEdge_StoreContext *SharedStore = WasmEdge_StoreCreate();
WasmEdge_VMContext *VM1 = WasmEdge_VMCreate(nullptr, SharedStore);
WasmEdge_VMContext *VM2 = WasmEdge_VMCreate(nullptr, SharedStore); // Runs successfully, but spams logs
```

```bash
[2026-05-18 23:51:43.156] [error]     At AST node: module
[2026-05-18 23:51:43.156] [error] instantiation failed: module name conflict, Code: 0x300
[2026-05-18 23:51:43.156] [error]     At AST node: module
[2026-05-18 23:51:43.156] [error] instantiation failed: module name conflict, Code: 0x300
[2026-05-18 23:51:43.156] [error]     At AST node: module
```

Even with a default configuration, the high-level `WasmEdge_VMCreate` automatically instantiates and registers global plugins or built-in mock modules (like `spectest` or WASI) into the assigned `Store`. When `VM2` boots up with the same `SharedStore`, it attempts to re-register these exact same built-in modules, triggering internal name conflicts that print to standard error.

Since the test runner still passes, should we care about these noisy conflict logs in the terminal, or is it acceptable to leave them as-is?

q82419:
> I have organized the VM-related test cases and added 4 new low-level core API test cases to `ModuleDeletion` to thoroughly verify the dependency tree's lifecycle behavior and improve Codecov branch coverage.
> 
> Here is a quick summary of the 4 new cases:
> 
> 1. Delete a host provider module and re-register a new one under the same name.
> 2. Delete a WASM provider module and re-register a new one under the same name.
> 3. Delete the store manager while leaving instantiated modules active.
> 4. Delete a shared provider module and destroy one of its associated stores.
> 
> Please let me know if you have any feedback or suggestions!

Please implement the above tests.

q82419:
> Hi @q82419,
> 
> I'm currently reviewing the [Codecov](https://app.codecov.io/gh/WasmEdge/WasmEdge/pull/4726?dropdown=coverage&src=pr&el=continue&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=WasmEdge) report for the new module deletion APIs and would like to get your advice on a few uncovered lines before digging deeper into the implementation:
> 
> 1. `BuiltInModInsts` & `PluginModInsts` Loops: Codecov flags the deletion loops for these two containers as uncovered. Conceptually, they process module instances similarly to RegModInsts. Do you think it is necessary to introduce dedicated unit tests (e.g., mock a plugin or enable WASI) to cover these specific paths, or is the current core module validation sufficient for this PR?

I think the current module validation is sufficient.

> 2. Component Model Lines (`Comp.reset()` & `ActiveCompInst.reset()`): There are a few red lines regarding the Component Model state resets. Since this PR focuses strictly on the Core Module dependency tree lifecycle, should I ignore these component-specific paths, or do we want to cross-test them here?

The component model would not have the dependency issue, because the component model already collected the inline modules in the instances and used inline linking. So you can ignore the cases here.

> 3. Guard Clauses (`Invalid store context`): The defensive guard clauses for handling invalid or `nullptr` contexts are currently uncovered. Would you prefer me to add quick negative test cases (passing nullptr) to clear these Codecov alerts?

Sure, you can add the quick negative test cases here.

> 
> While implementing Case 3 to verify cross-VM/borrower dependency survival within a shared store, I noticed that the test actually passes successfully, but it spams the console with multiple `0x300` error logs during `VM2` construction:
> 
> ```c++
> WasmEdge_StoreContext *SharedStore = WasmEdge_StoreCreate();
> WasmEdge_VMContext *VM1 = WasmEdge_VMCreate(nullptr, SharedStore);
> WasmEdge_VMContext *VM2 = WasmEdge_VMCreate(nullptr, SharedStore); // ⚠️ Runs successfully, but spams logs
> ```
> 
> ```shell
> [2026-05-18 23:51:43.156] [error]     At AST node: module
> [2026-05-18 23:51:43.156] [error] instantiation failed: module name conflict, Code: 0x300
> [2026-05-18 23:51:43.156] [error]     At AST node: module
> [2026-05-18 23:51:43.156] [error] instantiation failed: module name conflict, Code: 0x300
> [2026-05-18 23:51:43.156] [error]     At AST node: module
> ```
> 
> Even with a default configuration, the high-level `WasmEdge_VMCreate` automatically instantiates and registers global plugins or built-in mock modules (like `spectest` or WASI) into the assigned `Store`. When `VM2` boots up with the same `SharedStore`, it attempts to re-register these exact same built-in modules, triggering internal name conflicts that print to standard error.
> 
> Since the test runner still passes, should we care about these noisy conflict logs in the terminal, or is it acceptable to leave them as-is?

The conflict logs is related to the plugin loading workflow issue.
As the similar issue discovered by another mentorship program about the plugins loading, I will refine this in the future. Please ignore this.

q82419:
@Wang-Yan-Hao 
Please fix for the misc linter (spelling), and errors on Windows platforms.
You can easily fix that according to the CI error messages.

Wang-Yan-Hao:
> Please implement the above tests.

Thanks for the advice! I have implement the tests.

Wang-Yan-Hao:
> @Wang-Yan-Hao
> Please fix for the misc linter (spelling), and errors on Windows platforms.
> You can easily fix that according to the CI error messages.

I have fixed the misc linter warnings and resolved the Windows MSVC compilation errors.

Wang-Yan-Hao:
> Sure, you can add the quick negative test cases here.

Regarding the defensive guard clauses, I decided to remove the `if (!StoreCxt)` check from `WasmEdge_VMDeleteRegisteredModule` and `WasmEdge_VMForceDeleteRegisteredModule`. Even if `StoreCxt` were hypothetically null under some corrupted state, the subsequent API `WasmEdge_StoreFindModule(StoreCxt, ModuleName)` handles null pointers internally and would safely return `nullptr`, causing the execution to bypass the store unregistration loop without crashing.

```cpp
WASMEDGE_CAPI_EXPORT void
WasmEdge_VMDeleteRegisteredModule(const WasmEdge_VMContext *Cxt,
                                  const WasmEdge_String ModuleName) noexcept {
  if (!Cxt || !ModuleName.Buf) {
    return; // Invalid input
  }

  // Cast away const to match WasmEdge_VMGetStoreContext signature
  WasmEdge_StoreContext *StoreCxt =
      WasmEdge_VMGetStoreContext(const_cast<WasmEdge_VMContext *>(Cxt));
/* Remove this
  if (!StoreCxt) {
    return; // Invalid store context
  }
*/

  // Unregister from VM
  (const_cast<WasmEdge_VMContext *>(Cxt)->VM)
      .unregisterModule(genStrView(ModuleName));

  const WasmEdge_ModuleInstanceContext *ModInst =
      WasmEdge_StoreFindModule(StoreCxt, ModuleName);
  if (ModInst) {
    // Unregister from StoreManager
    fromStoreCxt(StoreCxt)->unregisterModule(genStrView(ModuleName));
  }
}
```

Regardless of whether the VM has an associated Store or not, we can still safely unregister a specific module from the VM context.

lib/api/wasmedge.cpp
```diff
@@ -2665,7 +2665,12 @@ WASMEDGE_CAPI_EXPORT void WasmEdge_ModuleInstanceAddGlobal(
 
 WASMEDGE_CAPI_EXPORT void
 WasmEdge_ModuleInstanceDelete(WasmEdge_ModuleInstanceContext *Cxt) noexcept {
-  delete fromModCxt(Cxt);
+  if (Cxt) {
+    auto *Mod = fromModCxt(Cxt);
```
q82419: You've checked the `Cxt` not nullptr here, therefore can use `fromModCxt(Cxt)->terminate()` directly.

q82419 (changes requested):
Besides you mentioned about removing `StoreCxt` null checking above, please help to refine the `WasmEdge_VMDeleteRegisteredModule` implementation to not to use CAPI in function body, but to use internal APIs. For example, to use `fromStoreCxt(Cxt)->findModule()` instead of `WasmEdge_StoreFindModule()`.

q82419:
> > Sure, you can add the quick negative test cases here.
> 
> Regarding the defensive guard clauses, I decided to remove the `if (!StoreCxt)` check from `WasmEdge_VMDeleteRegisteredModule` and `WasmEdge_VMForceDeleteRegisteredModule`. Even if `StoreCxt` were hypothetically null under some corrupted state, the subsequent API `WasmEdge_StoreFindModule(StoreCxt, ModuleName)` handles null pointers internally and would safely return `nullptr`, causing the execution to bypass the store unregistration loop without crashing.
> 
> ```c++
> WASMEDGE_CAPI_EXPORT void
> WasmEdge_VMDeleteRegisteredModule(const WasmEdge_VMContext *Cxt,
>                                   const WasmEdge_String ModuleName) noexcept {
>   if (!Cxt || !ModuleName.Buf) {
>     return; // Invalid input
>   }
> 
>   // Cast away const to match WasmEdge_VMGetStoreContext signature
>   WasmEdge_StoreContext *StoreCxt =
>       WasmEdge_VMGetStoreContext(const_cast<WasmEdge_VMContext *>(Cxt));
> /* Remove this
>   if (!StoreCxt) {
>     return; // Invalid store context
>   }
> */
> 
>   // Unregister from VM
>   (const_cast<WasmEdge_VMContext *>(Cxt)->VM)
>       .unregisterModule(genStrView(ModuleName));
> 
>   const WasmEdge_ModuleInstanceContext *ModInst =
>       WasmEdge_StoreFindModule(StoreCxt, ModuleName);
>   if (ModInst) {
>     // Unregister from StoreManager
>     fromStoreCxt(StoreCxt)->unregisterModule(genStrView(ModuleName));
>   }
> }
> ```
> 
> Regardless of whether the VM has an associated Store or not, we can still safely unregister a specific module from the VM context.

I think this work is almost done with the above requested changes. Please help to rebase to the latest master.

Wang-Yan-Hao:
> Besides you mentioned about removing `StoreCxt` null checking above, please help to refine the `WasmEdge_VMDeleteRegisteredModule` implementation to not to use CAPI in function body, but to use internal APIs. For example, to use `fromStoreCxt(Cxt)->findModule()` instead of `WasmEdge_StoreFindModule()`.

I have re-implement with internal API and fix the redundant statement. All the test also passed.

lib/api/wasmedge.cpp
```diff
@@ -3532,28 +3535,31 @@ WASMEDGE_CAPI_EXPORT void WasmEdge_VMCleanup(WasmEdge_VMContext *Cxt) noexcept {
...
   }
 
-  // Cast away const to match WasmEdge_VMGetStoreContext signature
-  WasmEdge_StoreContext *StoreCxt =
-      WasmEdge_VMGetStoreContext(const_cast<WasmEdge_VMContext *>(Cxt));
-  if (!StoreCxt) {
-    return; // Invalid store context
-  }
+  auto &VM = const_cast<WasmEdge_VMContext *>(Cxt)->VM;
+  auto *StoreCxt = &(VM.getStoreManager());
```
q82419: `auto &StoreMgr`. The getter returns reference already, don't convert to pointer again.

q82419 (changes requested):

lib/api/wasmedge.cpp
Wang-Yan-Hao: Fix it.

q82419:
@Wang-Yan-Hao Still some errors on Windows.

Wang-Yan-Hao:
> @Wang-Yan-Hao Still some errors on Windows.

I have updated the patch to resolve the compilation errors and warnings on Windows (MSVC):
1. Resolved `C4458` (declaration hides class member): Changed the local variable name from `Mod` to `ModInst` in `unsafeUnregisterModule` to avoid shadowing the class member variable VM::Mod.
2. Resolved `[-Wshorten-64-to-32]`: Added `static_cast<uint32_t>` to explicitly convert `.size()` (64-bit `size_type`) to a 32-bit integer when passing it to the C API buffers.
3. Resolved `[-Wshadow]` (variable shadowing): Removed the redundant inner-scope declarations of `Params` and `Returns` arrays in `APIUnitTest.cpp`, reuse the outer-scope variables instead.

Wang-Yan-Hao:
Regarding the MSVC build failure on Windows (`fatal error C1128: number of sections exceeded object file format limit`), it seems that the template function `cleanupModInstContainer` is causing template code bloating (code expansion) and generating too many debug sections in `vm.cpp.obj` when instantiated across multiple unit tests.

I may explicitly overload `cleanupModInstContainer` into two separate, regular functions: one dedicated to the `std::unordered_map` container and the other for the `std::vector` container?

q82419:
> Regarding the MSVC build failure on Windows (`fatal error C1128: number of sections exceeded object file format limit`), it seems that the template function `cleanupModInstContainer` is causing template code bloating (code expansion) and generating too many debug sections in `vm.cpp.obj` when instantiated across multiple unit tests.
> 
> I may explicitly overload `cleanupModInstContainer` into two separate, regular functions: one dedicated to the `std::unordered_map` container and the other for the `std::vector` container?

Maybe you cannot use the template to extend the cleanup function. Try the overload way.

Wang-Yan-Hao:
> Maybe you cannot use the template to extend the cleanup function. Try the overload way.

Fix it with overloading.

```cpp
  void cleanupModInstContainer(
      std::vector<std::unique_ptr<Runtime::Instance::ModuleInstance>>
          &Container) {
    for (auto &Item : Container) {
      if (auto *ModInst = Item.release()) {
        ModInst->terminate();
      }
    }
    Container.clear();
  }

  void cleanupModInstContainer(
      std::unordered_map<std::string,
                         std::unique_ptr<Runtime::Instance::ModuleInstance>>
          &Container) {
    for (auto &Item : Container) {
      if (auto *ModInst = Item.second.release()) {
        ModInst->terminate();
      }
    }
    Container.clear();
  }

  void cleanupModInstContainer(
      std::unordered_map<HostRegistration,
                         std::unique_ptr<Runtime::Instance::ModuleInstance>>
          &Container) {
    for (auto &Item : Container) {
      if (auto *ModInst = Item.second.release()) {
        ModInst->terminate();
      }
    }
    Container.clear();
  }
```

Wang-Yan-Hao:
I have refactored the template functions into explicit overloads (`cleanupModInstContainer`) to reduce the number of template instantiations and generated sections. However, even with the template completely removed, the build still fails on Windows Server 2025 with `fatal error C1128: number of sections exceeded object file format limit`. This indicates that the sheer size of `vm.cpp` inherently exceeds the default MSVC COFF section limit (65,536).

Sholud I add enforce the `/bigobj` flag for windows 2025 server to compile or maybe I sholud make `vm.cpp` smaller (split into some small files)?

By the way I think the [StepSecurity Harden-Runner](https://github.com/WasmEdge/WasmEdge/pull/4726/checks?check_run_id=77580391026) fail is because the network issue, not my code.

hydai:
Hi @Wang-Yan-Hao 
Yes, please add the `/bigobj` flag to make the compilation work.

Wang-Yan-Hao:
> Hi @Wang-Yan-Hao Yes, please add the `/bigobj` flag to make the compilation work.

Because it is the  `lib/vm.cpp` file problem, I add

```make
if(MSVC)
  target_compile_options(wasmedgeVM PRIVATE "/bigobj")
endif()
```

in `lib/vm/CMakeLists.txt`.

hydai:
Hi @Wang-Yan-Hao,  
It seems that the Windows CI passed. However, since I merged another LFX mentorship PR first, you will need to resolve some conflicts before your PR can be merged.

Wang-Yan-Hao:
> Hi @Wang-Yan-Hao, It seems that the Windows CI passed. However, since I merged another LFX mentorship PR first, you will need to resolve some conflicts before your PR can be merged.

The conflict was indeed due to the data structure migration to the dual-container design (`RegModInsts` as a vector and `RegModMap` as the index tracker).

```cpp
  /// Registered module instances by user.
  std::vector<std::unique_ptr<Runtime::Instance::ModuleInstance>> RegModInsts;
  /// Map from ID to the index in RegASTModules and RegModInsts.
  std::unordered_map<std::string, std::array<size_t, 2>> RegModMap;
```

To align with this architecture, I have refactored the deletion logic in `VM::unsafeUnregisterModule`. Instead of calling `erase()` on the `RegModInsts` vector (which would shift subsequent elements and invalidate all cached indices in `RegModMap`), I chose to preserve the slot by releasing the ownership:

1. Look up the element's index from `RegModMap`.
2. Call `.release()` on that specific vector slot. This safely turns that unique_ptr slot into `nullptr` while keeping the vector's size and other modules' indices perfectly intact.
3. Call `terminate()` and explicitly `delete` the raw pointer to prevent memory leaks.
4. Erase the name entry from `RegModMap`.

Here is the updated code snippet for this part:

```cpp
Expect<void> VM::unsafeUnregisterModule(std::string_view Name) {
  auto RegMapIt = RegModMap.find(std::string(Name));
  if (RegMapIt != RegModMap.end()) {
    size_t ModInstIdx = RegMapIt->second[1];

    if (ModInstIdx < RegModInsts.size()) {
      auto *ModInst = RegModInsts[ModInstIdx].release();
      if (ModInst) {
        ModInst->terminate();
      }
    }

    RegModMap.erase(RegMapIt);
    return {};
  }
  // ... rest of the built-in and plugin module checks
```

We erase the Module form RegModamp, the vector RegModinsta still live the slot in there and just reslease (become null). We don't erse the modul in vecotr is because if we erase the vecotr will move and let the index become not correct.

hydai (approved):

q82419 (approved):

Wang-Yan-Hao:
Hi @q82419 and @hydai,

Just a quick heads-up: I actually haven't addressed the issue caught by Codex earlier. I'm currently looking into the data structures to figure out the best way to fix it.

```cpp
  /// Registered module instances by user.
  std::vector<std::unique_ptr<Runtime::Instance::ModuleInstance>> RegModInsts;
  /// Map from ID to the index in RegASTModules and RegModInsts.
  std::unordered_map<std::string, std::array<size_t, 2>> RegModMap;
```

Codex correctly pointed out that `unsafeUnregisterModule` searches `RegModMap` using the passed `Name`, but the map is actually populated using `Module.getID()` during registration. Because of this, using `auto RegMapIt = RegModMap.find(std::string(Name));` is incorrect, as find expects the internal ID rather than the registered alias.

The simplest (but slowest) fix would be to traverse the `RegModInsts` vector to find the `ModuleInstance` with the matching name.

I'm a bit unsure about the intended design here. It seems like we originally used the module `Name` as the unique key. As I'm looking into the new features today, I'm wondering: is it now possible for two different module IDs to share the same registered name.

I initially considered changing `RegModInsts` to an `unordered_map<std::string, ModuleInstance>` using the registered `Name` as the key. However, since the new `RegModMap` structure relies on RegModInsts being a `vector` (it stores the vector index in `array[1]`), I can't simply change the type of `RegModInsts` to a map.

image

Above is the origin advice of the desgin of the `RegModMap`.

Wang-Yan-Hao:
I may use below code as temporally solution, I have push to feature branch.
```cpp
  auto InstIt = std::find_if(
      RegModInsts.begin(), RegModInsts.end(),
      [&Name](const std::unique_ptr<Runtime::Instance::ModuleInstance> &Inst) {
        return Inst && Inst->getModuleName() == Name;
      });
  if (InstIt != RegModInsts.end()) {
    auto ModId = InstIt->get()->getID();
    auto *ModInst = InstIt->release();
    if (ModInst) {
      ModInst->terminate();
    }

    RegModMap.erase(ModId);
    return {};
  }
```

hydai:
Hi @Wang-Yan-Hao 
Just create a follow-up PR for fixing it.

Wang-Yan-Hao:
> Hi @Wang-Yan-Hao Just create a follow-up PR for fixing it.

PR #4905 is create.

PR #4706, opened 2026-03-10
feat: refactor component model canonical section

kalamkaar9404:
## Description

Refactored the component model canonical section implementation to address issue #4334 requirements:

1. **Unified component function instance** - Combined lifted (core WASM) and host (native component) functions into a single `FunctionInstance` class with `FunctionKind` discriminator
2. **Integrated component value types** - Updated `HostFunctionBase` to use `AST::Component::FuncType` throughout the function hierarchy
3. **Complete canonical ABI details** - Added `CanonicalOptions` struct and implemented full option parsing for lift/lower operations (memory, realloc, post-return, encoding, async)

### Key Changes:
- Unified `FunctionInstance` class supporting both lifted and host functions
- Added `CanonicalOptions` struct for canonical ABI parameters
- Implemented dual-path function invocation (lifted vs host)
- Enhanced canonical section instantiation with complete option parsing
- Added post-return cleanup invocation for proper memory management
- Fixed realloc alignment parameter (now 1 for strings per Canonical ABI spec)

Fixes #4334

## Checklist

- [x] **DCO Signed-off**: All commits are signed-off (`git commit -s`)
- [x] **Commit Messages**: Follows Conventional Commit standards
- [x] **Local Tests Passed**: Linter tests passed on fork

PR #4730, opened 2026-03-21
fix(component): pass correct alignment to realloc in canonical lifting

LFsWang:
The canonical ABI spec requires realloc to receive alignment=1 for UTF-8 strings, but WasmEdge was passing alignment=0. This caused realloc implementations using the align parameter to compute incorrect pointers, potentially writing string data to offset 0 in linear memory.

This bug causes string input to be written at offset 0 in linear memory, corrupting data segments and allocator state, because any spec-compliant realloc that aligns pointers with (ptr + align - 1) & ~(align - 1) will collapse to 0 when align=0.

Bug wat:

```wat
;; Minimal reproducer: WasmEdge passes align=0 to realloc in canonical lifting.
;;
;; The Component Model canonical ABI spec requires that when the runtime calls
;; realloc to allocate space for a string argument, it must pass align=1.
;; WasmEdge passes align=0 instead.
;;
;; This component dumps the realloc parameters received from the runtime.
;; Expected: old_ptr=0  old_size=0  align=1  new_size=<len of input>
;; Actual:   old_ptr=0  old_size=0  align=0  new_size=<len of input>

(component
  (core module $M
    (memory (export "memory") 1)

    (func (export "realloc") (param i32 i32 i32 i32) (result i32)
      ;; Store params at known offsets for inspection
      (i32.store (i32.const 100) (local.get 0))
      (i32.store (i32.const 104) (local.get 1))
      (i32.store (i32.const 108) (local.get 2))
      (i32.store (i32.const 112) (local.get 3))
      i32.const 1024  ;; return a safe pointer
    )

    ;; Return the align parameter (param 2) that realloc received
    (func (export "get-align") (param i32 i32) (result i64)
      (i64.extend_i32_u (i32.load (i32.const 108)))
    )
  )
  (core instance $m (instantiate $M))
  (alias core export $m "memory" (core memory $mem))
  (alias core export $m "realloc" (core func $realloc))
  (alias core export $m "get-align" (core func $f))
  (type $t (func (param "s" string) (result u64)))
  (func $lift (type $t) (canon lift (core func $f) (memory $mem) (realloc $realloc)))
  (export "get-align" (func $lift))
)

```

## Checklist

## Test Evidence

---


LFsWang:
I found it is also fixed in #4706

LFsWang:
It can be continued.

hydai (approved):

PR #4731, opened 2026-03-23
feat(hash): replace rapidhash with a5hash v5.25

ibmibmibm:
Replace rapidhash with a5hash for improved small-key hashing performance (~11.5 vs ~18 cycles for 0-64 byte keys).

Changes:
- Replace rapidMum/rapidMix with a5hashMul128 (128-bit multiply)
- Replace RandomEngine with a5rand two-seed PRNG
- Remove Miller-Rabin secret generation (~200 lines)
- Use fixed PI-mantissa seed constants with per-process randomization
- Rename Hash::rapidHash to Hash::a5Hash

### DIFF
--- a/include/common/hash.h
+++ b/include/common/hash.h
@@ -9,7 +9,7 @@
 ///
 /// \file
 /// This file contains the functions about hashing data.
-/// Currently using rapidhash.
+/// Currently using a5hash.
 ///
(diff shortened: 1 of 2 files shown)


q82419:
Not plan to support this currently.

PR #4758, opened 2026-04-06
fix(validator): populate instance exports and fix export index aliasing

LFsWang:
- Register component exports on the instance after instantiation, so that subsequent `alias export` can resolve export names.
- Register inline exports on the instance after bounds checking.
- Introduce InstanceExport record (SortType + optional ExternDesc*) to support both externdesc-bearing exports and inline exports that only carry a sortidx.
- Simplify alias export sort matching: read SortType directly from the InstanceExport record instead of re-deriving it from ExternDesc.
- Fix export validation to increment core sort index sizes per spec: "All exports (of all sorts) introduce a new index."
- Add descTypeToSortType helper with correct ValueBound → Value mapping.
- Enable validate() in LoadAndRun_MultiComponentBinary regression test.

## Description
Repleace #4713 because the base has been changed.

## Checklist

## Test Evidence

---


lib/validator/component_validator.cpp
```diff
@@ -13,6 +13,142 @@ namespace Validator {
...
+  for (const auto &Decl : IT.getDecl()) {
+    if (Decl.isExportDecl()) {
+      const auto &Exp = Decl.getExport();
+      const auto &ED = Exp.getExternDesc();
+      // TODO(GAP-I-3): InstanceExport cannot represent a `(core module)`
+      // export because its SortType field is component-side only; skip
+      // such exports rather than crashing.
+      auto ST = descTypeToSortType(ED.getDescType());
+      if (!ST.has_value()) {
+        continue;
```
dannypsnl, line 66: should we add a warning log here?

q82419:
@LFsWang  please check does this PR conflict with #4762 . Thanks!

lib/validator/component_validator.cpp
LFsWang: fixed, add a debug msg.

dannypsnl (approved):

q82419 (approved):

PR #4771, opened 2026-04-09
fix: use unique_lock in StoreManager::reset() to prevent data race

### LINKED ISSUE #4770: bug: StoreManager::reset() uses wrong lock type
### Summary

Hi, I was reading through `include/runtime/storemgr.h` and noticed `reset()` grabs a `shared_lock` but then calls `NamedMod.clear()`. That feels wrong since a shared lock still lets other threads read the map at the same time.

### Current State

`reset()` uses `shared_lock` while it writes to `NamedMod` via `clear()`. 

What's interesting is that `registerModule()` right below it correctly uses `unique_lock` when it writes to the same map. Looks like `reset()` just has the wrong lock type.

### Expected State

Change shared_lock to unique_lock :
```cpp
void reset() noexcept {
    std::unique_lock Lock(Mutex);   
    for (auto &&[Name, ModInst] : NamedMod) {
      (const_cast<Instance::ModuleInstance *>(ModInst))
          ->unlinkStore(this, Name);
    }
    NamedMod.clear(); 
  }
```

Happy to send a PR if this looks right to you!

### Reproduction steps

To verify this, I wrote a small test file that mirrors the exact same 
pattern: `shared_lock` on a map, then clearing it while another thread 
reads from it.
Compiled it with `-fsanitize=thread` and `TSAN` flagged it immediately as 
a confirmed data race between the `clear()` and the concurrent read.

The same issue is in `storemgr.h — reset()`and `findModule()` have this 
exact same relationship.

### Screenshots

_No response_

_No response_

### Components

Others

main branch (latest)

macOS

### Hardware Architecture

arm64

_No response_

barry3406:
## Summary

- `reset()` in `storemgr.h` uses `std::shared_lock` but performs write operations (`unlinkStore` + `NamedMod.clear()`), causing a data race with concurrent readers like `findModule()`
- Changed to `std::unique_lock` to match all other mutating methods (`registerModule`, `unregisterModule`, `registerComponent`)

## Root Cause

`shared_lock` allows concurrent readers while `reset()` is clearing the map. This is undefined behavior — TSAN confirms the race between `clear()` and concurrent `find()` calls.

## Test Plan

- [x] Verified all other write methods in `StoreManager` use `unique_lock`
- [x] Verified `findModule()` (read path) correctly uses `shared_lock`
- [ ] Existing CI tests should pass — this only tightens locking

Fixes #4770

### DIFF
--- a/include/runtime/storemgr.h
+++ b/include/runtime/storemgr.h
@@ -70,7 +70,7 @@ public:
 
   /// Reset this store manager and unlink all the registered module instances.
   void reset() noexcept {
-    std::shared_lock Lock(Mutex);
+    std::unique_lock Lock(Mutex);
     for (auto &&[Name, ModInst] : NamedMod) {
       (const_cast<Instance::ModuleInstance *>(ModInst))
           ->unlinkStore(this, Name);


dannypsnl (approved):

barry3406:
The macOS 14 (arm64) failure is an unrelated timeout in `wasmedgeLLVMCoreTests` (1500s limit). My change only touches `storemgr.h` (lock type), which has no connection to the LLVM test suite. All other 39 checks passed. Could a maintainer re-run the failed job?

barry3406:
@dannypsnl Could you re-run the failed macOS 14 job when you get a chance? Thanks!

mrinalchaturvedi27:
Hey, I was the one who opened issue #4770 and mentioned I'd work on the fix. Happy to see it's getting addressed either way, just noting for context.

barry3406:
> Hey, I was the one who opened issue #4770 and mentioned I'd work on the fix. Happy to see it's getting addressed either way, just noting for context.

@mrinalchaturvedi27 Thanks for opening the issue and for taking a look at it! I noticed the issue and put together a fix, but didn’t catch that you were planning to work on it — sorry about that. Happy to collaborate or adjust anything if you had a specific approach in mind

q82419 (approved):

PR #4777, opened 2026-04-11
fix(executor): prevent silent override of error code in  registerModule() & registerComponent()

mrinalchaturvedi27:
### Summary
- In 
https://github.com/WasmEdge/WasmEdge/blob/b836225c7816de1bea26478976036538024c87ef/lib/executor/executor.cpp#L45-L55
if `StoreMgr.registerModule()` fails, the code manually overwrites the true error with `ModuleNameConflict` before printing it to the logs.
- This ensures true error codes are logged accurately instead of always claiming there is a name conflict.
### Root Cause
When component registration fails for any reason, the code intercepts the real error and overwrites it. This forces the logs to say `ModuleNameConflict`, dropping the real error completely and making debugging difficult. The same fix is also made for `registerComponent()`

### Test Plan
- [x] Confirmed `StoreMgr::registerComponent()` already returns `ModuleNameConflict` on 
  real name conflicts, so the overwrite was always redundant.
- [x] Verified file formatting locally with `git clang-format` (no issues found).
- [x] Existing CI tests should pass this only removes a misleading log override.


q82419 (changes requested):
Please fix for the DCO and commit lint.

mrinalchaturvedi27:
> Please fix for the DCO and commit lint.

Thank you for the review @q82419. I fixed the DCO and commitlint issues. Could you please approve the pending workflows for this PR?

q82419 (approved):

PR #4795, opened 2026-04-15
fix(executor): return error when Inst is null

mrinalchaturvedi27:
## Description
Replace `assuming(Inst)` with `Unexpect(..)` return in `executor.cpp.`
https://github.com/WasmEdge/WasmEdge/blob/5cb30cf81f8179650144f73914ae59c5b13eed8c/lib/executor/executor.cpp#L179-L188

In release builds , `assuming(...)` expands to `__builtin_unreachable()`. 
https://github.com/WasmEdge/WasmEdge/blob/5cb30cf81f8179650144f73914ae59c5b13eed8c/include/common/errcode.h#L29-L36
If `Inst` is unexpectedly null, that can lead to undefined behavior instead of a safe error. This change makes the executor return an explicit error in that case.

## Checklist
- [x]  Local Tests Passed: Local tests were run successfully.

## Test
Screenshot 2026-04-15 at 3 40 16 PM

### DIFF
--- a/lib/executor/executor.cpp
+++ b/lib/executor/executor.cpp
@@ -178,7 +178,10 @@ Executor::invoke(const Runtime::Instance::FunctionInstance *FuncInst,
         // already dynamic typed into the top abstract heap type.
         auto *Inst =
             Val.get<RefVariant>().getPtr<Runtime::Instance::CompositeBase>();
-        assuming(Inst);
+        if (unlikely(!Inst)) {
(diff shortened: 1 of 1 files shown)


q82419:
I believe the copilot AI has answered. Use assuming here is expected.

mrinalchaturvedi27:
> I believe the copilot AI has answered. Use assuming here is expected.

Thank you for the review @q82419. I understand that `assuming(Inst)` is expected here.

PR #4802, opened 2026-04-19
refactor(executor): extract SIMD superinstruction primitives into header

matthargett:
## Summary

- Introduces `include/executor/engine/simd_ops.h` — a single set of inline, free-function SIMD op templates (arithmetic, min/max, fmin/fmax, neg/abs/sqrt, ceil/floor/trunc/nearest, popcnt, v128 bitwise, narrow/widen conversions).
- Converts the `Executor::runVector*` member functions in `binary_numeric_vector.ipp` and `unary_numeric_vector.ipp` into thin wrappers that forward to these templates.
- Pure refactor — no behavior change, no new public API.

## Why: superinstruction enablement

Previously, each SIMD op lived inline inside its `Executor::runVector*` member function. That made the op bodies unreachable from anywhere except the member-function dispatch path — including from the engine's own fused / inline-dispatched code paths.

The extracted templates give us SIMD **superinstruction primitives**: small, inlineable op bodies that are callable from multiple dispatch strategies with zero overhead.

Concrete consumers this unlocks:

1. **Inline dispatch in `Engine.cpp`.** The v128 whole-register bitwise ops (`v128.not`, `v128.and`, `v128.andnot`, `v128.or`, `v128.xor`) are hot enough that inline-dispatching them from the main switch — rather than calling an `Executor` member function — saves an indirect call per op. The extracted `simdOps::v128*` helpers make this straightforward.

2. **Fused-opcode / superinstruction paths.** Any future fused opcode (e.g., `v128.load + f32x4.add`, broadcast-fma chains) can call the same `simdOps::*` template for the arithmetic component without duplicating the implementation.

3. **Load-time passes.** A constant-folding / IPCP pass (separate follow-up PR) can construct a temporary `ValVariant`, call the same template, and read back the result — reusing exactly the runtime semantics instead of reimplementing them.

All three uses share one canonical implementation, so semantic drift between them is structurally impossible.

## Implementation notes

- Templates are `inline` with no external linkage — the compiler inlines the thin wrappers away, so the runtime dispatch path compiles to the same code as before.
- In-place update convention: ops mutate `Val` / `Val1` directly. Since `ValVariant` is an unsafe union where all 16-byte SIMD types alias `uint128_t`, callers can `emplace<uint128_t>` raw bits and invoke a typed op without a copy.
- **Non-MSVC only.** MSVC has separate `runVector*` overrides in the existing `.ipp` files that use `std::array` loops. Those are intentionally left untouched so `simd_ops.h` stays free of MSVC-specific workarounds.

## Test plan

- [x] `wasmedgeExecutorTests` passes
- [x] `wasmedgeExecutorRegressionTests` passes (SIMD spec coverage)
- [x] `wasmedgeAPIUnitTests` / `wasmedgeAPIVMCoreTests` / `wasmedgeAPIStepsCoreTests` pass
- [x] Full `ctest` run (24/24 suites) passes locally on macOS arm64
- [ ] CI green on Linux x86_64 / arm64 / Windows

## Diff size

3 files changed, 328 insertions(+), 112 deletions(-)

### DIFF
--- a/include/executor/engine/simd_ops.h
+++ b/include/executor/engine/simd_ops.h
@@ -1,0 +1,237 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2019-2024 Second State INC
+
+//===-- wasmedge/executor/engine/simd_ops.h - SIMD op free templates ------===//
+//
+// Part of the WasmEdge Project.
+//
+//===----------------------------------------------------------------------===//
+///
+/// \file
+/// Free-function template implementations of SIMD operations.  Each op
+/// mutates its first argument (Val or V1) in-place.
+///
+//===----------------------------------------------------------------------===//
+#pragma once
+
+#if !defined(_MSC_VER) || defined(__clang__)
+
+#include "common/roundeven.h"
+#include "common/types.h"
+#include "executor/engine/vector_helper.h"
+
+#include <cmath>
+#include <cstdint>
+#include <type_traits>
+
+namespace WasmEdge {
+namespace Executor {
+namespace simdOps {
+
+// ---------------------------------------------------------------------------
+// Lane-wise binary arithmetic.  T is the lane element type
+// (e.g. uint32_t, float, double).
+// ---------------------------------------------------------------------------
+
+template <typename T>
+inline void vectorAdd(ValVariant &V1, const ValVariant &V2) noexcept {
+  using VT [[gnu::vector_size(16)]] = T;
+  V1.get<VT>() += V2.get<VT>();
+}
+
+template <typename T>
+inline void vectorSub(ValVariant &V1, const ValVariant &V2) noexcept {
+  using VT [[gnu::vector_size(16)]] = T;
+  V1.get<VT>() -= V2.get<VT>();
+}
+
+template <typename T>
+inline void vectorMul(ValVariant &V1, const ValVariant &V2) noexcept {
+  using VT [[gnu::vector_size(16)]] = T;
+  V1.get<VT>() *= V2.get<VT>();
+}
+
+template <typename T>
+inline void vectorDiv(ValVariant &V1, const ValVariant &V2) noexcept {
+  using VT [[gnu::vector_size(16)]] = T;
+  V1.get<VT>() /= V2.get<VT>();
+}
+
+// ---------------------------------------------------------------------------
+// Lane-wise integer min/max.  Use signed T for _s ops, unsigned T for _u ops.
+// ---------------------------------------------------------------------------
+
+template <typename T>
+inline void vectorMin(ValVariant &V1, const ValVariant &V2) noexcept {
+  using VT [[gnu::vector_size(16)]] = T;
+  VT &A = V1.get<VT>();
+  const VT &B = V2.get<VT>();
+  A = detail::vectorSelect(A > B, B, A);
+}
+
+template <typename T>
+inline void vectorMax(ValVariant &V1, const ValVariant &V2) noexcept {
+  using VT [[gnu::vector_size(16)]] = T;
+  VT &A = V1.get<VT>();
+  const VT &B = V2.get<VT>();
+  A = detail::vectorSelect(B > A, B, A);
+}
+
+// ---------------------------------------------------------------------------
+// Float min/max with NaN propagation.
+//
+//   fmin:  R = bits(A) | bits(B)    // merge NaN payloads
+//          if A < B: R = A
+//          if A > B: R = B
+//          if A is NaN: R = A
+//          if B is NaN: R = B
+//   fmax:  same but & instead of |, and reversed comparisons.
+//
+// NaN check uses (V == V) which is false for NaN lanes, so
+// vectorSelect(false, R, V) = V.
+// ---------------------------------------------------------------------------
+
+template <typename T>
+inline void vectorFMin(ValVariant &V1, const ValVariant &V2) noexcept {
+  static_assert(std::is_floating_point_v<T>);
+  using VT [[gnu::vector_size(16)]] = T;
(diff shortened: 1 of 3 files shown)


q82419:
Actually, the SIMD executions lived inside the `executor` because the `executor` is really the engine of interpreter/AOT/JIT.
The `runVector*` functions are already the dispatchers from the instruction vectors.
For the refactoring of using more templates make sense, but I don't understand why you want to dispatch the contents again.

matthargett:
> Actually, the SIMD executions lived inside the `executor` because the `executor` is really the engine of interpreter/AOT/JIT. The `runVector*` functions are already the dispatchers from the instruction vectors. For the refactoring of using more templates make sense, but I don't understand why you want to dispatch the contents again.

You're right that `runVector*` already dispatch from the instruction vectors, and I'm not changing where SIMD lives. simd_ops.h sits in `include/executor/engine/` with the existing .ipp files, same namespace, same module.
 the next PR tees up the next consumer I allude to in the description, I was trying to keep things incremental. let me know if you want bigger chunks. the next PR in my fork: https://github.com/rebeckerspecialties/WasmEdge/pull/2

it's based on work I helped with in GCC's devirtualization in ~2010, Luau bytecode interpreter, and darklua optimizing bundler. I created these patches locally so I could get the WASM interpreter workload fully onto Efficiency Cores on iPhone XS while maintaining a 60fps WASI-gpu rendering. At module load, before any Executor instance exists, the fold pass needs to evaluate a SIMD op on temporary ValVariants to see if it reduces to a constant. It can't call `Executor::runVectorAddOp<T>` because that's a non-static member function returning `Expect<void>` for the interpreter's dispatch protocol — both of which require an Executor instance and add plumbing we don't need for pure computation.

The extraction factors the pure computation out of the member function so it's callable from any context. The member function is now a thin wrapper that calls the template and returns `Expect<void>{}`. Both the wrapper and the template are `inline`, so the extra indirection is compiled away: the interpreter dispatch path produces the same machine
code as before (I verified this in the disassembly on XCode 26.4's clang). 

If you'd rather see a different factoring — e.g., making `runVector*` static free functions in a detail namespace directly,
without the `simdOps::` intermediate — I'm happy to restructure. The only hard requirement is that the SIMD op bodies be callable from outside an Executor instance. 

Do you have a preference? Just let me know! :D

include/executor/engine/simd_ops.h
```diff
@@ -0,0 +1,279 @@
...
+  const TOut Part = static_cast<TOut>(Val.get<TIn>());
+  using VTOut [[gnu::vector_size(16)]] = TOut;
+  if constexpr (!std::is_floating_point_v<TOut>) {
+    Val.emplace<VTOut>(VTOut{} + Part);
+  } else if constexpr (sizeof(TOut) == 4) {
+    Val.emplace<VTOut>(VTOut{Part, Part, Part, Part});
+  } else {
+    Val.emplace<VTOut>(VTOut{Part, Part});
+  }
+}
```
hydai: This function is not used in ipp.

include/executor/engine/simd_ops.h
```diff
@@ -0,0 +1,279 @@
...
+#include <cmath>
+#include <cstdint>
+#include <type_traits>
+
+namespace WasmEdge {
+namespace Executor {
+namespace simdOps {
+
+// ---------------------------------------------------------------------------
+// V128 whole-register bitwise operations
```
hydai: The `v128` series of functions is not used in the executor.

hydai (changes requested):

include/executor/engine/simd_ops.h
matthargett: You're right — the v128 bitwise series (`v128.not/and/andnot/or/xor`) is inline-dispatched in `engine.cpp` (lines 1231-1287) and never goes through `Executor::runVector*`, so the helpers here had no in-tree consumers. Removed in the amended commit. If a future PR refactors those cases in `engine.cpp` to share helpers (which would also let us drop the duplicated `#ifdef _MSC_VER` branches there), the templates can be re-added alongside the consumer migration.

include/executor/engine/simd_ops.h
matthargett: Confirmed — `splatOp` had no consumer in the `.ipp` files. `Executor::runSplatOp` in `unary_numeric_vector.ipp` carries its own inlined body and `engine.cpp` calls it directly. Removed in the amended commit.

include/executor/engine/simd_ops.h
```diff
@@ -0,0 +1,237 @@
...
+template <typename T> inline void vectorSqrt(ValVariant &Val) noexcept {
+  static_assert(std::is_floating_point_v<T>);
+  using VT [[gnu::vector_size(16)]] = T;
+  VT &Result = Val.get<VT>();
+  if constexpr (sizeof(T) == 4) {
+    Result = VT{std::sqrt(Result[0]), std::sqrt(Result[1]),
+                std::sqrt(Result[2]), std::sqrt(Result[3])};
+  } else {
+    Result = VT{std::sqrt(Result[0]), std::sqrt(Result[1])};
+  }
```
hydai: I found that in the original implementation, there will have a check like this:

```cpp
else if constexpr (sizeof(T) == 8)
```

What's the reason to remove all of them?

hydai (changes requested):

include/executor/engine/simd_ops.h
matthargett: sorry, I was going for congruence I think. the original check is acutally better since it will correctly deal with `long double` or `_Float16` float formats whereas my style alignment would actually cause wrong semantics to be used. fixed.

hydai (approved):

hydai:
Thanks.

PR #4818, opened 2026-04-27
fix(executor): pin SIMD float add/mul operand order to preserve NaN payload

q82419:
## Description

`f32x4.add` / `f64x2.add` (and the corresponding multiplies) were implemented with the gnu-vector compound-assignment forms

    V1 += Val2.get<VT>();
    V1 *= Val2.get<VT>();

LLVM models `fadd` and `fmul` as commutative at the IR level, so after inlining and LTO clang is free to swap the operands and emit the equivalent of `fadd dst, V2, V1`. On platforms whose hardware add picks the first NaN operand's payload (ARM64 FADD, x86 ADDPD), this swap silently changes which NaN payload propagates -- causing NaN results to differ from other engines such as Wasmtime even though the inputs were `lhs op rhs`.

For example, the module

    (func (result v128)
      v128.const i32x4 0xffffffff 0xffffffff 0xffffffff 0xffffffff
      v128.const i32x4 0xffffffff 0xffffffff 0xffffffff 0xffff60ff
      f64x2.add)

returned the rhs's NaN bits (0xffff60ffffffffff in the high lane) instead of the lhs's bits (0xffffffffffffffff) that Wasmtime and the WAT-level operand order suggest.

Pin the operand order for the floating-point specialisations by adding lane by lane through a general-register asm barrier. The empty `asm volatile("" : "+r"(...))` clobbers prevent the optimiser from re-vectorising the loop or treating the add/mul as commutative, so the lhs's NaN payload is preserved consistently. Integer vector adds/muls keep the original vector form since NaN payload semantics do not apply.

Assisted-by: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

## Checklist

## Test Evidence

---

### DIFF
--- a/include/executor/engine/binary_numeric_vector.ipp
+++ b/include/executor/engine/binary_numeric_vector.ipp
@@ -164,7 +164,28 @@ Expect<void> Executor::runVectorAddOp(ValVariant &Val1,
                                       const ValVariant &Val2) const {
   using VT [[gnu::vector_size(16)]] = T;
   VT &V1 = Val1.get<VT>();
-  V1 += Val2.get<VT>();
+  const VT &V2 = Val2.get<VT>();
+  if constexpr (std::is_floating_point_v<T>) {
+    // LLVM models fadd as commutative at the IR level, so after
+    // inlining/LTO it can swap the operands and emit the equivalent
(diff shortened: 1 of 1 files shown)


q82419:
This will not plan to fix.
#4819

PR #4832, opened 2026-05-02
fix(executor): correct WASI socket v2 fallback matching

mrinalchaturvedi27:
## Description
Fix WASI socket `_v2` fallback handling during function import matching.
https://github.com/WasmEdge/WasmEdge/blob/494c90f39f5478fd89b2dce90a68a12daa3f0ace/lib/executor/instantiate/import.cpp#L189-L203
The executor now only selects a `_v2` socket fallback when the export exists and its signature matches the Wasm module’s expected import type.

## Root Cause

The previous fallback logic could incorrectly accept the `_v2` path after a v1 type mismatch because it:
- did not check whether `findFuncExports(... "_v2")` returned `nullptr`
- rechecked the original v1 function type instead of the v2 candidate type
- used inverted match logic, causing fallback to be accepted incorrectly

## Changes
- Add a null check for `ImpInstV2`.
- Match against `ImpInstV2->getTypeIndex()`.
- Only assign `ImpInst = ImpInstV2` when the v2 signature matches `TypeIdx`.
- Preserve the existing `IncompatibleImportType` path when no compatible fallback exists.

## Checklist

## Test Evidence
Screenshot 2026-05-02 at 12 06 47 PM

Screenshot 2026-05-02 at 12 07 12 PM


q82419 (changes requested):
Please fix your commit lint

mrinalchaturvedi27:
> Please fix your commit lint

Hi @q82419!, Thanks for the review. I have done the changes, it should run green now.

q82419 (approved):

PR #4834, opened 2026-05-02
fix: null funcref crash in proxyCallRef

ashnaaseth2325-oss:
### **SUMMARY**

This PR fixes a null dereference in the AOT execution path of `Executor::proxyCallRef()` by adding a missing null check for function references. It aligns AOT behavior with the interpreter, ensuring proper WebAssembly trap handling instead of a host crash.

Primary changes are in `lib/executor/engine/proxy.cpp`.

---

### **FIX**

```cpp
// Before
const auto *FuncInst = retrieveFuncRef(Ref);
const auto &FuncType = FuncInst->getFuncType();  // potential null deref

// After
const auto *FuncInst = retrieveFuncRef(Ref);
if (unlikely(FuncInst == nullptr)) {
  return Unexpect(ErrCode::Value::AccessNullFunc);
}
const auto &FuncType = FuncInst->getFuncType();
```

### DIFF
--- a/lib/executor/engine/proxy.cpp
+++ b/lib/executor/engine/proxy.cpp
@@ -208,6 +208,9 @@ Expect<void> Executor::proxyCallRef(Runtime::StackManager &StackMgr,
                                     const ValVariant *Args,
                                     ValVariant *Rets) noexcept {
   const auto *FuncInst = retrieveFuncRef(Ref);
+  if (unlikely(FuncInst == nullptr)) {
+    return Unexpect(ErrCode::Value::AccessNullFunc);
+  }
   const auto &FuncType = FuncInst->getFuncType();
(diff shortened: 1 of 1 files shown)


ashnaaseth2325-oss:
Hello @q82419 
Kindly review this PR whenever you have time.
Thanks!

q82419:
> Hello @q82419 Kindly review this PR whenever you have time. Thanks!

Please provide your tests in AOT mode which causes crash.
In the AOT compilation here: https://github.com/WasmEdge/WasmEdge/blob/master/lib/llvm/compiler.cpp#L4484-L4485
I think we've done the nullref checking.

q82419 (changes requested):
As comment

q82419:
Close due to no response.

PR #4839, opened 2026-05-03
fix(cli): reject mismatched argument count

LFsWang:
## Description

`wasmedge` accepts mismatched CLI arg counts for the invoked function.
Component lifted calls in release builds **silently return wrong answers with exit 0**.

### Minimal repro

```wat
;; add.wat
(component
  (core module $M
    (func (export "add") (param i32 i32) (result i32)
      (i32.add (local.get 0) (local.get 1))))
  (core instance $m (instantiate $M))
  (alias core export $m "add" (core func $add))
  (type $t (func (param "a" s32) (param "b" s32) (result s32)))
  (func (export "add") (type $t) (canon lift (core func $add))))
```

```bash
wasm-tools parse add.wat -o add.wasm
wasmedge --enable-component add.wasm add 5    # prints "5",  exit 0  ← silent
wasmedge --enable-component add.wasm add      # prints "0",  exit 0  ← silent
wasmedge --enable-component add.wasm add 5 7  # prints "12", exit 0  (correct)
```

Debug build hits `stackmgr.h:68` assertion. 
Release build: `assuming(...)` becomes `__builtin_unreachable()`, optimizer silently produces a wrong answer. `wasmtime`/`wasmer` reject the same input with a clear error and non-zero exit.

### `assuming(...)` is not a release-build guard

under NDEBUG `assuming(R)` reduces to `__builtin_unreachable()` — a debug-only assertion **and** an optimization hint, not a runtime check. 
Treating `assuming` as an input-validation barrier (as this code path implicitly did) means a caller that violates the precondition can produce silently wrong results in release.

### Question

Commit 3 removes a fallback present since at least 2021-10-03  (`14d6d553`): extra argv beyond declared params were silently appended to `FuncArgs` as `u64`. No tests, docs, or CLI flags reference it.

Strict `!=` in commit 2 makes it unreachable, so commit 3 is pure dead-code removal — but if anyone downstream relied on the lenient behavior (e.g. shoving raw stack values for ad-hoc testing), this is a CLI breaking change. 

## Checklist

## Test Evidence

---

### DIFF
--- a/lib/driver/runtimeTool.cpp
+++ b/lib/driver/runtimeTool.cpp
@@ -72,6 +72,14 @@ ToolOnModule(WasmEdge::VM::VM &VM, const std::string &FuncName,
   std::vector<ValVariant> FuncArgs;
   std::vector<ValType> FuncArgTypes;
 
+  const size_t Expected = FuncType.getParamTypes().size();
+  const size_t Got = Opt.Args.value().size() - 1;
+  if (Got < Expected) {
+    spdlog::error("function `{}` expects {} argument(s), got {}"sv, FuncName,
+                  Expected, Got);
+    return EXIT_FAILURE;
+  }
+
   for (size_t I = 0;
        I < FuncType.getParamTypes().size() && I + 1 < Opt.Args.value().size();
        ++I) {
@@ -201,6 +209,14 @@ ToolOnComponent(WasmEdge::VM::VM &VM, const std::string &FuncName,
                 const AST::Component::FuncType &FuncType) noexcept {
   std::vector<ComponentValVariant> FuncArgs;
   std::vector<ComponentValType> FuncArgTypes;
+
+  const size_t Expected = FuncType.getParamList().size();
+  const size_t Got = Opt.Args.value().size() - 1;
+  if (Got < Expected) {
+    spdlog::error("function `{}` expects {} argument(s), got {}"sv, FuncName,
+                  Expected, Got);
+    return EXIT_FAILURE;
+  }
 
   for (size_t I = 0;
        I < FuncType.getParamList().size() && I + 1 < Opt.Args.value().size();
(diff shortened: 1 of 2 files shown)


q82419 (changes requested):
In the caller which call this function in reactor mode, I think the function name checking has been done.
On the other hand, currently we fill the insufficient args with `uint64_t(0)`, so we didn't check for the return arity,

LFsWang:
I am not sure the where to fill the zero on missing args. I trace the code do not find that. 
And in fact, it will raise UB if I compile with `-fsanitize=undefined -fno-sanitize-recover=undefined` in release mode. The run path do not be blocked by `assuming` and continue running until raising the out of bound access on stack.

```
$ wasmedge --enable-component too_few_args_crash.wasm add 5
include/runtime/stackmgr.h:68:5: runtime error: execution reached an unreachable program point
    #0 runLocalGetOp
    #1 execute (instruction lambda)
    #2 execute
    #3 runFunction
    #4 Executor::invoke (FunctionInstance, module overload)
    #5 Executor::invoke (Component::FunctionInstance, component overload)
    #6 VM::executeComponent
    #7 std::thread async
```

q82419:
> I am not sure the where to fill the zero on missing args. I trace the code do not find that. And in fact, it will raise UB if I compile with `-fsanitize=undefined -fno-sanitize-recover=undefined` in release mode. The run path do not be blocked by `assuming` and continue running until raising the out of bound access on stack.
> 
> ```
> $ wasmedge --enable-component too_few_args_crash.wasm add 5
> include/runtime/stackmgr.h:68:5: runtime error: execution reached an unreachable program point
>     #0 runLocalGetOp
>     #1 execute (instruction lambda)
>     #2 execute
>     #3 runFunction
>     #4 Executor::invoke (FunctionInstance, module overload)
>     #5 Executor::invoke (Component::FunctionInstance, component overload)
>     #6 VM::executeComponent
>     #7 std::thread async
> ```

Sorry that I found it just only append the redundant args onto stack with extracting as `u64`.
Then I think we can return fail when `if (Got < Expected)`, for the giving args more than expected case, we can let it have the original behavior (or drop the redundant args, to avoid the `assuming` at the end of execution).

q82419 (approved):

PR #4842, opened 2026-05-04
feat(component,validator): validate canonical built-ins

q82419:
## Description

Implement structural validation for component-model canonical built-ins (canon lift, canon lower, resource.new/rep/drop/drop_async) and the shared canonical-options rules. Replaces the prior stub that only bumped index-space counters.

ComponentContext infrastructure:
- Replace FuncCount/CoreFuncCount counters with Funcs/CoreFuncs vectors of type pointers so canon lift/lower can bind and resolve func types by index.
- Track locally-defined resources via a DefinedResources set populated by addType(); imports and outer aliases go through a new addTypeImported() path that registers the resource type without marking it local. Factor the two into a shared addTypeImpl helper.
- Add isLocalResource(), getFunc(), getCoreFunc() accessors.

Canonical validation:
- Split validate(Canonical) into per-opcode helpers; both Resource__drop and Resource__drop_async route to the same validator.
- validateCanonOptions takes the Canonical::OpCode directly (no extra context enum) and enforces: at-most-one per option (string-encoding slot shared across UTF-8/UTF-16/Latin-1), per-opcode whitelist (lift accepts every option, lower rejects the lift-only options post-return/callback/always-task-return, every other built-in rejects all options), structural rules (callback requires async, always-task-return requires async, post-return/async mutually exclusive, realloc requires memory), and index bounds on memory/realloc/callback/post-return. Unknown codes emit UnknownCanonicalOption.
- resource.new / resource.rep require bounds + isResourceType + isLocalResource; resource.drop permits imported resources.
- canon lift resolves the target type to a FuncType via getDefType() and binds the pointer on addFunc(); unresolved type slots emit a dedicated diagnostic distinct from wrong-kind.
- Other canonical opcodes continue to return ComponentNotImplValidator.

Tests: 26 new unit tests in ComponentValidatorTest covering bounds, wrong-kind, imported/local resource rules, async variant, option whitelist, structural rules, and happy paths, via reusable helpers makeCompWithLocalResource / appendCanonSection / mkOpt / makeCompWithImportedFunc / makeCompWithCoreFuncAndFuncType.

Deferred: canon lift/lower ABI signature match against flatten_functype (for the core func's flattened params/results); realloc/callback/ post-return signature-body shape checks; post-MVP canon opcodes (backpressure.set, task.*, context.*, stream.*, future.*, error-context.*, waitable.*, thread.*).

Assisted-by: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

## Checklist

## Test Evidence

---

### DIFF
--- a/include/validator/component_context.h
+++ b/include/validator/component_context.h
@@ -36,7 +36,7 @@ public:
     std::vector<std::unordered_map<std::string, ExternalType>>
         CoreInstances;                                 // core:instance
     std::vector<const AST::SubType *> CoreTypes;       // core:type
-    uint32_t CoreFuncCount = 0;                        // core:func
+    std::vector<const AST::SubType *> CoreFuncs;       // core:func
     std::vector<const AST::TableType *> CoreTables;    // core:table
     std::vector<const AST::MemoryType *> CoreMemories; // core:memory
     std::vector<const AST::GlobalType *> CoreGlobals;  // core:global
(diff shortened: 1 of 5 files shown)


include/validator/component_context.h
```diff
@@ -343,6 +368,25 @@ class ComponentContext {
...
+    if (DT != nullptr) {
+      if (DT->isInstanceType()) {
+        Ctx.InstanceTypes[Idx] = &DT->getInstanceType();
+      }
+      if (DT->isResourceType()) {
+        Ctx.ResourceTypes[Idx] = &DT->getResourceType();
+        if (IsLocal) {
+          Ctx.DefinedResources.insert(Idx);
+        }
+      }
```
dannypsnl, line 384: So something can be both an instance type and a resource type?

q82419: Fixed

dannypsnl (approved):

hydai (approved):

PR #4860, opened 2026-05-09
fix(socket): correct ipv6 address copy size in windows wasi connect

omda777:
## Description
While validating IPv6 socket connections on Windows, I found that the WASI socket connect path was copying IPv6 addresses using the IPv4 address size, which could lead to truncated or malformed sin6_addr data.

This change updates the copy logic to use the correct in6_addr size.
I also added IPv6 connection tests for the Windows socket path.
image

## Checklist


omda777:
hi @q82419 , can review this PR

hydai (approved):
Nice catch!

PR #4861, opened 2026-05-10
fix(executor,component): validate invoke args and replace assuming() with FuncNotFound

mrinalchaturvedi27:
## What I fixed

I fixed the component function invoke path so it now rejects wrong arguments early with `FuncSigMismatch`, instead of letting them go through silently.

The checks are simple:

- how many values are passed
- how many value types are passed
- whether the passed component types actually match the function parameters

I also changed the `assuming(CoreFuncInst)` line. Earlier, if the lowered core function was missing, the whole process used to abort. Now it returns `FuncNotFound` properly, so the caller can handle it.

## How I found it

I was going through `lib/executor/executor.cpp` and saw the component `Executor::invoke` path. The argument type matching was a commented-out TODO, and the arity check was only `ParamTypes.size() < ExpectedArity`, which is very loose. It was not even comparing `Params.size()` against the function's arity. So you could send wrong number of values, or wrong types, and the call would just continue into ABI lowering.

Then I checked `include/runtime/instance/component/function.h`. The lower function is stored as a plain raw pointer there, so it can be null in real cases. That means `assuming(CoreFuncInst)` was not guarding an internal invariant, it was actually a possible runtime case. Using `assuming` for something that can be null at runtime is not the correct error path.

## What I changed

- Added proper argument count check. All three sizes must be equal: `Params.size()`, `ParamTypes.size()`, and the function's own param list size.
- Added a per-parameter `ComponentValType` check. The error message also tells which parameter index failed, and prints expected vs got type using `ComponentTypeCodeStr`, so the user can actually see what went wrong.
- Returned `FuncNotFound` when `getLowerFunction()` is null. Also moved this check before `convValsToCoreWASM`, so a bad call does not run any realloc work before failing.
- Fixed the existing `Component.LoadAndRun_SimpleBinary` test. It was passing `U64`, but the fixture's `mdup` function is actually `s64 -> s64`. Earlier this mismatch was hidden because type matching itself was off. With matching turned on, the test correctly failed, so I updated it to `S64` / `int64_t`, which is what it should have been from the start.
- Added two regression tests in `test/component/ComponentRegressionTest.cpp`:
  - `Component.Invoke_BadArguments` — covers four cases: too few args, too many args, value/type count mismatch, and wrong type.
  - `Component.Invoke_MissingLowerFunction` — directly creates a `FunctionInstance` with a null lower pointer and calls `Executor::invoke`. Without my fix, this test would have aborted the whole process.

Please note: anyone who was earlier passing a wrong `ComponentValType` and getting a successful return will now get `FuncSigMismatch`. This is the whole purpose of the fix, but I am calling it out so reviewers are not surprised.
## Validation

- Screenshot 2026-05-10 at 6 47 42 PM
 
- Screenshot 2026-05-10 at 6 48 03 PM

- Full gtest binary — 90/90 passed across 4 suites.
- `clang-format --dry-run --Werror` on both touched files — clean.
- `git diff --check` — clean.
- Forced rebuild of `executor.cpp.o` — no new compiler warnings.

## Notes

I only added exact type matching for now. Full subtype matching for component types is a bigger task and can be done separately. I have kept the existing `// TODO: COMPONENT - implement full component subtype matching.` comment for that.

I also did not touch the `assuming(Returns.size() == ReturnTypes.size())` line on the result side. That one is checking something we control internally, not user input. Different issue, can be looked at separately if needed.

## Checklist


q82419:
Close this due to out-of-date from current latest branch.
Component-model now is planned to base on the active `dev/component` branch.
And besides, I think it's propper to have a full version of type matching for this issue instead of partitially implemented.
Please update and re-create PR to that branch if this is still in active development.

PR #4863, opened 2026-05-11
feat(component,validator): resource, start, externdesc, tag validation

q82419:
## Description

Tag in core-export alias and inline-export
- validate(Alias) core-export branch maps CoreSortType::Tag to ExternalType::Tag instead of falling into the unsupported default.
- validate(CoreInstance) inline-export switch picks up Tag, so an inline-exported tag is registered as Tag rather than rejected.

Resource destructor signature [i32] -> []
- Anonymous-namespace helpers coreFuncSubType(params, results) and isResourceDtorSig(SubType): the first caches and returns a stable SubType* for a synthesised core function signature; the second checks the signature is [i32] -> [].
- validateCanonResource{New,Rep} record their synthesised core func as [i32] -> [i32]; validateCanonResourceDrop records it as [i32] -> [].
- validate(ResourceType) verifies the destructor's recorded signature is exactly [i32] -> [].

Annotated-name resource-in-scope check
- New ResourceLabels: unordered_map<string, uint32_t> on ComponentContext::Context, with addResourceLabel / hasResourceLabel.
- validate(Import) for a TypeBound import with a Label name registers the label -> type idx in the resource labels map.
- validate(Import) for Constructor/Method/Static names extracts the resource label and rejects the import with ComponentInvalidName when no resource of that name is in scope.

validate(StartSection)
- Bounds-check the function index against the component func space.
- When the func type is known, check arg arity equals param count and declared result count equals the func's result arity.
- Bounds-check each argument's value index against the value space.
- Append result fresh values to the value index space.

ExternDesc bounds and ExportDecl delegation
- validate(ExternDesc) bounds-checks the referenced type index against the right space: CoreType against core:type; FuncType / ComponentType / InstanceType against the component-level type space.
- validate(ExportDecl) delegates to validate(ExternDesc) instead of reimplementing a partial dispatch that skipped bounds.

UnknownCoreTag error code
- New ErrCode value UnknownCoreTag = 0x02B5, message "unknown tag".
- validate(Alias) core-export and validate(CoreInstance) inline-export emit UnknownCoreTag when the failing sort is Tag.

Spec tests
- test/spec/spectest.cpp: tags Validate flag flipped to true.

Tests
- 8 new ComponentValidatorTest cases covering the above.

## Checklist

## Test Evidence

---

### DIFF
--- a/lib/validator/validator.cpp
+++ b/lib/validator/validator.cpp
@@ -18,6 +18,21 @@ namespace Validator {
 
 namespace {
 
+// One-shot builder for the pre-defined core function SubTypes.
+AST::SubType makeCoreFuncType(std::initializer_list<TypeCode> Params,
+                              std::initializer_list<TypeCode> Results) {
+  AST::FunctionType FT;
+  for (auto T : Params) {
+    FT.getParamTypes().emplace_back(T);
+  }
+  for (auto T : Results) {
+    FT.getReturnTypes().emplace_back(T);
+  }
+  AST::SubType ST;
+  ST.getCompositeType().setFunctionType(std::move(FT));
+  return ST;
+}
+
 static constexpr uint32_t MaxSubtypeDepth = 63;
 
 // TODO: make the super type depth table instead of recursively querying.
@@ -64,6 +79,12 @@ checkSubtypeDepth(const uint32_t BaseIdx, uint32_t TestIdx,
 }
 
 } // namespace
+
+// Validator constructor. See "include/validator/validator.h".
+Validator::Validator(const Configure &Conf) noexcept
+    : Conf(Conf),
+      CoreFuncType_I32_I32(makeCoreFuncType({TypeCode::I32}, {TypeCode::I32})),
+      CoreFuncType_I32_Void(makeCoreFuncType({TypeCode::I32}, {})) {}
 
 // Validate Module. See "include/validator/validator.h".
 Expect<void> Validator::validate(const AST::Module &Mod) {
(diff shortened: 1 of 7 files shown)


q82419:
@hydai 
This PR is the first component validator PR, and should be merged after 0.17.0 release.

lib/validator/validator.cpp
```diff
@@ -65,6 +80,12 @@ checkSubtypeDepth(const uint32_t BaseIdx, uint32_t TestIdx,
 
 } // namespace
 
+// Validator constructor. See "include/validator/validator.h".
+Validator::Validator(const Configure &Conf) noexcept
+    : Conf(Conf),
+      CoreFuncType_I32_I32(makeCoreFuncType({TypeCode::I32}, {TypeCode::I32})),
+      CoreFuncType_I32_Void(makeCoreFuncType({TypeCode::I32}, {})) {}
```
dannypsnl, line 87: Why are we create this instance at the point? This means if someone create another constructor for `Validator` then

1. one could set `CoreFuncType_I32_I32` to something else
2. or one forgot to configure it

q82419: 1. The `CoreFuncType_I32_I32` is `const`. How to set to something else?
2. The reason is the same as other members. If someone creates another ctor, other members are also possible to forget to initialize. It's the ctor's responsibility.

dannypsnl: I mean can we use delegating constructors here?

q82419: Can you give some example? I don't think we would have several types of validator ctors in the future.

dannypsnl: I see.

dannypsnl (approved):

PR #4894, opened 2026-05-23
feat: add Fish shell env script support

### LINKED ISSUE #4838: feat: Supporting env script for Fish shell
### Summary

After installing `wasmedge` which does work with Fish shell, `source ~/.wasmedge/env` failed due to this error:

```
~/.wasmedge/env (line 4): 'case' builtin not inside of switch block
case :"$PATH": in
^~~^
from sourcing file ~/.wasmedge/env
source: Error while reading file '/Users/quatton/.wasmedge/env'
```

Would be nice if we
1. have a separate `env.fish`
or
2. make the original `env` fish compatible (which I believe would be more complicated)

I'd be willing to help if you could provide some sort of directions we wanna go with!

### Details

e.g. we might be able to do something like 

`source ~/.wasmedge/env.fish` instead like how we do it in `source .venv/bin/activate.fish`

### Appendix

_No response_

hariprakazz:
Fixes #4838

## Changes
- Generate `env.fish` alongside `env` during installation
- Auto-hook into `~/.config/fish/config.fish` when fish shell detected
- Fish-compatible syntax using `fish_add_path` and `set -gx`

## Testing
Verified `env.fish` is generated correctly with proper fish syntax.

Closes #4838

### DIFF
--- a/.github/scripts/installer-test-helpers.sh
+++ b/.github/scripts/installer-test-helpers.sh
@@ -25,3 +25,25 @@ verify_plugin() {
     exit 1
   fi
 }
+
+# Verify env.fish was created (Fish shell users)
+verify_fish_env() {
+  local ipath=${1:-~/.wasmedge}
+  if [ -f "$ipath/env.fish" ]; then
+    echo "✓ env.fish found: $ipath/env.fish"
+  else
+    echo "✗ env.fish not found: $ipath/env.fish"
+    exit 1
+  fi
+}
+
+# Verify env.fish was NOT created (non-Fish shell users)
+verify_no_fish_env() {
+  local ipath=${1:-~/.wasmedge}
+  if [ ! -f "$ipath/env.fish" ]; then
+    echo "✓ env.fish correctly absent for non-Fish shell"
+  else
+    echo "✗ env.fish unexpectedly created for non-Fish shell"
+    exit 1
+  fi
+}
--- a/utils/install_v2.sh
+++ b/utils/install_v2.sh
@@ -609,7 +609,8 @@ main() {
 
 	echo "$ENV" >"$IPATH/env"
 	echo "# Please do not edit comments below this for uninstallation purpose" >> "$IPATH/env"
-	# Generate env.fish for Fish shell users
+	# Generate env.fish only if user is running Fish shell
+	if echo "$SHELL" | grep -qi "fish"; then
 	local FISH_ENV="# wasmedge shell setup for fish
 fish_add_path -p \"$IPATH/bin\"
 if test -n \"\$DYLD_LIBRARY_PATH\"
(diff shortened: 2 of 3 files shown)


utils/install_v2.sh
```diff
@@ -609,7 +609,31 @@ main() {
...
+else
+    set -gx C_INCLUDE_PATH \"$IPATH/include\"
+end
+if test -n \"\$CPLUS_INCLUDE_PATH\"
+    set -gx CPLUS_INCLUDE_PATH \"$IPATH/include\" \$CPLUS_INCLUDE_PATH
+else
+    set -gx CPLUS_INCLUDE_PATH \"$IPATH/include\"
+end"
+	echo "$FISH_ENV" >"$IPATH/env.fish"
+	echo "# Please do not edit comments below this for uninstallation purpose" >> "$IPATH/env.fish"
```
hydai: This will always produce an `env.fish` without checking whether the users are using Fish or not.

hydai (changes requested):
Since this will inject a new environment into a Fish user's environment, a test covering it must be submitted in the same PR.

ravindra-RKB:
Hi @hydai ,

I noticed this pr  has requested changes regarding Fish-shell detection and test coverage....  If the original author is no longer working on it, I'd be interested in helping address the feedback and preparing an updated implementation.....

hariprakazz:
Hi @ravindra-RKB,

Thanks for offering to help! Yes, please feel free to pick this up — the
two key things @hydai flagged are:

1. **Fish shell detection** — `env.fish` should only be generated when the
user is actually running Fish (e.g. check `$SHELL` or detect Fish in the
environment).
2. **Test coverage** — a test for the Fish shell env setup needs to be
included in the same PR.

Happy to review once you have a patch ready. Thanks again!

hariprakazz:
Hi @hydai, I've addressed both review points:

env.fish is now only generated when Fish shell is detected via $SHELL
Added verify_fish_env and verify_no_fish_env helper functions to .github/scripts/installer-test-helpers.sh

Please take another look when you get a chance. Thanks!

hariprakazz:
@ravindra-RKB — Quick update: I went ahead and pushed the fix myself just now. No action needed from you — but thanks again for offering!

.github/scripts/installer-test-helpers.sh
```diff
@@ -25,3 +25,25 @@ verify_plugin() {
     exit 1
   fi
 }
+
+# Verify env.fish was created (Fish shell users)
+verify_fish_env() {
```
hydai, line 30: I don't think this is correct.  
The CI now uses `bash` to run the script. Your verification won't work.  
https://github.com/WasmEdge/WasmEdge/blob/master/.github/workflows/test-installers.yml

hydai (changes requested):

hariprakazz:
Hi @hydai, thanks for catching that!

I've added a dedicated CI step `Test Fish Shell Detection` in
`.github/workflows/test-installers.yml` that explicitly overrides `$SHELL`
to test both cases:

- `SHELL=/usr/bin/fish bash utils/install_v2.sh` → verifies `env.fish`
**is** created
- `SHELL=/bin/bash bash utils/install_v2.sh` → verifies `env.fish` is
**not** created

This way the tests run correctly under bash in CI without needing fish to
be installed. Please take another look!

hydai:
Closing this due to a lack of activity. Please feel free to re-open it if you still want to work on it.

PR #4902, opened 2026-05-27
fix(wasi-nn,MLX): prevent shell injection in whisper audio input

hydai:
## Description

The MLX whisper backend copied guest-controlled wasi-nn tensor bytes verbatim into a file-path string, then decoded the audio by building an ffmpeg command run through popen(). Because popen() invokes /bin/sh, a malicious guest could embed shell metacharacters in the path (e.g. "/tmp/a.wav; id #") to execute arbitrary host commands, crossing the Wasm sandbox boundary; it could also name any host file ffmpeg opens.

Take decoded FP32 PCM samples directly instead: setInput rejects non-FP32 or malformed tensors and copies the samples into the audio array, and transcribe() consumes the samples directly. loadAudio() and all shell command construction are removed, so no shell is invoked and no host path is opened. Audio is now handled in-process, like the GGML whisper backend.

## Checklist


grorge123:
It looks good, but I am service in military to 6/10, I can't review it now. If it is emergency fix, please assign other people review it, thanks.

hydai:
No problem, this is not urgent.

dm4 (approved):

PR #4905, opened 2026-05-28
fix: fix lookup error while vm unregister module

Wang-Yan-Hao:
## Description

This PR fixes a key mismatch bug in `unsafeUnregisterModule` where it incorrectly attempts to look up `RegModMap` using the registered module `Name` (alias) instead of its internal `Module.getID()`.

During module registration, `RegModMap` is populated using `Module.getID()` as the key. However, `unsafeUnregisterModule` receives the registered `Name` (alias) and previously called `RegModMap.find(std::string(Name))`. This caused a lookup failure since the alias does not match the internal module ID.

As a temporary solution to unblock development while we reconsider the overall data structure design (as discussed), this PR introduces a `std::find_if` traversal over `RegModInsts` to locate the target `ModuleInstance` by matching `getModuleName() == Name`.

## Checklist

## Test Evidence

image

### DIFF
--- a/lib/vm/vm.cpp
+++ b/lib/vm/vm.cpp
@@ -345,19 +345,19 @@ VM::unsafeRegisterModule(std::string_view Name,
 }
 
 Expect<void> VM::unsafeUnregisterModule(std::string_view Name) {
-  auto RegMapIt = RegModMap.find(std::string(Name));
-  if (RegMapIt != RegModMap.end()) {
-    size_t ModInstIdx = RegMapIt->second[1];
+  auto InstIt = std::find_if(
+      RegModInsts.begin(), RegModInsts.end(),
+      [&Name](const std::unique_ptr<Runtime::Instance::ModuleInstance> &Inst) {
+        return Inst && Inst->getModuleName() == Name;
+      });
+  if (InstIt != RegModInsts.end()) {
+    auto ModId = (*InstIt)->getID();
+    auto *ModInst = (*InstIt).release();
 
-    if (ModInstIdx < RegModInsts.size()) {
-      auto *ModInst = RegModInsts[ModInstIdx].release();
-      if (ModInst) {
-        ModInst->terminate();
-      }
+    if (ModInst) {
+      ModInst->terminate();
     }
-
-    RegModMap.erase(RegMapIt);
-
+    RegModMap.erase(ModId);
     return {};
   }
   for (auto It = BuiltInModInsts.begin(); It != BuiltInModInsts.end(); ++It) {


Wang-Yan-Hao:
Just to document the rationale behind this temporary `std::find_if` fix for future reference:

image

As Yiying previously pointed out, a VM might register a module under an alias (a different name), meaning the registered name and the actual internal module name can be different. This discrepancy is why I initially intended to replace the `std::vector` with an `unordered_map` for `RegModInsts` in PR #4726.

However, because the current `RegModMap` structure relies heavily on storing and looking up the specific vector index (in `array[1]`), we cannot easily change `RegModInsts` to a map at this stage without breaking the indexing mechanism. 

Therefore, this PR introduces the `std::find_if` traversal as a temporary workaround to unblock the bug. The broader redesign of these data structures—including breaking the vector index dependency and refactoring the module lifecycle management—will be left for a dedicated discussion in a future issue.

---

As previously discussed (and noted by Yiying), the architecture implies that a module can be registered under an alias (`Name`), which might differ from its actual internal WebAssembly module name. 

This is explicitly visible in `storemgr.h`, where `StoreManager::registerModule` uses the passed `Name` (the alias/given name) as the key for its `NamedMod` map, treating it as the primary identifier for linking:
```cpp
Expect<void> registerModule(const Instance::ModuleInstance *ModInst,
                            std::string_view Name) {
    std::unique_lock Lock(Mutex);
    auto Iter = NamedMod.find(Name);
    if (likely(Iter != NamedMod.cend())) {
      return Unexpect(ErrCode::Value::ModuleNameConflict);
    }
    
    // StoreManager internally binds the specific registered alias to the instance pointer
    NamedMod.emplace(std::string(Name), ModInst);
    // ...
```

Although `StoreManager` supports distinct alias names in theory, in WasmEdge's current implementation, the registered name and the internal module name are identical during registration. Specifically, in `VM::unsafeRegisterModule(Name, Module)`, the instantiated `ModuleInstance` is constructed such that its internal module name is explicitly assigned to match the registered `Name`:

```cpp
// Within VM::unsafeRegisterModule
EXPECTED_TRY(auto ModInst, ExecutorEngine.registerModule(StoreRef, Module, Name));
RegModInsts.push_back(std::move(ModInst));
```

Because the `Executor` forces the instance's internal name to be identical to the registration `Name` at this stage, there is currently no discrepancy between the two values in the active codebase. Therefore, traversing `RegModInsts` and matching against `Inst->getModuleName() == Name` is completely deterministic and guaranteed to find the correct module right now.

dannypsnl (approved):

Wang-Yan-Hao:
Oh I think the macOS test failed is because version mismatch?

hydai:
I am updating the homebrew formula, will re-trigger it once the new version is published.

Issue #1448, opened 2022-05-13
The installer should copy plugins into the installed folder


hydai:
## Motivation

The file structure of the latest pre-release 0.10.0-alpha.1:

```
.
├── bin
│   ├── wasmedge
│   └── wasmedgec
├── include
│   └── wasmedge
│       ├── dense_enum_map.h
│       ├── enum_configure.h
│       ├── enum_errcode.h
│       ├── enum.inc
│       ├── enum_types.h
│       ├── int128.h
│       ├── spare_enum_map.h
│       ├── version.h
│       └── wasmedge.h
└── lib64
    ├── libwasmedge_c.so
    └── wasmedge
        └── libwasmedgePluginWasmEdgeProcess.so
```

We have moved the WasmEdge process host functions into an external shared library. The default installed path should be `~/.wasmedge/plugin`.

It's better that the installer and uninstaller to support copy and remove these files.

hydai:
Hi @SAtacker,

Could you please help with this issue?

SAtacker:
> Hi @SAtacker,
> 
> Could you please help with this issue?

Yes, thanks for notifying me.

SAtacker:
Can you please point me to the resource where the details of these are mentioned?
Details like why was it done? How the user achieves it?
Also, I need to make the default directory to `~/.wasmedge/plugin` for this version and beyond. Is that it?

hydai:
Yes, after 0.10.0, the default directory will have a new folder called `plugin` which contains all the default installed plugins from the tarball. You may need to check the tarball on every distros to make sure the resource path.

SAtacker:
> Yes, after 0.10.0, the default directory will have a new folder called `plugin` which contains all the default installed plugins from the tarball. You may need to check the tarball on every distros to make sure the resource path.

Thanks. Another thing is, with the new plugins, shouldn't the directory tree you have mentioned be something like
```
.wasmedge/
                  ├── bin
                  │   ├── wasmedge
                  │   └── wasmedgec
                  ├── include
                  │   └── wasmedge
                  │       ├── dense_enum_map.h
                  │       ├── enum_configure.h
                  │       ├── enum_errcode.h
                  │       ├── enum.inc
                  │       ├── enum_types.h
                  │       ├── int128.h
                  │       ├── spare_enum_map.h
                  │       ├── version.h
                  │       └── wasmedge.h
                  └── lib64
                      └── libwasmedge_c.so
.wasmedge/plugin/
[...]

PR #4909, opened 2026-05-28
feat(executor): add constant-folding / IPCP optimization pass

matthargett:
Builds on the merged precursor #4802 (SIMD superinstruction primitives in `simd_ops.h`). The const-folding pass consumes `simdOps::v128*` and `simdOps::splatOp` from that header — those helpers are re-introduced in this PR (PR #4802 trimmed them to only what its in-tree `.ipp` consumers needed, per review).

## Summary

Adds a load-time bytecode optimizer for the interpreter path, in two stages:

- **Per-function arithmetic folding** runs during `instantiate` for O1+, folding
  const+const+binop / const+unary / const+relop / identity patterns. Uses
  Nop-fill so jump offsets and `InstrVec` size are invariant across the pass.
- **Module-wide IPCP** runs during `VM::instantiate` when no AOT/JIT symbol is
  present. Pure call sites with constant arguments are evaluated at load time
  via a bounded mini-interpreter; step/depth budgets scale by `-O` level.

Wasm trap semantics are preserved by per-op `isSafe*` predicates; impure
callees (memory.grow, memory.size, table mutation, host calls, transitively-
impure call graphs) are skipped. Multi-return sites are folded only when the
result count fits the available instruction slots.

Adds a move-taking `FunctionInstance` constructor so the instantiator can
hand off the optimized `InstrVec` without a copy.

## Tests

- `wasmedgeConstFoldTests` (214) — fold-primitive coverage: int/float
  arithmetic, SIMD, trap-semantics preservation, NaN canonicalization,
  boundary conditions
- `wasmedgeIPCPEdgeCaseTests` (36) — end-to-end including 7 multi-return cases
- `wasmedgeConstFoldBenchmark` — load-time fold ratios over fib / factorial /
  mandelbrot fixtures

Test cases that reproduce prior-art bugs reference their source in an inline
comment: wasm3 #220 (multi-return reordering), Wasmtime CVE-2022-31169
(AArch64 sign-extension), Wasmer #2215 (`i64.rotr(_, 0)` LLVM UB),
SpiderMonkey 1669964 / 1647288 (SIMD pmin/pmax / min/max NaN), LLVM #101335
(IPSCCP cross-site contamination), LLVM #167620 / Wasmtime #1448 (`clz(0)`
UB in C vs. well-defined in Wasm).

## Test plan

- [x] `ctest` 26/26 suites pass on macOS arm64 (excluding env-dependent
      `CompileSubcommand` which requires `WASMEDGE_USE_LLVM=ON`)
- [x] `clang-format --dry-run -Werror` clean
- [x] Fork CI matrix green (44 pass / 1 Codecov-token infra fail / 1 skipping)

### DIFF
--- a/include/executor/engine/const_fold.h
+++ b/include/executor/engine/const_fold.h
@@ -1,0 +1,151 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2019-2024 Second State INC
+
+//===-- wasmedge/executor/engine/const_fold.h - Constant folding ----------===//
+//
+// Part of the WasmEdge Project.
+//
+//===----------------------------------------------------------------------===//
+///
+/// \file
+/// This file contains the declaration of the interpreter constant folding pass
+/// and its associated cost model.
+///
+/// ## Cost Model Background
+///
+/// This optimization pass uses a stack-thrash-aware cost model inspired by the
+/// Luau compiler's approach, adapted for WasmEdge's interpreter architecture.
+///
+/// Reference cost models from production compilers:
+///
+/// GCC (gcc/ipa-inline-analysis.cc): Uses instruction count as primary cost
+///   metric. Functions under ~40 instructions are auto-inlined. Cost is
+///   weighted by loop nesting (10x per level).
+///
+/// LLVM (llvm/include/llvm/Analysis/InlineCost.h): InlineCost class with
+///   default threshold of 225 cost units. Each LLVM IR instruction costs 5
+///   units base. Memory ops and calls cost more. Bonus for single-callsite
+///   and small functions.
+///
+/// ChakraCore (lib/Backend/InliningDecider.cpp): Uses bytecode size + call
+///   depth + loop presence. Threshold scales with optimization level.
+///   Small functions (<30 bytecodes) inlined aggressively.
+///
+/// Luau (Compiler/src/CostModel.cpp): Packed 64-bit parallel cost model.
+///   Byte 0 = baseline cost, bytes 1-7 = discount when variable 0-6 is
+///   constant. Uses saturating byte arithmetic to compute all 8 scenarios
+///   simultaneously. Literals cost 0, calls cost 3, allocations cost 10.
+///
+/// ## WasmEdge Stack-Thrash Cost Model
+///
+/// WasmEdge's interpreter uses std::vector<ValVariant> as its operand stack.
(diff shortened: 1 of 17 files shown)


test/executor/mandelbrot_c.wasm
dannypsnl, line 1: No binary file

hydai (changes requested):
We will no longer accept any binary files in this project; please handle test cases as we did before.

matthargett:
Fixed in the amended commit:

- Removed all 6 `.wasm` files from the repo (`fibonacci_bench`, `factorial_bench`, `mandelbrot_bench`, `mandelbrot_c`, `constFoldBench`, `ipcp_edge_cases`).
- Added a `mandelbrot_c.wat` decompiled from the previous `mandelbrot_c.wasm` (clang -O1 output) so all 6 fixtures now have a `.wat` source of truth.
- `test/executor/CMakeLists.txt` now finds `wat2wasm` via `find_program` and compiles each `.wat` → `.wasm` in the build directory at build time via `add_custom_command`. The benchmark target (`wasmedgeConstFoldBenchmark`) is gated on `wat2wasm` being available; if it's not, only the benchmark target is skipped and a `STATUS` message is emitted (tests are unaffected).

`IPCPEdgeCaseTest` already used an inline `std::array<Byte>` for its module per the existing repo convention (see `ExecutorRegressionTest.cpp`), so no consumer change was needed there.

All 26/26 local `ctest` suites pass on macOS arm64 with the new build-time `.wat` → `.wasm` pipeline.

hydai (changes requested):
Hi @matthargett,  
Thanks for this PR. I am still running the benchmarks across multiple OSs and hardware to ensure there are no performance regressions. It still takes time since I am using QEMU for the RISC-V and s390x platforms, which will be quite slow due to the limitations of my hardware. I apologize for the delay.  

Additionally, could you please reorganize your files:  

1. The `*Bench.wat` files should belong to `test/executor/benchmarks`.  
2. The SPDX copyright should now be `// SPDX-FileCopyrightText: Copyright The WasmEdge Authors`.

matthargett:
heya, I'm closing this as I don't have more time to spread across as many WASM projects as I used to. I still think it's a good change, but as the wasmtime maintainers told me in PR review, maybe this class of optimization belongs in a separate wasm-opt tool outside of the runtime itself.

hydai:
Hi @matthargett ,
Thanks for your contributions!

PR #4914, opened 2026-06-01
fix(plugin/image): compute output buffer size in 64-bit in decodeImgToSize

aizu-m:
In the image plugin, decodeImgToSize checks DstBuf against W * H * 3 * sizeof(...) computed in 32-bit, but W and H come straight from the guest, so the product wraps (e.g. W=2^30, H=4 gives 0) and the check passes even for a zero-length output buffer before stbir_resize_ writes the real multi-GB image. Rework the check to compare DstBuf.size() / BytesPerPixel against the 64-bit pixel count, so no multiplication of guest-controlled values can overflow (a plain 64-bit product can still wrap, e.g. W=H=2^31 with RGB32F gives 2^62 * 12 = 3 * 2^64 = 0).

### DIFF
--- a/plugins/wasmedge_image/image_func.cpp
+++ b/plugins/wasmedge_image/image_func.cpp
@@ -62,12 +62,12 @@ bool decodeImgToSize(Span<const uint8_t> Buf, uint32_t W, uint32_t H,
   }
 
   // Resize.
-  if (unlikely(DstBuf.size() <
-               W * H * 3 * (IsU8 ? sizeof(uint8_t) : sizeof(float)))) {
+  const uint64_t NeededSize = static_cast<uint64_t>(W) * H * 3 *
+                              (IsU8 ? sizeof(uint8_t) : sizeof(float));
+  if (unlikely(DstBuf.size() < NeededSize)) {
     spdlog::error("[WasmEdge-Image] Output buffer size {} not enough. "sv
                   "At least need {} bytes."sv,
-                  DstBuf.size(),
-                  W * H * 3 * (IsU8 ? sizeof(uint8_t) : sizeof(float)));
+                  DstBuf.size(), NeededSize);
     return false;
   }
   if (IsU8) {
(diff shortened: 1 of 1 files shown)


aizu-m:
DCO was red because the commit had no sign-off. Added it and reworked the header to the conventional-commit format. Green now.

plugins/wasmedge_image/image_func.cpp
```diff
@@ -62,12 +62,12 @@ bool decodeImgToSize(Span<const uint8_t> Buf, uint32_t W, uint32_t H,
   }
 
   // Resize.
-  if (unlikely(DstBuf.size() <
-               W * H * 3 * (IsU8 ? sizeof(uint8_t) : sizeof(float)))) {
+  const uint64_t NeededSize = static_cast<uint64_t>(W) * H * 3 *
```
hydai: What if `W = UINT32_MAX` and `H = UINT32_MAX`? Will this still cause an overflow?

hydai (changes requested):

plugins/wasmedge_image/image_func.cpp
aizu-m: Good catch. Traced it through: W*H alone still fits, since (2^32-1)^2 < 2^64. The wrap comes from the *3 and *4 steps. With W = H = UINT32_MAX the wrapped value happens to stay huge, so the check still rejects by luck. The bad case is W = H = 2^31 with RGB32F: 2^62 pixels * 12 bytes = 3 * 2^64, which wraps to exactly 0, and an empty buffer passes.

Reworked the check to divide instead of multiply: DstBuf.size() / BytesPerPixel < NumPixels. Nothing left that can overflow, whatever W and H are. Verified the 2^31 case is rejected now and wasmedgeImageTests still passes.

hydai (approved):

hydai:
You need rebase the branch yourself; otherwise, it will trigger the squashed commit, which will block the merge process.

hydai:
Or you can wait for this PR, https://github.com/WasmEdge/WasmEdge/pull/5163, to be merged, which will relax the commit lint checks. Thanks.

aizu-m:
Rebased. The three merge commits from the update-branch button are gone; the branch is now just the two fix commits on top of current master. Tree is identical to the head you approved.

aizu-m:
Thanks for merging this, and for the pointer on the rebase.

PR #4920, opened 2026-06-02
fix: reject non-funcref tables/refs in call_indirect and call_ref

q82419:
## Description

A table whose element type is a concrete non-funcref reference (e.g. `(ref null $s)` for a struct type) passed validation because the checks used `ValType::isFuncRefType()`, which treats any concrete type index as a funcref. Such a module would then reinterpret the struct reference as a `FunctionInstance *` at runtime -- a type confusion where attacker- controlled field bytes stand in for function metadata.

Resolve the concrete type via `TypeMatcher::matchType(.., FuncRef, ..)` in the table.set/call_indirect validator path, and harden the executor so `retrieveFuncRef` returns nullptr for non-funcref references, making `call_ref`/`call_indirect` raise a null-function trap instead of dereferencing a forged pointer.

Add a regression test covering call_indirect on a struct-typed table.

## Checklist

## Test Evidence

---

### DIFF
--- a/include/common/types.h
+++ b/include/common/types.h
@@ -689,6 +689,9 @@ inline ValVariant ValueFromType(ValType Type) noexcept {
 
 inline const Runtime::Instance::FunctionInstance *
 retrieveFuncRef(const RefVariant &Val) {
+  if (!Val.getType().isFuncRefType()) {
+    return nullptr;
+  }
   return Val.getPtr<Runtime::Instance::FunctionInstance>();
 }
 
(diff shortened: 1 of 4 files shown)


include/common/types.h
```diff
@@ -689,6 +689,9 @@ inline ValVariant ValueFromType(ValType Type) noexcept {
 
 inline const Runtime::Instance::FunctionInstance *
 retrieveFuncRef(const RefVariant &Val) {
+  if (!Val.getType().isFuncRefType()) {
+    return nullptr;
+  }
```
dannypsnl, line 694: What if we report "this is not a function"?

q82419: return nullptr, such as cast fail.

dannypsnl (approved):

PR #4923, opened 2026-06-02
fix(plugins/opencvmini): bound-check guest buffer length in host funcs

aizu-m:
## Description

The `wasmedge_opencvmini` host functions `imdecode`, `imshow`, `imwrite` and `imencode` fetched the guest input buffer with `MemInst->getPointer<char *>(Ptr)`. That call only bounds-checks one byte at `Ptr`, then the code read a fully guest-controlled length (`BufLen` / `WindowNameLen` / `TargetFileNameLen` / `ExtLen`) starting from that pointer.

A length larger than the guest allocation reads past the end of the WebAssembly linear memory into adjacent host process memory, i.e. an out-of-bounds read / information leak.

Fixed at all four sites by fetching the input with `getSpan<char>(Ptr, Len)` and returning `HostFuncError` when the returned span size does not equal the requested length. This mirrors the output-buffer `getSpan` check that already exists in `imencode`.

## Checklist

- [x] **DCO Signed-off**: the commit carries a `Signed-off-by` line, DCO check is green.
- [x] **Commit Messages**: `commitlint` passes on the commit (conventional commit, `fix(plugins/opencvmini): ...`).
- [x] **Local Tests Passed**: lint/format checks run clean locally, logs below.

## Test Evidence

`clang-format-20` dry-run on the changed file (this was the failing linter):

```
$ clang-format-20 -style=file -Werror --dry-run plugins/wasmedge_opencvmini/opencvmini_func.cpp
$ echo $?
0
```

`commitlint` on the commit message:

```
$ git log -1 --format=%B | npx commitlint --verbose
⧗   input: fix(plugins/opencvmini): bound-check guest buffer length in host funcs
✔   found 0 problems, 0 warnings
```

The plugin needs OpenCV to compile, so the full functional build is left to CI.

### DIFF
--- a/plugins/wasmedge_opencvmini/opencvmini_func.cpp
+++ b/plugins/wasmedge_opencvmini/opencvmini_func.cpp
@@ -24,9 +24,12 @@ WasmEdgeOpenCVMiniImdecode::body(const Runtime::CallingFrame &Frame,
     return Unexpect(ErrCode::Value::HostFuncError);
   }
 
-  char *Buf = MemInst->getPointer<char *>(BufPtr);
+  auto Buf = MemInst->getSpan<char>(BufPtr, BufLen);
+  if (unlikely(Buf.size() != BufLen)) {
+    return Unexpect(ErrCode::Value::HostFuncError);
+  }
 
-  std::vector<char> Content(Buf, Buf + BufLen);
+  std::vector<char> Content(Buf.begin(), Buf.end());
   cv::Mat Img = cv::imdecode(cv::InputArray(Content), cv::IMREAD_COLOR);
 
   return Env.insertMat(Img);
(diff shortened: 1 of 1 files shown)


dannypsnl (commented):
The changes is ok, but you need to put your email (in the right format) in every commit message to pass DCO check.

aizu-m:
Good catch, the sign-off was missing. Amended with a proper Signed-off-by line and force-pushed, DCO should pass now.

dannypsnl (dismissed):

hydai:
The basic linter issue has not been resolved, and the PR is not following our template.

aizu-m:
Sorted both points.

Linter: the violation was `opencvmini_func.cpp:199`, the `std::copy_n` line went one column past 80 after the getSpan change. clang-format-20 `-Werror --dry-run` on the file is clean now. commitlint was failing with type-empty/subject-empty because neither the commit nor the PR title carried a conventional-commit type. Retitled to `fix(plugins/opencvmini): ...` and amended the commit to match, both pass commitlint locally.

Template: filled in the description, checklist and test evidence.

The refreshed workflow runs are sitting in action_required, so they'll need approval to actually execute.

hydai (approved):

PR #4935, opened 2026-06-05
Fix dangling sd_image_t buffers in readControlImage and readMaskImage

aizu-m:
Spotted this reading through the stable-diffusion image readers. Two of them hand back an sd_image_t whose data pointer is already dead.

readControlImage frees the decoded buffer unconditionally:

    ControlImg->data = ControlImageBuffer;   // CannyPreprocess == false
    free(ControlImageBuffer);                // frees the buffer it returns
    return ControlImg;                        // ->data now dangles

img2img and txt2img then read ControlImage->data. readMaskImage is the same shape in its empty-mask branch: it returns Arr.data() of a std::vector local to the function, so the buffer is gone the moment it returns.

A quick clobber after each reader confirms it: the default mask reads 0xAB instead of 255, and the control buffer reads back garbage once the freed chunk is reused.

Free the decoded buffer only when preprocess_canny actually replaced it, and malloc the default mask so it outlives the reader.

### DIFF
--- a/plugins/wasmedge_stablediffusion/sd_func.cpp
+++ b/plugins/wasmedge_stablediffusion/sd_func.cpp
@@ -104,8 +104,9 @@ sd_image_t *readControlImage(Span<uint8_t> ControlImage, int Width, int Height,
     ControlImg->data =
         preprocess_canny(ControlImg->data, ControlImg->width,
                          ControlImg->height, 0.08f, 0.08f, 0.8f, 1.0f, false);
+    // preprocess_canny returns a fresh buffer, so the decoded one is now unused.
+    free(ControlImageBuffer);
   }
-  free(ControlImageBuffer);
   return ControlImg;
 }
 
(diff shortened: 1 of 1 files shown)


aizu-m:
gentle ping

hydai:
Just wanted to know if you are still working on this.

aizu-m:
Yes, still on it. Just rebased onto master and pushed an update.

Went back over the two readers with the ownership in mind. The Copilot pass was right that the first cut only moved the bug: once the readers keep their buffers alive, nobody frees them after the generation call, so it traded a UAF for a per-request leak.

Traced who owns what through stable-diffusion.cpp:
- txt2img/img2img only copy the pixels in (sd_image_to_tensor / sd_mask_to_tensor). They never take ownership or free the inputs.
- The upstream CLI confirms it: it frees control_image_buffer and input_image_buffer itself after generating.

So this push:
- readControlImage frees the decoded buffer only when preprocess_canny actually replaced it, same as before the UAF was there.
- txt2img and img2img now free ControlImage->data and (in img2img) MaskImage.data once the generation call returns, alongside the existing free(ControlImage)/free(InputImageBuffer). That covers the malloc'd default mask and the stbi-loaded mask, which was already leaking.
- Dropped the narrating inline comment per the style note.

On the channel observation: the non-empty branches decode 3-channel while the field reports 1, but that predates this change and sd_mask_to_tensor keys off the buffer, so I left it out to keep this focused on the lifetime bug.

Built the plugin locally with -Werror, clean.

PR #4936, opened 2026-06-05
fix(common): resolve Clang compilation error in uint128_fallback subtraction

pramila-bishnoi:
## Description

This PR fixes a compilation failure encountered when building WasmEdge with developer tests enabled on arm64 macOS using a standard LLVM/Clang toolchain. 

The custom `uint128_fallback` structure lacks a native unary bitwise NOT (`~`) operator, which causes modern strict compilers to fail with an `invalid argument type` error when trying to evaluate the original `~Result` expression. 

This fix explicitly extracts and applies the bitwise NOT operator to the underlying 64-bit halves (`~RHS.high()` and `~RHS.low()`), correctly calculating the Two's Complement subtraction values. This bypasses the missing unary expression entirely and prevents an infinite template recursion loop during evaluation.

## Checklist

## Test Evidence

The project compilation and full test runner linking phases were executed locally on an Apple Silicon (M1/M2/M3) architecture target. The environment was successfully cleared of any library path pollution, and the entire build completed without compilation or symbol resolution failures:

```text
[244/244] Linking CXX executable test/api/wasmedgeAPIAOTCoreTests
Build completed successfully.

### DIFF
--- a/include/common/int128.h
+++ b/include/common/int128.h
@@ -581,8 +581,8 @@ operator+(detail::uint128_fallback LHS, unsigned int RHS) {
 
 inline constexpr detail::uint128_fallback
 operator-(unsigned int LHS, detail::uint128_fallback RHS) {
-  uint128_fallback Result = RHS;
-  return (~Result) + 1 + LHS;
+  detail::uint128_fallback Negated = {~RHS.high(), ~RHS.low()};
+  return Negated + 1 + LHS;
 }
 
 inline constexpr detail::uint128_fallback &


hydai (changes requested):
Please provide the environment you used to encounter these errors.  
We do have an arm64 macOS CI environment without any issues.

pramila-bishnoi:
Hi @hydai,

To clarify, my local machine had an active Miniconda distribution installed and its path overrides polluted the toolchain discovery during the CMake configuration phase. this accidentally forced the build toolchain to use the `uint128_fallback` structure instead of native `__int128` types.

While a clean arm64 macOS setup bypasses this fallback file entirely, dropping into this pathway because of Miniconda exposed a genuine bug in the fallback subtraction logic. Because `uint128_fallback` doesn't have an overloaded unary `~` operator, using `~Result` breaks the build on modern strict compilers. 

This PR simply patches that low-level fallback arithmetic so that if any platform, toolchain, or conda-polluted configuration ever needs to rely on the fallback pathway in the future, it works reliably without crashing.

hydai:
I see. Thank you for the clarification.

hydai (approved):

hydai:
Thanks!

pramila-bishnoi:
Thank you for the review and for merging! Glad I could help improve the fallback path. Looking forward to contributing more! 😊

PR #4941, opened 2026-06-08
fix(plugin/wasi_nn): clamp get_output copy to guest buffer size

aizu-m:
## Description

Found this while reading the wasi-nn `get_output` / `get_output_single` path. The output buffer reaches each backend as a `Span<uint8_t>` whose size is the guest-supplied `OutBufferMaxSize`. Most backends clamp the copy to that size, a few do not.

1. `get_output` hands the backend `OutBuffer`, sized to the guest's `OutBufferMaxSize`.
2. GGML, MLX, ChatTTS, OpenVINO and OpenVINOGenAI pass the source length to `std::copy_n` (LlamaOutputs, output metadata, the MLX answer/text bytes, the OpenVINO output tensor, the OpenVINOGenAI string output) rather than `OutBuffer.size()`.
3. A guest that hands over a buffer smaller than the produced output gets a write past the span, off the end of the linear memory into adjacent host memory.

Observation: the RPC branch in `wasinnfunc.cpp` and the bitnet/whisper/piper/torch/tfl backends already guard this with `std::min(OutBuffer.size(), ...)`. These five were the ones left unclamped, so this brings them in line.

`BytesWritten` still reports the full output length so a caller can spot the truncation, resize and read again.

## Checklist

- [x] **DCO Signed-off**: commit is signed off.
- [x] **Commit Messages**: conventional-commit subject, checked with commitlint.
- [x] **Local Tests Passed**: see below.

## Test Evidence

Reduced the construct to the same shape (4-byte destination span, 4096-byte source) to confirm the overrun and the clamp:

```
OutBuffer.size()=4 ModelOutput.size()=4096
buggy copy_n count=4096 -> writes 4092 bytes past the span
clamped copy_n count=4 -> in bounds
```

### DIFF
--- a/plugins/wasi_nn/GGML/core/output_generator.cpp
+++ b/plugins/wasi_nn/GGML/core/output_generator.cpp
@@ -27,7 +27,9 @@ Expect<ErrNo> getOutputSingle(WasiNNEnvironment &Env, uint32_t ContextId,
   // Use index 1 for output metadata.
   if (Index == 1) {
     std::string Metadata = buildOutputMetadata(CxtRef);
-    std::copy_n(Metadata.data(), Metadata.length(), OutBuffer.data());
+    const size_t BytesToCopy =
+        std::min(static_cast<size_t>(OutBuffer.size()), Metadata.length());
+    std::copy_n(Metadata.data(), BytesToCopy, OutBuffer.data());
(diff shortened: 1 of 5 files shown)


hydai:
Thanks for this PR. This will be replaced by #5091, which unified all wasi-nn backends in the same way to handle this kind of issues.

PR #4953, opened 2026-06-09
fix: correct table.copy and memory.init/copy index encoding

suhr25:
## Description

This PR fixes a few serializer issues where the emitted WASM bytes didn't match what the loader expects. While testing bulk memory instructions, I found that `table.copy` was missing its source table index, `memory.init` was writing the wrong index, and `memory.copy` wasn't validating the source index correctly.
The fix brings the serializer back in sync with the loader and adds regression tests to cover all three cases.

## Checklist

* [x] DCO Signed-off
* [x] Commit message follows project guidelines
* [x] Local tests passed

## Test Evidence

Added regression tests for `table.copy`, `memory.init`, and `memory.copy`. All new tests fail before the fix and pass after it, with existing serializer tests remaining green.


q82419:
Nice catch.

q82419 (approved):

PR #4956, opened 2026-06-10
feat(installer): add --no-modify-shell-profile flag

### LINKED ISSUE #4955: feat: provide option to not write to .bashrc
### Summary

I think it would be good to allow the user to decide if they want the source env line in their home profile file or not.

### Details

There is already a path flag, to allow the user to not install into their home directory, potentially allowing a local installation within a project with a specific version.

The change to the bashrc in the user's home folder does not align well with the idea of fully localized installation. It still assumes one installation per user. 

If the user tries to install to multiple path, multiple entries are added to the bashrc.

```bash
# install twice
$ cd $(mktemp -d)
$ curl -sSf https://raw.githubusercontent.com/WasmEdge/WasmEdge/master/utils/install.sh | bash -s - -p .
$ cd $(mktemp -d)
$ curl -sSf https://raw.githubusercontent.com/WasmEdge/WasmEdge/master/utils/install.sh | bash -s - -p .
# check path
$ grep /tmp/tmp ~/.bashrc
. "/tmp/tmp.6MVvmXvAaw/env"
. "/tmp/tmp.HtgET3Y0tN/env"
```

Additionally, if not careful, the user might not be fully aware of this and run into bugs where another version is used than the one they think.

### Appendix

_No response_

bluebrown:
## Description
Closes #4955


hydai (changes requested):
Since this will extend the behavior of the installer, could you also add a test to cover this new feature?

utils/install.py
```diff
@@ -1523,6 +1528,13 @@ def main(args):
         default=PATH,
         help="Installation path for WasmEdge",
     )
+    parser.add_argument(
+        "--shell-config",
+        dest="configure_shell",
+        default=[True],
+        help="Run uninstaller script before installing",
```
bluebrown: oops need to change that too. Help text is not correct.

bluebrown:
> Since this will extend the behavior of the installer, could you also add a test to cover this new feature?

I added a test, but I only tested it locally on my machine. The code could fail on older versions of Python.
But I am a bit hesitant to enable the actions on my fork because there are so many.

Maybe I can add a manual workflow dispatch to the test-installer job. Not sure.

hydai:
I will trigger it from my side. If it fails, I think an alternative way is to add a temporary commit that removes all other workflows except for the installer. This will be easier for you to verify.

bluebrown:
i got the workflows to run. You dont need to trigger it. Will post screenshot once all are green

bluebrown:
@hydai , I managed to make all tests green. So many issues due to different shell syntax.. https://github.com/bluebrown/WasmEdge/actions/runs/27328565250/job/80735152366

I have restored the original workflow trigger and rebased, after the test.

.github/scripts/installer-test-helpers.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
hydai: I don't think this should be changed.

.github/workflows/test-installers.yml
```diff
@@ -5,6 +5,7 @@ concurrency:
   cancel-in-progress: true
 
 on:
+  workflow_dispatch: {}
```
hydai: Remove this line.

hydai (changes requested):

.github/scripts/installer-test-helpers.sh
bluebrown: regarding shell issues. I did change this here, but actually now I see this is still wrong.
It should be regular sh for max POSIX compliance as this runs also on mac and likely on zsh.
I had to install zsh locally to find some differences even to something like dash which is POSIX compliant.

So while a regular sh shebang is still not 100% accurate its still better than bash which is definitly wrong.

.github/workflows/test-installers.yml
bluebrown: i would recommend to leave it, because i hard such hard time getting this one to trigger. Next time someone wants to make a PR they have better times. But If you want i will remove it.

.github/scripts/installer-test-helpers.sh
bluebrown: sorry I was typing at the same time as you https://github.com/WasmEdge/WasmEdge/pull/4956#discussion_r3393838687.

Generally going over `#!//usr/bin/env <your shell>` is better for MAC compatibility.

However, this script is only ever sourced, so the shebang is not doing anything in the CI. But for local
development, the shebang can guide to find errors with tools like shellcheck. As mention in the other comment
I think it should be a `sh` for max compat.

.github/workflows/test-installers.yml
hydai: Sounds good. Let's keep it :-)

bluebrown: lol now I removed it again. I also reverted the shebang. Trying to go for minimum invasion.
Perhaps I can make another PR for CI fix improvement. Treating this only for a feature.

bluebrown:
@hydai , I think those message from copilot are out of scope, for this PR. I would say It is ready for merging.

.github/workflows/test-installers.yml
```diff
@@ -478,6 +478,21 @@ jobs:
           echo "Test 3: Deprecated version handling"
           ${python_cmd} utils/install.py -v ${DEPRECATED_TEST_VERSION} -D 2>&1 | grep -q "Version not supported. Min Version: 0.13.0" && \
             echo "[PASS] Version validation works" || (echo "[FAIL] Version validation failed" && exit 1)
+
+          # Test shell config
+          printf "\n=== Test 4: Shell config (default) ===\n"
+          tempdir="$(mktemp -d)"
+          ${python_cmd} utils/install.py -p "$tempdir"
+          verify_shell_config "$tempdir" "1"
+          bash utils/uninstall.sh -q -V -p "$tempdir"
```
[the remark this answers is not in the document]
bluebrown: If we fix this one, then we should fix all tests in the pipeline, I guess. This is the established
pattern. They all use the default $HOME. Hence, I think it's out of scope. If we think it's an issue,
Then this could be worth its own issue + PR to address it in the entire CI.

utils/install.py
```diff
@@ -1393,14 +1393,16 @@ def main(args):
         logging.debug("CONST_urls: %s", CONST_urls)
         logging.debug("CONST_lib_dir: %s", CONST_lib_dir)
 
-        if getenv("SHELL") != SHELL:
-            logging.warning("SHELL variable not found. Using %s as SHELL", SHELL)
+        if args.configure_shell:
+            env_shell = getenv("SHELL")
+            if env_shell != SHELL:
+                logging.warning(f"Unexpected SHELL: {env_shell}. Using {SHELL} as SHELL")
```
hydai: Is this compatible with Python 2?

hydai (changes requested):

utils/install.py
bluebrown: hm, got point. I dont think so. I changed to to something thats for sure compatible.

utils/install.py
```diff
@@ -1523,6 +1529,13 @@ def main(args):
         default=PATH,
         help="Installation path for WasmEdge",
     )
+    parser.add_argument(
+        "--no-shell-config",
```
hydai: I prefer `--no-modify-shell-profile`.

utils/install.py
```diff
@@ -1523,6 +1529,13 @@ def main(args):
...
+        dest="configure_shell",
```
hydai: use `modify_shell_profile`

utils/install.py
```diff
@@ -1523,6 +1529,13 @@ def main(args):
...
+        default=True,
+        help="Do not add env file source entry to detected shell i.e. .bashrc",
```
hydai: `Do not modify shell profile/startup files; source the generated env file manually`

hydai (changes requested):

hydai:
I think this PR should be ready to merge once the above three comments regarding the naming issue are resolved.

CC @SAtacker, would you like to take a look at this PR?

utils/install.py
bluebrown: sorry didnt see this one. Done

hydai (approved):
LGTM

PR #4962, opened 2026-06-11
fix: serialize tag export index correctly

### LINKED ISSUE #4961: bug: Incorrect 0x00 attribute byte injection in Tag ExportDesc serialization
### Summary

The C++ core serializer is incorrectly injecting a `0x00` attribute byte into `Tag` type `ExportDesc` serializations, causing the output binary to violate the WebAssembly exception handling specification.

### Current State

In `lib/loader/serialize/serial_description.cpp`, when serializing an `AST::ExportDesc` of type `ExternalType::Tag`, the code executes:
`return serializeType(Desc.getExternalIndex(), OutVec);`
Since `getExternalIndex()` returns a `uint32_t`, C++ implicitly constructs an `AST::TagType` from the integer and dispatches to the `serializeType(const AST::TagType &)` overload. This overload purposefully injects an `0x00` attribute byte into the binary. While this is mathematically correct for an `importdesc`, it is strictly invalid for an `exportdesc`.

### Expected State

According to the Wasm Exception Handling proposal, the `exportdesc` for a Tag must solely contain `0x04 tagidx` (a raw `u32` representing the index, with absolutely no attribute byte). The serializer should bypass the `serializeType` dispatch and directly serialize the `u32` index.

### Reproduction steps

1. Initialize a `WasmEdge::Configure` object with the `ExceptionHandling` proposal enabled.
2. Create an `AST::ExportDesc` object of type `Tag` with a valid index (e.g., `0x02`).
3. Serialize the `ExportDesc` using the `Serializer` class.
4. Observe that the serialized binary output contains an erroneous `0x00` byte preceding the `u32` index (e.g., outputting `0x04 0x00 0x02` instead of the spec-compliant `0x04 0x02`).

KrishnaParihar1:
Fixes #4961
## Description

Fix tag export serialization in `serializeDesc(const AST::ExportDesc &Desc, ...)`.

The tag export path incorrectly used `serializeType()` for the export index. This change replaces it with `serializeU32()` so tag exports are serialized consistently with other export kinds and according to the expected export index encoding.

## Checklist

* [x] **DCO Signed-off**: All commits are signed-off (`git commit -s`). CI workflows will only be approved if DCO check is passed.
* [x] **Commit Messages**: Run `commitlint` on your commit messages to ensure they meet the Conventional Commit standards.
* [x] **Local Tests Passed**: Provide a screenshot or logs below to prove that you have passed the test suites locally.

## Test Evidence

```bash
./build/test/loader/wasmedgeLoaderSerializerTests --gtest_filter=SerializeDescriptionTest.*
```

```text
[ RUN      ] SerializeDescriptionTest.SerializeImportDesc
[       OK ] SerializeDescriptionTest.SerializeImportDesc

[ RUN      ] SerializeDescriptionTest.SerializeExportDesc
[       OK ] SerializeDescriptionTest.SerializeExportDesc

[  PASSED  ] 2 tests.
```


test/loader/serializeDescriptionTest.cpp
```diff
@@ -235,16 +213,17 @@ TEST(SerializeDescriptionTest, SerializeExportDesc) {
...
   EXPECT_TRUE(Ser.serializeSection(createExportSec(Desc), Output));
   Expected = {
-      0x07U,                                           // Export section
-      0x0BU,                                           // Content size = 10
-      0x01U,                                           // Vector length = 1
-      0x06U, 0x4CU, 0x6FU, 0x61U, 0x64U, 0x65U, 0x72U, // External name: Loader
-      0x04U,                                           // Tag type
-      0x00U, 0x02U                                     // TypeIdx value
+      0x07U,
+      0x0AU,
```
dannypsnl, line 217: Sorry, it seems you only change this line and rest are removing comments without reason.

Then why is this a fix of any problem?

hydai:
Closing this and marking it as a spam PR because it attempts to remove existing comments that indicate the test case meanings without any reason.

Issue #4441, opened 2025-12-13
feat: Component Model: validator implementation tracking issue


LFsWang:
### Summary

This issue tracks the implementation and testing status of validator in component-model proposal.

### Details

Because the validator is under the developing, we can add the test but pass it before we finish the validator.

- [x] Convert the [WAST test files](https://github.com/WebAssembly/component-model/tree/main/test/wasm-tools) into the test suite format in [WasmEdge-spectest repo](https://github.com/WasmEdge/wasmedge-spectest) such as the core WASM tests (self-defined json format allowed), and create PR to the WasmEdge-spectest repo.
- [x] Add the spec test codes in this repo and pass the tests.

Refer to the parent issue #4236 for the validator implementation status for every sections.

Implementation guide:

1. Task issue base on Section or more detailed structure.
2. List the required spec test entries in that issue to pass.
3. For the PR, modify the boolean of validation phase about the required spec tests in `ComponentModelFolders` map to `true` in the `test/spec/spectest.cpp`.
4. Run the component model spec test and pass to prove your implementation.
5. For the trap messages not reasonable in the spec test suite, it's fine to standardize them and update on the wasmedge-spectest repo.

The passing status of spec tests in [WasmEdge-spectest repo](https://github.com/WasmEdge/wasmedge-spectest):
* 🆎 : the tests do not pass with on error message mismatch only.

- [x] [`adapt.wast`](https://github.com/WebAssembly/component-model/blob/main/test/wasm-tools/adapt.wast)
- [x] [`alias.wast`](https://github.com/WebAssembly/component-model/blob/main/test/wasm-tools/alias.wast)
- [x] [`big.wast`](https://github.com/WebAssembly/component-model/blob/main/test/wasm-tools/big.wast)
- [x] [`definedtypes.wast`](https://github.com/WebAssembly/component-model/blob/main/test/wasm-tools/definedtypes.wast)
- [x] [`empty.wast`](https://github.com/WebAssembly/component-model/blob/main/test/wasm-tools/empty.wast)
- [x] [`example.wast`](https://github.com/WebAssembly/component-model/blob/main/test/wasm-tools/example.wast)
- [x] [`export-ascription.wast`](https://github.com/WebAssembly/component-model/blob/main/test/wasm-tools/export-ascription.wast)
- [x] [`export-introduces-alias.wast`](https://github.com/WebAssembly/component-model/blob/main/test/wasm-tools/export-introduces-alias.wast)
[...]

PR #4967, opened 2026-06-11
feat(validator): enable component-model naming spec tests

parthdagia05:
This enables the `naming` folder of the component-model spec tests (#4441).

While digging into why it was disabled, I found that WasmEdge was already rejecting 14 of the 15 invalid cases it just reported generic error codes ("invalid component name", "duplicate name in component", "export not found") whose text didn't line up with what the spec tests expect. The harness matches the expected message as a prefix of our ErrCodeStr, so those cases failed even though validation itself was correct.

So most of this PR is making the diagnostics specific. I added four error codes:

  - ComponentNameNotKebab            -> "not in kebab case"
  - ComponentPackageNameNotLowercase -> "not lowercase in package name/namespace"
  - ComponentImportNameConflict      -> "import name conflicts with previous name"
  - ComponentUnknownExport           -> "unknown export"

and routed the label / interface-name / package / import-uniqueness / alias-export validation sites to them.

There was also one real bug: instance inline-export names weren't being validated at all, so an invalid name like "1" was accepted. I added a validateExportName() check in the component-instance inline-export loop, which
is what naming.1 exercises.

Testing:
  - the `naming` spec test passes now
  - the full spec test suite still passes (525 tests), including every other
    component-model folder, so no regression
  - clang-format (20) is clean

I kept the scope to load+validate for this folder; instantiate/execute stay off.

Part of #4441.

### DIFF
--- a/include/common/enum.inc
+++ b/include/common/enum.inc
@@ -1120,6 +1120,16 @@ E(InvalidExportName, 0x02B3, "not a valid export name")
 E(InvalidExternName, 0x02B4, "not a valid extern name")
 // Core tag referenced by a core-export alias or inline-export does not exist
 E(UnknownCoreTag, 0x02B5, "unknown tag")
+// Component name label is not in kebab case
+E(ComponentNameNotKebab, 0x02B6, "not in kebab case")
+// Component package name/namespace is not lowercase
+E(ComponentPackageNameNotLowercase, 0x02B7,
+  "not lowercase in package name/namespace")
+// Component import name conflicts with a previous import name
+E(ComponentImportNameConflict, 0x02B8,
+  "import name conflicts with previous name")
+// Component alias references an export the instance does not provide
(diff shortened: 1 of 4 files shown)


parthdagia05:
@dannypsnl @ibmibmibm @q82419 can you review this PR

dannypsnl (approved):

parthdagia05:
@dannypsnl i have rebased the branch, and the check failure was not from my code

hydai:
@q82419  
You should ask the contributor to cherry-pick their commits to `dev/component`.  
Directly changing the merge target may not work because `dev/component` is behind the master branch.

q82419:
@dannypsnl  @parthdagia05 
Currently we will merge the component model related PRs into `dev/component`, and regularly merge the branch into master once the changes are stable.

PR #4978, opened 2026-06-12
fix(serializer): correctly encode sub clause for open SubTypes without supertypes & adding a new test case (`SubType3`)

### LINKED ISSUE #4977: bug: Serializer incorrectly drops 'sub' prefix for open GC SubTypes without supertypes
### Summary

When the WasmEdge Serializer encounters an open (non-final) WebAssembly GC SubType that has zero supertypes (i.e. `sub [] <comptype>`), it incorrectly skips serializing the `0x50 0x00` prefix. This forces the type to be abbreviated to just `<comptype>`, which fundamentally changes its semantic meaning from `open` to `final`.

### Current State

In the WebAssembly GC Proposal, an open type with no parents is defined as `sub [] <comptype>` and encoded with prefix `0x50 0x00`. A final type is `sub final [] <comptype>`, which the spec allows to be abbreviated to just `<comptype>`.

Currently, in `lib/loader/serialize/serial_type.cpp`, the Serializer only checks if `SType.getSuperTypeIndices().size() > 0` before deciding to serialize the `sub` prefix encoding. 

https://github.com/WasmEdge/WasmEdge/blob/master/lib/loader/serialize/serial_type.cpp#L169-L170
```cpp
  // Sub type: vec(typeidx)
  if (SType.getSuperTypeIndices().size() > 0) {
    if (!Conf.hasProposal(Proposal::GC)) {
```

Because an open type with no supertypes has a size of `0`, the serializer skips the `0x50 0x00` encoding entirely. WasmEdge therefore mutates a developer's open, inheritable type into a locked, final type during serialization, which will break all subsequent subtyping validation when the binary is reloaded.

### Expected State

The Serializer should correctly evaluate if the type is explicitly open (`!SType.isFinal()`) and forcibly write the `0x50 0x00` encoding prefix, regardless of whether the supertypes vector is empty. The corrected logic should be:   

```cpp
 // Sub type: vec(typeidx)
- if (SType.getSuperTypeIndices().size() > 0) {
+ if (SType.getSuperTypeIndices().size() > 0 || !SType.isFinal()) {
    if (!Conf.hasProposal(Proposal::GC)) {
```

### Reproduction steps

1. Parse a Wasm module containing an open GC type with no parents (e.g. sub [] <comptype>).
2. Serialize the module back to binary using WasmEdge's serialization API.
3. Reload the serialized module.
4. The open type will be incorrectly deserialized as `final` because the required 0x50 0x00 prefix was dropped.

Sukuna0007Abhi:
## Description

This PR fixes a critical semantic bug in the WasmEdge Serializer related to the WebAssembly GC Proposal, where open (non-final) types with no supertypes were being incorrectly forced into becoming `final` types. 

When defining an open GC type with no supertypes (`sub [] <comptype>`), the spec requires the `0x50 0x00` prefix in the binary. Previously, `Serializer::serializeType()` only wrote this prefix if the `SuperTypeIndices` vector had a size greater than 0. Because our open type had exactly 0 supertypes, the serializer skipped the prefix and just outputted `<comptype>`. Since the spec explicitly defines a bare `<comptype>` as an abbreviation for a *final* type (`sub final [] <comptype>`), any Wasm engine reading the resulting binary would misinterpret the type as final—completely destroying the developer's intent and breaking subsequent subtyping validation!

I updated the serialization condition to explicitly check if the type is open (`if (SType.getSuperTypeIndices().size() > 0 || !SType.isFinal())`), logically guaranteeing that the required prefix is forcibly written. I also added a new test case (`SubType3`) to prove that open types with empty supertypes now correctly serialize to their exact expected byte sequences.

Resolves #4977 

## Checklist


Sukuna0007Abhi:
Plz also review this after that pr, thanks @hydai @q82419 @dannypsnl

dannypsnl:
You mean the serializer make a non-final type be final, am I read this correct?

Sukuna0007Abhi:
> You mean the serializer make a non-final type be final, am I read this correct?

Yes, exactly! 

If a user defines an explicitly open type with 0 supertypes (`sub [] <comptype>`), the Serializer was skipping the `0x50 0x00` prefix entirely because `SType.getSuperTypeIndices().size()` was `0`. 

Because the Serializer outputs the bare `<comptype>` without the `0x50` prefix, WasmEdge's own Deserializer (`lib/loader/ast/type.cpp`) will fall into the `default` case and explicitly force the type to become final:

```cpp
  switch (static_cast<TypeCode>(B)) {
  default:
    // Case: comptype.
    SType.setFinal(true);
    return loadCompositeType(SType.getCompositeType());
  case TypeCode::Sub:
    // Case: 0x50 vec(typeidx) comptype.
    SType.setFinal(false);
    break;
```
Because of this abbreviation rule in the spec, the serializer was silently mutating explicitly open, non-final types into final types!

dannypsnl (dismissed):

q82419:
Your CI failed.

Sukuna0007Abhi:
> Your CI failed.

I fixed it now, pls check

Sukuna0007Abhi:
During CI investigation, I found that the test failures on Linux and macOS  (`SerializeSubType`, `SerializeCompositeType`) were caused by undefined behavior in `AST::SubType`: `bool IsFinal` had no default initializer.

On Linux/macOS, the stack memory is zeroed so `IsFinal` defaults to `false`, causing the serializer to incorrectly emit a `0x50` Sub wrapper for every default-constructed `SubType`, even ones with no supertypes. On Windows the 
stack is dirty, so the bug was masked locally.

Fixed by adding `= true` as the default initializer for `IsFinal` in `include/ast/type.h`. This aligns with the WebAssembly GC spec where types without an explicit `sub` declaration are implicitly **final**, and matches the behavior of the existing `SubType(const FunctionType &FT)` constructor which already explicitly sets `IsFinal(true)`.

All 49 serializer tests now pass across platforms.

Sukuna0007Abhi:
Ci fixed , pls check @dannypsnl  @q82419

Sukuna0007Abhi:
#5001 dupilcate pr of my pr

q82419:
#5001 has correct comments on it's test case patterns.

Sukuna0007Abhi:
> #5001 has correct comments on it's test case patterns.

This is unfair, I reported the issue 1st then give a fix, and now someone duplicates and mine got closed because of comments, If I add comments can my pr be merged

PR #4981, opened 2026-06-12
feat(validator): enable component-model invalid spec tests

parthdagia05:
A core:moduledecl import that references an out-of-range core type index now reports "type index out of bounds" instead of the generic "invalid index", matching the spec diagnostic. Flip the invalid folder to validate=true.
This enables the `invalid` folder of the component-model spec tests (targeting dev/component).

The folder was already almost there  WasmEdge rejects all the invalid cases correctly. The only mismatch was one diagnostic: when a core:moduledecl import references a core type index that's out of range, e.g.

    (core type (module (import "" "" (func (type 1)))))   ;; no type 1 exists

WasmEdge reported the generic "invalid index", but the spec test expects "type index out of bounds".

So I added a CoreTypeIndexOutOfBounds error code and used it at that check in the CoreImportDesc validation, then flipped the invalid folder to validate=true.

Testing:
  - the invalid spec test passes
  - the full spec test suite still passes (525 tests), no regression
  - clang-format (20) clean

Signed-off-by: Parth Dagia <parth.24bcs10414@sst.scaler.com>

### DIFF
--- a/include/common/enum.inc
+++ b/include/common/enum.inc
@@ -1130,6 +1130,8 @@ E(ComponentImportNameConflict, 0x02B8,
   "import name conflicts with previous name")
 // Component alias references an export the instance does not provide
 E(ComponentUnknownExport, 0x02B9, "unknown export")
+// Core type index referenced in a component core:moduledecl is out of bounds
+E(CoreTypeIndexOutOfBounds, 0x02BA, "type index out of bounds")
 // @}
 
 // Instantiation phase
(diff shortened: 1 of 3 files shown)


parthdagia05:
@q82419 @dannypsnl can you take a look, also have commited to dev/component .

dannypsnl (approved):

dannypsnl:
@parthdagia05 BTW, you don't have to pin people account, the code owner rule will assign reviewer, that mean we already have a notification once.

parthdagia05:
> @parthdagia05 BTW, you don't have to pin people account, the code owner rule will assign reviewer, that mean we already have a notification once.

sure !

PR #4983, opened 2026-06-12
fix: correct SIMD immediate byte order for v128.const and i8x16.shuffle

### LINKED ISSUE #4973: bug: Serializer outputs backwards memory byte layout for i8x16.shuffle instruction
### Summary

When serializing an AST back into a WebAssembly binary, WasmEdge writes the 16 immediate bytes of the `i8x16.shuffle` instruction in reverse order. This results in an invalid WebAssembly binary where the SIMD lane shuffle indices are completely reversed, silently corrupting SIMD vector mathematical operations.

### Current State

In `lib/loader/serialize/serial_instruction.cpp` at line 529, the serializer unpacks the `uint128_t` memory layout backwards:
```cpp
  case OpCode::I8x16__shuffle: {
    uint128_t Value = Instr.getNum().get<uint128_t>();
    const std::uint8_t *Ptr = reinterpret_cast<const uint8_t *>(&Value);
    for (uint32_t I = 0; I < 16; ++I) {
      OutVec.push_back(Ptr[15 - I]); // <--- BUG: outputs backwards
    }
    return {};
  }
```
Because the `Loader` packs the bytes into `uint128_t` sequentially (`Value |= byte << (I * 8U)`), the serializer must also unpack and emit them sequentially. `Ptr[15 - I]` causes the highest-memory byte to be emitted first.

### Expected State

The serializer should loop forward sequentially, emitting `Ptr[I]`, exactly matching the parsing specification, to ensure the binary is not corrupted. In addition, the test suite (`test/loader/serializeInstructionTest.cpp`) is currently missing a `SerializeSIMDInstruction` block to catch this regression. 

### Reproduction steps

1. Use `WasmEdge::Loader::Serializer` to serialize an AST containing `i8x16.shuffle` with a specific lane sequence (e.g. `0x00` to `0x0F`).
2. Inspect the generated binary block.
3. Observe that the 16 bytes immediately following the `0xFD 0x0D` opcode are emitted in `0x0F` to `0x00` reverse order.

master

suhr25:
## Description

This PR fixes byte-order handling for SIMD immediates in `v128.const` and `i8x16.shuffle`.The loader and serializer were using inconsistent encoding logic, causing serialized SIMD immediates to be written in reverse order. This could change constant values or shuffle lane indices after a load, serialize and reload round trip. 
The fix aligns the serializer with the loader's encoding logic to ensure correct round-trip behavior.

## Checklist

* [x] DCO Signed-off
* [x] Commit message follows project guidelines
* [x] Local tests passed

## Test Evidence

Added regression tests for `v128.const` and `i8x16.shuffle` using a non-symmetric 16-byte pattern. The tests fail before the fix and pass after it, with existing serializer tests remaining green.

### DIFF
--- a/lib/loader/serialize/serial_instruction.cpp
+++ b/lib/loader/serialize/serial_instruction.cpp
@@ -524,9 +524,8 @@ Serializer::serializeInstruction(const AST::Instruction &Instr,
   // SIMD Shuffle Instruction.
   case OpCode::I8x16__shuffle: {
     uint128_t Value = Instr.getNum().get<uint128_t>();
-    const std::uint8_t *Ptr = reinterpret_cast<const uint8_t *>(&Value);
     for (uint32_t I = 0; I < 16; ++I) {
-      OutVec.push_back(Ptr[15 - I]);
+      OutVec.push_back(static_cast<uint8_t>(Value >> (I * 8)));
     }
     return {};
   }
--- a/test/loader/serializeInstructionTest.cpp
+++ b/test/loader/serializeInstructionTest.cpp
@@ -1338,6 +1338,61 @@ TEST(SerializeInstructionTest, SerializeConstInstruction) {
   EXPECT_EQ(Output, Expected);
 }
 
+TEST(SerializeInstructionTest, SerializeSIMDConstInstruction) {
+  std::vector<uint8_t> Expected;
+  std::vector<uint8_t> Output;
+  std::vector<WasmEdge::AST::Instruction> Instructions;
+
+  // 13. Test SIMD const and shuffle instructions.
+  //
+  //   1.  Serialize V128__const instruction.
+  //   2.  Serialize I8x16__shuffle instruction.
+
+  WasmEdge::AST::Instruction V128Const(WasmEdge::OpCode::V128__const);
+  WasmEdge::AST::Instruction I8x16Shuffle(WasmEdge::OpCode::I8x16__shuffle);
+  WasmEdge::AST::Instruction End(WasmEdge::OpCode::End);
+
+  // Use an asymmetric 16-byte pattern so byte-order bugs are detectable.
+  WasmEdge::uint128_t Value = 0U;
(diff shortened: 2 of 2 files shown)


Sukuna0007Abhi:
Hey @suhr25, great minds think alike! I actually ran into this exact same serialization byte-order bug with SIMD immediates, raised issue #4973 yesterday, and submitted a fix for it earlier yesterday in PR #4975. 

cc @dannypsnl @hydai @q82419 - I wanted to gently bump this since PR #4975 was submitted earlier. If you feel it's appropriate, would it be possible to review that one PR #4975 first to avoid duplicate effort? 

I noticed @suhr25 also added a really great test for `v128.const`. If the codeowners & maintainers agree, maybe we could merge the core code fix from my PR (#4975), and then reduce this PR #4983 to just adding the extra `v128.const` test coverage? By this we both get our contribution,

Btw, I am happy to defer to whatever the team thinks is the best approach!

suhr25:
Thanks for pointing this out, and nice catch on the bug.
I wasn't aware of #4973 or #4975 when I was investigating this issue. I am happy to follow whatever approach the maintainers think makes the most sense.

q82419 (approved):

q82419:
> Hey @suhr25, great minds think alike! I actually ran into this exact same serialization byte-order bug with SIMD immediates, raised issue #4973 yesterday, and submitted a fix for it earlier yesterday in PR #4975.
> 
> cc @dannypsnl @hydai @q82419 - I wanted to gently bump this since PR #4975 was submitted earlier. If you feel it's appropriate, would it be possible to review that one PR #4975 first to avoid duplicate effort?
> 
> I noticed @suhr25 also added a really great test for `v128.const`. If the codeowners & maintainers agree, maybe we could merge the core code fix from my PR (#4975), and then reduce this PR #4983 to just adding the extra `v128.const` test coverage? By this we both get our contribution,
> 
> Btw, I am happy to defer to whatever the team thinks is the best approach!

I'd merge this PR because the earlier #4975 CI failed.

PR #4987, opened 2026-06-13
feat(validator): enable component-model memory64 spec tests

parthdagia05:
The validator tracked core export kinds but not types, so it ignored core memory index types (i32 vs i64). Check that each provided memory matches its import on instantiation, reject a non-32-bit canonical memory option, and carry memory types through core:instance exports and alias core export to support both. Flip memory64 to validate=true.

This enables the memory64 folder of the component-model spec tests (targeting dev/component).

These cases check that the validator respects core memory index types (i32 vs i64), which it currently ignores it tracks core export kinds but not their types, so:

  - instantiating a module that imports an i32 memory with a provider exporting an i64 memory (or vice versa) was wrongly accepted, and
  - a canon lift/lower could point its `memory` option at a 64-bit memory.

To fix it I carry a bit of the type info that GAP-CI-1 leaves out today:
  - core:instance exports now record the MemoryType for memory exports (not just the kind),
  - `alias core export` of a memory carries that type into the core-memory sort space,
  - core-module instantiation subtype-checks each provided memory against the import, and
  - validateCanonOptions rejects a non-32-bit canonical memory.

Scoped to memories (what memory64 exercises); the same plumbing is the natural place to extend to full func/table/global instance subtyping later.

Testing:
  - the memory64 spec test passes
  - the full spec test suite still passes (525 tests), no regression
  - clang-format (20) clean

### DIFF
--- a/include/common/enum.inc
+++ b/include/common/enum.inc
@@ -1130,6 +1130,12 @@ E(ComponentImportNameConflict, 0x02B8,
   "import name conflicts with previous name")
 // Component alias references an export the instance does not provide
 E(ComponentUnknownExport, 0x02B9, "unknown export")
+// Core instance arg memory index type does not match the import (i32 vs i64)
+E(ComponentMemoryIndexTypeMismatch, 0x02BA,
+  "mismatch in index type used for memories")
+// Canonical option `memory` must reference a 32-bit linear memory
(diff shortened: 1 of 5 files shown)


dannypsnl (approved):

q82419:
cc @LFsWang

PR #4990, opened 2026-06-14
feat(validator): check core global/table types on component instantiation

parthdagia05:
Extends the core-instance instantiation subtype checking (GAP-CI-1) from memory (added in #4987) to **global** and **table** for imported core module types:

  - global: reject if the provided global's value type or mutability differs from the import
  - table:  reject if the provided table's element type or limits don't satisfy the import

To do this, core:instance exports now carry the global/table type (stored by value the module-type descriptor getter returns a temporary, so a pointer would dangle).

Verified against the component-model `instantiate` spec tests these cases now pass: "type mismatch in global type", "mismatch in table limits", "type mismatch in table element type".

Notes:
- **Stacked on #4987** (memory64). Until that merges, this PR's diff includes #4987's commit;  I'll rebase once it lands so only this change remains.
- This does **not** flip the `instantiate` spec folder yet that folder also needs the component value-type subtyping (variant/record/tuple/flags/result), which is the larger GAP-CI-1 work. A focused unit test + the folder-enable will follow with that.
- Full spec suite still passes (525 tests); clang-format clean.

### DIFF
--- a/include/validator/component_context.h
+++ b/include/validator/component_context.h
@@ -81,7 +81,30 @@ public:
     // (GAP-CI-1).
     struct CoreInstanceExport {
       ExternalType Kind;
-      const AST::MemoryType *Mem = nullptr;
+      // Stored by value (not pointer): the module-type descriptor getter
+      // returns a temporary, so a pointer into it would dangle. Each optional
+      // is populated only for its matching Kind, for instantiation subtype
+      // checks (GAP-CI-1).
+      std::optional<AST::MemoryType> Mem;
+      std::optional<AST::TableType> Tab;
+      std::optional<AST::GlobalType> Glob;
+
+      CoreInstanceExport() noexcept = default;
+      // Copies whichever core extern type is provided (the others stay empty).
+      CoreInstanceExport(ExternalType K, const AST::MemoryType *M,
+                         const AST::TableType *T,
+                         const AST::GlobalType *G) noexcept
+          : Kind(K) {
+        if (M != nullptr) {
+          Mem = *M;
+        }
+        if (T != nullptr) {
+          Tab = *T;
+        }
+        if (G != nullptr) {
+          Glob = *G;
+        }
+      }
     };
 
     // ---- Scope identity ----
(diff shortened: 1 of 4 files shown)


parthdagia05:
i have built this PR on top of #4987 so once that gets merged i will rebase this

include/validator/component_context.h
```diff
@@ -243,14 +256,29 @@ class ComponentContext {
...
+    E.Kind = ET;
+    if (Mem != nullptr) {
+      E.Mem = *Mem;
+    }
+    if (Tab != nullptr) {
+      E.Tab = *Tab;
+    }
+    if (Glob != nullptr) {
+      E.Glob = *Glob;
+    }
```
dannypsnl: Maybe gives a constructor?

parthdagia05: done added the constructors and also improved the coverage

dannypsnl (approved):

q82419:
cc @LFsWang

parthdagia05:
@q82419 rebased this branch

parthdagia05:
@q82419 @dannypsnl can you take look?

hydai:
It seems like there are unresolved conflicts.
Google Chrome 2026-06-27 03 42 16

parthdagia05:
> It seems like there are unresolved conflicts. Google Chrome 2026-06-27 03 42 16

image
@hydai

hydai:
Could you try rebasing onto the branch again without creating a merge commit, like this?

CleanShot 2026-06-27 at 03 50 42

It could be an issue causing GitHub to misread the conflicts. The current state does not allow merging; at least, I cannot press the merge button.

parthdagia05:
is fedora check resolved?

> Could you try rebasing onto the branch again without creating a merge commit, like this?
> 
> CleanShot 2026-06-27 at 03 50 42
> It could be an issue causing GitHub to misread the conflicts. The current state does not allow merging; at least, I cannot press the merge button.

yeah rebased it

PR #4991, opened 2026-06-15
fix(executor): guard realloc result in component string lowering

WHOIM1205:
## Description

This PR fixes a crash in the Component Model canonical ABI string-lowering path caused by unchecked assumptions about the return value of the guest-provided `realloc` function.

The component validator currently defers `realloc` signature verification (GAP-C-5b), allowing components to reference a core function that does not return the expected pointer value. During string lowering, the executor invoked `realloc` and unconditionally accessed the first element of the returned result vector. If the referenced function returned zero results, the executor would dereference an empty vector, resulting in an out-of-bounds access and potential host process crash.

Additionally, failures from `MemoryInstance::setBytes()` were ignored, allowing invalid `(ptr, size)` pairs to be returned even when the string data was never successfully copied into guest memory.

This fix hardens the runtime by validating that the `realloc` invocation produced a result before accessing it and by propagating memory-copy failures through the existing error path. Malformed or non-conforming components now safely fall back to the existing `(0, 0)` failure behavior instead of crashing the runtime.

As a result, crafted components can no longer trigger a crash through malformed `realloc` definitions, while behavior for valid components remains unchanged.

## Checklist

## Test Evidence

- Valid component functions using string lowering continue to work as expected.
- Empty or malformed `realloc` results no longer cause out-of-bounds access.
- Runtime safely falls back to the existing failure path when allocation results are invalid.
- `MemoryInstance::setBytes()` failures are properly handled and no longer return invalid `(ptr, size)` pairs.
- Modified translation unit compiles successfully using the project's existing build configuration.

Example malformed component scenario used during validation:

```text
realloc: (func (param i32 i32 i32 i32))

Expected behavior after fix:
String lowering gracefully falls back to the existing failure path.

No out-of-bounds access occurs.
No host process crash is triggered.
Malformed realloc results are handled safely.

### CODE REVIEW

**Comment by WHOIM1205** [comment]
hey @q82419 
This PR prevents crashes in component string lowering by validating realloc results and handling memory write failures safely.

**Comment by q82419** [comment]
For the component model, please base on that branch.

**Comment by WHOIM1205** [comment]
> For the component model, please base on that branch.

fixed

**Review by q82419 — APPROVED** [review]

**Review by q82419 — CHANGES_REQUESTED** [review]
lint issue

**Comment by WHOIM1205** [comment]
> lint issue

fixed

File: lib/executor/instantiate/component/component_canon.cpp [hunk 1]
```diff
@@ -66,13 +66,24 @@ std::vector<ValVariant> Executor::convValsToCoreWASM(
...
+            Res) {
+          CoreVals.push_back(PtrInMem);
+          CoreVals.push_back(StrSize);
+        } else {
+          CoreVals.push_back(0);
+          CoreVals.push_back(0);
+        }
       } else {
         CoreVals.push_back(0);
         CoreVals.push_back(0);
```
dannypsnl: Maybe you can reverse the logic to guard logic?

1. if !AllocRes => push_back(0)
2. if AllocRes->empty => push_back(0)
3. etc

WHOIM1205: Thanks for the suggestion! I've updated the implementation to use guard-style early failure handling as suggested:

- `!AllocRes` → fallback to `(0, 0)`
- `AllocRes->empty()` → fallback to `(0, 0)`
- `setBytes()` failure → fallback to `(0, 0)`

This removes the nesting while preserving the same behavior. PR has been updated.

dannypsnl (approved):

lib/executor/instantiate/component/component_canon.cpp
```diff
@@ -66,17 +66,33 @@ std::vector<ValVariant> Executor::convValsToCoreWASM(
...
-        CoreVals.push_back(PtrInMem);
-        CoreVals.push_back(StrSize);
-      } else {
+      // The 'realloc' core function signature is not yet checked during
+      // validation (GAP-C-5b). It must return exactly one i32 pointer; a
+      // component declaring a 'realloc' with no result would otherwise make
+      // (*AllocRes)[0] an out-of-bounds access on an empty vector. Guard the
+      // failure cases first, and propagate the setBytes failure instead of
+      // silently handing the guest a (ptr, size) pair for bytes that were
+      // never copied.
```
q82419: Shorten your comments. Too lengthy.

lib/executor/instantiate/component/component_canon.cpp
```diff
@@ -66,17 +66,33 @@ std::vector<ValVariant> Executor::convValsToCoreWASM(
...
+      if (!AllocRes) {
```
q82419: `if (!AllocRes || AllocRes->empty())`

q82419 (changes requested):

q82419:
If you use the AI tools about codes, you should follow the [AI disclosure](https://github.com/WasmEdge/WasmEdge/blob/master/AGENTS.md#ai-assistance-disclosure) in your commits.

WHOIM1205:
> If you use the AI tools about codes, you should follow the [AI disclosure](https://github.com/WasmEdge/WasmEdge/blob/master/AGENTS.md#ai-assistance-disclosure) in your commits.

Thanks for the review.

- Shortened the comments.
- Merged the checks into a single `if (!AllocRes || AllocRes->empty())` guard.
- Added the AI disclosure to the commit.

The behavior remains unchanged and the file builds successfully.

q82419 (approved):

Issue #4992, opened 2026-06-15
bug: WasmEdge serializer refinement and fixing


q82419:
# We would refactor the serializer currently, therefore no new PRs acceptable before completion of refactoring

### Summary

WasmEdge implemented the WASM serializer. But due to supporting WASM 3.0 with several proposals, the serializer is out-of-date and may have encoding bugs due to lack of tests.
This issue collects the PRs, summarizes of the serializer fixing, and test coverages.

### Current State

- `v128.const` / `i8x16.shuffle` 16-byte immediate is byte-reversed #4973 #4983
- Memory64 (and table64) limit flag bit `0x04` is never emitted #5178
- Non-final GC subtype with zero supertypes loses its `sub` marker #4977 #5001
- Active element segment at table 0 with a non-funcref element type is downgraded to funcref
- Element-segment proposal gate is inverted (!= `Passive` should be != `Active`) #5269

### Test Coverage

  serializeModuleTest.cpp — 1 case

  - ✅ Magic + version header
  - ✅ Custom-section ordering by start offset
  - ❌ No populated multi-section module; ❌ no round-trip

  serializeSectionTest.cpp — 14 cases

  - ✅ Custom, Type, Import, Function, Table, Memory, Global, Export, Start, Element, Code, Data, DataCount, Tag (one happy-path each)
  - 🟡 Each is minimal framing only (section id / size / vec-count)

  serializeTypeTest.cpp — 7 cases

  - ✅ FunctionType, GlobalType (incl. mut-global negative), CompositeType
  - 🟡 TableType / MemoryType limits — only HasMin, HasMinMax
  - 🟡 SubType / Rec / Field — final+supers, non-final+supers, array, struct, rec, rec-without-GC negative
  - ❌ Shared limits (0x03), Memory64 i64 limits (0x04–0x07)
  - ❌ Non-final SubType with zero supertypes

  serializeDescriptionTest.cpp — 2 cases

  - ✅ ImportDesc (func/table/mem/global/tag), mut-global negative
  - ✅ ExportDesc (incl. tag), EH negative

  serializeSegmentTest.cpp — 4 cases

  - ✅ GlobalSegment, CodeSegment, DataSegment
  - 🟡 ElementSegment — all 9 modes happy-path (proposals on) + 1 negative
  - ❌ TableSegment (no standalone test; init-expr 0x40 form never exercised)
  - ❌ Element: active-idx0 non-funcref; ❌ MVP-only (proposals-off) active path

[...]

PR #4998, opened 2026-06-16
fix(serializer): correct element segment proposal gate logic

### LINKED ISSUE #4997: bug: Serializer accepts Passive and Declarative element segments without Bulk Memory Operations enabled
### Summary

The WasmEdge serializer logic for Element Segments incorrectly allows `Passive` and `Declarative` element segments to bypass the proposal checks when the Bulk Memory Operations and Reference Types proposals are disabled (e.g., in `WASM_1` standard configuration).

Subpart of #4992 

### Current State

In `lib/loader/serialize/serial_segment.cpp`, the proposal gate incorrectly checks `getMode() != AST::ElementSegment::ElemMode::Passive` instead of `Active`. This inverted logical condition means that if an element segment is `Passive` or `Declarative`, it will not trigger the expected `ErrCode::Value::ExpectedZeroByte` error when bulk memory is disabled, resulting in invalid serialization into the binary.

### Expected State

The serializer should properly check `getMode() != AST::ElementSegment::ElemMode::Active`, and return an `ExpectedZeroByte` error to reject both `Passive` and `Declarative` element segments when the Bulk Memory Operations proposal is disabled.

### Reproduction steps

1. Configure `WasmEdge::Loader::Serializer` with standard `WASM_1` (Bulk Memory and Reference Types disabled).
2. Attempt to serialize an `AST::ElementSection` containing a `Passive` or `Declarative` element segment (with `idx` = 0).
3. Observe that the serializer erroneously returns true and serializes the unsupported segment, instead of rejecting it.

Sukuna0007Abhi:
## Description

This PR fixes an inverted logic gate in the element segment serializer. Previously, the serializer checked `getMode() != ElemMode::Passive`, which incorrectly allowed `Declarative` and `Passive` segments to bypass the Bulk Memory Operations and Reference Types proposal gates. 
The condition has been corrected to `getMode() != ElemMode::Active`, properly restricting older Wasm configurations to only `Active` segments at table `0`.

Unit tests have been added to `test/loader/serializeSegmentTest.cpp` to verify that `WASM_1` configurations correctly reject both `Passive` and `Declarative` segment serialization.

fixes #4997 
Subpart of #4992 

## Checklist

### DIFF
--- a/lib/loader/serialize/serial_segment.cpp
+++ b/lib/loader/serialize/serial_segment.cpp
@@ -51,7 +51,7 @@ Serializer::serializeSegment(const AST::ElementSegment &Seg,
   // vec(u32) + vec(expr)
   if (!Conf.hasProposal(Proposal::BulkMemoryOperations) &&
       !Conf.hasProposal(Proposal::ReferenceTypes) &&
-      (Seg.getMode() != AST::ElementSegment::ElemMode::Passive ||
+      (Seg.getMode() != AST::ElementSegment::ElemMode::Active ||
        Seg.getIdx() != 0)) {
     return logNeedProposal(ErrCode::Value::ExpectedZeroByte,
                            Proposal::BulkMemoryOperations,
--- a/test/loader/serializeSegmentTest.cpp
+++ b/test/loader/serializeSegmentTest.cpp
@@ -305,6 +305,16 @@ TEST(SerializeSegmentTest, SerializeElementSegment) {
   };
   EXPECT_EQ(Output, Expected);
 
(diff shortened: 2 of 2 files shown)


hydai (changes requested):
If you are trying to add multiple fixes to the serializer, you need to include all of them in a single PR. I don't want to merge numerous fragments of serializer PRs, as it wastes CI waiting time and is also annoying for a series of fairly small PRs.

q82419:
Squash your element segment fixing with your PR #5000

PR #5001, opened 2026-06-16
fix(serializer): emit sub prefix for non-final GC subtypes with no supertypes

### LINKED ISSUE #4977: bug: Serializer incorrectly drops 'sub' prefix for open GC SubTypes without supertypes

ANAMASGARD:
## Description

Fixes the GC SubType serialization bug where the `sub` prefix (`0x50`) was silently dropped when a non-final subtype had zero supertypes.

### The Problem

The WASM-GC binary format encodes a `SubType` in one of these four forms:

| Subtype kind | Correct binary encoding |
|---|---|
| final, no supertypes | bare `comptype` (shorthand) |
| **non-final, no supertypes** | `0x50 0x00 <comptype>` ← was broken |
| non-final, with supertypes | `0x50 <vec(typeidx)> <comptype>` |
| final, with supertypes | `0x4F <vec(typeidx)> <comptype>` |

The serializer only emitted the `0x50`/`0x4F` prefix when the supertype list was non-empty. So a non-final subtype with zero supertypes silently fell through to the bare `comptype` shorthand — dropping the `sub` marker entirely. The loader then reads those bytes back as **final**, making the round-trip lossy and incorrect.

### Root Cause

The condition guarding the prefix emission only checked supertype count, completely ignoring the `IsFinal` flag. A non-final subtype with zero supertypes has `size() == 0`, so it skipped the prefix entirely.

### Fix

Changed the condition so the prefix is emitted whenever the subtype is non-final **or** has supertypes covering all four encoding cases correctly.

### Files Changed

- **`lib/loader/serialize/serial_type.cpp`** — fixed the prefix emission condition
- **`include/ast/type.h`** — defaulted `AST::SubType::IsFinal` to `true` so a default-constructed SubType has a well-defined final state matching the spec shorthand and the loader's default case
- **`test/loader/serializeTypeTest.cpp`** — added a test case for a non-final subtype with zero supertypes, asserting the correct `0x50 0x00 <comptype>` output

Closes #4977
Part of #4992


hydai (changes requested):
If you are trying to add multiple fixes to the serializer, you need to include all of them in a single PR. I don't want to merge numerous fragments of serializer PRs, as it wastes CI waiting time and is also annoying for a series of fairly small PRs.

q82419 (dismissed):

hydai (dismissed):

ANAMASGARD:
@hydai  & @q82419 the Fedora Rawhide failure is a pre-existing CI environment issue - protobuf3-compiler and protobuf-c-compiler conflict during package installation, unrelated to this PR's changes. All other 47 checks pass including Fedora Latest. Also I asked the copilot for another review it suggested to resolve the Doxygen comment which I did not please re-reun the Github Action Workflow and give your feedback if any changes are needed , Thanks .

dannypsnl (approved):

PR #5008, opened 2026-06-16
fix(plugin/wasi-logging): validate full string extent in Log::body

Yashika0724:
## Description

`getPointer<char *>` checks `sizeof(char)` — 1 byte — at the given offset. The context and message strings are then consumed as `CxtLen` / `MsgLen` bytes from those pointers: first through the `string_view` comparisons (`== "stdout"`, `!= getLogFileName()`) and then via the spdlog formatting calls.

When a Wasm module passes `CxtPtr + CxtLen` or `MsgPtr + MsgLen` past the linear memory boundary, the subsequent reads hit the `PROT_NONE` guard page and crash the host process (`SIGSEGV` on mmap platforms, access violation on Windows).

Replaced both `getPointer<char *>` + manual `string_view` construction with `getStringView(ptr, len)`, which calls `checkAccessBound(offset, len)` over the full extent and returns an empty view on failure — same pattern used for string arguments elsewhere in the codebase.

## Checklist

## Test Evidence

Added two cases to `test/plugins/wasi_logging/wasi_logging.cpp` (`func_log_oob`):

- **OOB context**: `CxtPtr=65535`, `CxtLen=100` — extends 99 bytes past the single-page boundary. Before the fix this crashes the host process; after the fix `run()` returns `HostFuncError`.
- **OOB message**: valid `CxtPtr` (`"stdout"`), `MsgPtr=65535`, `MsgLen=100`. Same outcome.

All existing `func_log` cases continue to pass — `getStringView` returns the same `string_view` as before for in-bounds inputs.


hydai (changes requested):
Please provide a proof of concept to verify this behavior.

Yashika0724:
> Please provide a proof of concept to verify this behavior.

Hi @hydai, I've added regression tests in func_log_oob covering both the OOB context and OOB message cases. 
Let me know if you'd like any changes!

Yashika0724:
Hi @hydai, here's the proof of concept, including a minimal WAT reproducer and the precise crash path.

## Crash path (unpatched code)

`Log::body` in `lib/plugin/wasi_logging/func.cpp` (lines 27–35 before the fix):

```cpp
char *MsgBuf = MemInst->getPointer<char *>(MsgPtr);   // validates sizeof(char) = 1 byte only
// ...
std::string_view MsgSV(MsgBuf, MsgLen);                // constructs a view of MsgLen bytes
```

`getPointer<char*>(offset)` calls `checkAccessBound(offset, sizeof(char))` — it validates only **1 byte** at `MsgPtr`.
With `MsgPtr = 65535` on a 1-page memory (65 536 bytes): `65535 + 1 = 65536 ≤ 65536` → check passes, returns `&DataPtr[65535]`.

`string_view MsgSV(MsgBuf, 100)` then constructs a view of **100 bytes** starting at that pointer.
spdlog reads bytes `[65535, 65634)` - **99 bytes past the Wasm page boundary** - which lands in the OS guard page → **SIGSEGV** on Linux/macOS, access violation on Windows.

## Reproducer: `test/plugins/wasi_logging/poc_oob_log.wat`

I've committed a minimal WAT file to the branch. Steps to verify:

```bash
# Compile the WAT to Wasm
wat2wasm test/plugins/wasi_logging/poc_oob_log.wat -o /tmp/poc_oob_log.wasm

# --- Without the fix (parent commit d8d8df9^) ---
git checkout d8d8df9^
cmake --build build --target wasmedge -j$(nproc)
./build/tools/wasmedge/wasmedge /tmp/poc_oob_log.wasm
# → Segmentation fault (core dumped)

# --- With the fix ---
git checkout fix/wasi-logging-log-oob-read
cmake --build build --target wasmedge -j$(nproc)
./build/tools/wasmedge/wasmedge /tmp/poc_oob_log.wasm
# → [error] ErrCode: HostFuncError  (clean trap, no crash)
```

## After the fix

`getStringView(MsgPtr, MsgLen)` calls `checkAccessBound(65535, 100)` → `65535 + 100 = 65635 > 65536` → fails → returns `{}` → `MsgSV.data() == nullptr` → returns `HostFuncError`. No memory is read past the page boundary.

The `func_log_oob` test cases exercise exactly this path at the unit-test level.

test/plugins/wasi_logging/poc_oob_log.wat
```diff
@@ -0,0 +1,42 @@
+;; Proof-of-concept: OOB read in wasi:logging/logging#log (pre-fix)
```
q82419: This WAT is not used in your test file and unrelated. remove it.

q82419 (changes requested):

test/plugins/wasi_logging/poc_oob_log.wat
Yashika0724: Done, removed in the latest commit.

q82419 (approved):

Yashika0724:
Hi @hydai, following up here - I've added a minimal WAT reproducer and the exact crash path above (https://github.com/WasmEdge/WasmEdge/pull/5008#issuecomment) showing the unpatched code segfaulting (`getPointer<char*>` only bounds-checks 1 byte, then the host reads `MsgLen` bytes past it into the guard page) and the fix returning a clean `HostFuncError` instead. @q82419 has since reviewed the latest commit and approved. Could you take another look when you have a chance? Happy to add anything else needed to verify.

hydai (approved):

PR #5011, opened 2026-06-17
fix(wasi): map Windows WSANO_DATA in fromWSAError to avoid crash

### LINKED ISSUE #4375: bug: Assertion failed: false && "unreachable", file D:\a\WasmEdge\WasmEdge\lib\host\wasi\win.h, line 303
### Summary

when run rust http client example on windows 11

### Current State

_No response_

### Expected State

_No response_

### Reproduction steps
use release prebuilt
```console
git clone https://github.com/WasmEdge/wasmedge_reqwest_demo
cd wasmedge_reqwest_demo

# Build the Rust code
$env:RUSTFLAGS="--cfg wasmedge --cfg tokio_unstable"
cargo build --target wasm32-wasip1 --release
cd target/wasm32-wasip1/release
wasmedge http.wasm
```

### Screenshots

DESCRIPTION

_No response_

### Components

CLI

0.15.0

Windows 11

### Hardware Architecture

amd64

_No response_

ANAMASGARD:
## Description

On Windows, `getaddrinfo()` returns its error code as the return value, which `INode::getAddrinfo` (`lib/host/wasi/inode-win.cpp:1530`) forwards to `fromWSAError()` in `lib/host/wasi/win.h`. The code `WSANO_DATA` (11004 — the name resolves in DNS but has no A/AAAA record) was not handled in the switch, so it fell through to `default: assumingUnreachable()` and crashed the runtime:

Assertion failed: false && "unreachable", lib/host/wasi/win.h, line 303

This is a normal, common DNS result (mail-only / record-less domains, transient resolver states), so a `reqwest`/`tokio` HTTP client hits it regularly on Windows.

This PR:
- Defines the missing `WSANO_DATA_` constant in `include/system/winapi.h`.
- Adds `case WSANO_DATA_: return __WASI_ERRNO_AINODATA;`, matching the existing
  Linux/macOS mapping `EAI_NODATA -> __WASI_ERRNO_AINODATA` in `fromEAIErrNo()`
  (`lib/host/wasi/linux.h`, `lib/host/wasi/macos.h`).
- Replaces the unreachable `default` with `return __WASI_ERRNO_AIFAIL`, so any
  other unmapped `getaddrinfo()` code degrades to a generic name-resolution
  failure instead of crashing. This mirrors the defensive `default` already used
  by `fromLastError()` in the same file, and is appropriate because the input is
  an OS syscall return that the runtime does not fully control.

Fixes #4375

## Checklist

- [x] **DCO Signed-off**: commit created with `git commit -s`.
- [x] **Commit Messages**: follows Conventional Commits (`fix(wasi): ...`).
- [ ] **Local Tests Passed**: see Test Evidence — this code path is Windows-only and
  cannot be built/run on a non-Windows host; validated via Windows CI.

## Test Evidence

`lib/host/wasi/win.h` is guarded by `#if !WASMEDGE_OS_WINDOWS \n #error \n #endif`, so
the changed `fromWSAError()` only compiles on Windows and cannot be exercised on a
Linux/macOS developer machine. Correctness was verified by:

1. Tracing the crash path: `INode::getAddrinfo` → `fromWSAError(Res)` → unmapped
   `WSANO_DATA` → `default: assumingUnreachable()` at `win.h:303`, which matches the
   assertion line in the report exactly.
2. Parity with the already-correct POSIX implementation (`fromEAIErrNo()` maps
   `EAI_NODATA -> __WASI_ERRNO_AINODATA`).

Relying on the **Windows CI job** in this PR for end-to-end compilation and test coverage. Happy to add a Windows-guarded unit test asserting
`fromWSAError(WSANO_DATA_) == __WASI_ERRNO_AINODATA` if maintainers prefer.


hydai (changes requested):
Since the original WASI tests have passed, you should add a new test case to verify this behavior.

lib/host/wasi/win.h
```diff
@@ -293,14 +293,20 @@ inline constexpr __wasi_errno_t fromWSAError(int WSAError) noexcept {
...
     return __WASI_ERRNO_AISERVICE;
   case WSAESOCKTNOSUPPORT_:
     return __WASI_ERRNO_AISOCKTYPE;
   default:
-    assumingUnreachable();
+    // getaddrinfo() may return WSA error codes that are not part of the
+    // documented set (e.g. version-specific). Fall back to a generic name
+    // resolution failure instead of crashing the runtime, mirroring the
+    // defensive default used by fromLastError() above. See issue #4375.
+    return __WASI_ERRNO_AIFAIL;
```
hydai: This is completely the wrong way to handle it.  
We should cover all of the errors, providing an exact error number.  
If not, this unreachable code should be triggered, creating an issue for it to be fixed.

hydai (changes requested):

lib/host/wasi/win.h
ANAMASGARD: Thanks @hydai, updated:

- Reverted the `default:` back to `assumingUnreachable()`. Since `WSANO_DATA` is
  now mapped, all 9 documented `getaddrinfo()` WSA error codes are covered, so the
  unreachable stays as the intended tripwire for any future unmapped code.
- Added `test/host/wasi/winTest.cpp` → `TEST(WinTest, fromWSAError)` asserting every
  WSA → WASI errno mapping (including `WSANO_DATA → __WASI_ERRNO_AINODATA`), following
  the same approach as the existing `LinuxTest.fromEAIErrNo`.

I'm on Linux so the Windows-only test can't run locally; `wasiTests` builds and passes
here, and I'm relying on the Windows CI for the `fromWSAError` coverage.

One note: since the test is Windows-only, after you push, check the Windows CI job on the PR to confirm WinTest.fromWSAError actually runs and passes there.

hydai: > One note: since the test is Windows-only, after you push, check the Windows CI job on the PR to confirm WinTest.fromWSAError actually runs and passes there.

What do you mean by "after I push"? Another issue is that, how can you confirm this is getting work without setting up an Windows environment for the verification? Are you only using an AI tool for it?

ANAMASGARD: Sorry Sir @hydai  for confusion what i meant by "after I push" was that like the test which I wrote for the WINDOWS , the github CI action the pass or fail of it can easily be a determining factor for this issue  here, since the whole thing is Windows-only.

**How I verified without Windows:** 
- The fix itself is verifiable by code inspection:
  - `win.h` line 5 has `#if !WASMEDGE_OS_WINDOWS` with `#error`, so this code literally cannot compile on Linux - it's architecturally Windows-only
  - The crash path matches the bug report exactly: getaddrinfo returns `WSANO_DATA` (11004) -> falls through to unhandled `default:` at line 303
  - The mapping I added (`WSANO_DATA_` -> `__WASI_ERRNO_AINODATA`) mirrors exactly what Linux/macOS already do
  
  test/host/wasi/winTest.cpp  This test catches that exact bug.
  
Once you approve the workflows, the Windows CI will run automatically. The test will 
execute on the Windows server, and we'll see if `WinTest.fromWSAError` passes. That's 
the real proof.

Can you please run the Github Actions CI so that we can see the results , I do use Perplexity AI to understand the issue like where in codebase which files to investigate and which line is causing bug , stuff like that .

q82419: You didn't disclose your AI assistance. Please squash your commits and follow the [AI disclosure](https://github.com/WasmEdge/WasmEdge/blob/master/AGENTS.md#ai-assistance-disclosure).

hydai:
Your Windows CI failed. Have you verified it yourself? It seems like you are trying to send a PR that you never tested yourself and are asking us to review it.

ANAMASGARD:
@hydai please give me some time I am testing it and will fix it.

hydai:
Since this PR is not ready and the build process has failed, I am marking it as a draft.

hydai:
Replaced by #5373

PR #5019, opened 2026-06-17
fix(vm):unsafeUnregisterModule silently succeeds when module is not found

aniket866:
### Summary

**File:** [`lib/vm/vm.cpp`](file:///c:/Users/Hp/WasmEdge/lib/vm/vm.cpp#L180-L222)

### What is wrong?

When you call `unsafeUnregisterModule("some-name")` and the module with that name does not exist in any list, the function simply returns `{}` (success) — as if it deleted something. This is misleading. The caller has no way to know the module was never found.

### Why fix this?

This breaks the **API contract**: a function named `unregister` should either confirm it removed something or tell you it could not find it. If calling code relies on the return value to decide whether to retry or log an error, it will incorrectly assume the module was removed. This kind of silent false-success is a classic source of hard-to-reproduce bugs in production.

### Flow diagram

```
unsafeUnregisterModule("foo")
        │
        ├─► search RegModInsts  ──── not found ──►┐
        ├─► search BuiltInModInsts ─ not found ──►│
        ├─► search PlugInModInsts ── not found ──►│
        │                                         ▼
        └─────────────────────────── return {}  ← BUG: should be an error
```

### Code — Before (line 221–222)

```cpp
  }

  return {};   // ← always succeeds even if nothing was removed
}
```

### Code — After 

```cpp
  }

  spdlog::error(ErrCode::Value::UnknownImport);
  return Unexpect(ErrCode::Value::UnknownImport);
  
}
```

## Checklist

## Test Evidence

---

### DIFF
--- a/lib/vm/vm.cpp
+++ b/lib/vm/vm.cpp
@@ -219,7 +219,8 @@ Expect<void> VM::unsafeUnregisterModule(std::string_view Name) {
     }
   }
 
-  return {};
+  spdlog::error(ErrCode::Value::WrongInstanceAddress);
+  return Unexpect(ErrCode::Value::WrongInstanceAddress);
 }
 
 VM::WasmUnitKind


aniket866:
@q82419 @hydai 
do checkout this fix whenever you get time

hydai (changes requested):
You need to add a test to cover this case.

hydai:
After checking the test, I believe this behavior is expected. Please check the existing test: https://github.com/WasmEdge/WasmEdge/blob/master/test/vm/VMTest.cpp#L141-L145.

Trying to delete a non-existing module, which will be a no-op, is fine to report as success here.

PR #5028, opened 2026-06-18
 fix(plugin/wasi-nn): guard deleteGraph against double-free of recycled graph IDs

### LINKED ISSUE #5023: Bug:  Out-of-Bounds Memory Corruption & Crash via Double Free of Graph IDs
### Why this is a bug:
In `wasinnenv.h`, `deleteGraph` does not verify if the Graph ID is already in `NNGraphRecycle`. If `deleteGraph(0)` is called twice (a double free of the Graph ID):
1. Graph 0 gets reset and its index `0` is added to `NNGraphRecycle`.
2. On the second call, since `Id < NNGraph.size()` is still true, `NNGraph.pop_back()` is triggered, reducing the size of the graph vector to `0`.
3. The index `0` remains in `NNGraphRecycle`.
4. When a new graph is created, `newGraph` retrieves `ID = 0` from `NNGraphRecycle` and attempts to write to `NNGraph[0]`. Since the vector is empty, this leads to an out-of-bounds memory access and a crash.

### Before and After Code Snippet

**Before:**
```cpp
// Located in plugins/wasi_nn/wasinnenv.h
void deleteGraph(const uint32_t Id) noexcept {
  std::unique_lock Lock(GraphMutex);
  if (Id < NNGraph.size()) {
    auto &G = NNGraph[Id];
    G.setFinalized();
    if (G.getContextCount() == 0) {
      if (Id == NNGraph.size() - 1) {
        NNGraph.pop_back();
      } else {
        G.reset();
        NNGraphRecycle.insert(Id);
      }
    }
  }
}
```

**After:**
```cpp
// Located in plugins/wasi_nn/wasinnenv.h
void deleteGraph(const uint32_t Id) noexcept {
  std::unique_lock Lock(GraphMutex);
  if (Id < NNGraph.size() && NNGraphRecycle.find(Id) == NNGraphRecycle.end() && !NNGraph[Id].isFinalized()) {
    auto &G = NNGraph[Id];
    G.setFinalized();
    if (G.getContextCount() == 0) {
      if (Id == NNGraph.size() - 1) {
        NNGraph.pop_back();
      } else {
        G.reset();
        NNGraphRecycle.insert(Id);
      }
    }
  }
}
```

### Current State

_No response_

### Expected State

_No response_

### Reproduction steps

# WasmEdge Double Free Vulnerability Simulation

## The Setup

A WebAssembly guest application loads two models.
* The host allocates two graphs: **Graph 0** and **Graph 1**.
* `NNGraph` vector has a size of **2** (indices 0 and 1).
* `NNGraphRecycle` (the recycle bin) is **empty**.

---

### Step 1: Guest unloads Graph 0
The guest calls `wasi_nn::unload` on Graph 0.
* **Host State Check:** `Id = 0`. Since `0 < 2` (size is 2) and `0 != 2 - 1` (not the last index), WasmEdge does not shrink the vector.
* **Recycle Bin:** Graph 0 is reset, and index `0` is added to `NNGraphRecycle`.
* **Result:** `NNGraph` size is still **2**. `NNGraphRecycle` has `{0}`.

### Step 2: Guest unloads Graph 1
The guest calls `wasi_nn::unload` on Graph 1.
* **Host State Check:** `Id = 1`. Since `1 < 2` and `1 == 2 - 1` (is the last index), WasmEdge shrinks the vector.
* **Action:** `NNGraph.pop_back()` is called.
* **Result:** `NNGraph` size is reduced to **1** (only index 0 remains). `NNGraphRecycle` still has `{0}`.

### Step 3: Guest unloads Graph 0 again (The Double Free)
Due to a bug in the guest application (e.g. double destruction), the guest calls `wasi_nn::unload` on Graph 0 a second time.
* **Host State Check:** `Id = 0`. WasmEdge does not check if `0` is already in the recycle bin.
* **Action:**
  1. `0 < 1` (size is 1) is **true**.
  2. `0 == 1 - 1` (is the last index now) is **true**.
  3. WasmEdge calls `NNGraph.pop_back()`, shrinking the vector size to **0**.
* **Result:** `NNGraph` size is now **0** (empty). `NNGraphRecycle` still has `{0}`.

### Step 4: Guest loads a new model (The Crash)
The guest calls `wasi_nn::load` to load a new model.
* **Host State Check:** The host tries to allocate a new Graph ID. It sees that `NNGraphRecycle` is not empty (contains `{0}`).
* **Action:** The host assigns `ID = 0` and attempts to initialize the graph:
  ```cpp
  NNGraph[0].init(BE);
  ```
* **Result:** Since `NNGraph` is empty (size is 0), accessing index `0` causes an **out-of-bounds memory access** and WasmEdge crashes instantly.

ANAMASGARD:
Fixes #5023

## What was wrong

In the WASI-NN plugin, unloading the same graph twice could crash WasmEdge with an out-of-bounds memory access.

WASI-NN keeps loaded graphs in a `NNGraph` vector and remembers freed slots in a `NNGraphRecycle` set so their IDs can be reused. The `deleteGraph` function never checked whether an ID had **already** been freed. So a guest that unloads the same graph twice (a common double-free bug in guest code) could shrink the `NNGraph` vector below an index still sitting in the recycle set. The next
`load` would then reuse that stale ID and write into an empty vector — out of bounds, and a crash.

Concretely, the sequence from the issue:

1. `unload(0)` → graph 0 (not the last) is reset and its index goes into the recycle set.
2. `unload(1)` → graph 1 (the last) is popped; the vector shrinks but `{0}` stays in the recycle set.
3. `unload(0)` again → with no guard, index 0 is now "the last", so it gets popped too; the vector is now empty but `{0}` is still in the recycle set.
4. `load(...)` → reuses ID 0 and does `NNGraph[0].init(...)` on an empty vector → **out-of-bounds crash**.

## What this PR changes

One-line fix in `plugins/wasi_nn/wasinnenv.h`: `deleteGraph` now ignores IDs that are already in `NNGraphRecycle`, making a repeated unload a harmless no-op. This mirrors the guard that the sibling `deleteContext` function already had — the two had simply drifted apart, which is what allowed the bug.

A regression test (`WasiNNTest.GraphIdDoubleFree`, GGML-gated) reproduces the exact double-free sequence and asserts that the following allocation stays in bounds.

## Result

- The double-free / double-unload sequence is now safe — the second unload does nothing instead of corrupting the recycle bookkeeping.
- The invariant "every recycled ID is a valid index into `NNGraph`" is preserved, so `newGraph` can never index out of range.
- No behavior change for any correct (non-double-free) usage.

## Verification

- Reproduced the bug and confirmed the fix against a standalone model of the
  allocation/recycle logic: the original code reaches an empty-vector
  out-of-bounds state, the patched code returns a valid reused ID with no OOB.
- Added the in-tree GGML-gated regression test for CI coverage.

## Note for maintainers (follow-up, not in this PR)

The root cause is that the graph/context slot-recycling logic is hand-duplicated several times across `newGraph`, `newContext`, `deleteGraph`, and `deleteContext`, and the copies drifted (only `deleteContext` had the recycle guard). I kept this PR minimal and backport-friendly, but I'm happy to follow up with a small `RecyclingPool<T>` abstraction that collapses those duplicates and makes this class of bug structurally impossible.

---

This contribution and PR description made with  assistance of Perplexity AI


hydai:
Thanks for the contribution. I am closing this because I believe a better solution is to do a complete redesign to prevent such issues from happening again. This will be replaced by #5100.

PR #5035, opened 2026-06-18
fix(OSS-Fuzz): add the missing PropComponent in the driver fuzzer

arthurscchan:
This PR adds the missing PropComponent that cause OSS-Fuzz fuzzing build failing.


hydai (changes requested):
LGTM.

~~Please refine the commit message to: `fix(OSS-Fuzz): add the missing PropComponent in the driver fuzzer` to pass the commitlint CI.~~

Never mind; I found that I can skip the check and squash the commit directly.

hydai (approved):

hydai:
Thanks for the fix. I plan to create a follow-up CI workflow to ensure that we can achieve a successful build for the fuzzer.

PR #5040, opened 2026-06-22
feat: add configurable call stack size limit with CallStackExhausted trap

### LINKED ISSUE #2079: Unexpected behavior

q82419:
## Description

Previously unbounded recursion grew the interpreter frame stack or the native stack without limit, so the WebAssembly assert_exhaustion spec tests could not be supported. Add a configurable maximum stack size (RuntimeConfigure::MaxStackSize, in bytes); exceeding it now traps with a new execution error, CallStackExhausted ("call stack exhausted").

Interpreter: Executor::enterFunction bounds the heap value and frame stacks in bytes before entering a function, because the interpreter keeps its activation on the heap while the native stack stays flat. A tail call reuses the current frame and is exempt. A limit of 0 selects a built-in default so the interpreter is always bounded.

AOT/JIT: the compiled function prologue compares the native stack pointer against a precomputed limit (the entry stack address minus the configured size) and traps when the pointer descends past it. A limit of 0 disables the check, so compiled code keeps its existing behavior by default. The limit is reached through the read-only execution context.

C API: WasmEdge_ConfigureSetMaxStackSize and
WasmEdge_ConfigureGetMaxStackSize expose the new option.

Spec tests: the assert_exhaustion command is now handled in every harness, including skip-stack-guard-page, whose large per-frame locals are bounded by the byte-based limit in both the interpreter and compiled modes.

Measured on a fib micro-benchmark (a worst-case tiny function body): the interpreter heap check is within noise (~1%, ~0.6 ns/call); the compiled native-stack check adds ~0.25 ns/call (~28% on this micro-benchmark, far less on realistic function bodies).

Assisted-by: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

## Checklist

## Test Evidence

---

close #2079


q82419:
Should also fix #4422 and #4410

lib/executor/helper.cpp
```diff
@@ -58,6 +84,28 @@ Expect<AST::InstrView::iterator> Executor::enterFunction(
...
+        LocalBytes;
+    if (unlikely(StackSizeNeeded > StackSizeLimit)) {
+      spdlog::error(ErrCode::Value::CallStackExhausted);
+      return Unexpect(ErrCode::Value::CallStackExhausted);
+    }
+  }
+
   // Get the function type for the parameter and return counts.
   const auto &FuncType = Func.getFuncType();
   const uint32_t ArgsN = static_cast<uint32_t>(FuncType.getParamTypes().size());
```
ibmibmibm, line 129:
```c++
const uintptr_t SP = []() -> uintptr_t {
#if defined(__GNUC__) || defined(__clang__)
        return reinterpret_cast<uintptr_t>(__builtin_stack_address());
#elif defined(_MSC_VER)
        return reinterpret_cast<uintptr_t>(_AddressOfReturnAddress());
#else
        volatile char StackProbe;
        return reinterpret_cast<uintptr_t>(&StackProbe);
#endif
}();
```

ibmibmibm (changes requested):

lib/executor/helper.cpp
q82419: Fixed

PR #5044, opened 2026-06-22
feat(OSS-Fuzz): add new fuzzer targets component

arthurscchan:
This PR creates a new OSS-Fuzz fuzzer targets component processing, together with the build update in OSS-Fuzz (https://github.com/google/oss-fuzz/pull/15790).


q82419:
Component model fuzzing already exists and not in here.

q82419:
Maybe duplicated in `test/component/fuzzing`.

hydai:
It seems that OSS-Fuzz would like to enable the new fuzzer. Perhaps we should move the current component fuzzing from the test folder to the tool folder so we can trace the fuzzing-related parts in the same location.

hydai:
I am going to merge this PR first, and then create a follow-up PR to clean up the existing testing fuzzer.

hydai (approved):

PR #5045, opened 2026-06-22
fix(wasi): bound sock_getaddrinfo copies to guest buffer capacity

WHOIM1205:
## Description

This PR fixes a guest-triggerable out-of-bounds write in the WASI `sock_getaddrinfo` implementation caused by copying resolver-generated data into guest linear memory without validating guest-provided buffer capacities.

The current implementation validates that guest buffers are in-bounds during initialization but later copies resolver output using lengths derived from host-side data rather than the guest-declared capacities. During `sock_getaddrinfo`, the executor unconditionally copies `ai_canonname` and `sa_data` into guest memory after overwriting the original capacity fields, allowing writes beyond the validated destination buffers.

Additionally, the same issue affects socket address handling where fixed-size `sa_data` structures may exceed the guest-provided `sa_data_len`, resulting in the same class of memory corruption vulnerability.

This fix hardens the runtime by preserving the guest-declared capacities before they are overwritten and validating all copy operations against those capacities. When resolver output exceeds the available buffer size, the runtime now returns `__WASI_ERRNO_OVERFLOW` instead of performing an unsafe write.

As a result, crafted guests can no longer trigger host-side memory corruption, crashes, or silent guest-memory corruption through malformed `sock_getaddrinfo` inputs, while behavior for correctly sized buffers remains unchanged.

## Checklist

## Test Evidence

- Canonical name copies succeed when the guest-provided buffer is sufficiently large.
- Oversized `ai_canonname` results now return `__WASI_ERRNO_OVERFLOW` instead of writing past guest memory.
- Socket address copies validate `sa_data_len` before performing `memcpy`.
- Oversized socket address data returns `__WASI_ERRNO_OVERFLOW` and safely aborts the operation.
- Linux implementation compiles successfully with the existing build configuration.
- Changes were applied consistently across Linux, macOS, and Windows implementations.

Example failure scenario handled after validation:

```cpp
// Guest-provided capacity
addrinfo.ai_canonname_len = 1;

// Resolver returns a longer canonical name
resolver_canonname = "very-long-canonical-hostname.example.com";

// Expected behavior after fix:
// - Capacity check fails
// - __WASI_ERRNO_OVERFLOW is returned
// - No out-of-bounds write occurs
// - Host process remains stable
```

Expected behavior after fix:

- Guest buffer capacities are always respected.
- Resolver output larger than the destination buffer is rejected.
- No host-side memory corruption occurs.
- No guest-triggerable crash is possible through this code path.
- Existing behavior remains unchanged for valid inputs.

### DIFF
--- a/lib/host/wasi/inode-linux.cpp
+++ b/lib/host/wasi/inode-linux.cpp
@@ -857,10 +857,20 @@ WasiExpect<void> INode::getAddrinfo(std::string_view Node,
 
     // process ai_canonname in addrinfo
     if (SysResItem->ai_canonname != nullptr) {
-      CurAddrinfo->ai_canonname_len = std::strlen(SysResItem->ai_canonname);
+      // The guest-declared capacity of the canonical-name buffer. It must not
+      // be exceeded: the destination points into the guest linear memory and
+      // was only validated to be in-bounds for this many bytes.
+      const __wasi_size_t CanonnameCap = CurAddrinfo->ai_canonname_len;
+      const size_t SysCanonnameLen = std::strlen(SysResItem->ai_canonname);
+      if (SysCanonnameLen + 1 > CanonnameCap) {
+        ::freeaddrinfo(SysResPtr);
+        return WasiUnexpect(__WASI_ERRNO_OVERFLOW);
+      }
+      CurAddrinfo->ai_canonname_len =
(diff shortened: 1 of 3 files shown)


q82419:
Your PR template is out-of-date and lack of some checklists.
Review will be pending until you update the PR description and your earlier PR finished.

hydai:
Replaced by #5344

PR #5058, opened 2026-06-24
perf(validator): optimize subtype depth validation with memoization

### LINKED ISSUE #5017: Subtype Hierarchy Validation Complexity Optimization via Memoization
### Location
- File: [lib/validator/validator.cpp](file:///c:/Users/Hp/WasmEdge/lib/validator/validator.cpp#L39-L79)

### Technical Analysis
In the WebAssembly Garbage Collection (WasmGC) validator module, type subsumption rules necessitate the verification of subtype hierarchies via the `checkSubtypeDepth` recursive algorithm. This routine traverses defined subtyping relations to enforce compliance with the `MaxSubtypeDepth` (63) architectural threshold and detect cyclic inheritance graphs. 

For deep or complex type-inheritance graphs, this recursive formulation scales quadratically O(N * D)relative to defined types. Introducing a memoization lookup table (e.g. mapping verified type indices to their absolute inheritance depths via a flat registry inside the `Validator` class) will prevent redundant recursive traversals, flattening lookup complexity to $O(1)$ while preserving cycle-detection invariants.

---

#### Before Code Snippet
```cpp
// TODO: make the super type depth table instead of recursively querying.
Expect<void>
checkSubtypeDepth(const uint32_t BaseIdx, uint32_t TestIdx,
                  std::unordered_set<uint32_t> &VisitedNodes,
                  const std::vector<const WasmEdge::AST::SubType *> &TypeVec,
                  uint32_t Depth) {
  if (VisitedNodes.count(TestIdx)) {
    spdlog::error(ErrCode::Value::InvalidSubType);
    spdlog::error("    Cycle detected in subtype hierarchy for type {}."sv,
                  BaseIdx);
    return Unexpect(ErrCode::Value::InvalidSubType);
  }

  if (Depth >= MaxSubtypeDepth) {
    spdlog::error(ErrCode::Value::InvalidSubType);
    spdlog::error("    Subtype depth for type {} exceeded the limits of {}"sv,
                  BaseIdx, MaxSubtypeDepth);
    return Unexpect(ErrCode::Value::InvalidSubType);
  }

  VisitedNodes.insert(TestIdx);
  const auto &TestType = *TypeVec[TestIdx];
  for (const auto SuperIdx : TestType.getSuperTypeIndices()) {
    if (unlikely(SuperIdx >= TypeVec.size())) {
      spdlog::error(ErrCode::Value::InvalidSubType);
      return Unexpect(ErrCode::Value::InvalidSubType);
    }
    EXPECTED_TRY(
        checkSubtypeDepth(BaseIdx, SuperIdx, VisitedNodes, TypeVec, Depth + 1)
            .map_error([=](auto E) {
              return E;
            }));
  }
  return {};
}
cpp ```

---
```

#### After Code Snippet (Highlighted lines of change in yellow)

<pre><code class="cpp">
Expect&lt;uint32_t&gt;
checkSubtypeDepth(const uint32_t BaseIdx, uint32_t TestIdx,
                  std::unordered_set&lt;uint32_t&gt; &amp;VisitedNodes,
                  const std::vector&lt;const WasmEdge::AST::SubType *&gt; &amp;TypeVec,
                  <mark>std::vector&lt;int32_t&gt; &amp;MemoizedDepths</mark>) {
  if (VisitedNodes.count(TestIdx)) {
    spdlog::error(ErrCode::Value::InvalidSubType);
    return Unexpect(ErrCode::Value::InvalidSubType);
  }
  <mark>if (MemoizedDepths[TestIdx] != -1) {
    return static_cast&lt;uint32_t&gt;(MemoizedDepths[TestIdx]);
  }</mark>
  VisitedNodes.insert(TestIdx);
  const auto &amp;TestType = *TypeVec[TestIdx];
  uint32_t MaxDepth = 0;
  for (const auto SuperIdx : TestType.getSuperTypeIndices()) {
    if (unlikely(SuperIdx &gt;= TypeVec.size())) {
      return Unexpect(ErrCode::Value::InvalidSubType);
    }
    EXPECTED_TRY(uint32_t SuperDepth,
        checkSubtypeDepth(BaseIdx, SuperIdx, VisitedNodes, TypeVec, <mark>MemoizedDepths</mark>));
    if (SuperDepth + 1 &gt;= MaxSubtypeDepth) {
      return Unexpect(ErrCode::Value::InvalidSubType);
    }
    MaxDepth = std::max(MaxDepth, SuperDepth + 1);
  }
  <mark>VisitedNodes.erase(TestIdx);
  MemoizedDepths[TestIdx] = static_cast&lt;int32_t&gt;(MaxDepth);</mark>
  return MaxDepth;
}

</code></pre>

### Workflow Diagram
```mermaid
graph TD
    Start["Invoke checkSubtypeDepth()"] --> CacheCheck{"Is Depth for TestIdx Cached?"}
    CacheCheck -- "Yes" --> ReturnCached["Return Memoized Depth (O(1))"]
    CacheCheck -- "No" --> CycleCheck{"Is TestIdx in VisitedNodes Set?"}
    CycleCheck -- "Yes (Cycle)" --> Err["Return InvalidSubType Error"]
    CycleCheck -- "No" --> Recurse["Recursively Traverse SuperType Indices"]
    Recurse --> SaveCache["Memoize Traversed Node Depth in Cache Vector"]
    SaveCache --> Success["Return Depth Verification Success"]
```

ravindra-RKB:
This PR addresses #5017 by memoizing subtype depth computation during validation.

Previously, `checkSubtypeDepth` recursively traversed the subtype hierarchy for every subtype being validated. In hierarchies with shared supertypes, the same depth calculations could be repeated multiple times.

This change introduces a `SubTypeDepthMap` cache that stores computed subtype depths and reuses them across validations. The existing validation behavior is preserved, including:

Maximum subtype depth enforcement (`MaxSubtypeDepth = 63`)
Cycle detection in subtype hierarchies
Existing validation error reporting

The cache is cleared when validating a new `TypeSection` or component `CoreTypeSection` to ensure state does not leak across validation boundaries.

Testing

Executed validator regression tests locally:

`ValidatorRegressionTest.ExactMaxSubtypeDepth`
`ValidatorRegressionTest.MaxSubtypeDepthExceeded`
Full `wasmedgeValidatorRegressionTests` suite

Result:

```text
[==========] 11 tests from 1 test suite ran.
[  PASSED  ] 11 tests.
```

Fixes #5017


hydai (changes requested):
This is a PR related to performance. Please show the benchmark results comparing the original design with your PR.

ravindra-RKB:
@hydai

Thanks for the feedback. I ran a local benchmark to compare the original implementation against this PR.

Since WasmEdge currently limits subtype depth to 63 and the current subtype hierarchy is restricted to single inheritance, I generated synthetic modules containing multiple maximum-depth subtype chains and measured validation time during type validation.

The benchmark harness was used locally for measurement only and is not included in this PR.

### Benchmark Results

* 126 types: 1392 µs → 148 µs (~9.4× faster)
* 504 types: 5168 µs → 445 µs (~11.6× faster)
* 1008 types: 10948 µs → 618 µs (~17.7× faster)
* 2016 types: 20279 µs → 1249 µs (~16.2× faster)

### Analysis

The improvement comes from caching previously computed subtype depths and avoiding repeated traversal of the same subtype chains. The memoized depth map also eliminates repeated construction of temporary `std::unordered_set<uint32_t>` instances during validation.

### Validation

I also re-ran the validator regression tests locally:

* `ValidatorRegressionTest.ExactMaxSubtypeDepth`
* `ValidatorRegressionTest.MaxSubtypeDepthExceeded`
* Full `wasmedgeValidatorRegressionTests` suite (**11/11 passed**)

If useful, I can also share the benchmark harness and raw measurement logs used to produce these measurements.

ravindra-RKB:
@q82419 @hydai

Thanks for taking the time to review the PR.

I have already addressed the benchmark request and incorporated the cache pre-allocation optimization suggested by the low-confidence Copilot comment (`SubTypeDepthMap.assign(...)`).

Regarding the two high-confidence Copilot comments about restoring the top-down `Depth` guard, my understanding is that they propose a behavioral change rather than a pure optimization. Before making another update, I'd like to confirm the expected direction with the maintainers.

Would you prefer that I restore the early depth guard while preserving the memoization approach, or would you rather keep the current implementation and evaluate it as-is?

I'd appreciate your guidance before making further changes.

hydai:
If you fix the issues reported by Copilot, please resolve the threads.

lib/validator/component_validator.cpp
```diff
@@ -364,6 +364,7 @@ Expect<void> Validator::validate(
 
 Expect<void>
 Validator::validate(const AST::Component::CoreTypeSection &TypeSec) noexcept {
+  SubTypeDepthMap.clear();
```
dannypsnl: It's not a good idea to have to clean an instance before use it in an algorithm.

ravindra-RKB:
@maintainers,

Thanks for the review and the helpful feedback.

I've updated the implementation based on the review comments:

- Restored the early recursion depth guard while preserving the memoization optimization.
- Restored the original `OwnTypeIdx` validation semantics to match the current `master` behavior.
- Localized the subtype depth cache to the type section validation flow instead of keeping it as `Validator` state.
- Removed the unnecessary component validator cache handling.
- Rebased the branch onto the latest `master` to incorporate the recent validator changes.

The build completes successfully in my local environment after rebasing.

I've addressed the Copilot-reported issues in the latest revision and will resolve the corresponding review threads.

Please let me know if there are any further changes you'd like me to make.

dannypsnl (dismissed):

ravindra-RKB:
@maintainers,

Thanks again for reviewing this PR.

After addressing the previous review comments, I investigated the remaining Codecov report in detail.

The only uncovered branches now appear to be the defensive checks for:

- `TestIdx >= DepthMap.size()`
- `DepthMap[TestIdx] == Visiting`

From my investigation, these branches seem to be structurally unreachable through the normal validator flow because of the current validator invariants:

- `SubTypeDepthMap` is always initialized to match the current `TypeSection` size.
- The validator enforces `Index < OwnTypeIdx` for supertype references, so subtype traversal always proceeds toward smaller indices and cannot revisit the same node.

I wasn't able to find a valid black-box test that exercises these branches without bypassing encapsulation or relying on white-box testing techniques, which I'd prefer to avoid.

Could you please advise whether you would prefer:
- an alternative testing approach for these defensive branches, or
- leaving them uncovered as defensive code?

I'm happy to follow whichever direction you recommend.

ravindra-RKB:
Thanks for the review.

I've updated the implementation so that cycle detection is performed before the depth-limit check, restoring the original behavior while preserving the memoization optimization.

I also re-ran the relevant validator regression tests locally after the change, and they continue to pass.

Please let me know if there's anything else you'd like me to adjust.

ravindra-RKB:
@hydai

Thanks for taking the time to review this PR.

I've addressed the requested changes, all CI checks are now passing, and the latest implementation preserves the original behavior while keeping the memoization optimization.

Whenever you have a chance, I'd appreciate another review.

hydai (approved):

Issue #5067, opened 2026-06-26
feat: support loading the component model value section


parthdagia05:
### Summary

While reviewing my value-linearity PR (#5014), @q82419 pointed out that the loader side isn't there yet:

> Component model loader haven't support the value section yet.

#5014 added validation for the value index space, but the loader never parses section 12 (the value section), so that index space is currently only populated by start results and value imports, never by a defined value. Worse, a component that actually contains a value section is rejected outright as MalformedSection today.

This adds loader support for the value section so the loader → validator path is consistent end to end.

### Details

Spec grammar:

```
  v*:section_12(vec(<value>)) => v*
  value ::= t:<valtype> len:<core:u32> v:<val(t)> (where len = ||v||)
```

Plan:
1. Add a Value AST node holding the valtype + the raw len-prefixed payload bytes. The payload is kept as bytes on purpose, not decoded: the len prefix exists exactly so a value can be captured without resolving val(t) against a (possibly type-indexed) type at load time. Decoding into runtime values belongs to instantiation, not the loader.
2. Add a ValueSection node mirroring ExportSection, register it in the Component section variant, and add the Comp_Sec_Value AST node attr.
3. Wire section id 0x0C into the loader dispatch (component.cpp) and implement loadSection / loadValue, reusing loadType for the valtype and readU32 + readBytes for len and payload.
4. Validator: register each defined value into the value index space (CompCtx.addValue()) after validating its type, same as start results and valuebound imports already do. This is what makes #5014's index-space checks line up with reality.
5. Executor: leave value-section instantiation as a TODO for now. Turning defined values into live runtime values needs a value representation and is a separate piece of work.
6. Add a loader test that round-trips a component carrying one value.

parthdagia05:
marked as done

PR #5014, opened 2026-06-17
feat(validator): enforce component value linearity (GAP-EX-4)

parthdagia05:
## Background

The Component Model requires values in the value index space to be consumed exactly once. WasmEdge currently tracks only the number of values and does not validate linear consumption semantics (GAP-EX-4).

## Changes

* Replace value count tracking with per-value consumption state in `ComponentContext`
* Track value consumption when values are used by:

  * `start` arguments
  * value exports
* Reject double consumption of the same value
* Reject components that finish validation with unconsumed values
* Remove existing GAP-EX-4 TODOs
* Add unit tests covering valid and invalid consumption scenarios

## Notes

* This PR only wires consumption checks for currently reachable validation paths (`start` arguments and value exports).
* Support for value consumption through component instantiation arguments can be added when the corresponding value feature path is enabled.

PR #5068, opened 2026-06-26
feat(loader/component): support loading the value section

parthdagia05:
## Description

The component loader didn't parse section 12 (the value section), so a component carrying a defined value was rejected as `MalformedSection`. As noted on #5014, the loader doesn't support it yet, while #5014 already validates the value index space.

## Closes: #5067 

This parses `value ::= valtype + u32 len + val(t)`, storing the type and the raw length-prefixed payload bytes - decoding into runtime values is left to instantiation, which the `len` prefix is designed for. The validator now registers each defined value into the value index space, matching how start results and value-bound imports already do. Executor instantiation is left as a TODO.

## Checklist

- [x] **Commit Messages**: Conventional Commit format.
- [x] **Local Tests Passed**: See Test Evidence below.

- [ ] **AI Disclosure**: n/a
- [ ] **Human-in-the-loop**: n/a

## Test Evidence

Full component suite (`wasmedgeComponentRegressionTests`):


lib/executor/instantiate/component/component.cpp
```diff
@@ -19,6 +19,8 @@ Executor::instantiate(Runtime::StoreManager &StoreMgr,
       using T = std::decay_t<decltype(Sec)>;
       if constexpr (std::is_same_v<T, AST::CustomSection>) {
         // Do nothing to custom section.
+      } else if constexpr (std::is_same_v<T, AST::Component::ValueSection>) {
+        // TODO: COMPONENT - instantiate component value section.
```
dannypsnl, line 29: warning here?

lib/executor/instantiate/component/component.cpp
```diff
@@ -53,6 +55,8 @@ Executor::instantiate(Runtime::Instance::ComponentImportManager &ImportMgr,
...
```
hydai, line 67: We should show errors or warnings here to let users know this is not usable.

hydai (changes requested):

parthdagia05:
@hydai @dannypsnl added a warning

dannypsnl (approved):

hydai (changes requested):
The Windows Server CI failed the component model test. Could you please check if this is related to your modification?

parthdagia05:
> 

> The Windows Server CI failed the component model test. Could you please check if this is related to your modification?

I am having a check on it

parthdagia05:
The validator stored a raw pointer to a memory type that became invalid after a vector reallocation, causing it to read garbage memory and incorrectly treat a 32-bit memory as 64-bit, which led to a valid module being rejected.
@hydai

parthdagia05:
image
once ran it on my fork

dannypsnl:
Are Fedora and MacOS known issues? @hydai

hydai:
@dannypsnl Yup, these two should be fixed in the master branch. I think it's fine to ignore both of them temporarily.

dannypsnl (approved):

PR #5060, opened 2026-06-24
fix(runtime): add END sentinel to WasmFunction instructions

Manar-Elhabbal7:
## Description

This PR fixes a FIXME comment in the `FunctionInstance::WasmFunction` constructor regarding vector capacity reservation.

### Problem
- The constructor reserved `Expr.size() + 1` capacity but only copied `Expr.size()` instructions
- The extra `+1` had no clear purpose, causing potential issues when connecting two instruction vectors
- The `OpCode::END` sentinel was not explicitly added as the last instruction

### Solution
- Append `OpCode::END` as a sentinel after copying instructions
- The `+1` capacity now serves a clear purpose: to hold the END opcode
- Add missing include for `common/enum_ast.hpp` to resolve `OpCode` definition

### Changes
- `include/runtime/instance/function.h`: Add END sentinel and missing include
## Checklist

Before submitting this PR, please ensure the following:

- [x] **DCO Signed-off**: All commits are signed-off (`git commit -s`). CI workflows will only be approved if DCO check is passed.
- [x] **Commit Messages**: Run `commitlint` on your commit messages to ensure they meet the [Conventional Commit](https://www.conventionalcommits.org/en/v1.0.0/) standards.
- [x] **Local Tests Passed**: Provide a screenshot or logs below to prove that you have passed the test suites locally.

If you are using AI-assisted tools, please ensure the following:

- [x] **AI Disclosure**: Disclose it using the `Assisted-by:` trailer in your commits and note it in the PR description. Please refer to the CONTRIBUTING.md/AGENTS.md files for the detailed format.
- [x] **Human-in-the-loop**: Submitting commits, issues, and PRs without human verification is forbidden.

PR #5071, opened 2026-06-26
fix(runtime): remove redundant capacity reservation in wasm function

Manar-Elhabbal7:
## Description

The `WasmFunction` constructor in `function.h` contains a FIXME:

```cpp
// FIXME: Modify the capacity to prevent connecting 2 vectors.
Instrs.reserve(Expr.size() + 1);
Instrs.assign(Expr.begin(), Expr.end());
```

reserve(N+1) allocates capacity for N+1 Instructions, but assign() only copies N. The extra slot is never used.

## Checklist

## Test Evidence

---


q82419:
You ever didn't test for your changes.

Manar-Elhabbal7:
> You ever didn't test for your changes.

Hello, @q82419 
My PR was draft ! 
Draft pr = not ready for review 🙃
I was still working on it and I didn't test it but I will make before mark it as ready for review 

So why my pr is labeled as spam? I followed the instructions in contributing file and I was still working on it

hydai:
A draft PR means "work in progress," and that's fine. But it still must represent a plausible attempt at the real problem. I believe that YiYing marking this PR with the `SPAM` label is not because the PR is incomplete; it's about the approach.  
You already tried to send a similar PR (see #5060) before, which does not address the underlying FIXME. And it seems like this second attempt is still stuck on the same issue.

If you think the label is incorrect, please explain what you believe the root cause is and how your solution resolves it.

Manar-Elhabbal7:
> A draft PR means "work in progress," and that's fine. But it still must represent a plausible attempt at the real problem. I believe that YiYing marking this PR with the `SPAM` label is not because the PR is incomplete; it's about the approach. You already tried to send a similar PR (see #5060) before, which does not address the underlying FIXME. And it seems like this second attempt is still stuck on the same issue.
> 
> If you think the label is incorrect, please explain what you believe the root cause is and how your solution resolves it.

Hello @hydai,

I just wanted to explain that I didn't take it personally. I only wanted to better understand the project's rules and the reason behind the label. And to show i didn't mean to spam.  I really appreciate the time you and the other maintainers spend reviewing contributors' code. 
At first, I thought the `+1`  in the reserve function was there to reserve space for one extra instruction when concation two vectors, so I assumed it should be filled with an `END` instruction. Later, I learned that this is already handled by the executor, so that wasn't the right approach.
Then I thought the solution was simply to remove the extra `reserve()` since `assign()` already copies all the instructions.

Looking back, I realize I didn't fully understand the actual problem, so both of my solutions were incorrect. This is my first time working with the WasmEdge codebase, so I'm still learning.

Thank you for taking the time to explain it to me :)

hydai:
I have removed the SPAM label.

PR #5073, opened 2026-06-27
fix(runtime): change FailedMod to vector of unique_ptr

Manar-Elhabbal7:
## Description

Changed `FailedMod` from `std::unique_ptr<ModuleInstance>` to `std::vector<std::unique_ptr<ModuleInstance>>` to support multiple failed module instances as required by the WebAssembly specification.

### Changes:
1. Changed type of `FailedMod` to `vector<unique_ptr<ModuleInstance>>`.
2. Updated `recycleModule()` to move the `unique_ptr` directly.
3. Updated `reset()` to clear the vector.
4. **Updated the comment** to reflect the new design and removed the `FIXME.`
 
## Checklist

## Test Evidence

Screenshot from 2026-06-27 16-54-06

### DIFF
--- a/include/runtime/storemgr.h
+++ b/include/runtime/storemgr.h
@@ -89,6 +89,7 @@ public:
     }
     NamedMod.clear();
     NamedComp.clear();
+    FailedMod.clear();
   }
 
   /// Register a named module in this store.
@@ -150,13 +151,7 @@ private:
 
   /// Collect the instantiation failed module.
   void recycleModule(std::unique_ptr<Instance::ModuleInstance> &&Mod) {
-    if (FailedMod) {
-      auto *OldMod = FailedMod.release();
-      if (OldMod) {
-        OldMod->terminate();
-      }
-    }
-    FailedMod = std::move(Mod);
+    FailedMod.emplace_back(std::move(Mod));
   }
 
   /// \name Module name mapping.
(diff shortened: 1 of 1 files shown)


q82419:
Still considering whether this should be a vector, because this is a trick for spec test which would query for failed module (such as get last error) but useless in real cases.

Manar-Elhabbal7:
> Still considering whether this should be a vector, because this is a trick for spec test which would query for failed module (such as get last error) but useless in real cases.

I thought that if multiple failed instantiations occur before reset(), the current implementation immediately destroys the previous failed module when a new one is stored. Keeping them in a vector would instead tie their lifetime to the Store, and cleanup or this will be a rare scenario?

q82419:
> > Still considering whether this should be a vector, because this is a trick for spec test which would query for failed module (such as get last error) but useless in real cases.
> 
> I thought that if multiple failed instantiations occur before reset(), the current implementation immediately destroys the previous failed module when a new one is stored. Keeping them in a vector would instead tie their lifetime to the Store, and cleanup or this will be a rare scenario?

The original code is for the following situation in [spec test](https://github.com/WasmEdge/wasmedge-spectest/blob/main/wasm-1.0/linking/linking.wast#L356-L388):

```
(module $Ms
  (type $t (func (result i32)))
  (memory (export "memory") 1)
  (table (export "table") 1 funcref)
  (func (export "get memory[0]") (type $t)
    (i32.load8_u (i32.const 0))
  )
  (func (export "get table[0]") (type $t)
    (call_indirect (type $t) (i32.const 0))
  )
)
(register "Ms" $Ms)

(assert_trap
  (module
    (import "Ms" "memory" (memory 1))
    (import "Ms" "table" (table 1 funcref))
    (data (i32.const 0) "hello")
    (elem (i32.const 0) $f)
    (func $f (result i32)
      (i32.const 0xdead)
    )
    (func $main
      (unreachable)
    )
    (start $main)
  )
  "unreachable"
)

(assert_return (invoke $Ms "get memory[0]") (i32.const 104))  ;; 'h'
(assert_return (invoke $Ms "get table[0]") (i32.const 0xdead))
```

The first module "$Ms" instantiated succeeded, and the second anonymous module failed during its instantiation phase.
But in the second module instantiation phase, it imports the "table" from the "$Ms" module, and initialized the first element in "table" with the function reference of the function in the anonymous module.
Although the second module instantiation failed, and the caller would not get the returning module instance, the function reference left in the table of "$Ms" module. That's why the test of invoking ` $Ms "get table[0]"` would get `0xdead`.
Thus the situation is very rare, especially for multiple instantiation failed modules cases.

For refining this, I think it's not a good mechanism for just storing the failed modules.
The modules would left in the store forever. If apply this, every instantiation failed modules will stay in store, and it'll cause performance issue. The correct algorithm is to only store the modules which owning instances are referenced.

PR #5075, opened 2026-06-28
fix(executor): use matching address types for memory.copy and table.copy operands

### LINKED ISSUE #5074: bug: memory.copy and table.copy read source/destination operands with swapped address types
## Summary

`memory.copy` and `table.copy` extract the source and destination operands using
each other's address type. When the source and destination have different index
types — for example copying between an i32 memory and an i64 memory under the
memory64 + multiple-memories proposals — the two offsets are read with swapped
widths, so they get truncated or wrongly extended.

When both operands share the same index type (the usual case) the two types are
equal and the swap has no effect, which is presumably why this hasn't shown up.

## Details

The validator and the executor disagree on which operand carries which address
type.

The validator pushes the three operands as (bottom to top)
`dst : ATDst`, `src : ATSrc`, `len : min(ATDst, ATSrc)`.

`lib/validator/formchecker.cpp`, `Memory__copy` (L1265) and `Table__copy` (L1172):

```cpp
auto ATDst = Mems[Instr.getTargetIndex()];
auto ATSrc = Mems[Instr.getSourceIndex()];
auto ATMin = convTypeCodeToAddrType(ATDst) <= convTypeCodeToAddrType(ATSrc)
                 ? ATDst
                 : ATSrc;
return StackTrans({ValType(ATDst), ValType(ATSrc), ValType(ATMin)}, {});
```

The executor pops them back, but reads `Src` with the destination's type and
`Dst` with the source's type.

`lib/executor/engine/memoryInstr.cpp:70` (and the identical pattern in
`lib/executor/engine/tableInstr.cpp:84`):

```cpp
const auto AddrType1 = MemInstSrc.getMemoryType().getLimit().getAddrType();
const auto AddrType2 = MemInstDst.getMemoryType().getLimit().getAddrType();
uint64_t Len = extractAddr(StackMgr.pop(), std::min(AddrType1, AddrType2));
uint64_t Src = extractAddr(StackMgr.pop(), AddrType2); // dst's type
uint64_t Dst = extractAddr(StackMgr.pop(), AddrType1); // src's type
```

`AddrType1` is the source's address type and `AddrType2` is the destination's, so
`Src` ends up extracted with the destination's type and `Dst` with the source's —
the opposite of what the validator pushed. `extractAddr` reads the stack value as
`uint32_t` or `uint64_t` depending on the address type:

```cpp
inline uint64_t extractAddr(const ValVariant &Val, const AddressType AT) noexcept {
  switch (AT) {
  case AddressType::I32: return static_cast<uint64_t>(Val.get<uint32_t>());
  case AddressType::I64: return static_cast<uint64_t>(Val.get<uint64_t>());
  ...
```

so once the two types differ, the offsets are read with the wrong width.

The `len` operand is correct: it uses `min(AddrType1, AddrType2)`, which is
symmetric.

## When it bites

A module with two memories of different index types (multiple-memories +
memory64), e.g. memory 0 declared `i32` and memory 1 declared `i64`. A
`memory.copy` between them then reads the source/destination offsets with each
other's width. Copying from the i64 memory into the i32 memory with a source
offset that does not fit in 32 bits is the clearest case: the high bits of the
source offset are dropped, so the copy reads from the wrong place (or passes a
bounds check it should fail). `table.copy` between an i32-indexed and an
i64-indexed table has the same problem.

The bundled spec testsuite only exercises same-index-type copies, so it doesn't
catch this.

## Suggested fix

Extract `Src` with the source's address type and `Dst` with the destination's,
matching what the validator pushed:

```cpp
uint64_t Src = extractAddr(StackMgr.pop(), AddrType1); // source
uint64_t Dst = extractAddr(StackMgr.pop(), AddrType2); // destination
```

in both `runMemoryCopyOp` and `runTableCopyOp`. I'll open a PR.

## Environment

- master @ e4249cbe

tkshsbcue:
Fixes #5074.

### What

`memory.copy` and `table.copy` extract the source and destination offsets using
each other's address type. `Src` is read with the destination's type and `Dst`
with the source's. This is a no-op when both memories/tables share an index type,
but with the memory64 and multiple-memories proposals the two can differ (one
i32, one i64), and the offsets are then read with the wrong width — truncated or
wrongly extended.

### Why

The validator and the executor disagreed on which operand carries which type.

The validator (`lib/validator/formchecker.cpp`, `Memory__copy` / `Table__copy`)
pushes the operands as `dst : ATDst`, `src : ATSrc`, `len : min(ATDst, ATSrc)`:

```cpp
return StackTrans({ValType(ATDst), ValType(ATSrc), ValType(ATMin)}, {});
```

The executor (`runMemoryCopyOp` / `runTableCopyOp`) popped them back with `Src`
and `Dst` swapped:

```cpp
const auto AddrType1 = MemInstSrc...getAddrType(); // source
const auto AddrType2 = MemInstDst...getAddrType(); // destination
uint64_t Len = extractAddr(StackMgr.pop(), std::min(AddrType1, AddrType2));
uint64_t Src = extractAddr(StackMgr.pop(), AddrType2); // used dst's type
uint64_t Dst = extractAddr(StackMgr.pop(), AddrType1); // used src's type
```

`extractAddr` reads the value as `uint32_t` or `uint64_t` based on the address
type, so once the two types differ the offsets are misread. `len` was already
correct because it uses the symmetric `min` of both.

### Fix

Extract `Src` with the source's type and `Dst` with the destination's, in both
`runMemoryCopyOp` and `runTableCopyOp`.

### Notes

- The bundled spec testsuite only covers same-index-type copies, which is why
  this wasn't caught.
- Built locally (`wasmedgeExecutor`) to confirm it compiles.

---

This contribution was developed with assistance from Claude (Anthropic); the
analysis and change were reviewed before submitting.

### DIFF
--- a/lib/executor/engine/memoryInstr.cpp
+++ b/lib/executor/engine/memoryInstr.cpp
@@ -70,8 +70,8 @@ Executor::runMemoryCopyOp(Runtime::StackManager &StackMgr,
   const auto AddrType1 = MemInstSrc.getMemoryType().getLimit().getAddrType();
   const auto AddrType2 = MemInstDst.getMemoryType().getLimit().getAddrType();
   uint64_t Len = extractAddr(StackMgr.pop(), std::min(AddrType1, AddrType2));
-  uint64_t Src = extractAddr(StackMgr.pop(), AddrType2);
-  uint64_t Dst = extractAddr(StackMgr.pop(), AddrType1);
+  uint64_t Src = extractAddr(StackMgr.pop(), AddrType1);
+  uint64_t Dst = extractAddr(StackMgr.pop(), AddrType2);
 
   // Replace mem[Dst : Dst + Len] with mem[Src : Src + Len].
   // When source and destination are the same memory instance, overlapping
--- a/lib/executor/engine/tableInstr.cpp
+++ b/lib/executor/engine/tableInstr.cpp
@@ -84,8 +84,8 @@ Executor::runTableCopyOp(Runtime::StackManager &StackMgr,
   const auto AddrType1 = TabInstSrc.getTableType().getLimit().getAddrType();
   const auto AddrType2 = TabInstDst.getTableType().getLimit().getAddrType();
   uint64_t Len = extractAddr(StackMgr.pop(), std::min(AddrType1, AddrType2));
-  uint64_t Src = extractAddr(StackMgr.pop(), AddrType2);
-  uint64_t Dst = extractAddr(StackMgr.pop(), AddrType1);
+  uint64_t Src = extractAddr(StackMgr.pop(), AddrType1);
+  uint64_t Dst = extractAddr(StackMgr.pop(), AddrType2);
 
   // Replace tab_dst[Dst : Dst + Len] with tab_src[Src : Src + Len].
   return TabInstSrc.getRefs(0, Src + Len)


hydai:
Closing this and marking it as SPAM because it doesn't follow any PR or issue template, which indicate it could be machine-generated slop and lacks any human oversight.

tkshsbcue:
Hey hi @hydai, sorry to bring this up, but I noticed that the PR and issue templates weren't followed. Since maintaining a consistent format is important, would you mind letting me handle it using the proper template?

q82419:
SPAM not only because of wrong issue template, but wrong in WASM spec.
The order of address types for memory.copy and table.copy are CORRECT.

tkshsbcue:
@q82419 if you allow me i can fix that

q82419:
> @q82419 if you allow me i can fix that

@tkshsbcue Apologize that after tracing for optimization the instruction implementations, you are CORRECT.
This part may in wrong order when applying the memory64 proposal.
Please help to create another new PR for this change with correct formats and we would merge.

tkshsbcue:
alright will do it!

PR #5116, opened 2026-07-03
ci: deduplicate GitHub Actions caches to stay under the 10 GB quota

hydai:
## Description

After shipping the #5112, we are caching the WASI-NN models to avoid re-downloading these large files repeatedly and to resolve the flaky issues. However, the repository's actions cache is limited to a hard 10GB quota. This means that GitHub's LRU eviction policy keeps deleting the master-seeded caches. As a result, every PR run misses the cache, re-downloads the files, and saves a duplicate copy onto their own `refs/pull/<PR Number>/merge` ref, which can never be shared with other PRs. Worse yet, this process causes evictions of even more cache entries on the master seeds.

Observed duplication before this change:

| Cache | Waste | Cause |
| ----- | ----- | ----- |
| `ml-fixtures-wasmedge_stablediffusion-*` twice on master | ~2.0 GB | same key, two cache *versions*: zstd (hosted macOS runners) vs gzip (Linux containers without zstd) |
| `android-ndk-r23b-cmake-3.22.2` per PR ref | ~760 MB × N PRs | PR-side `actions/cache` save after the master seed was evicted |
| `llvm-21.1.3-windows-2025` + `llvm-21.1.3-windows-2025-msvc` | ~980 MB | two keys for the identical extracted LLVM zip |
| ML fixtures on PR refs (e.g. bitnet, 520 MB) | ~0.5 GB × N PRs | PR-side save |

Expected steady state: roughly 10.5 GB → 6 GB, master seeds stop being evicted, and the PR cache hit rate recovers.

One commit per fix:

1. `ci: save heavy caches only on master and unify the Windows LLVM key` —  
   Split every large `actions/cache` usage into `actions/cache/restore` plus  
   an explicit `actions/cache/save` gated on `github.ref == 'refs/heads/master'` and a cache miss (Android NDK, Windows LLVM in three jobs; ML model fixtures in four jobs). PR runs still restore the master-seeded entries; a miss now costs a re-download instead of a duplicated cache entry. The `-msvc` key suffix is dropped so all three Windows jobs share one LLVM entry, and the split actions are pinned to the `actions/cache` v6.1.0 revision already used elsewhere.

2. `ci: install zstd in Linux build images for consistent cache compression` —  
   `actions/cache` versions entries by compression method and falls back to gzip inside containers without zstd. Installing zstd in the Ubuntu base image and the manylinux_2_28 plugins images makes container jobs produce the same cache version as the hosted runners, collapsing the double-stored fixture caches into a single shared entry per key.

3. `ci: delete a pull request's Actions caches when it closes` —  
   New `cleanup-pr-caches.yml` on `pull_request_target` (closed): deletes the closed PR's merge-ref caches so they stop occupying quota for up to a week. The job checks out nothing and runs no PR-controlled code; the base-repo token is needed for the `actions: write` scope on fork PRs.

Verification of the compression split: `actions/cache` computes the cache version as `sha256(paths | compressionMethod | salt)`. Recomputing this for all four observed version hashes matches exactly (4/4)—see Test Evidence—and the pinned v6.1.0 source selects zstd purely based on the presence of the `zstd` binary on Linux.

Note on the coding style and local test checklist items: this PR touches only workflow YAML files and Dockerfiles, so there is no C/C++ code for `clang-format` or the test suites to cover; `actionlint`, `lineguard`, and `commitlint` runs are attached as Test Evidence instead.

This contribution was developed with assistance from Claude Fable 5 (Anthropic) via GitHub Copilot CLI.

## Checklist

## Test Evidence

`commitlint` on all three commits:

```console
=== ci: delete a pull request's Actions caches when it closes
PASS
=== ci: install zstd in Linux build images for consistent cache compression
PASS
=== ci: save heavy caches only on master and unify the Windows LLVM key
PASS
```

`actionlint` parity with master (only pre-existing shellcheck notes in
untouched scripts; the new workflow reports nothing):

```console
$ actionlint <master copies of edited workflows> | grep -c shellcheck
41
$ actionlint <edited workflows> | grep -c shellcheck
41
$ actionlint .github/workflows/cleanup-pr-caches.yml
(no output)
```

`lineguard` on all edited files:

```console
✓ All files passed lint checks!
```

Cache-version hash verification — recomputed
`sha256(paths | compression | salt)` against the four versions observed via
`GET /repos/WasmEdge/WasmEdge/actions/caches`:

```console
MATCH a193326089346c3a  computed as [fixtures + gzip]               == observed [SD/bitnet fixtures (linux container jobs)]
MATCH 63e10b2d4bc8cf29  computed as [android + gzip]                == observed [android NDK (container)]
MATCH 2cb93949004e5135  computed as [fixtures + zstd-without-long]  == observed [SD/mlx fixtures (macOS jobs)]
MATCH 2e557480df06ee1d  computed as [llvm + zstd-without-long + windows-only] == observed [windows LLVM (hosted runner)]
```

---

### DIFF
--- a/.github/workflows/cleanup-pr-caches.yml
+++ b/.github/workflows/cleanup-pr-caches.yml
@@ -26,11 +26,17 @@ jobs:
           REPO: ${{ github.repository }}
           REF: refs/pull/${{ github.event.pull_request.number }}/merge
         run: |
-          while :; do
+          for pass in 1 2 3 4 5 6 7 8 9 10; do
             ids=$(gh cache list --repo "$REPO" --ref "$REF" --limit 100 --json id --jq '.[].id')
             [ -z "$ids" ] && break
+            echo "Delete pass ${pass}"
             for id in $ids; do
               gh cache delete "$id" --repo "$REPO" || true
             done
           done
-          echo "All caches on $REF are deleted."
+          remaining=$(gh cache list --repo "$REPO" --ref "$REF" --limit 1 --json id --jq 'length')
+          if [ "$remaining" -gt 0 ]; then
+            echo "::warning::Caches remain on $REF after 10 delete passes; they expire automatically within 7 days."
+          else
+            echo "All caches on $REF are deleted."
+          fi
--- a/.github/workflows/cleanup-pr-caches.yml
+++ b/.github/workflows/cleanup-pr-caches.yml
@@ -1,0 +1,36 @@
+name: Cleanup PR caches
+
+on:
+  pull_request_target:
+    types:
+      - closed
+
+permissions:
+  contents: read
+
+jobs:
+  cleanup:
+    name: Delete caches of the closed PR
+    permissions:
+      actions: write
+    runs-on: ubuntu-latest
+    steps:
+      - name: Harden the runner (Audit all outbound calls)
+        uses: step-security/harden-runner@9af89fc71515a100421586dfdb3dc9c984fbf411 # v2.19.4
+        with:
+          egress-policy: audit
+
+      - name: Delete caches on the merge ref
+        env:
+          GH_TOKEN: ${{ github.token }}
+          REPO: ${{ github.repository }}
+          REF: refs/pull/${{ github.event.pull_request.number }}/merge
+        run: |
+          while :; do
+            ids=$(gh cache list --repo "$REPO" --ref "$REF" --limit 100 --json id --jq '.[].id')
+            [ -z "$ids" ] && break
+            for id in $ids; do
+              gh cache delete "$id" --repo "$REPO" || true
+            done
+          done
+          echo "All caches on $REF are deleted."
--- a/.github/workflows/reusable-build-on-windows.yml
+++ b/.github/workflows/reusable-build-on-windows.yml
@@ -52,6 +52,20 @@ jobs:
         with:
           path: LLVM-21.1.3-win64-MultiThreadedDLL
           key: llvm-21.1.3-windows-2025
+      - name: Download LLVM for Windows
+        if: steps.cache-llvm.outputs.cache-hit != 'true'
+        run: |
+          $uri = "https://github.com/WasmEdge/llvm-windows/releases/download/llvmorg-21.1.3/LLVM-21.1.3-win64-MultiThreadedDLL.zip"
+          $llvm = "LLVM-21.1.3-win64-MultiThreadedDLL.zip"
+          Invoke-WebRequest -Uri $uri -HttpVersion 2.0 -OutFile $llvm
+          Expand-Archive -Path $llvm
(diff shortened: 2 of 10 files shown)


.github/workflows/cleanup-pr-caches.yml
```diff
@@ -0,0 +1,36 @@
...
+        env:
+          GH_TOKEN: ${{ github.token }}
+          REPO: ${{ github.repository }}
+          REF: refs/pull/${{ github.event.pull_request.number }}/merge
+        run: |
+          while :; do
+            ids=$(gh cache list --repo "$REPO" --ref "$REF" --limit 100 --json id --jq '.[].id')
+            [ -z "$ids" ] && break
+            for id in $ids; do
+              gh cache delete "$id" --repo "$REPO" || true
```
0yi0, line 34: Potential issue detected by Claude Fable 5:
> The `while` : delete loop's only exit is an empty cache list, and `gh cache delete` failures are swallowed with `|| true`. If a PR is closed while its CI is still running (sccache keeps creating entries), or an entry persistently fails to delete, the loop spins until the 6-hour job timeout with a green log.
> Fix: bound it to a few passes and emit a warning if caches remain.

.github/workflows/reusable-build-on-windows.yml
```diff
@@ -89,6 +89,12 @@ jobs:
           $cmake_sys_version = "10.0.26100.0"
           cmake -Bbuild -GNinja -DCMAKE_SYSTEM_VERSION="$cmake_sys_version" -DCMAKE_MSVC_RUNTIME_LIBRARY=MultiThreadedDLL -DCMAKE_MT="$llvm_mt" -DLLVM_DIR="$llvm_dir" -DWASMEDGE_BUILD_TESTS="$Env:build_tests" -DWASMEDGE_BUILD_PACKAGE=ZIP .
           cmake --build build
+      - name: Save LLVM for Windows
+        if: ${{ github.ref == 'refs/heads/master' && steps.cache-llvm.outputs.cache-hit != 'true' }}
+        uses: actions/cache/save@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
```
0yi0: Is this cache save after `Build Wasmedge` intended? Can we do it before build? (the same question for msvc)

.github/workflows/cleanup-pr-caches.yml
hydai: Fixed.

.github/workflows/reusable-build-on-windows.yml
hydai: Fixed. Extract the download phase from the build into a standalone step, and move the cache save before the build.

ibmibmibm (approved):

Issue #310, opened 2021-07-22
Support dynamic host functions


juntao:
Today's WasmEdge host functions are statically compiled and linked to the VM binary.

In the future, we would like to support a dynamic mechanism to register custom host functions in a running WasmEdge instance.

hydai:
This feature is included in [Wasm-C-API PR](https://github.com/WasmEdge/WasmEdge/issues/306) #306

PR #5136, opened 2026-07-07
docs(examples): add exception handling repro README

### LINKED ISSUE #2406: Add example for doing exception handling
## Motivation

We should provide an example after adding the exception handling proposal for WasmEdge https://github.com/WasmEdge/WasmEdge/issues/2335.

## Details

According to https://github.com/emscripten-core/emscripten/discussions/19150, it seems that emscripten is able to generate standalone wasm from C++ with exception. 

`em++ -O1 -fwasm-exceptions a.cpp -o a.wasm`

```cpp
#include <stdio.h>

int main() {
  try {
    puts("throw...");
    throw 1;
    puts("(never reached)");
  } catch(...) {
    puts("catch!");
  }
  return 0;
}
```

shubham5080:
Fixes #2406

Adds minimal example README with Emscripten repro and link to the WasmEdge Book.

Docs PR: https://github.com/WasmEdge/docs/pull/310


hydai:
All docs and examples should belong to WasmEdge/docs

shubham5080:
@hydai Understood — moved the example content into WasmEdge/docs PR #310. Thanks for the clarification.

PR #5146, opened 2026-07-09
fix(component-model): fix validator gaps in canon lift/lower memory requir…

### LINKED ISSUE #4441: feat: Component Model: validator implementation tracking issue

Ankitdotraider:
Fixes #4441

## What this does

The component model validator was accepting some invalid modules it
shouldn't have — specifically cases where `canon lift`/`canon lower`
don't specify `memory`/`realloc` even though the function signature
actually needs them (strings, lists, or any aggregate type that wraps
them).

`validateCanonOptions` was only checking consistency between the options
themselves (e.g. realloc requires memory), never checking them against
the actual function signature. So a func type like `(param string)`
with zero canonical options would pass validation when it should've
been rejected.

## Changes

- Added `needsMemory()` on `ComponentValType`/`DefValType`, same pattern
  as the existing `containsBorrow()` — recurses through records, variants,
  tuples, options, results, etc.
- `validateCanonOptions` now takes the func's `FuncType` and checks params/
  results against it. `realloc` is only required when allocation actually
  happens in that direction (lift needs it for params, lower needs it for
  results).
- Added the `MAX_FLAT_PARAMS`/`MAX_FLAT_RESULTS` (16/1) canonical ABI
  limits as a simplified param/result count check — not full flattening
  yet, that's still a follow-up (see GAP-C-1b/GAP-C-2b comments in the
  file).
- Fixed a bug in `ComponentContext::getFunc()` — imported/aliased funcs
  were never resolving their type, which was also silently masking some
  of the above (things looked "valid" because there was no func type to
  check against). It now stores an optional type index and resolves
  lazily via `getDefType()`.
- Fixed duplicate parameter name detection, which was case-sensitive
  when the spec test expects case-insensitive comparison.

## Testing

Flipped `Validate: true` for the `func` folder in `test/spec/spectest.cpp`.
Fixes the previously-failing assert_invalid cases: func.2, func.3, func.7,
func.8, func.9.

Ran the full core test suite locally, all 530 tests pass:
Screenshot 2026-07-09 at 4 27 34 PM

> This contribution was developed with assistance from Cursor and Claude (Anthropic).


lib/validator/component_validator.cpp
```diff
@@ -2398,7 +2478,7 @@ Validator::validate(const AST::Component::DefValType &DVT) noexcept {
 
 Expect<void> Validator::validate(const AST::Component::FuncType &FT) noexcept {
   // Validate param names: kebab-case + unique
-  std::unordered_set<std::string_view> ParamNames;
+  std::unordered_set<std::string> ParamNames;
```
dannypsnl, line 2481: Is there any reason that we must store `string` here?

Ankitdotraider: since we lowercase the label before inserting (toLowerStr(P.getLabel())), we need to own the string. a string_view would point to a temporary that goes out of scope.

hydai:
I think all component model-related PRs should be based on the `dev/component` branch. Closing this.

Issue #4711, opened 2026-03-13
feat: Component Model: executor implementation tracking issue


q82419:
### Summary

This issue tracks the implementation and testing status of executor, includes the instantiation in component-model proposal.

### Details

TBD: support execution implementation status in the spec test

Refer to the parent issue #4236 for the instantiation and execution implementation status for every sections.

1. Task issue base on Section or more detailed structure.
2. List the required spec test entries in that issue to pass.
3. For the PR, modify the boolean of validation phase about the required spec tests in `ComponentModelFolders` map to `true` in the `test/component/spectest.cpp`.
4. Run the component model spec test and pass to prove your implementation.
5. For the trap messages not reasonable in the spec test suite, it's fine to standardize them and update on the wasmedge-spectest repo.

The passing status of spec tests in [WasmEdge-spectest repo](https://github.com/WasmEdge/wasmedge-spectest):

- [ ] [`adapt.wast`](https://github.com/WebAssembly/component-model/blob/main/test/wasm-tools/adapt.wast)
- [ ] [`alias.wast`](https://github.com/WebAssembly/component-model/blob/main/test/wasm-tools/alias.wast)
- [ ] [`big.wast`](https://github.com/WebAssembly/component-model/blob/main/test/wasm-tools/big.wast)
- [ ] [`definedtypes.wast`](https://github.com/WebAssembly/component-model/blob/main/test/wasm-tools/definedtypes.wast)
- [ ] [`empty.wast`](https://github.com/WebAssembly/component-model/blob/main/test/wasm-tools/empty.wast)
- [ ] [`example.wast`](https://github.com/WebAssembly/component-model/blob/main/test/wasm-tools/example.wast)
- [ ] [`export-ascription.wast`](https://github.com/WebAssembly/component-model/blob/main/test/wasm-tools/export-ascription.wast)
- [ ] [`export-introduces-alias.wast`](https://github.com/WebAssembly/component-model/blob/main/test/wasm-tools/export-introduces-alias.wast)
- [ ] [`export.wast`](https://github.com/WebAssembly/component-model/blob/main/test/wasm-tools/export.wast)
- [ ] [`func.wast`](https://github.com/WebAssembly/component-model/blob/main/test/wasm-tools/func.wast)
- [ ] [`import.wast`](https://github.com/WebAssembly/component-model/blob/main/test/wasm-tools/import.wast)
[...]

PR #5151, opened 2026-07-12
fix(component-model): track ResourceId for exports in raw-Component instantiation.

Ankitdotraider:
Related to #4711

was going through component_validator.cpp on dev/component and found the GAP-I-3 todo, resource typed exports werent getting a ResourceId tracked when a raw Component gets instantiated (the loop around line 1197). the alias export path already handles this correctly (added in #4891) but this specific loop was left as a followup

added resolveChildResourceType which walks the components own AST directly, since by the time this instantiation loop runs the components validation scope is already exited so calling CompCtx.getResource() on the index directly wouldnt work. it handles locally defined resources, imported resources, outer aliased resources and resource exports off an imported instance, mirroring resolveChildInstanceType's approach

build is clean and all 530 existing tests pass, but I havent added a dedicated regression test for this specific path yet. happy to add one if thats needed before this can be reviewed properly, just wanted to get this up and get feedback on the approach first
Screenshot 2026-07-12 at 5 49 46 PM

### DIFF
--- a/lib/validator/component_validator.cpp
+++ b/lib/validator/component_validator.cpp
@@ -138,6 +138,118 @@ resolveChildInstanceType(const AST::Component::Component &Comp,
   return nullptr;
 }
 
+// Like resolveChildInstanceType, but for resource types — needed because
+// Comp's own scope is already popped by the time this runs, so we can't
+// just call CompCtx.getResource() on the index directly.
+std::optional<uint64_t> resolveChildResourceType(
+    const AST::Component::Component &Comp, uint32_t TypeIdx,
+    Span<const AST::Component::InstantiateArg<AST::Component::SortIndex>> Args,
+    ComponentContext &CompCtx,
+    std::unordered_map<const AST::Component::DefType *, uint64_t>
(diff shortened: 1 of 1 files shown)


dannypsnl:
Please update your commit message format, and can you provide a test that previously failed without this commit?

q82419:
Close this due to not follow the contribute requirement

Ankitdotraider:
I'll fix the commit message format and add a test that fails without this fix. will push to this branch and follow up here once its ready, if it makes sense to reopen at that point let me know. also which specific requirement did I miss, just want to make sure I fix the right thing before pushing again

Ankitdotraider:
hey, addressed both points

fixed the commit message format (body lines were over 100 chars, wrapped them properly, confirmed clean with commitlint now)

added a regression test, InstantiationPropagatesResourceExport, in ComponentRegressionTest.cpp. verified it fails without this fix with "borrow type index 0 does not refer to a resource type" and passes with it, pushed to the same branch

let me know if this is good to reopen, or should I just open a fresh PR

PR #5152, opened 2026-07-12
chore: upgrade fmt to 12.1.0 and spdlog to v1.17.0

ibmibmibm:
## Description

Bumps **fmt 11.0.2 → 12.1.0** and **spdlog v1.13.0 → v1.17.0**, and gives both
dependencies the `-isystem` treatment that simdjson and gtest already receive.

### Why fmt stops at 12.1.0 and not 12.2.0

An earlier revision of this PR pinned 12.2.0 and CI rejected it for two independent,
upstream reasons. Both are recorded in a comment above the `FetchContent_Declare` so the
next person to bump fmt does not rediscover them:

1. **fmt 12.2.0 requires CMake ≥ 3.19.** It sets `VERSION`/`SOVERSION`/`DEBUG_POSTFIX` on
   its `fmt-header-only` **INTERFACE** target, and CMake rejects non-whitelisted properties
   on INTERFACE libraries before 3.19. This project requires 3.18, and Debian bullseye —
   which builds the static release — ships CMake 3.18.4, with no backport available.
2. **fmt 12.2.0 cannot format its own 128-bit fallback.** `do_format_decimal` guards a
   32-bit fast path with a plain `if` instead of `if constexpr`, so the body must still
   type-check for every `UInt`. It calls `write2digits_i(Char*, size_t)`, and fmt's
   `uint128` class has only an *explicit* integral conversion — so every target without a
   native `__int128` (MSVC among them) fails to compile it.

fmt 12.1.0 has neither problem, still provides the `detail::uint128_fallback` that
`include/common/int128.h`'s existing `FMT_VERSION` guards select, and — unlike 12.2.0 —
does not deprecate the conversion operator spdlog v1.17.0 still routes through.
**No source changes are needed.**

### spdlog has to move with fmt

Under `SPDLOG_FMT_EXTERNAL`, spdlog v1.13.0 includes `<fmt/core.h>`, which fmt 12 reduced
to a stub that no longer provides `fmt::format` without `FMT_DEPRECATED_HEAVY_CORE`.
spdlog v1.17.0 includes `<fmt/format.h>` directly and still guards its
`find_package(fmt CONFIG REQUIRED)` behind `if(NOT TARGET fmt::fmt)`, so the
FetchContent-provided fmt continues to satisfy it.

### On scoping the suppressions to `PRIVATE`

`target_compile_options(... PUBLIC)` reaches every source of every target that *links* the
dependency, not just files that include it. `wasmedgeCommon` links `spdlog::spdlog` PUBLIC
and spdlog links `fmt::fmt` PUBLIC, so fmt's suppression list was reaching **all 115 `lib/`
TUs — including files with no fmt include at all**.

Flipping to `PRIVATE` therefore *removes* suppressions from WasmEdge's own code, and
`-isystem` cannot compensate, because those warnings attach to our own source lines. So
four flags are **re-homed into `WASMEDGE_CFLAGS`** rather than dropped — under clang-cl
`-Wall` expands to `-Weverything`, which enables them:

| Flag | Where it fires | Action |
|---|---|---|
| `-Wpadded` | 5358 hits, 51 of our files | re-home (WIN32+Clang) |
| `-Wunique-object-duplication` | ~55 `static inline` members of `lib/llvm/llvm.h` | re-home (WIN32+Clang) |
| `-Wnrvo` | our code | re-home (WIN32+Clang) |
| `-Wmissing-noreturn` | our code, **every Clang platform** | re-home (general Clang) |
| `-Wduplicate-enum` | only dependency headers | stays `PRIVATE` — `-isystem` subsumes |

Nothing is deleted; flags move.

### Also

- `lib/llvm/llvm.h` calls `std::memcpy` but never included `<cstring>` — it compiled only
  because fmt 11's fat headers dragged it in transitively. fmt 12's slim `base.h` does not.
  Fixed as its own commit.
- `fossa-deps.yml` still named spdlog `1.9.1` and omitted fmt entirely.

## Checklist

- [x] **DCO Signed-off**: all 5 commits signed off; author == sign-off email.
- [x] **Commit Messages**: Conventional Commits (`fix:` x1, `chore:` x4); headers <= 69 chars.
- [x] **Coding Style**: `lineguard` clean. The only C++ change is a one-line `#include`.
- [x] **Local Tests Passed**: see below.

If you are using AI-assisted tools:

- [x] **AI Disclosure**: every commit carries `Assisted-by: Claude (Anthropic)`.
      This contribution was developed with assistance from Claude (Anthropic).
- [x] **Human-in-the-loop**: reviewed and submitted by a human maintainer.

## Test Evidence

```
GCC 15.2.0    (WASMEDGE_USE_LLVM=ON)  100% tests passed, 0 failed out of 33
Clang 21.1.8  (WASMEDGE_USE_LLVM=ON)  100% tests passed, 0 failed out of 33
WASI-NN GGML  (fmt/ranges.h surface)  build clean, 386/386
```

Verified after the bump, with `-Werror` in force throughout:

- `FMT_VERSION 120100`, `SPDLOG_VER_MINOR 17` in both build trees.
- `-Wno-missing-noreturn` on **115/115** `lib/` TUs — the flag survived the
  `PUBLIC`→`PRIVATE` flip, now sourced from `WASMEDGE_CFLAGS` instead of leaked through
  fmt. `-Wno-sign-conversion` on **0/115** (fmt-`PRIVATE`, correctly never leaks).
- Zero TUs reach fmt or spdlog through a plain `-I` — checked across default, plugin,
  test, `WASMEDGE_LINK_PLUGINS_STATIC=ON` and WASI-NN configurations.

The previous 12.2.0 revision of this branch was green on Fedora (the `find_package(spdlog)`
path), Alpine static, macOS, Ubuntu 20.04, Manylinux and Android, and failed **only** on the
two upstream fmt 12.2.0 bugs described above — both of which 12.1.0 avoids.


ibmibmibm:
Second-opinion review (GPT-5.6) raised two points. Adjudication, with what I verified:

**Taken — `fmt::fmt-header-only` on the `find_package` path.** A genuine gap.
`cmake/WASINNDeps.cmake:529` substitutes `fmt::fmt-header-only` into MLX's
`INTERFACE_LINK_LIBRARIES`, so anything linking MLX inherits its include dirs. On the
FetchContent path this is already covered — fmt applies `FMT_SYSTEM_HEADERS_ATTRIBUTE`
to both `fmt` (CMakeLists.txt:331) and `fmt-header-only` (:380) — but on the
`find_package` path only `fmt::fmt` was being marked. A system fmt plus MLX would
therefore have reached the WASI-NN sources through a plain `-I` under `-Werror`. Fixed
by marking it too; folded into the system-includes commit.

**Not taken — moving the marking above the `if(TARGET spdlog::spdlog) return()`.**
That early return is unchanged from master, and all three `target_compile_options` calls
live inside the `else()` FetchContent branch that path never enters — so no suppressions
were ever applied there, and this is a pre-existing gap rather than a regression from
this PR. The proposed fix also has an unstated cost: marking a *parent* project's targets
as system mutates targets WasmEdge does not own and silently changes the parent's own
build. Worth addressing for `add_subdirectory` consumers, but on its own terms, not here.

**Not taken — quoting the genex expansion in the helper.** No fmt/spdlog target uses the
`$<BUILD_INTERFACE:/a;/b>` shape (the review concedes this), and #5145 defines a
byte-identical helper; diverging now would conflict when whichever lands second drops the
duplicate.

Re-verified after the change: Clang build clean, `ctest` 100% (33/33), and the
FetchContent-path compile lines are byte-identical (the new line only affects the
`find_package` branch).

hydai:
@ibmibmibm please rebase the branch to resolve the conflicts

hydai (approved):

PR #5161, opened 2026-07-13
fix(coredump): bound corestack value serialization

### LINKED ISSUE #5160: bug: avoid overflowing the corestack value buffer
## Summary
`lib/executor/coredump.cpp` serializes corestack locals and stack values into a 4-byte temporary buffer, but both loops copy `sizeof(int64_t)` bytes into that buffer. When coredump generation runs after a trap, this causes a heap buffer overflow in the coredump recovery/debugging path.

## Current State
`createCorestack` emits `0x7F` (`i32`) for each serialized local or stack value, allocates a 4-byte temporary buffer, and then copies 8 bytes into it with `std::memcpy(..., sizeof(int64_t))` in both serialization loops. The existing regression only checked whether a `coredump.*` file was created, so this unsafe write was not covered by a focused payload-validation test.

## Expected State
Coredump generation should serialize each corestack value without writing past the temporary buffer boundaries, and the generated `corestack` payload should remain loadable and structurally valid.

## Reproduction steps
1. Build WasmEdge from `master` with AddressSanitizer enabled.
2. Enable coredump generation and execute a module that traps, for example through the existing `Coredump.generateCoredump` regression path in `test/executor/ExecutorTest.cpp`.
3. Let execution enter `WasmEdge::Coredump::createCorestack`.
4. Observe the invalid write at the `memcpy` used for local or value-stack serialization, or confirm statically from the copy into `std::vector<Byte>(4)` with `sizeof(int64_t)`.

## Any logs you want to share for showing the specific issue
Static evidence before the fix:
- `lib/executor/coredump.cpp` allocated `std::vector<Byte> ValueBytes(4)` and copied `sizeof(int64_t)` bytes into it in both the locals loop and the value-stack loop.

Runtime symptom when exercising the trapping path:
- coredump generation runs immediately after an out-of-bounds memory access trap and reaches the affected serialization path in `WasmEdge::Coredump::createCorestack`.
- The issue is directly observable under ASan builds of the executor coredump test.

Related context:
- PR #3948 and PR #4488 touched coredump code, but neither addressed this overflow.

## Components
- Core

## WasmEdge Version or Commit you used
Observed on upstream `master` before the fix in PR #5161. The contributing branch implementing the fix is based on current `master` and references issue #5160.

## Operating system information
Source-level bug that is platform-independent. Reproduced locally on macOS.

## Hardware Architecture
Source-level bug that is architecture-independent. Reproduced locally on arm64.

## Appendix
Affected files:
- `lib/executor/coredump.cpp`
- `test/executor/ExecutorTest.cpp`

Likely root cause:
- `createCorestack` hard-coded an i32-sized destination buffer while using `sizeof(int64_t)` as the copy size.

Proposed direction:
- Serialize exactly the emitted 32-bit value width, or use a correctly sized fixed buffer that matches the debugger format.
- Add a regression that validates the generated `corestack` payload instead of only checking file creation.

Acceptance criteria:
- No out-of-bounds write occurs during coredump generation under ASan.
- A regression test validates the generated coredump payload instead of only checking for file creation.
- Serialized locals and stack values match the intended debugger format.

rootp1:
## Summary
- prevent `createCorestack` from copying past the temporary 4-byte value buffer
- add a coredump regression that validates the generated `corestack` custom section payload
- keep the change scoped to coredump serialization and its existing executor test target

## Linked Issue
Fixes #5160

## What Changed
- replace the unsafe 8-byte copy into a 4-byte buffer with a helper that serializes exactly 4 bytes for each emitted `i32` corestack value
- update `Coredump.generateCoredump` to run in an isolated temp directory, reload the emitted coredump module, and validate the `corestack` section structure and encoded value payload size

## Verification
- `bash ./.github/scripts/clang-format.sh /Library/Developer/CommandLineTools/usr/bin/clang-format` — passed
- `cmake -S . -B build -DWASMEDGE_BUILD_TESTS=ON -DWASMEDGE_USE_LLVM=OFF` — passed
- `cmake --build build --target wasmedgeExecutorCoreTests -j$(sysctl -n hw.logicalcpu)` — passed
- `ctest -R wasmedgeExecutorCoreTests --output-on-failure` — passed
- `./wasmedgeExecutorCoreTests --gtest_filter=Coredump.generateCoredump` (in `build/test/executor`) — passed
- `/Users/rootp1/.cargo/bin/lineguard --config .lineguardrc --verbose lib/executor/coredump.cpp test/executor/ExecutorTest.cpp` — passed
- `git ls-files | grep -v '^thirdparty' | grep -v '/thirdparty/' | grep -v '/dist/' | xargs codespell --ignore-words .github/workflows/ignore_words 2>/dev/null` — passed
- `cmake -S . -B build-asan -DWASMEDGE_BUILD_TESTS=ON -DWASMEDGE_USE_LLVM=OFF -DCMAKE_C_FLAGS='-fsanitize=address -fno-omit-frame-pointer' -DCMAKE_CXX_FLAGS='-fsanitize=address -fno-omit-frame-pointer' -DCMAKE_EXE_LINKER_FLAGS='-fsanitize=address' -DCMAKE_SHARED_LINKER_FLAGS='-fsanitize=address'` — passed
- `cmake --build build-asan --target wasmedgeExecutorCoreTests -j$(sysctl -n hw.logicalcpu)` — passed
- `./wasmedgeExecutorCoreTests --gtest_filter=Coredump.generateCoredump` (in `build-asan/test/executor`) — passed
- `cd build-asan && ctest -R wasmedgeExecutorCoreTests --output-on-failure` — passed
- `npx --yes -p commitlint@20.2.0 -p @commitlint/config-conventional@20.2.0 commitlint --extends @commitlint/config-conventional --from HEAD~1 --to HEAD --verbose` — passed

## Scope
- only `lib/executor/coredump.cpp` and `test/executor/ExecutorTest.cpp` are included in the branch diff against `upstream/master`

### DIFF
--- a/lib/executor/coredump.cpp
+++ b/lib/executor/coredump.cpp
@@ -6,6 +6,7 @@
 #include "common/types.h"
 #include "runtime/stackmgr.h"
 #include "spdlog/spdlog.h"
+#include <array>
 #include <cstdint>
 #include <cstring>
 #include <vector>
@@ -14,6 +15,18 @@ using namespace std::literals;
 
 namespace WasmEdge {
 namespace Coredump {
+namespace {
+
+std::array<Byte, sizeof(uint32_t)>
+serializeCoreStackValue(const Runtime::StackManager::Value &Value) noexcept {
+  std::array<Byte, sizeof(uint32_t)> ValueBytes{};
+  const auto RawValue = Value.unwrap();
+  std::memcpy(ValueBytes.data(), &RawValue, ValueBytes.size());
+  return ValueBytes;
+}
+
+} // namespace
+
 void generateCoredump(const Runtime::StackManager &StackMgr,
                       bool ForWasmgdb) noexcept {
   spdlog::info("Generating coredump..."sv);
@@ -128,9 +141,7 @@ AST::CustomSection createCorestack(
       // 0x7F implies i32. i128 is not supported here, and wasmgdb does not
       // support i64.
       Content.push_back(0x7F);
-      auto Value = Iter.unwrap();
-      std::vector<Byte> ValueBytes(4);
-      std::memcpy(ValueBytes.data(), &Value, sizeof(int64_t));
+      auto ValueBytes = serializeCoreStackValue(Iter);
       Content.insert(Content.end(), ValueBytes.begin(), ValueBytes.end());
     }
     if (!ForWasmgdb) {
@@ -138,9 +149,7 @@ AST::CustomSection createCorestack(
         // 0x7F implies i32. i128 is not supported here, and wasmgdb does not
         // support i64.
         Content.push_back(0x7F);
-        auto Value = Iter.unwrap();
-        std::vector<Byte> ValueBytes(4);
-        std::memcpy(ValueBytes.data(), &Value, sizeof(int64_t));
+        auto ValueBytes = serializeCoreStackValue(Iter);
         Content.insert(Content.end(), ValueBytes.begin(), ValueBytes.end());
       }
     }
--- a/test/executor/ExecutorTest.cpp
+++ b/test/executor/ExecutorTest.cpp
@@ -20,6 +20,7 @@
 #include "../spec/hostfunc.h"
 #include "../spec/spectest.h"
 
+#include <algorithm>
 #include <gtest/gtest.h>
 
 #include <array>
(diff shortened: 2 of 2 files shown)


rootp1:
Hello, @hydai @q82419
I would like to propose this PR, wdyt?
Please review the PR whenever you get a chance to

hydai:
Closing this and marking it as SPAM because it does not follow our PR template, which indicates it is not human-in-the-loop and is a highly machine-generated PR.

PR #5164, opened 2026-07-13
 fix(coredump): bound corestack value serialization

### LINKED ISSUE #5162: bug: avoid overflowing the corestack value buffer
### Summary

`lib/executor/coredump.cpp` serializes corestack locals and stack values into a 4-byte temporary buffer, but both loops copy `sizeof(int64_t)` bytes into that buffer. 

When coredump generation runs after a trap, this causes a heap buffer overflow in the coredump recovery/debugging path.

### Current State

`createCorestack` emits `0x7F` (`i32`) for each serialized local or stack value, allocates a 4-byte temporary buffer, and then copies 8 bytes into it with `std::memcpy(..., sizeof(int64_t))` in both serialization loops. The existing regression only checked whether a `coredump.*` file was created, so this unsafe write was not covered by a focused payload-validation test.

### Expected State

Coredump generation should serialize each corestack value without writing past the temporary buffer boundaries, and the generated `corestack` payload should remain loadable and structurally valid.

### Reproduction steps

1. Build WasmEdge from `master` with AddressSanitizer enabled.
2. Enable coredump generation and execute a module that traps, for example through the existing `Coredump.generateCoredump` regression path in `test/executor/ExecutorTest.cpp`.
3. Let execution enter `WasmEdge::Coredump::createCorestack`.
4. Observe the invalid write at the `memcpy` used for local or value-stack serialization, or confirm statically from the copy into `std::vector<Byte>(4)` with `sizeof(int64_t)`.

Static evidence before the fix
- `lib/executor/coredump.cpp` allocated `std::vector<Byte> ValueBytes(4)` and copied `sizeof(int64_t)` bytes into it in both the locals loop and the value-stack loop.

runtime symptome:
- coredump generation runs immediately after an out-of-bounds memory access trap and reaches the affected serialization path in `WasmEdge::Coredump::createCorestack`.
- The issue is directly observable under ASan builds of the executor coredump test.

### Components

Core

The contributing branch implementing the fix is based on current `master` and references issue #5160.

locally on macOS.

### Hardware Architecture

Reproduced locally on arm64.

### Appendix

_No response_

rootp1:
Closes #5162

## Description
Fixes a heap buffer overflow in coredump corestack serialization.

The coredump path was allocating a 4-byte temporary buffer for each serialized corestack value but copying `sizeof(int64_t)` bytes into it. This change makes the serializer write exactly the emitted 32-bit payload width and adds a regression test that reloads the generated coredump and validates the `corestack` section structure.

## Checklist

## Test Evidence

image
image

### DIFF
--- a/lib/executor/coredump.cpp
+++ b/lib/executor/coredump.cpp
@@ -6,6 +6,7 @@
 #include "common/types.h"
 #include "runtime/stackmgr.h"
 #include "spdlog/spdlog.h"
+#include <array>
 #include <cstdint>
 #include <cstring>
 #include <vector>
@@ -14,6 +15,18 @@ using namespace std::literals;
 
 namespace WasmEdge {
 namespace Coredump {
+namespace {
+
+std::array<Byte, sizeof(uint32_t)>
+serializeCoreStackValue(const Runtime::StackManager::Value &Value) noexcept {
+  std::array<Byte, sizeof(uint32_t)> ValueBytes{};
+  const auto RawValue = Value.unwrap();
+  std::memcpy(ValueBytes.data(), &RawValue, ValueBytes.size());
+  return ValueBytes;
+}
+
+} // namespace
+
 void generateCoredump(const Runtime::StackManager &StackMgr,
                       bool ForWasmgdb) noexcept {
   spdlog::info("Generating coredump..."sv);
@@ -128,9 +141,7 @@ AST::CustomSection createCorestack(
       // 0x7F implies i32. i128 is not supported here, and wasmgdb does not
       // support i64.
       Content.push_back(0x7F);
-      auto Value = Iter.unwrap();
-      std::vector<Byte> ValueBytes(4);
-      std::memcpy(ValueBytes.data(), &Value, sizeof(int64_t));
+      auto ValueBytes = serializeCoreStackValue(Iter);
       Content.insert(Content.end(), ValueBytes.begin(), ValueBytes.end());
     }
     if (!ForWasmgdb) {
@@ -138,9 +149,7 @@ AST::CustomSection createCorestack(
         // 0x7F implies i32. i128 is not supported here, and wasmgdb does not
         // support i64.
         Content.push_back(0x7F);
-        auto Value = Iter.unwrap();
-        std::vector<Byte> ValueBytes(4);
-        std::memcpy(ValueBytes.data(), &Value, sizeof(int64_t));
+        auto ValueBytes = serializeCoreStackValue(Iter);
         Content.insert(Content.end(), ValueBytes.begin(), ValueBytes.end());
       }
     }
--- a/test/executor/ExecutorTest.cpp
+++ b/test/executor/ExecutorTest.cpp
@@ -15,6 +15,7 @@
 //===----------------------------------------------------------------------===//
 
 #include "common/spdlog.h"
+#include "experimental/scope.hpp"
 #include "vm/vm.h"
 
 #include "../spec/hostfunc.h"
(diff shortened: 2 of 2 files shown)


rootp1:
This is a pr, related to issue #5162 
@q82419, @hydai please have a look whenever you have a chance to

rootp1:
I updated it with the evidence screenshot @hydai

test/executor/ExecutorTest.cpp
```diff
@@ -40,6 +41,24 @@ using namespace std::literals;
...
+  ~ScopedWorkingDirectory() {
+    std::filesystem::current_path(Original);
+    std::filesystem::remove_all(Current);
+  }
+
+private:
+  std::filesystem::path Original;
+  std::filesystem::path Current;
+};
+
```
hydai: What's the reason for creating a new class if you're only going to use it once?

hydai (changes requested):

test/executor/ExecutorTest.cpp
rootp1: I added it to keep the coredump test isolated
so the test switches to a temporary directory and needs to restore the original path even if an assertion exits early.

since it is only used in this one test, I agree a file scope class is probably more than necessary.
I’ll simplify it.

test/executor/ExecutorTest.cpp
```diff
@@ -321,15 +336,60 @@ TEST(Coredump, generateCoredump) {
...
+        return 0U;
+      }
+      const uint8_t Byte = Content[Offset++];
+      Result |= static_cast<uint32_t>(Byte & 0x7FU) << Shift;
+      if ((Byte & 0x80U) == 0) {
+        return Result;
+      }
+      Shift += 7U;
+    }
+  };
```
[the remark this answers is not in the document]
rootp1: Good catch, thanks. Fixed by bounding the loop to 5 iterations (28 bit shift max) with a for loop, and calling ADD_FAILURE if the encoding still hasn't terminated by then, so we no longer risk undefined behavior from an oversized shift on malformed input.

lib/executor/coredump.cpp
```diff
@@ -14,6 +15,18 @@ using namespace std::literals;
...
 namespace Coredump {
+namespace {
+
+std::array<Byte, sizeof(uint32_t)>
+serializeCoreStackValue(const Runtime::StackManager::Value &Value) noexcept {
+  std::array<Byte, sizeof(uint32_t)> ValueBytes{};
+  const auto RawValue = Value.unwrap();
+  std::memcpy(ValueBytes.data(), &RawValue, ValueBytes.size());
+  return ValueBytes;
+}
```
q82419, line 26: For making a helper function just return an array not make sense. you can refine at the original place.

q82419 (changes requested):

test/executor/ExecutorTest.cpp
```diff
@@ -321,15 +336,60 @@ TEST(Coredump, generateCoredump) {
...
+  EXPECT_EQ(Content[Offset++], 0x00U);
+  EXPECT_EQ(Content[Offset++], 0x04U);
+  EXPECT_EQ(
+      std::string_view(reinterpret_cast<const char *>(&Content[Offset]), 4),
+      "main");
+  Offset += 4U;
+  EXPECT_EQ(ReadU32(), 1U);
+  EXPECT_EQ(Content[Offset++], 0x00U);
+  static_cast<void>(ReadU32());
+  static_cast<void>(ReadU32());
```
Ankitdotraider, line 388: Noticed the Funcidx/Codeoffset reads here are discarded instead of asserted. Is that because Funcidx isn't right yet? I see a TODO about main's Funcidx being 0 in createCorestack.

PR #5167, opened 2026-07-14
feat(wasi_crypto): implement ML-KEM post-quantum key encapsulation

### LINKED ISSUE #5110: feat: implement ML-KEM post-quantum key encapsulation in the WASI-Crypto kx module
### Summary

The WASI-Crypto key exchange module exposes `kx_encapsulate` and `kx_decapsulate`, but they currently return `NOT_IMPLEMENTED`, so only classical Diffie-Hellman key exchange works today.

The wasi-crypto spec already names ML-KEM (FIPS 203), and OpenSSL 3.5 provides it through the same EVP_PKEY interface the plugin already uses, so we can add it without a new dependency. This covers ML-KEM-512, 768, and 1024 behind those two functions.

### Details

The change is additive and gated on OpenSSL 3.5, so older builds are unaffected and nothing existing breaks.

Impact: WasmEdge gains post-quantum key encapsulation, which is already becoming a default in production TLS (OpenSSL 3.5, nginx, HAProxy), and the runtime stays current with the spec instead of falling behind once applications start expecting it.

Timeline, about one month:

- [ ] Week 1: design note, confirm the approach and key encoding with maintainers
- [ ] Weeks 1 to 3: implement ML-KEM 512/768/1024 gated behind OpenSSL 3.5, with round-trip and negative tests
- [ ] Week 4: FIPS 203 known-answer tests, short docs, open the PR

The OpenSSL 3.5 upgrade for the Linux build images is tracked separately, since republishing the dependency images needs maintainer access.

### Appendix

wasi-crypto spec naming ML-KEM: https://github.com/WebAssembly/wasi-crypto/commit/d1e85c621c7538439a58ad12e889c9e4e6874c9b

I have already run the OpenSSL 3.5 upgrade as an experiment on my fork to confirm what it affects, and I would like to own this one.

parthdagia05:
## Description

Implements ML-KEM (FIPS 203) key encapsulation in the WASI-Crypto `kx` module. Closes #5110.

`kx_encapsulate` and `kx_decapsulate` previously returned `NOT_IMPLEMENTED`, so only classical Diffie-Hellman worked. This adds `ML-KEM-512`, `ML-KEM-768` and `ML-KEM-1024` through OpenSSL's `EVP_PKEY` KEM interface, which the plugin already uses. No new dependency.

Everything is gated behind `#if OPENSSL_VERSION_NUMBER >= 0x30500000L`. On older OpenSSL the algorithms stay unregistered and behaviour is unchanged, so no minimum OpenSSL version is raised. macOS CI verifies the ML-KEM path, Linux compiles it out.

## Key encodings

The spec does not pin the RAW encodings, so these are choices:

* Public key: `ek`, 800 / 1184 / 1568 bytes by parameter set, length validated.
* Secret key: import accepts 64 bytes as the seed `d||z`, or the expanded `dk`. Export emits `dk`. OpenSSL only keeps a seed for keys generated or imported from a seed, so a key imported from `dk` has none and seed-only export would fail for it.
* Keypair: `ek || dk`. Import validates that both halves belong to the same keypair: it reconstructs the key from `dk`, derives the encapsulation key, and rejects the input with `INVALID_KEY` if it does not match the supplied `ek`.

`dh()`, `verify()` and `toKeyPair()` return `NOT_IMPLEMENTED`. A KEM has no Diffie-Hellman step; `dh()` exists only because the generic `kx` dispatchers read the peer key type from its signature. `verify()` matches the existing X25519 behaviour.

## Note for reviewers

Decapsulating a corrupted ciphertext succeeds and returns a different pseudorandom secret. That is FIPS 203 implicit rejection, not a bug. The tests assert success with a mismatched secret for corrupted input, and failure only for wrong-length input.

`test/plugins/wasi_crypto/helper.cpp` has one unrelated-looking fix: `kxEncapsulate` wrote its two 4-byte output handles at byte offsets 0 and 1, so they overlapped and the secret handle was clobbered. It was never noticed because `kx_encapsulate` always returned `NOT_IMPLEMENTED`, so the handles were never read. Changed to offsets 0 and 8.

## Tests

Nine cases added to the existing `test/plugins/wasi_crypto/kx.cpp`, gated on OpenSSL 3.5. No new test binary. Every case runs against all three parameter sets.

* keypair generation, exported `ek` length, two generations differ
* secret key and public key accessors, derived public key matches the keypair's
* encapsulation, ciphertext and shared secret lengths, two encapsulations differ
* full round trip, encapsulate then decapsulate, secrets match
* corrupted ciphertext decapsulates successfully to a different secret, wrong-length ciphertext fails
* public key import, re-export is identical, imported key interoperates with the original secret key
* secret key export, the encapsulation key appears at the FIPS 203 offset inside `dk`
* secret key import from both `dk` and a 64-byte seed, seed import is deterministic
* keypair export layout, and keypair import rejecting mismatched or corrupted halves
* wrong lengths and unsupported encodings rejected throughout

## Checklist

- [x] **DCO Signed-off**: commits are signed off (`git commit -s`).
- [x] **Commit Messages**: follow Conventional Commits.
- [x] **Local Tests Passed**: 18/18 `wasiCryptoTests` pass against OpenSSL 3.5.7.


parthdagia05:
https://github.com/parthdagia05/WasmEdge/pull/5

parthdagia05:
Hi @hydai, I have completed most of the work from my side and have listed all the relevant commits in the PR description. Let me know if there's anything else you'd like me to add or any further changes you'd like me to make.

parthdagia05:
@hydai I also noticed that ArrayOutput's std::vector<uint8_t>&& constructor can still silently copy secret data if a future caller passes a vector. A SecretVec(std::vector<uint8_t>&&) constructor could eliminate that class of mistakes project-wide. Since that's a broader API change, what should I be doing here?

hydai:
I don't have a quick answer at the moment; I need to investigate it first.

parthdagia05:
@hydai i pushed the chnages

hydai:
Please resolve the fixed threads.

plugins/wasi_crypto/kx/kem/mlkem.cpp
```diff
@@ -0,0 +1,321 @@
...
+                 __WASI_CRYPTO_ERRNO_ALGORITHM_FAILURE);
+
+  std::vector<uint8_t> Ciphertext(CtSize);
+  SecretVec Secret(SecretSize);
+  size_t CiphertextLen = CtSize;
+  size_t SecretLen = SecretSize;
+  ensureOrReturn(EVP_PKEY_encapsulate(EncCtx.get(), Ciphertext.data(),
+                                      &CiphertextLen, Secret.data(),
+                                      &SecretLen) > 0,
+                 __WASI_CRYPTO_ERRNO_INVALID_KEY);
```
hydai: **[Suggestion, non-blocking] Errno semantics on encapsulate/decapsulate failure**

`EVP_PKEY_encapsulate` (here) and `EVP_PKEY_decapsulate` (L199-202) failures map to `INVALID_KEY`. However, length errors are already caught earlier as `INVALID_LENGTH`, and the key itself was validated at import time — a failure at this point is most likely a provider/RNG error, so `ALGORITHM_FAILURE` would be semantically more accurate.

*— Review agent powered by Copilot (Claude Fable 5). AI-generated feedback; may contain mistakes.*

plugins/wasi_crypto/kx/kem/mlkem.cpp
```diff
@@ -0,0 +1,321 @@
...
+    std::vector<uint8_t> DerivedPk(PkSize);
+    size_t Size = PkSize;
+    ensureOrReturn(
+        EVP_PKEY_get_octet_string_param(Res.get(), OSSL_PKEY_PARAM_PUB_KEY,
+                                        DerivedPk.data(), PkSize, &Size) > 0,
+        __WASI_CRYPTO_ERRNO_ALGORITHM_FAILURE);
+    ensureOrReturn(Size == PkSize, __WASI_CRYPTO_ERRNO_ALGORITHM_FAILURE);
+    ensureOrReturn(
+        std::equal(DerivedPk.begin(), DerivedPk.end(), Encoded.begin()),
+        __WASI_CRYPTO_ERRNO_INVALID_KEY);
```
hydai: **[Suggestion, non-blocking] OpenSSL can validate ek↔dk natively**

OpenSSL 3.5's `ml_kem_key_fromdata` accepts `OSSL_PKEY_PARAM_PUB_KEY` and `OSSL_PKEY_PARAM_PRIV_KEY` together in a single `EVP_PKEY_fromdata` call and internally `memcmp`s the supplied ek against the copy embedded in dk (see `providers/implementations/keymgmt/ml_kem_kmgmt.c`). The manual `DerivedPk` export + `std::equal` here is behaviorally equivalent but could be dropped in favor of the built-in check.

*— Review agent powered by Copilot (Claude Fable 5). AI-generated feedback; may contain mistakes.*

test/plugins/wasi_crypto/kx.cpp
```diff
@@ -124,6 +130,590 @@ TEST_F(WasiCryptoTest, KxDh) {
...
+    WASI_CRYPTO_EXPECT_FAILURE(keypairImport(__WASI_ALGORITHM_TYPE_KEY_EXCHANGE,
+                                             Alg, BadEk,
+                                             __WASI_KEYPAIR_ENCODING_RAW),
+                               __WASI_CRYPTO_ERRNO_INVALID_KEY);
+
+    std::vector<uint8_t> BadDk = Raw;
+    BadDk[PkSize] ^= 0xff;
+    WASI_CRYPTO_EXPECT_FAILURE(keypairImport(__WASI_ALGORITHM_TYPE_KEY_EXCHANGE,
+                                             Alg, BadDk,
+                                             __WASI_KEYPAIR_ENCODING_RAW),
```
hydai, line 644: **[Informational] BadDk rejection relies on OpenSSL's import PCT**

Flipping a `dk_PKE` byte is rejected thanks to OpenSSL's pairwise consistency test on import (default `random`; also mandated in the FIPS provider). If a deployment explicitly sets `ml-kem.import_pct_type=none`, a corrupted `dk_PKE` with intact `ek`/`H(ek)` would import successfully. Default behavior is correct — noting this only for awareness.

*— Review agent powered by Copilot (Claude Fable 5). AI-generated feedback; may contain mistakes.*

hydai:
> @hydai I also noticed that ArrayOutput's std::vector<uint8_t>&& constructor can still silently copy secret data if a future caller passes a vector. A SecretVec(std::vector<uint8_t>&&) constructor could eliminate that class of mistakes project-wide. Since that's a broader API change, what should I be doing here?

This should be a standalone PR. I don't think this API is exposed to the public interface; it is only used internally within the wasi_crypto plugin.

plugins/wasi_crypto/kx/kem/mlkem.cpp
parthdagia05: Changed both to ALGORITHM_FAILURE. Length is checked before the call and the key is validated at import, so a failure here is a provider or RNG error rather than a bad key.

plugins/wasi_crypto/kx/kem/mlkem.cpp
parthdagia05: Done. I now pass OSSL_PKEY_PARAM_PUB_KEY and OSSL_PKEY_PARAM_PRIV_KEY in a single EVP_PKEY_fromdata call and dropped the manual export and compare. Checked on OpenSSL 3.6.3 that a mismatched or corrupted ek is rejected during import, and the keypair import tests still pass.

test/plugins/wasi_crypto/kx.cpp
parthdagia05: Noted, thanks. The test relies on the default import PCT, which is what a normal deployment uses. Setting ml-kem.import_pct_type=none is an explicit opt out, so I have left the test as it is.

parthdagia05:
Understood. I will open a separate PR for the SecretVec(std::vector<uint8_t>&&) constructor and keep this one to ML-KEM.

I have also pushed 109a8043 addressing the two non-blocking suggestions above, and resolved the threads that are now fixed.

hydai:
I think we are pretty close to merging this. Could you please squash all your commits into a single one?

parthdagia05:
@hydai Before you approve the workflow, I’ll first check that all the checks are passing on my fork.

parthdagia05:
Screenshot 2026-08-27 at 9 56 00 PM
@hydai

parthdagia05:
@hydai should I rebase this branch?

parthdagia05:
The only red check is Harden-Runner, unrelated to this change

hydai (approved):

hydai:
Thanks for your contribution. We plan to ship this in the 0.18.0 release.

parthdagia05:
@hydai Thanks a lot for your guidance and help throughout this! I learned a lot about cryptography while working on this.

PR #5196, opened 2026-07-23
fix: correct Ubuntu 20.04 fallback to read /etc/lsb-release in instal…

deepak0x:
…l.py

The Python 3 distro detection has a fallback that reads /etc/lsb_release with an underscore. On Ubuntu the file is /etc/lsb-release with a hyphen, so the underscore path never exists. The 2>/dev/null hides the missing-file error, so run_shell_command returns an empty string and "Ubuntu 20.04" in "" is always false, leaving the fallback unable to detect anything.

The other two reads in the same block already use the hyphenated path, so this was a typo. Point the fallback at /etc/lsb-release so it can actually match an Ubuntu 20.04 host when the primary RELEASE-line check misses.

## Checklist

- [ ] **Coding Style**: Run `clang-format` on your code to ensure it meets the coding style guidelines.

## Test Evidence

---

### DIFF
--- a/utils/install.py
+++ b/utils/install.py
@@ -1253,7 +1253,7 @@ class Compat:
                     if VersionString(__lsb_rel).compare(
                         "20.04"
                     ) >= 0 or "Ubuntu 20.04" in run_shell_command(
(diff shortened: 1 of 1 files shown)


hydai:

Issue #5153, opened 2026-07-13
LFX Mentorship (2026, Term 3): Support for the Wide Arithmetic Proposal


hydai:
### Project Title

Support for Wide Arithmetic, a Phase 3 WebAssembly proposal

### Description

Add support for the WebAssembly Wide Arithmetic proposal to WasmEdge by implementing the `i64.add128`, `i64.sub128`, `i64.mul_wide_s`, and `i64.mul_wide_u` instructions. These instructions enable efficient 128-bit addition and subtraction, as well as signed and unsigned 64×64-bit widening multiplication.

Since this is not a large proposal, the mentee must implement it in all execution modes, including interpreter, JIT, and AOT.

### Expected Outcome

1. WasmEdge must be able to correctly parse, validate, and execute all four Wide Arithmetic instructions while preserving the semantics defined by the proposal.
2. The implementation must pass the corresponding WebAssembly specification tests and provide consistent behavior across the interpreter, AOT compiler, and JIT engine.
3. The document must be completed within the WasmEdge/docs repository.

### Recommend skills

C++, WebAssembly, LLVM

### Pre-tests

https://hyd.ai/lfx-wasmedge-pretests/2026-lfx-term-3-5153.html

### Mentor(s)

Hung-Ying Tai (@hydai, hydai@secondstate.io)
Yi-Ying He (@q82419, yiying@secondstate.io)

### Apply Link

https://mentorship.lfx.linuxfoundation.org/project/14d8a870-ed14-4a49-8b8c-8ac50c4605cc

### Appendix

Wide Arithmetic Proposal: https://github.com/WebAssembly/wide-arithmetic

Aditya7880900936:
Hi @hydai,

I'm very interested in this LFX mentorship project and would love to start contributing to WasmEdge before the application period.

I'm an active open source contributor with contributions merged in CNCF projects including Envoy Gateway, Dapr, Istio, and Kuadrant, where I've primarily worked on backend and systems-related features and bug fixes. I'm now looking to deepen my understanding of WebAssembly runtimes and compiler/runtime internals, and this project aligns well with that goal.

I'll start by working through the Wide Arithmetic proposal, the pre-test, and setting up the development environment. If there are any beginner-friendly issues or tasks that would help me become familiar with the relevant parts of the runtime, I'd really appreciate your guidance.

Looking forward to contributing to WasmEdge. Thanks!

akansh-sinha:
Hey!
[...]

PR #5202, opened 2026-07-23
fix: erase exception payload on catch_all in interpreter

anusha19murthy:
## Description

While looking into the exception handling code in `lib/executor/helper.cpp` (added in #5168, part of the wide arithmetic mentorship area I'm exploring, see #5153), I found a bug where catch_all and catch_all_ref don't discard the exception payload from the value stack. This leaves stale values behind that get used later in the function, so you get a wrong result with no trap or error at all.

The issue is in `Executor::throwException`. `popTopHandler` in `stackmgr.h` always keeps the payload values on top of the stack:

```cpp
ValueStack.erase(ValueStack.begin() + TopHandler.VPos,
                 ValueStack.end() - AssocValSize);
```

that's correct for catch/catch_ref since they're supposed to deliver the payload to the label. But catch_all/catch_all_ref are supposed to discard the payload per spec, and nothing in throwException was doing that. The validator's recordJump also computes Arity=0 for catch_all (or 1 for catch_all_ref, just the exnref), because it never saw the payload during validation. So branchToLabel only erases what the validator told it to, and the payload just sits there and corrupts whatever comes after it.

Repro:

```wat
(module
  (tag $e (param i32))
    (i32.const 1)
    (block $L
      (try_table (catch_all $L)
        (i32.const 42)
        (throw $e))
      unreachable)))
```

Before the fix, this returns 42 through the interpreter, should return 1. AOT is not affected because it stores the payload in `PendingExn` separately, not on the operand stack, so `proxyCatchPop` just clears it for catch_all.

Fix, erase the payload before branching when the clause is catch_all/catch_all_ref:

```cpp
if (C.IsRef) {
  const Runtime::Instance::ExceptionInstance *Inst = ExnInst;
  if (Inst == nullptr) {
    auto Payload = StackMgr.getTopSpan(AssocValSize);
    std::vector<ValVariant> Vec(Payload.begin(), Payload.end());
    auto *ModInst = const_cast<Runtime::Instance::ModuleInstance *>(
        StackMgr.getModule());
    Inst = ModInst->newException(&TagInst, std::move(Vec));
  }
  if (C.IsAll) {
    // catch_all_ref only delivers the exnref, payload itself is discarded
    StackMgr.eraseValueStack(AssocValSize, 0);
  }
  StackMgr.push(
      RefVariant(ValType(TypeCode::Ref, TypeCode::ExnRef), Inst));
} else if (C.IsAll) {
  // plain catch_all discards the payload entirely
  StackMgr.eraseValueStack(AssocValSize, 0);
}
```

For catch_all_ref I erase the payload after building the exception instance from it (need those values to construct the exnref) but before pushing the exnref itself. For plain catch_all I just erase it before the branch. catch/catch_ref paths never hit the IsAll branch so they're untouched.

Added a regression test in `ExecutorRegressionTest.cpp`, `TEST(ExecutorRegression, CatchAllDiscardsPayload)`, covering both cases:
- Part 1, catch_all: pushes i32.const 1, throws a tag with payload 42 inside try_table (catch_all $L). Without the fix the function returns 42, with the fix it returns 1.
- Part 2, catch_all_ref: same setup but catch_all_ref, exnref gets pushed then dropped. Without the fix the leftover payload sits underneath the exnref so the drop removes the wrong thing and 42 leaks through. With the fix it returns 1.

I verified this both ways, built with the fix reverted and ran just this test, it fails with `(*Result)[0] == 42, expected 1` for both parts, exactly reproducing the corruption. Then restored the fix and reran, goes green. So the test is catching the real bug, not a false positive.

Built with `-DWASMEDGE_USE_LLVM=OFF` since this is interpreter-only, AOT path was never affected.

## Checklist

## Test Evidence
$ ninja wasmedgeExecutorRegressionTests
[3/3] Linking CXX executable test/executor/wasmedgeExecutorRegressionTests

$ ./test/executor/wasmedgeExecutorRegressionTests
[==========] Running 7 tests from 1 test suite.
[ RUN ] ExecutorRegression.RefTestNullAbility
[ OK ] ExecutorRegression.RefTestNullAbility (1 ms)
[ RUN ] ExecutorRegression.RefCastNullAbility
[ OK ] ExecutorRegression.RefCastNullAbility (0 ms)
[ RUN ] ExecutorRegression.BrOnCastNullAbility
[ OK ] ExecutorRegression.BrOnCastNullAbility (0 ms)
[ RUN ] ExecutorRegression.NullLocalConcreteType
[ OK ] ExecutorRegression.NullLocalConcreteType (0 ms)
[ RUN ] ExecutorRegression.ThrowRefPreservesPayload
[ OK ] ExecutorRegression.ThrowRefPreservesPayload (0 ms)
[ RUN ] ExecutorRegression.CatchAllDiscardsPayload
[ OK ] ExecutorRegression.CatchAllDiscardsPayload (0 ms)
[ RUN ] ExecutorRegression.CallIndirectNonFuncTable
[ OK ] ExecutorRegression.CallIndirectNonFuncTable (0 ms)
[==========] 7 tests from 1 test suite ran. (4 ms total)
[ PASSED ] 7 tests.


q82419 (approved):
Nice catch.

PR #5206, opened 2026-07-23
fix(coredump): fix heap buffer overflow and file size truncation in coredump generation

### LINKED ISSUE #5205: bug: heap buffer overflow and 32-bit truncation in coredump serialization
### Summary

Three defects were identified in lib/executor/coredump.cpp: a heap buffer overflow in createCorestack during local/stack value serialization, a 32-bit truncation cast on file output, and a hardcoded thread name length byte.

### Current State

While reviewing `lib/executor/coredump.cpp`, three defects were identified in the coredump generation pipeline:

1. **Heap Buffer Overflow in `createCorestack`**: `ValueBytes` vector was allocated with 4 bytes (`sizeof(uint32_t)`), but `std::memcpy` copied `sizeof(int64_t)` (8 bytes) from the stack value into it, overwriting 4 bytes past the vector storage.
2. **Truncating Cast in `generateCoredump`**: `File.write` received `Res->size()` cast to `uint32_t`. `std::ofstream::write` takes `std::streamsize` (signed 64-bit on 64-bit platforms), silently truncating large coredump writes (>4 GiB).
3. **Hardcoded Thread Name Length Byte**: The length prefix byte for the main thread name was hardcoded as `0x04` equal to `strlen("main")` instead of deriving from `ThreadName.size()`.

### Expected State

- `ValueBytes` vector allocation size should match `sizeof(int64_t)` (8 bytes).
- Coredump binary file write size should be cast to `std::streamsize`.
- Thread length byte should be calculated dynamically via `ThreadName.size()`.

### Reproduction steps

1. Build WasmEdge from source: `cmake -S . -B build -DWASMEDGE_BUILD_TESTS=ON && cmake --build build -j$(nproc)`
2. Trigger coredump generation on a WebAssembly module with local variables or stack values.
3. Inspect coredump serialization in `lib/executor/coredump.cpp`.

Heap buffer overflow detected during createCorestack:
AddressSanitizer: heap-buffer-overflow on address 0x602000000104
WRITE of size 8 at 0x602000000104 thread T0
    #0 in WasmEdge::Coredump::createCorestack lib/executor/coredump.cpp:132

### Components

Core

master branch (commit ed7b3722ed)

Ubuntu 22.04 / Linux

### Hardware Architecture

x86_64

### Appendix

Fix branch pushed with commit DCO and ready for PR.

om7057:
This PR fixes a few issues I noticed while tracing the coredump generation path in `lib/executor/coredump.cpp`.

The main issue was in `createCorestack`, where `ValueBytes` was allocated using `sizeof(uint32_t)` but `std::memcpy` copied an `int64_t` into it. That results in writing past the end of the vector whenever stack or local values are serialized. The allocation now matches the copied type.

While looking through the same code, I also fixed two related issues:

* Use `std::streamsize` when passing the buffer size to `std::ofstream::write` instead of casting to `uint32_t`, avoiding truncation for large coredumps.
* Derive the thread name length from `ThreadName.size()` instead of using the hardcoded `0x04` value.

I built WasmEdge with `-DWASMEDGE_BUILD_TESTS=ON` and ran the unit tests locally to verify that the coredump generation path still behaves as expected.

Closes #5205.

## Checklist

## Test Evidence

---


om7057:
Hi @hydai @yiying,

Whenever you have some time, I'd appreciate a review of this PR. I've addressed the issue and verified the changes with the existing unit tests locally.

Thanks!

lib/executor/coredump.cpp
```diff
@@ -84,10 +85,8 @@ AST::CustomSection createCorestack(
   // thread-info type 0x00 for wasmedbg
   Content.push_back(0x00);
 
-  // Thread name size
```
hydai, line 87: You should keep this comment.

hydai (changes requested):

lib/executor/coredump.cpp
om7057: Sure, I've restored the comment.

om7057:
@hydai Is there anything else in the PR that needs to be reconsidered or verified?
Please let me know; I'm open to more feedback to make the PR ready to merge.

Ankitdotraider (commented):
The memcpy overflow is real. But the fix grows the buffer to 8 bytes instead of shrinking the copy to 4, and the comment right above says these should be 4 byte i32s (tag is 0x7F). So now every value writes 8 bytes while its tag says 4, crash's gone but the file format's broken, misaligned by 4 bytes per entry.
Would just keep ValueBytes(4) and memcpy sizeof(uint32_t) instead, matches what the comment already says, and you shouldn't tag the maintainers before a week once you already tagged them. good practice imo. 

cc @hydai

om7057:
@Ankitdotraider You're right, thanks for catching that. Pushed a fix that keeps the buffer and memcpy at `sizeof(uint32_t)` for both the locals and stack loops, so it lines up with the `0x7F` tag again. Also noted on the tagging, will space that out going forward.

Ankitdotraider:
looks fine to me, good work.

om7057:
@hydai, could you please check this PR?

om7057:
@q82419 @hydai Once you get free, could you please check this PR's status for merging?

lib/executor/coredump.cpp
```diff
@@ -85,9 +86,8 @@ AST::CustomSection createCorestack(
   Content.push_back(0x00);
 
   // Thread name size
-  Content.push_back(0x04);
-
   std::string ThreadName = "main";
+  Content.push_back(static_cast<uint8_t>(ThreadName.size()));
```
q82419: `static_cast` to `uint8_t` doesn't make sense here. This works only because the size of `"main"` is 4, which in 1-byte size. My seggestion is use `serializeU32` here.

q82419 (changes requested):

om7057:
Good catch @q82419, that was leftover from before this used the serializer helpers. Switched it to `Ser.serializeU32(...)`, same as the rest of the corestack fields.

om7057:
@q82419 @hydai small reminder on this one!

q82419 (approved):

q82419 (dismissed):

om7057:
Pushed a fix for the Codecov patch coverage gap. Added ExecutorRegression.CoredumpCorestackValueWidth in ExecutorRegressionTest.cpp, which parses the generated coredump's corestack section back out and walks it entry by entry (thread name length, per-value tag byte + 4-byte width, frame boundaries all consumed exactly). Verified it locally by reverting the fix, which makes it fail with a clear tag/offset mismatch, then restoring it, which passes; full local suite still green.
CC: @hydai @q82419

q82419:
This was resolved in #5284, therefore close this PR.

PR #5214, opened 2026-07-26
fix(system): keep the embedder's signal handlers and alternate stack usable

### LINKED ISSUE #2400: fatal error: non-Go code set up signal handler without SA_ONSTACK flag
## Description

Since adding support for executing AOT compiled modules, with wasmedge in Go, I have tests that fail with 
```
signal 16 received but handler not on signal stack
fatal error: non-Go code set up signal handler without SA_ONSTACK flag
```

### Current State

When running the project's tests, the above error obfuscates any test failures. Some tests that passed before, now fail with the above error. Some tests are flaky - sometimes they pass, and other times I get this error.

### Expected

Not having this `signal handler` error obfuscate the real failure.

What I've found is:

https://pkg.go.dev/os/signal#hdr-Go_programs_that_use_cgo_or_SWIG
> If the non-Go code installs any signal handlers, it must use the SA_ONSTACK flag with sigaction. Failing to do so is likely to cause the program to crash if the signal is received. Go programs routinely run with a limited stack, and therefore set up an alternate signal stack.

This might be solved by `Action.sa_flags = SA_SIGINFO | SA_ONSTACK;` here https://github.com/WasmEdge/WasmEdge/blob/4c37361fa44a6c2860b922ef3af3af594bb760c8/lib/system/fault.cpp#L129

## Environment

- Hardware Architecture: (e.g. x86\_64 or aarch64): arm64
- Operating system: (e.g. Ubuntu 20.04 or CentOS 7.6): macOS 13.3.1
- WasmEdge 0.11.2

- C++ Compiler version:
- CMake version:
- CMake flags: (e.g. `-DCMAKE_BUILD_TYPE=Release -DWASMEDGE_BUILD_AOT_RUNTIME=ON`)
- Boost version:

iemAnshuman:
## Description

Fixes #2400.

WasmEdge installs POSIX fault handlers while running WebAssembly. The old path had a few problems that show up under embedders like Go/cgo:

1. Handlers were registered without SA_ONSTACK, so Go can abort with "non-Go code set up signal handler without SA_ONSTACK flag".
2. Even with SA_ONSTACK, we left the handler via longjmp and never hit sigreturn. On Darwin that can leave SS_ONSTACK latched, so later signals (including ones we never register, like SIGURG) land off the alternate stack.
3. enableHandler overwrote the embedder's handlers, and disableHandler reset them to SIG_DFL instead of restoring them.
4. A fault on a thread with no Fault context hit assuming(localHandler != nullptr), which is UB under NDEBUG.

This PR:

- sets SA_ONSTACK on the fault handlers
- resumes faults through a ucontext rewrite + normal return so the kernel sigreturn clears SS_ONSTACK and restores the pre signal mask, then longjmps from a trampoline on the original stack
- saves and restores the previous sigactions
- chains to the previous handler when there is no local Fault (Windows VEH returns EXCEPTION_CONTINUE_SEARCH_ in the matching cases)
- uses _setjmp for PREPARE_FAULT on POSIX so the fast path does not pay for a mask save

## Tests

- FaultHandler.PreservesPreviousSignalActions
- FaultHandler.ChainsSignalFromThreadWithoutFaultContext
- LazyJITTest.RepeatedFaultsPreserveAlternateSignalState (real sigaltstack, two JIT traps, mask and alt stack state unchanged)

## Checklist

* [x] DCO Signed-off: All commits are signed-off (`git commit -s`).
* [x] Commit Messages: Conventional Commits.
* [x] Coding Style: clang-format clean on the touched sources.
* [x] Local Tests Passed: logs below.
* [x] AI Disclosure: Assisted-by trailer in commits; noted here.
* [x] Human-in-the-loop: reviewed and verified locally before submit.

This contribution was developed with assistance from Claude Code (Anthropic) and earlier from OpenAI GPT-5.6.

## Test Evidence

macOS arm64, WASMEDGE_USE_LLVM=ON (RelWithDebInfo) and OFF (Debug):

```text
wasmedgeThreadTests: 4/4 passed
wasmedgeLazyJITTests: 22/22 passed
```

Also checked:

```text
clang-format: clean on the changed sources
git diff --check: clean
codespell: clean
```

## Notes

I have not yet confirmed the original Go/cgo reproducer end to end. The new tests cover the mechanism Go trips over, but Fixes #2400 should not be treated as fully demonstrated until a real WasmEdge-go path is run. Linux ucontext paths are in the code and need CI; local exercise was on macOS arm64.

### DIFF
--- a/include/system/fault.h
+++ b/include/system/fault.h
@@ -14,9 +14,14 @@
 //===----------------------------------------------------------------------===//
 #pragma once
 
+#include "common/defines.h"
 #include "common/errcode.h"
 #include <array>
 #include <csetjmp>
+
+#if !WASMEDGE_OS_WINDOWS
+#include <setjmp.h>
+#endif
 
 namespace WasmEdge {
 
@@ -38,9 +43,13 @@ private:
   Fault *Prev = nullptr;
   std::jmp_buf Buffer;
   std::array<void *, 256> StackTraceBuffer;
-  size_t StackTraceSize;
+  size_t StackTraceSize = 0;
 };
 
 } // namespace WasmEdge
 
+#if WASMEDGE_OS_WINDOWS
 #define PREPARE_FAULT(f) (static_cast<uint32_t>(setjmp((f).buffer())))
+#else
+#define PREPARE_FAULT(f) (static_cast<uint32_t>(_setjmp((f).buffer())))
+#endif
--- a/include/system/winapi.h
+++ b/include/system/winapi.h
@@ -1369,6 +1369,7 @@ static inline constexpr const DWORD_ EXCEPTION_INT_DIVIDE_BY_ZERO_ =
 static inline constexpr const DWORD_ EXCEPTION_INT_OVERFLOW_ = 0xC0000095L;
 static inline constexpr const LONG_ EXCEPTION_CONTINUE_EXECUTION_ =
     static_cast<LONG_>(0xffffffff);
+static inline constexpr const LONG_ EXCEPTION_CONTINUE_SEARCH_ = 0L;
 
 using EXCEPTION_RECORD_ = struct _EXCEPTION_RECORD {
   DWORD_ ExceptionCode;
--- a/lib/system/fault.cpp
+++ b/lib/system/fault.cpp
@@ -18,6 +18,7 @@
 
 #if defined(SA_SIGINFO)
 #include <pthread.h>
+#include <sys/ucontext.h>
 #endif
 
 #if WASMEDGE_OS_WINDOWS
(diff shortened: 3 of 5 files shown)


ibmibmibm (changes requested):
`SA_ONSTACK` is the right flag, but it isn't safe to merge alone: it changes *where* the handler runs without changing *how* it leaves.

## 1. `SA_ONSTACK` needs a mask-restoring escape (blocking)

We never call `sigaltstack()`, so this flag is inert standalone. It only goes live in the case this PR targets: an embedder (Go) provides an alternate stack, and our handler moves from the 8 MB thread stack to Go's 32 KB `gsignal` stack — while still leaving via plain `longjmp()` (`lib/system/fault.cpp:121`), never `sigreturn`.

Modelled against a 32 KB alt stack on Linux/glibc:

| variant | result |
|---|---|
| `longjmp`, no unblock hack | 2nd fault fatal — glibc `longjmp` doesn't restore the mask |
| `longjmp` + unblock-at-entry (this PR) | survives repeat faults, but a fault *inside* the handler re-entered 9x, exhausted the alt stack, clobbered 153 bytes below it |
| `sigsetjmp(buf,1)`/`siglongjmp` | signal stays blocked through the body; nested fault dies at the real site |

- The unblock at `fault.cpp:31-36` happens at handler *entry*, so the whole body — including `stackTrace()` unwinding AOT/JIT frames — runs re-entrant.
- That was survivable with an 8 MB recursion budget. On the embedder's alt stack it's ~9 frames, and overflow runs off the bottom. **This PR is what makes it reachable.**

Two ways to fix:

- **A — `sigsetjmp`/`siglongjmp`.** Lets the `pthread_sigmask` block go away entirely. Needs `sigjmp_buf` in the public `include/system/fault.h` and a `PREPARE_FAULT` split, since Windows VEH has no `sigsetjmp`. Cost: `sigsetjmp(buf,1)` is a `rt_sigprocmask` syscall — I measured 94.4 ns vs 2.7 ns for `setjmp`, and `PREPARE_FAULT` is on every interpreter→AOT/JIT entry (`lib/executor/helper.cpp:175-176`).
- **B — restore `uc_sigmask` at the jump.** Keeps `setjmp`/`longjmp`, no fast-path cost. The handler already gets the `ucontext_t` and discards it; drop the entry unblock and instead, immediately before the `longjmp`:

  ```cpp
  pthread_sigmask(SIG_SETMASK, &static_cast<ucontext_t *>(Context)->uc_sigmask,
                  nullptr);
  ```

  Restores the full pre-fault mask rather than one signal, and keeps the body non-re-entrant.

I lean B for the fast path, but A is easier to argue correct — either is fine.

- Please also check on macOS, since that's where you tested: on BSD kernels `SS_ONSTACK` is sticky state cleared by `sigreturn(2)`, which we never call. If it survives our `longjmp`, every later signal on that thread lands off the alternate stack — the same symptom #2400 reports. I couldn't verify this from Linux; if it holds, neither option fixes it alone.

## 2. We destroy the embedder's handlers (need-fix)

- `enableHandler()` passes `nullptr` for `oldact`; `disableHandler()` resets to `SIG_DFL` (`fault.cpp:53-61`).
- Under Go/cgo: the first `Fault` overwrites Go's `SIGSEGV`/`SIGBUS`/`SIGFPE` handlers, and scope exit sets `SIG_DFL` instead of restoring them.
- After the first Wasm call, a nil deref anywhere in the Go program hard-crashes instead of panicking. Please save and restore `oldact`.

## 3. Faults on threads without a `Fault` are UB (need-fix)

- `handlerCount` is process-global but `localHandler` is `thread_local` (`fault.cpp:25-26`), so one live `Fault` installs handlers process-wide.
- A fault on any other thread still reaches `emitFault`'s `assuming(localHandler != nullptr)`, which is `__builtin_unreachable()` under `NDEBUG` (`include/common/errcode.h:30-32`).
- An unrelated genuine segfault in the host process therefore becomes UB rather than a crash or Go panic.
- With `oldact` saved from item 2, the fix falls out: chain to the previous handler when `localHandler == nullptr`. Same for `assuming(Siginfo->si_code == FPE_INTDIV)`.

Items 2 and 3 look closer to the actual #2400 bug than the flag is — note the reported failure is **signal 16**, `SIGURG` on Darwin, which we never register.

## 4. The test can't fail for the right reason

- It reads back the `sa_flags` it just wrote, so it restates the assignment and passes against every bug in items 1-3.
- It never installs a `sigaltstack`, so it never exercises the path this PR is about.
- Worth having instead: install an alt stack, trap through the AOT path, assert we recover, recover *again* on a second trap, and leave the mask and `sigaltstack` state unchanged.
- Minor: `WasmEdge::Fault` without `PREPARE_FAULT` leaves `Buffer` uninitialised, and its destructor resets all three signals to `SIG_DFL` process-wide, leaking into later tests in the same binary.

Since `Fixes #2400` is claimed, can you confirm the original reproducer (`go test` with cgo on macOS arm64) passes now? The unit test doesn't show that.

## Nits

- Title says "use alternate signal stack", but we never call `sigaltstack()` — this opts our handlers into one *if the embedder provides it*. Maybe `fix(system): set SA_ONSTACK on fault handlers for cgo compatibility`, and likewise for the test name.
- `struct sigaction Action{}` zero-initialises `sa_mask` instead of `sigemptyset()`. Fine on glibc/Darwin, but POSIX wants the explicit call — cheap to fix while you're in there.
- Pre-existing but amplified: `stackTrace()` in the handler isn't async-signal-safe (`backtrace` → `dl_iterate_phdr` takes a loader lock), and now runs on a much smaller stack.

iemAnshuman:
Thanks for the detailed review. Agreed on all of it. The first version only added SA_ONSTACK and left the harder problems alone. That is fixed in the latest commits.

### 1. Escape path / sticky SS_ONSTACK (blocking)

Confirmed on macOS arm64 against the previous tip (ed6d9583). After the first JIT/AOT fault with an embedder provided alt stack, SS_ONSTACK stayed latched, sigaltstack reinstall failed with EPERM, and the next fault was delivered off the alternate stack. That matches the #2400 failure mode (SIGURG = 16 on Darwin, handler not on signal stack).

Neither plain option A nor B alone is enough on Darwin if we still leave the handler without sigreturn. The new path:

1. stores the ErrCode in TLS,
2. rewrites the resume PC in the ucontext to a small trampoline (and aligns SP),
3. returns normally so the kernel's sigreturn restores the pre signal mask and clears SS_ONSTACK,
4. runs the trampoline on the original stack and _longjmps from there.

PREPARE_FAULT is now _setjmp / setjmp (no mask save), so there is no rt_sigprocmask on the interpreter to AOT/JIT boundary. The mask is still correct because sigreturn restores it. On the fallback path (unsupported target, or re entry before the trampoline ran), we restore uc_sigmask first (your option B) then longjmp.

stackTrace() now runs in the trampoline, outside the signal handler and off the small alt stack.

### 2 & 3. Save/restore + chain

enableHandler() saves oldact per signal. disableHandler() restores it instead of SIG_DFL. The install/refcount path is under a mutex so a concurrent 1 to 0 to 1 transition cannot save our own handler as the previous one.

When localHandler is null, the signal is chained to the saved previous action (SIG_IGN / SIG_DFL / custom, including SA_SIGINFO / mask / SA_NODEFER). The assuming(si_code == FPE_INTDIV) is gone. Non INTDIV SIGFPE forwards. On Windows, the VEH returns EXCEPTION_CONTINUE_SEARCH_ when there is no Fault or the code is unhandled.

### 4. Tests

Replaced the old flag read back test with:

- FaultHandler.PreservesPreviousSignalActions: save/restore exact handler, flags, and mask; asserts SA_ONSTACK while a Fault is live.
- FaultHandler.ChainsSignalFromThreadWithoutFaultContext: death test that a fault on a thread without a Fault reaches the pre existing handler.
- LazyJITTest.RepeatedFaultsPreserveAlternateSignalState: real sigaltstack, two consecutive JIT traps, asserts recovery plus unchanged alt stack state and signal mask.

Every Fault in these tests pairs with PREPARE_FAULT, and a guard restores signal actions so later tests in the same binary are not polluted. sigemptyset() is explicit in enableHandler().

### Verification (local, macOS arm64)

wasmedgeLazyJITTests: 22/22
wasmedgeThreadTests: 4/4
clang format / git diff --check / codespell clean

### Still open / not claiming yet

I have not yet confirmed the original Go/cgo reproducer on macOS arm64. The unit and JIT tests cover the mechanism the Go runtime trips over, but Fixes #2400 should not be treated as demonstrated until a real WasmEdge go + go test path is run. Happy to follow up with that.

Linux aarch64/x86_64 ucontext rewrites are implemented but only exercised here on macOS arm64. Relying on CI for Linux.

Title and description updated to match the broader fix. It is no longer just "use an alternate signal stack".

Happy to reshape any of the commits or take a different approach on the sigreturn path if you would rather.

PR #5225, opened 2026-07-28
fix(common): avoid span underflow in rapidHash tail read

### LINKED ISSUE #5212: bug: rapidHash reads out of bounds via span::last<N>() when the remaining span is shorter than N
### Summary

`Hash::rapidHash` calls `Data.last<16>()` and `Data.last<8>()` on a span that can be shorter than 16 (resp. 8) bytes. `span::last<Count>()` computes `data() + (size() - Count)` with unsigned arithmetic, so `size() - Count` wraps and the pointer arithmetic is undefined behaviour.

It is reachable from ordinary module validation  - `Validator::validate(const AST::ExportSection &)` hashes every export name of every module WasmEdge loads, so any Wasm file with a 49-byte export name triggers it. UBSan flags it in four separate test binaries and via the `wasmedge` CLI.

### Current State

`lib/common/hash.cpp:59-84` — the `size() > 16` branch advances `Data` by 48 bytes per loop iteration, then unconditionally takes the last 16 and last 8 bytes of what remains:

```cpp
} else {
  if (Data.size() > 48) {
    uint64_t See1 = Seed, See2 = Seed;
    do {
      ...
      Data = Data.subspan<48>();
    } while (Data.size() > 48);
    Seed ^= See1;
    Seed ^= See2;
  }
  if (Data.size() > 16) {
    ...
  }
  A = read(Data.last<16>().first<8>()) ^ Data.size();   // line 82
  B = read(Data.last<8>());                             // line 83
}
```

After the loop `Data.size()` can be as low as 1, so **both line 82 and line 83** are affected.

`include/experimental/span.hpp:205-208` has no runtime bounds check:

```cpp
template <size_t Count> constexpr span<T, Count> last() const {
  static_assert(Count <= Extent);
  return span<T, Count>(data() + (size() - Count), Count);
}
```

The `static_assert` is vacuous here because `Extent == dynamic_extent`, so nothing checks the *runtime* size.

The trigger condition is `size() > 48 && 1 <= size() % 48 <= 15` — 120 distinct lengths in 17..399, e.g. any export name of 49–63 bytes. (Length 64 is fine: `64 % 48 == 16`.)

**Root cause.** Upstream rapidhash (`Nicoshev/rapidhash`, `rapidhashNano_internal`) ends its `> 48` branch with:

```c
p += 48;
i -= 48;
...
a = rapid_read64(p+i-16) ^ i;  b = rapid_read64(p+i-8);
```

`p` is the **advanced** pointer, `i` the **remaining** length. When `i < 16`, `p+i-16` is a *negative* offset from `p` — it deliberately steps backwards into bytes the loop already consumed, which is in bounds because `p` advanced by at least 48. `Span::last<16>()` cannot express that; it clamps to the current sub-span, so the equivalent expression underflows instead.

Since `p + i == len` at that point, upstream's window is exactly the last 16 bytes of the **original** buffer:

| len | upstream `orig[p+i-16 : p+i]` | `Orig.last<16>()` | same |
|-----|------------------------------|-------------------|------|
| 49  | `[33, 49)`                   | `[33, 49)`        | yes  |
| 50  | `[34, 50)`                   | `[34, 50)`        | yes  |
| 63  | `[47, 63)`                   | `[47, 63)`        | yes  |
| 64  | `[48, 64)`                   | `[48, 64)`        | yes  |
| 97  | `[81, 97)`                   | `[81, 97)`        | yes  |
| 100 | `[84, 100)`                  | `[84, 100)`       | yes  |
| 145 | `[129, 145)`                 | `[129, 145)`      | yes  |

So the current code is both UB *and* reading a different window than the algorithm specifies, for the affected lengths.

### Expected State

No undefined behaviour, and the tail read should match the reference algorithm's window.

Keeping a reference to the original span achieves both:

```diff
 WASMEDGE_EXPORT uint64_t Hash::rapidHash(Span<const std::byte> Data) noexcept {
+  const Span<const std::byte> Orig = Data;
   uint64_t Seed = RandomSeed;
@@
-    A = read(Data.last<16>().first<8>()) ^ Data.size();
-    B = read(Data.last<8>());
+    A = read(Orig.last<16>().first<8>()) ^ Data.size();
+    B = read(Orig.last<8>());
```

`Data.size()` in the `^` is deliberately left unchanged — upstream mixes the *remaining* length `i` there, not the total length. In this branch `Orig.size() > 16` always holds, so both tail reads are in bounds by construction.

With this applied: `HashTest.*` passes with 0 UBSan reports, the CLI reproduction produces 0 reports, and the full spec suite passes 530/530 with 0 sanitizer reports. Reverting the fix brings back exactly 2 reports on both paths, so the check is not passing vacuously.

Separately, `span::last<Count>()` / `span::first<Count>()` may deserve a real runtime guard, since the `static_assert` cannot catch anything on a `dynamic_extent` span. Happy to open that as a follow-up if you'd prefer it handled independently.

### Reproduction steps

1. Build with UBSan:

```bash
cmake -S . -Bbuild-asan -GNinja -DCMAKE_BUILD_TYPE=Debug \
  -DWASMEDGE_BUILD_TESTS=ON -DWASMEDGE_USE_LLVM=OFF -DWASMEDGE_BUILD_PLUGINS=OFF \
  -DCMAKE_CXX_FLAGS="-fsanitize=address,undefined -fno-omit-frame-pointer -g1" \
  -DCMAKE_EXE_LINKER_FLAGS="-fsanitize=address,undefined"
cmake --build build-asan
```

2. **The existing unit test already triggers it** — `HashTest.SizeClasses` (`test/common/hashTest.cpp:57-58`) hashes lengths 49 and 100:

```bash
./build-asan/test/common/wasmedgeCommonTests --gtest_filter=HashTest.SizeClasses
```

3. **End-to-end through the CLI.** Generate a 79-byte module whose single export name is 49 bytes (`49 % 48 == 1`):

```python
def leb(n):
    out = b''
    while True:
        b = n & 0x7f; n >>= 7
        if n: out += bytes([b | 0x80])
        else: return out + bytes([b])

name = b'x' * 49
ty   = b'\x01' + b'\x60\x00\x00'
fn   = b'\x01' + b'\x00'
ex   = b'\x01' + leb(len(name)) + name + b'\x00' + b'\x00'
body = b'\x00\x0b'
code = b'\x01' + leb(len(body)) + body
mod  = (b'\x00asm\x01\x00\x00\x00'
        + b'\x01' + leb(len(ty))   + ty
        + b'\x03' + leb(len(fn))   + fn
        + b'\x07' + leb(len(ex))   + ex
        + b'\x0a' + leb(len(code)) + code)
open('longname.wasm', 'wb').write(mod)
```

```bash
./build-asan/tools/wasmedge/wasmedge --enable-all longname.wasm
```

4. Both produce UBSan "addition of unsigned offset ... overflowed" errors at `span.hpp:207` (see log section).

Note the unit test still reports `[ OK ]` — the UB is silent without a sanitizer.

$ ./build-asan/test/common/wasmedgeCommonTests --gtest_filter=HashTest.SizeClasses
include/experimental/span.hpp:207:34: runtime error: addition of unsigned offset to 0x606000000a10 overflowed to 0x606000000a01
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior include/experimental/span.hpp:207:34 in
include/experimental/span.hpp:207:34: runtime error: addition of unsigned offset to 0x606000000a10 overflowed to 0x606000000a09
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior include/experimental/span.hpp:207:34 in
[ OK ] HashTest.SizeClasses (1476 ms)

Two reports per triggering length: 0xa10 -> 0xa01 is -15 (line 82, last<16> on a
1-byte span) and 0xa10 -> 0xa09 is -7 (line 83, last<8>). SizeClasses hashes both
49 and 100, so the full binary emits 4 reports:

$ ./build-asan/test/common/wasmedgeCommonTests 2>&1 | grep -c "span.hpp:207"
4

$ ./build-asan/tools/wasmedge/wasmedge --enable-all longname.wasm
include/experimental/span.hpp:207:34: runtime error: addition of unsigned offset to 0x606000006590 overflowed to 0x606000006581
    #1 WasmEdge::Hash::Hash::rapidHash(cxx20::span<std::byte const, 18446744073709551615ul>) hash.cpp:82
    #7 WasmEdge::Validator::Validator::validate(WasmEdge::AST::ExportSection const&) validator.cpp:850
    #8 WasmEdge::Validator::Validator::validate(WasmEdge::AST::Module const&) validator.cpp:160
    #10 WasmEdge::VM::VM::validate() vm.h:166
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior include/experimental/span.hpp:207:34 in
include/experimental/span.hpp:207:34: runtime error: addition of unsigned offset to 0x606000006590 overflowed to 0x606000006589
    #1 WasmEdge::Hash::Hash::rapidHash(cxx20::span<std::byte const, 18446744073709551615ul>) hash.cpp:83

Reproduced independently in four test binaries. Note each must be run from its own
build directory, since the spec-test binaries resolve their test data relative to cwd:

  build-asan/test/common/wasmedgeCommonTests          4 reports
  build-asan/test/executor/wasmedgeExecutorCoreTests  2 reports
  build-asan/test/api/wasmedgeAPIVMCoreTests          2 reports
  build-asan/test/api/wasmedgeAPIStepsCoreTests       2 reports

--- With the proposed fix applied ---

$ ./build-asan/test/common/wasmedgeCommonTests --gtest_filter='HashTest.*'
[ PASSED ] 9 tests. (0 UBSan reports)

$ ./build-asan/tools/wasmedge/wasmedge --enable-all longname.wasm 2>&1 | grep -c "runtime error"
0

$ cd build-asan/test/executor && ./wasmedgeExecutorCoreTests
[==========] 530 tests from 6 test suites ran. (61568 ms total)
[ PASSED ] 530 tests. (0 sanitizer reports)

--- Negative control: fix reverted and rebuilt ---

$ ./build-asan/test/common/wasmedgeCommonTests 2>&1 | grep -c "span.hpp:207"
4
$ ./build-asan/tools/wasmedge/wasmedge --enable-all longname.wasm 2>&1 | grep -c "runtime error"
2

### Components

CLI Tools, Core

8cadcc547

macOS 15 (Darwin 24.6.0)

### Hardware Architecture

arm64

### Appendix

_No response_

prerak09:
Fixes #5212

## Description

`rapidHash` advances `Data` by 48 bytes per loop iteration in the `> 48` path, then took the tail read from the *remaining* span:

```cpp
A = read(Data.last<16>().first<8>()) ^ Data.size();
B = read(Data.last<8>());
```

When the remainder is shorter than 16 bytes, `span::last<16>()` computes `data() + (size() - 16)` with unsigned arithmetic, so the offset wraps. Affected lengths are those where `size % 48` falls in `[1, 15]` — e.g. any input of 49–63 bytes.

Reachable from ordinary module validation: `Validator::validate(const AST::ExportSection &)` hashes every export name of every module loaded, so a module with a 49-byte export name triggers it.

**Scope:** the wrapped pointer lands on the same bytes, so hash values are unchanged and there is no wrong-result bug. This removes the undefined behaviour only. `span::last<Count>()`'s `static_assert(Count <= Extent)` is vacuous for a `dynamic_extent` span, so nothing catches the runtime case today.

The fix keeps a reference to the original span and takes the tail from that. In this branch `Orig.size() > 16` always holds, so both reads are in bounds by construction. `^ Data.size()` is deliberately unchanged — upstream mixes the *remaining* length there, not the total.

If you'd prefer broader coverage, a runtime bounds guard on `span::last<Count>()` / `first<Count>()` would catch any other caller with the same pattern. Happy to open that separately.

This contribution was developed with assistance from Claude (Anthropic).

## Checklist

## Test Evidence

Built with UBSan:

```console
$ cmake -S . -Bbuild-asan -GNinja -DCMAKE_BUILD_TYPE=Debug \
    -DWASMEDGE_BUILD_TESTS=ON -DWASMEDGE_USE_LLVM=OFF -DWASMEDGE_BUILD_PLUGINS=OFF \
    -DCMAKE_CXX_FLAGS="-fsanitize=address,undefined -fno-omit-frame-pointer -g1" \
    -DCMAKE_EXE_LINKER_FLAGS="-fsanitize=address,undefined"
$ cmake --build build-asan
```

**With this change** — the new case and the whole common suite are clean:

```console
$ cd build-asan/test/common && ./wasmedgeCommonTests
[==========] 84 tests from 9 test suites ran.
[  PASSED  ] 84 tests.

$ ./wasmedgeCommonTests 2>&1 | grep -c "span.hpp:207"
0
```

**Negative control** — reverting only `lib/common/hash.cpp` and keeping the new test reproduces the UB, so the case is not passing vacuously:

```console
$ ./wasmedgeCommonTests --gtest_filter=HashTest.TailReadAfterLoopIsInBounds 2>&1 | grep -c "span.hpp:207"
4
```

Sample report from that run:

```console
include/experimental/span.hpp:207:34: runtime error: addition of unsigned offset to 0x606000000d70 overflowed to 0x606000000d61
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior include/experimental/span.hpp:207:34 in
```

`-15` from `last<16>()` and `-7` from `last<8>()`, for each of the two triggering lengths the case exercises. The test still reports `OK` because UBSan runs in non-halting mode — the defect is sanitizer-detectable rather than value-detectable, since the wrapped pointer reads the same bytes.

**Full spec suite** with the change, under ASan + UBSan:

```console
$ cd build-asan/test/executor && ./wasmedgeExecutorCoreTests
[==========] 530 tests from 6 test suites ran. (64762 ms total)
[  PASSED  ] 530 tests.
```
0 sanitizer reports.

**Release build** (no sanitizers), to confirm nothing regresses:

```console
$ cd build/test/common && ./wasmedgeCommonTests
[==========] 84 tests from 9 test suites ran.
[  PASSED  ] 84 tests.
```

Tested on macOS 15 (Darwin 24.6.0), arm64, against master @ `c30edc5ef`.

### DIFF
--- a/lib/common/hash.cpp
+++ b/lib/common/hash.cpp
@@ -57,6 +57,7 @@ WASMEDGE_EXPORT uint64_t Hash::rapidHash(Span<const std::byte> Data) noexcept {
       A = B = 0;
     }
   } else {
+    const Span<const std::byte> Orig = Data;
     if (Data.size() > 48) {
       uint64_t See1 = Seed, See2 = Seed;
       do {
@@ -79,8 +80,8 @@ WASMEDGE_EXPORT uint64_t Hash::rapidHash(Span<const std::byte> Data) noexcept {
                         read(Data.subspan<24>().first<8>()) ^ Seed);
       }
     }
-    A = read(Data.last<16>().first<8>()) ^ Data.size();
-    B = read(Data.last<8>());
+    A = read(Orig.last<16>().first<8>()) ^ Data.size();
+    B = read(Orig.last<8>());
   }
   A ^= Secret[1];
   B ^= Seed;
--- a/test/common/hashTest.cpp
+++ b/test/common/hashTest.cpp
@@ -107,4 +107,23 @@ TEST(HashTest, IntegralHash) {
   EXPECT_EQ(Seen.size(), 1000u);
 }
 
+// Tail read must stay in bounds when the >48-byte loop leaves a short
+// remainder; run under -fsanitize=undefined to detect the underflow.
+TEST(HashTest, TailReadAfterLoopIsInBounds) {
+  // Lengths where size % 48 falls in [1, 15], plus in-bounds controls.
+  const std::vector<size_t> Affected = {49, 50, 63, 97, 100, 145};
+  const std::vector<size_t> Controls = {17, 32, 48, 64, 96, 128};
+
+  std::unordered_set<uint64_t> Seen;
+  for (auto Size : Affected) {
+    std::string Input(Size, 'x');
+    Seen.insert(hashBytes(Input.data(), Input.size()));
+  }
+  for (auto Size : Controls) {
+    std::string Input(Size, 'x');
+    Seen.insert(hashBytes(Input.data(), Input.size()));
+  }
+  EXPECT_EQ(Seen.size(), Affected.size() + Controls.size());
+}
+
 } // namespace


lib/common/hash.cpp
ibmibmibm, line 59: Move definition of `Orig` into this block

ibmibmibm (changes requested):

lib/common/hash.cpp
prerak09: Done ! moved into the `else` block. Re-verified: UBSan clean, 84/84 common tests and 530/530 spec tests passing.

ibmibmibm (approved):

PR #5226, opened 2026-07-28
fix(executor): handle NaN in f64x2 trunc_sat interpreter path

anusha19murthy:
## Description
i found this while going through the SIMD trunc_sat instructions in `include/executor/engine/unary_numeric_vector.ipp`. The `runVectorTruncSatOp<TIn, TOut>` template has two paths depending on whether TIn and TOut are the same size:

The f32x4->i32x4 path (same size) has a NaN guard before the conversion result gets used:

```cpp
VTOut Y = __builtin_convertvector(X, VTOut);
Y = detail::vectorSelect(X == X, Y, VTOut{});  // zero out NaN lanes
```

The f64x2->i32x4 path (different sizes, goes through an intermediate 64 bit type) doesn't have this guard at all:

```cpp
VTIn X = {std::trunc(V[0]), std::trunc(V[1])};
VTOut2 Y = __builtin_convertvector(X, VTOut2);
Y = detail::vectorSelect(X <= FMin, IMin, Y);
Y = detail::vectorSelect(X >= FMax, IMax, Y);
```

Converting a NaN through `__builtin_convertvector` to an integer type is undefined behavior in C++, since NaN doesn't fit `X <= FMin` or `X >= FMax` (both are false for NaN), nothing catches it, whatever the UB conversion produces just flows through as the result. Per the WebAssembly spec, `trunc_sat` on NaN should always give 0, this isn't a spec-conformant guarantee right now on this path, it's just whatever the compiler's NaN-to-int conversion happens to produce.

I checked the MSVC interpreter variant (`unary_numeric_vector_msvc.ipp:149`) and it does `std::isnan(V[I])` explicitly in a loop covering both cases, so that one's fine. Checked the AOT compiler too (`vectorInstr.cpp:1585`, uses `createFCmpORD(V, V)`), also fine. So this is specifically a gap in the GCC/Clang interpreter path for `i32x4.trunc_sat_f64x2_s_zero`, `i32x4.trunc_sat_f64x2_u_zero`, and their relaxed variants.

Fix is one line, add the same guard the f32x4 path already has:

```cpp
VTOut2 Y = __builtin_convertvector(X, VTOut2);
Y = detail::vectorSelect(X == X, Y, VTOut2{});
Y = detail::vectorSelect(X <= FMin, IMin, Y);
Y = detail::vectorSelect(X >= FMax, IMax, Y);
```

One caveat I want to be upfront about since I actually tested this both ways: on this x86_64 machine, the bug doesn't currently change the observable output. I wrote a small standalone probe converting a NaN vector through `__builtin_convertvector` directly and it already comes out as 0 on this GCC/x86_64 combo, so my regression test passes whether the fix is present or not here, I confirmed this by literally removing the fix and rerunning, same result. I also tried building with `-fsanitize=undefined,float-cast-overflow` to catch the UB directly and it didn't flag anything either, `__builtin_convertvector` seems to not get instrumented by UBSan the same way a normal cast would. So I don't have a locally reproducible failing case, I'm relying on the fact that this is genuinely undefined behavior per the language standard (relying on unspecified NaN-to-int conversion results is not something the code should do regardless of what today's compiler happens to produce) and that it brings this path in line with the other two implementations that already guard against it. Happy to have a maintainer double check this reasoning if it seems off.

Added a regression test anyway (`TruncSatF64x2NaNProducesZero`) since it documents the expected spec behavior even though it can't currently prove the difference between buggy/fixed on x86.

## Checklist

## Test Evidence
$ ninja wasmedgeExecutorRegressionTests
[18/18] Linking CXX executable test/executor/wasmedgeExecutorRegressionTests

$ ./test/executor/wasmedgeExecutorRegressionTests
[==========] Running 8 tests from 1 test suite.
[ RUN ] ExecutorRegression.RefTestNullAbility
[ OK ] ExecutorRegression.RefTestNullAbility (6 ms)
[ RUN ] ExecutorRegression.RefCastNullAbility
[ OK ] ExecutorRegression.RefCastNullAbility (0 ms)
[ RUN ] ExecutorRegression.BrOnCastNullAbility
[ OK ] ExecutorRegression.BrOnCastNullAbility (0 ms)
[ RUN ] ExecutorRegression.NullLocalConcreteType
[ OK ] ExecutorRegression.NullLocalConcreteType (0 ms)
[ RUN ] ExecutorRegression.ThrowRefPreservesPayload
[ OK ] ExecutorRegression.ThrowRefPreservesPayload (0 ms)
[ RUN ] ExecutorRegression.CatchAllDiscardsPayload
[ OK ] ExecutorRegression.CatchAllDiscardsPayload (0 ms)
[ RUN ] ExecutorRegression.TruncSatF64x2NaNProducesZero
[ OK ] ExecutorRegression.TruncSatF64x2NaNProducesZero (0 ms)
[ RUN ] ExecutorRegression.CallIndirectNonFuncTable
[ OK ] ExecutorRegression.CallIndirectNonFuncTable (0 ms)
[==========] 8 tests from 1 test suite ran. (7 ms total)
[ PASSED ] 8 tests.
---

### DIFF
--- a/include/executor/engine/unary_numeric_vector.ipp
+++ b/include/executor/engine/unary_numeric_vector.ipp
@@ -159,6 +159,7 @@ Expect<void> Executor::runVectorTruncSatOp(ValVariant &Val) const {
     const VTOut2 IMax = VTOut2{} + std::numeric_limits<TOut>::max();
     VTIn X = {std::trunc(V[0]), std::trunc(V[1])};
     VTOut2 Y = __builtin_convertvector(X, VTOut2);
+    Y = detail::vectorSelect(X == X, Y, VTOut2{});
     Y = detail::vectorSelect(X <= FMin, IMin, Y);
     Y = detail::vectorSelect(X >= FMax, IMax, Y);
     using VTOut22 [[gnu::vector_size(32)]] = TOut2;


hydai (changes requested):
Is this covered by a test? I guess not, because the original test passed without this PR.

anusha19murthy:
Couldn't get this to fail on x86_64. i tried it on both GCC and Clang Release builds, and both hit the same thing: cvttsd2si produces INT64_MIN for NaN, and its low 32 bits happen to be 0, which is the spec correct answer anyway. So the bug ends up masked by the hardware here no matter which compiler I use.

wwhat i do have is a minimal reproducer where i isolated just __builtin_convertvector(NaN, int64_t), and under -O3 Clang treats it as poison, i read the same variable twice and got two different bit patterns back. that's real UB actually showing up, not just something I'm citing from the standard. UBSan doesn't catch it either since it doesn't instrument this builtin the same way it does a normal cast.

So the test i added won't fail here, but it documents the behavior the spec actually requires and would catch a regression on a platform where NaN to int conversion doesn't zero the low bits, RISC-V being the likely one since it doesn't have this INT64_MIN convention. I'm fine dropping the test if you'd rather not keep one that can't fail locally.

q82419:
Thanks for digging into this — the asymmetry you found is real, and your cross-references check out. I verified both: the MSVC path does the explicit `std::isnan` check (`unary_numeric_vector_msvc.ipp:148`), and the AOT path does `createFCmpORD(V, V)` → `createSelect(Normal, V, IntZeroV)` (`lib/llvm/compiler/vectorInstr.cpp:1576-1581`). The GCC/Clang interpreter widening path is indeed the odd one out.

However, I'd like the fix to take a different shape, because the one-liner as written does not actually remove the UB.

## Why the guard-after-convert doesn't remove the UB

`detail::vectorSelect` is not an `LLVM select` — look at `include/executor/engine/vector_helper.h:15`:

```cpp
template <typename U, typename V>
[[gnu::always_inline]] constexpr inline std::enable_if_t<
    !std::is_arithmetic_v<V>, V>
vectorSelect(U Cond, V A, V B) noexcept {
#if defined(__clang__)
  return reinterpret_cast<V>((Cond & reinterpret_cast<U>(A)) |
                             (~Cond & reinterpret_cast<U>(B)));
#else
  return Cond ? A : B;
#endif
}
```

On Clang it lowers to a **bitwise blend**, not a `select`. `select` is one of the few instructions that block poison; `and`/`or` propagate it, and `Cond` here is a runtime comparison result so no `and X, 0 -> 0` fold applies. The poison lane produced by `fptosi(NaN)` therefore survives the blend. Adding `vectorSelect(X == X, Y, VTOut2{})` after the conversion documents the intent and matches the AOT backend, but the poison has already been created by then.

This is also why it isn't NaN-specific. The pre-existing saturation clamps rely on exactly the same unsound assumption — `fptoui` of a negative double in the `_u_zero` case is equally UB, and `vectorSelect(X <= FMin, IMin, Y)` "repairs" it the same way after the fact. So a fix that addresses only NaN leaves half the hole open.

## Suggested shape: sanitize the input before the conversion

Make the conversion unreachable for any value it cannot represent, so no poison is ever produced:

```diff
   if constexpr (sizeof(TIn) == sizeof(TOut)) {
     const VTOut IMin = VTOut{} + std::numeric_limits<TOut>::min();
     const VTOut IMax = VTOut{} + std::numeric_limits<TOut>::max();
     VTIn X = {std::trunc(V[0]), std::trunc(V[1]), std::trunc(V[2]),
               std::trunc(V[3])};
-    VTOut Y = __builtin_convertvector(X, VTOut);
-    Y = detail::vectorSelect(X == X, Y, VTOut{});
+    const VTIn Z = detail::vectorSelect((X > FMin) & (X < FMax), X, VTIn{});
+    VTOut Y = __builtin_convertvector(Z, VTOut);
     Y = detail::vectorSelect(X <= FMin, IMin, Y);
     Y = detail::vectorSelect(X >= FMax, IMax, Y);
     Val.emplace<VTOut>(Y);
   } else {
     using TOut2 = std::conditional_t<std::is_signed_v<TOut>, int64_t, uint64_t>;
     using VTOut2 [[gnu::vector_size(16)]] = TOut2;
     const VTOut2 IMin = VTOut2{} + std::numeric_limits<TOut>::min();
     const VTOut2 IMax = VTOut2{} + std::numeric_limits<TOut>::max();
     VTIn X = {std::trunc(V[0]), std::trunc(V[1])};
-    VTOut2 Y = __builtin_convertvector(X, VTOut2);
+    const VTIn Z = detail::vectorSelect((X > FMin) & (X < FMax), X, VTIn{});
+    VTOut2 Y = __builtin_convertvector(Z, VTOut2);
     Y = detail::vectorSelect(X <= FMin, IMin, Y);
     Y = detail::vectorSelect(X >= FMax, IMax, Y);
```

`(X > FMin) & (X < FMax)` is false for NaN, so the separate `X == X` guard is no longer needed — the NaN lane converts `0.0`, and neither saturation select overwrites it, which gives the spec-required `0`. Out-of-range lanes also convert `0.0` and are then overwritten by `IMin`/`IMax` exactly as today.

Note that clamping `X` to `FMax` instead of `0` would not work: in the f32 path `static_cast<float>(INT32_MAX)` rounds *up* to 2^31, so converting that value would be UB again. Substituting `0` sidesteps the boundary problem in all four instantiations.

`detail::vectorSelect` with a floating-point `V` is already established in the codebase (`include/executor/engine/simd_ops.h:102-107`), so this introduces no new pattern.

## Verification

I checked the proposed form against the current form on clang/arm64 at `-O2`, over all four instantiations (`f32x4_s`, `f32x4_u`, `f64x2_s_zero`, `f64x2_u_zero`) and these inputs: `+nan`, `-nan`, `±inf`, fractional values, exact `±2^31`, `2^32`, `INT32_MAX`/`UINT32_MAX`, and the largest representable float strictly below each bound. All 18 cases produce the spec-required result in both forms.

Codegen is identical between the two forms — 20 instructions for `<float, int32_t>` and 25 for `<double, int32_t>`. There is no cost to doing this the sound way.

One thing I did not test: the `&` of two comparison masks on the GCC `Cond ? A : B` path. It is standard GNU vector-extension semantics, but please confirm it on a GCC build before pushing.

## On testing

To answer my earlier question directly: this behaviour **is** already covered. `wasm-2.0/simd_i32x4_trunc_sat_f64x2.wast` has eight NaN assertions expecting `0` (lines 96-102 and 200-206), and the same cases exist in the `wasm-3.0-simd` suite. They pass today on x86 and ARM because the hardware conversion happens to yield the spec-correct low 32 bits.

So no new unit test is needed, and no test can be written that fails on any CI-supported target. Please drop the claim that a test was added and reference the existing spec coverage instead.

PR #5234, opened 2026-07-31
fix(common): prevent uint64_t overflow bypass in addCost (#5228)

### LINKED ISSUE #5228: bug: addCost integer overflow bypasses cost limit
### Summary

`addCost` in `include/common/statistics.h` (line 91) computes `NewCostSum = OldCostSum + Cost` using unchecked `uint64_t` addition. When the true sum exceeds `UINT64_MAX`, C++ unsigned integer wraparound produces a small value, which then passes the `NewCostSum > Limit` guard on line 92. The result is that the cost limit is silently bypassed and `CostSum` is reset to the wrapped, near-zero value.

### Current State

```cpp
// include/common/statistics.h:85-99
bool addCost(uint64_t Cost) {
    const auto Limit = CostLimit;
    uint64_t NewCostSum;
    do {
      NewCostSum = OldCostSum + Cost;   // line 91 — no overflow check
      if (unlikely(NewCostSum > Limit)) {
        spdlog::error("Cost exceeded limit. Force terminate the execution."sv);
        return false;
      }
                                            std::memory_order_relaxed));
    return true;
}
```

`uint64_t` addition wraps modulo 2⁶⁴ on overflow. The limit check on line 92 is evaluated **after** the wrap, so the wrapped value is compared instead of the true mathematical sum.

**Worked example with real numbers:**

- `CostLimit = UINT64_MAX` (18446744073709551615)
- `OldCostSum = UINT64_MAX - 5` (18446744073709551610)
- `Cost = 100`
- True sum = 18446744073709551710, which exceeds `UINT64_MAX`
- Actual `NewCostSum` after wrap = `(UINT64_MAX - 5 + 100) mod 2⁶⁴ = 94`
- Guard evaluates `94 > 18446744073709551615` → **false** → limit not exceeded
- `addCost` returns `true`, `CostSum` is stored as `94`

A Wasm module that has legitimately consumed nearly all of its gas budget can add further cost, trigger the overflow, and continue executing with `CostSum` reporting near-zero gas usage. Subsequent calls will similarly pass until the module exhausts the full budget again, effectively multiplying the effective gas budget by a large factor.

### Expected State

When `OldCostSum + Cost` would exceed `CostLimit`, `addCost` should return `false` and log the limit-exceeded error, regardless of whether the addition would overflow `uint64_t`. The overflow must be detected before performing the addition so that the comparison is made against the true mathematical sum.

### Reproduction steps

1. Build WasmEdge at commit `c30edc5e` or reproduce the logic in isolation using the standalone test below.
2. Create a `Statistics` instance and call `setCostLimit(UINT64_MAX)`.
3. Manually set (or drive via repeated `addCost` calls) `CostSum` to `UINT64_MAX - 5`.
4. Call `addCost(100)`.
5. Observe that the call returns `true` (accepted) and `getTotalCost()` returns `94`.

The expected result is that `addCost(100)` returns `false` because the true cost sum would exceed `CostLimit`.

**Commands run:**

```bash
# Compile
g++ -std=c++17 -O2 -o /tmp/test_addcost_overflow /tmp/test_addcost_overflow.cpp

# Run
/tmp/test_addcost_overflow
```

**Output:**
```
CostLimit  = 18446744073709551615
CostSum    = 18446744073709551610  (UINT64_MAX - 5)
Adding Cost = 100 ...
addCost(100) returned: true (ACCEPTED)
CostSum after        = 94

FAIL: overflow bypass confirmed. CostSum wrapped to 94, limit check evaded.
```

**FAIL** — `addCost(100)` with `CostSum = UINT64_MAX-5` and `CostLimit = UINT64_MAX` returns `true` and leaves `CostSum = 94`, exactly matching the described overflow bypass.

### Components

Core

c30edc5e

Ubuntu 24.04.3 LTS

### Hardware Architecture

x86_64 (amd64)

### Appendix

Affected code path: `Statistics::addCost` → line 91 (`NewCostSum = OldCostSum + Cost`) → line 92 (`NewCostSum > Limit`). The overflow occurs at line 91 before the guard on line 92 is reached.

https://github.com/WasmEdge/WasmEdge/blob/c30edc5ef2dd6980cc414499ad1f909de197ea3c/include/common/statistics.h#L91

DhruvMalviya0:
## fix(common): prevent uint64_t overflow bypass in `addCost`

Fixes #5228

### Problem

`addCost` in `include/common/statistics.h` computes `NewCostSum = OldCostSum + Cost` using unchecked `uint64_t` addition. When the true mathematical sum exceeds `UINT64_MAX`, C++ unsigned integer wraparound produces a small value that passes the `NewCostSum > Limit` guard. This silently bypasses the cost limit and resets `CostSum` to the wrapped near-zero value.

**Example:**
| Variable | Value |
|---|---|
| `CostLimit` | `18446744073709551615` (`UINT64_MAX`) |
| `OldCostSum` | `18446744073709551610` (`UINT64_MAX - 5`) |
| `Cost` | `100` |
| **Expected** `NewCostSum` | `18446744073709551710` (exceeds limit → reject) |
| **Actual** `NewCostSum` | `94` (wrapped → passes guard → accepted ❌) |

A Wasm module that has consumed nearly all of its gas budget can trigger the overflow and continue executing with near-zero reported gas usage, effectively multiplying the gas budget by a large factor.

### Fix

Replace the post-addition comparison with a pre-addition overflow-safe check:

```diff
     do {
-      NewCostSum = OldCostSum + Cost;
-      if (unlikely(NewCostSum > Limit)) {
+      if (unlikely(OldCostSum > Limit || Cost > Limit - OldCostSum)) {
         spdlog::error("Cost exceeded limit. Force terminate the execution."sv);
         return false;
       }
+      NewCostSum = OldCostSum + Cost;
```

**How it works:**
- `OldCostSum > Limit` — catches the case where the accumulated cost already exceeds the limit
- `Cost > Limit - OldCostSum` — mathematically equivalent to `OldCostSum + Cost > Limit`, but uses subtraction which cannot underflow (guarded by the first condition). This avoids the `uint64_t` wraparound entirely.

The addition on the next line is now guaranteed to be safe since we've verified `OldCostSum + Cost <= Limit`.

### Testing

The fix was verified against the reproduction scenario from the issue:
- `setCostLimit(UINT64_MAX)`, `CostSum = UINT64_MAX - 5`, `addCost(100)` → now correctly returns `false`

### DIFF
--- a/include/common/statistics.h
+++ b/include/common/statistics.h
@@ -88,11 +88,11 @@ public:
     uint64_t OldCostSum = CostSum.load(std::memory_order_relaxed);
     uint64_t NewCostSum;
     do {
-      NewCostSum = OldCostSum + Cost;
-      if (unlikely(NewCostSum > Limit)) {
+      if (unlikely(OldCostSum > Limit || Cost > Limit - OldCostSum)) {
         spdlog::error("Cost exceeded limit. Force terminate the execution."sv);
         return false;
       }
+      NewCostSum = OldCostSum + Cost;
     } while (!CostSum.compare_exchange_weak(OldCostSum, NewCostSum,
                                             std::memory_order_relaxed));
     return true;


deepak0x:
hi @ibmibmibm 
This pr has been raised for mine issue without asking me.. i was already working on it.. posted comment already https://github.com/WasmEdge/WasmEdge/issues/5228#issuecomment-5129096817 .. u can check mine pr #5236

DhruvMalviya0:
Hi @ibmibmibm — for context: this PR fixes #5228 with a verified overflow-safe check (tested against the exact repo in the issue). Happy to address any review feedback. Whichever approach you prefer is completely fine by me — just flagging that mine is ready and tested.

hydai:
I am closing this because there is already a working PR.

PR #5235, opened 2026-07-31
fix(loader): reject reserved br_on_cast/br_on_cast_fail flag bits using dedicated MalformedCastFlags error

### LINKED ISSUE #5232: bug: loader accepts reserved br_on_cast flag bits
### Summary

The loader accepts reserved flag values for the `br_on_cast` and
`br_on_cast_fail` instructions.

The WebAssembly binary grammar permits only four cast flag values:
`0x00`, `0x01`, `0x02`, and `0x03`. However, WasmEdge currently reads the
flag byte and only examines bits `0x01` and `0x02`, without rejecting other
bits.

Consequently, malformed values such as `0x04`, `0x05`, and `0xFF` can be
silently interpreted as valid flag combinations.

### Current State

In `Loader::loadInstruction`, the `Br_on_cast` and `Br_on_cast_fail`
decoding path reads a `uint8_t Flag` and uses only its two lowest bits:

- bit 0 determines the source nullability
- bit 1 determines the destination nullability

The loader does not verify that the remaining bits are zero. For example,
`0x04` is treated the same as `0x00`.

Relevant implementation:
https://github.com/WasmEdge/WasmEdge/blob/77a76b72f4239ee9be37808f32d2975b4e5e50d7/lib/loader/ast/instruction.cpp#L454-L472

### Expected State

The loader should reject cast flag values greater than `0x03`.

Legal values `0x00` through `0x03` should continue to decode normally.
Reserved values such as `0x04` and `0xFF` should produce a loader error.

### Reproduction steps

1. Check out commit `77a76b72f4239ee9be37808f32d2975b4e5e50d7`.
2. Add the following regression case to `test/loader/instructionTest.cpp`:

```cpp
TEST(InstructionTest, RejectReservedBrOnCastFlags) {
  const std::vector<uint8_t> Vec = {
      0x0AU, 0x0AU, 0x01U, 0x08U, 0x00U, 0xFBU,
      0x18U, 0x04U, 0x00U, 0x6DU, 0x6DU, 0x0BU,
  };
  EXPECT_FALSE(Ldr.parseModule(prefixedVec(Vec)));
}
```

3. Build the existing `wasmedgeLoaderASTTests` target.
4. Run the test.

The module contains a `br_on_cast` instruction with the reserved flag value
`0x04`. The loader accepts it even though the binary grammar does not define
that value.

Value of: Ldr.parseModule(prefixedVec(Vec))
  Actual: true
Expected: false

### Components

Core

77a76b72f4239ee9be37808f32d2975b4e5e50d7

Linux Mint 22.3

### Hardware Architecture

x86_64

### Appendix

The WebAssembly binary instruction grammar defines only four castop
encodings: `0x00`, `0x01`, `0x02`, and `0x03`.

Specification:
https://webassembly.github.io/spec/core/binary/instructions.html

I also searched the existing open issues and pull requests but did not find
one reporting the acceptance of reserved cast flag bits.

DhruvMalviya0:
## Summary

Fixes #5232.

The loader for `br_on_cast` / `br_on_cast_fail` reads a single flag
byte but only checks bits `0x01` and `0x02` (source/destination
nullability). It never validates that the remaining bits are zero, so
reserved values such as `0x04` or `0xFF` are silently accepted instead
of being rejected, even though the WebAssembly binary grammar only
defines four legal castop flag values: `0x00`–`0x03`.

This PR adds a check in `Loader::loadInstruction` that rejects any
flag bits outside `0x03` for both `br_on_cast` and `br_on_cast_fail`,
and adds regression tests covering both the accepted range and
reserved values.

## Changes

- `lib/loader/ast/instruction.cpp`: reject `Flag & ~0x03U` before
  decoding the heap types, for both `Br_on_cast` and `Br_on_cast_fail`.
- Added a new `ErrCode::Value::MalformedCastFlags` error code and its
  associated message mapping.
- `test/loader/instructionTest.cpp`:
  - Positive tests confirming flag values `0x00`–`0x03` still decode
    correctly for both opcodes.
  - Regression test for `br_on_cast` with reserved flag `0x04` (from
    the issue's repro case).
  - Equivalent regression test for `br_on_cast_fail` with a reserved
    flag.

## Error code — updated

**Previous approach:** this PR originally used the existing
`ErrCode::Value::IllegalGrammar` to report the rejection, as the
closest available "binary doesn't match spec grammar" code, since no
merged precedent existed at the time — a similar issue (#4631,
reserved bits in memarg) was closed as *not planned* without a fix
landing.

**Current approach:** per @uda18's pointer on #5232, PR #4676 already
established a precedent for this exact class of bug — it introduced
`MalformedCatchFlags` when `try_table` accepted catch flag values
outside `0x00`–`0x03`. This PR now follows that same convention and
introduces a matching `MalformedCastFlags` error code instead, so the
two reserved-flag-bits fixes (`try_table` catch flags and
`br_on_cast`/`br_on_cast_fail` cast flags) stay consistent.

## Testing

- `wasmedgeLoaderASTTests` passes locally, including the new tests.
- No other test regressions observed.

### DIFF
--- a/include/common/enum.inc
+++ b/include/common/enum.inc
@@ -945,6 +945,8 @@ E(MalformedMemoryOpFlags, 0x011C, "malformed memop flags")
 E(MalformedLimitFlags, 0x011D, "malformed limits flags")
 // Malformed catch flag (Exception-handling proposal).
 E(MalformedCatchFlags, 0x011E, "malformed catch flags")
+// Malformed cast flag (GC proposal).
+E(MalformedCastFlags, 0x011F, "malformed cast flags")
 // @}
 
 // Component model load phase
--- a/lib/loader/ast/instruction.cpp
+++ b/lib/loader/ast/instruction.cpp
@@ -456,6 +456,10 @@ Expect<void> Loader::loadInstruction(AST::Instruction &Instr) {
     // Read the flag.
     uint8_t Flag = 0U;
     EXPECTED_TRY(readU8(Flag).map_error(ReportError));
+    if (unlikely(Flag > 0x03U)) {
+      return logLoadError(ErrCode::Value::MalformedCastFlags,
+                          FMgr.getLastOffset(), ASTNodeAttr::Instruction);
+    }
     // Read the label index.
     uint32_t LabelIdx = 0U;
     EXPECTED_TRY(readU32(LabelIdx).map_error(ReportError));
--- a/test/loader/instructionTest.cpp
+++ b/test/loader/instructionTest.cpp
@@ -1307,4 +1307,111 @@ TEST(InstructionTest, LoadTryTable) {
   EXPECT_FALSE(Ldr.parseModule(Vec));
 }
 
+TEST(InstructionTest, LoadBrOnCastFlags) {
+  std::vector<uint8_t> Vec;
+
+  // 15. Test br_on_cast and br_on_cast_fail flag validation.
+  //
+  //   1.  Load br_on_cast with all valid flag values (0x00-0x03).
+  //   2.  Load br_on_cast_fail with all valid flag values (0x00-0x03).
+  //   3.  Reject br_on_cast with reserved flag 0x04 (issue #5232 regression).
+  //   4.  Reject br_on_cast with reserved flag 0xFF.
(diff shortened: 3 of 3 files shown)


q82419:
Already resolved

PR #5236, opened 2026-07-31
fix(common): prevent uint64_t overflow bypass in Statistics::addCost

### LINKED ISSUE #5228: bug: addCost integer overflow bypasses cost limit

### LINKED ISSUE #5242: bug: subCost() rejects reducing cost to zero due to off-by-one in condition
### Summary

`subCost()` in `include/common/statistics.h` uses `<=` on line 106 to guard against underflow, but this condition incorrectly rejects the case where the refund amount equals the current running total. When `CostSum == Cost`, the subtraction would produce exactly zero — a valid result — but the function returns `false` without modifying `CostSum`, silently losing the refund.

### Current State

```cpp
// include/common/statistics.h:102-113
bool subCost(uint64_t Cost) {
    uint64_t NewCostSum;
    do {
      if (unlikely(OldCostSum <= Cost)) {  // line 106 — off-by-one
        return false;
      }
      NewCostSum = OldCostSum - Cost;
                                            std::memory_order_relaxed));
    return true;
  }
```

The `<=` condition fires whenever `OldCostSum <= Cost`. When `OldCostSum == Cost`, `NewCostSum` would be `0`, which is a valid `uint64_t` value — not an underflow. The guard is only needed when `OldCostSum < Cost` (which would underflow to a large positive value).

**Worked example:**

| Step | Call | `OldCostSum` before | Condition `OldCostSum <= Cost` | Result | `CostSum` after |
|------|------|---------------------|-------------------------------|--------|-----------------|
| 1 | `addCost(10)` | 0 | — | `true` | 10 |
| 2 | `subCost(5)` | 10 | `10 <= 5` → false | `true` | 5 |
| 3 | `subCost(5)` | 5 | `5 <= 5` → **true** | **`false`** | **5** (unchanged) |

Step 3 should reduce `CostSum` to `0` but instead returns `false` and leaves `CostSum` at `5`.

### Expected State

`subCost(N)` should return `true` and set `CostSum` to `0` when `CostSum == N`. Reducing the running cost total to zero is a valid operation — it represents a full refund of accounted cost. The function should only reject calls where `Cost` strictly exceeds `CostSum`, as those would cause unsigned integer underflow.

### Reproduction steps

1. Check out commit `77a76b72` of WasmEdge.
2. Create a standalone reproducer that mirrors the `subCost` CAS loop from `include/common/statistics.h:102-113` using a `std::atomic<uint64_t>` initialized to `0`.
3. Call the equivalent of `addCost(10)` — verify `CostSum` becomes `10`.
4. Call `subCost(5)` — verify it returns `true` and `CostSum` becomes `5`.
5. Call `subCost(5)` again with `CostSum == 5` — observe that it returns `false` and `CostSum` remains `5` instead of becoming `0`.

---

**Commands run:**

```bash
g++ -std=c++17 /tmp/repro_subcost_standalone.cpp -o /tmp/repro_subcost_standalone
/tmp/repro_subcost_standalone
```

**Output:**

```
addCost(10), subCost(5): true, total=5 (expected true, 5)
subCost(5) when total=5: false, total=5 (expected true, 0)
FAIL
```

**FAIL** — `subCost(5)` when `TotalCost == 5` returns `false` and leaves total at `5` instead of reducing to `0`, exactly as described. Bug is the `<=` on `include/common/statistics.h:106`.

### Components

Core

`77a76b72`

Ubuntu 24.04.3 LTS

### Hardware Architecture

x86_64 (amd64)

### Appendix

Affected code path: `include/common/statistics.h`, `Statistics::subCost()`, line 106 — the `unlikely` guard condition inside the CAS retry loop.

https://github.com/WasmEdge/WasmEdge/blob/77a76b72f4239ee9be37808f32d2975b4e5e50d7/include/common/statistics.h#L106

deepak0x:
## Description

`Statistics::addCost` in `include/common/statistics.h` compared the post-addition result `OldCostSum + Cost` against the cost limit. When the true sum exceeds `UINT64_MAX`, unsigned wraparound yields a small value that passes the guard, silently bypassing the limit and resetting `CostSum` to a near-zero wrapped value (the exact scenario reported in #5228).

This replaces the unchecked addition with `__builtin_add_overflow`, which reports overflow as a distinct condition. Both the wraparound case and the genuine limit-exceeded case are now caught before any value is stored, with no arithmetic that can wrap.

The fix uses the canonical overflow-safe builtin instead of the subtraction trick `Cost > Limit - OldCostSum`, which relies on subtle underflow-avoidance reasoning. It also adds regression tests (the duplicate PR added none).

This PR also resolves #5242 (closed as a duplicate of #5228): `Statistics::subCost` used `OldCostSum <= Cost`, which incorrectly rejected the valid case `OldCostSum == Cost` (a full refund reducing the total to zero). The guard now uses `OldCostSum < Cost`. `subInstrCost` delegates to `subCost`, so it is corrected by the same change. Regression tests updated to reflect that a full refund reduces the total to zero. Combined into this PR per the maintainer's request to keep the Statistics cost fixes together.

Fixes: #5228
Related: #5242

## Checklist

- [x] **AI Disclosure**: I wrote this code entirely myself. No AI-assisted tools were used to produce, review, or generate this patch or its commits.

## Test Evidence

Built `wasmedgeCommonTests` and ran the statistics suite:

```text
[ RUN      ] StatisticsTest.AddCostRespectsLimit
[       OK ] StatisticsTest.AddCostRespectsLimit (0 ms)
[ RUN      ] StatisticsTest.AddCostNoOverflowBypass
[       OK ] StatisticsTest.AddCostNoOverflowBypass (0 ms)
[ RUN      ] StatisticsTest.SubCostRejectsCostAboveTotal
[       OK ] StatisticsTest.SubCostRejectsCostAboveTotal (0 ms)
[ RUN      ] StatisticsTest.SubCostReducesToZeroOnFullRefund
[       OK ] StatisticsTest.SubCostReducesToZeroOnFullRefund (0 ms)
[ RUN      ] StatisticsTest.CustomCostTableAndSubInstrCost
[       OK ] StatisticsTest.CustomCostTableAndSubInstrCost (0 ms)
...
[  PASSED  ] 11 tests.
```

Standalone reproduction confirming the addCost fix:

```text
OLD: addCost(100) @ 18446744073709551610 -> returned=1, CostSum=94
NEW: addCost(100) @ 18446744073709551610 -> returned=0, CostSum=18446744073709551610
NEW: addCost(100) @ (MAX-100) -> returned=1, CostSum=18446744073709551615 (expect 1, 18446744073709551615)
```

The OLD (vulnerable) logic wraps to 94 and returns accepted; the NEW logic returns rejected and leaves `CostSum` intact.

### DIFF
--- a/include/common/statistics.h
+++ b/include/common/statistics.h
@@ -88,8 +88,8 @@ public:
     uint64_t OldCostSum = CostSum.load(std::memory_order_relaxed);
     uint64_t NewCostSum;
     do {
-      NewCostSum = OldCostSum + Cost;
-      if (unlikely(NewCostSum > Limit)) {
+      if (unlikely(__builtin_add_overflow(OldCostSum, Cost, &NewCostSum) ||
+                   NewCostSum > Limit)) {
         spdlog::error("Cost exceeded limit. Force terminate the execution."sv);
         return false;
       }
--- a/include/common/statistics.h
+++ b/include/common/statistics.h
@@ -103,7 +103,7 @@ public:
     uint64_t OldCostSum = CostSum.load(std::memory_order_relaxed);
     uint64_t NewCostSum;
     do {
-      if (unlikely(OldCostSum <= Cost)) {
+      if (unlikely(OldCostSum < Cost)) {
         return false;
       }
       NewCostSum = OldCostSum - Cost;
--- a/test/common/statisticsTest.cpp
+++ b/test/common/statisticsTest.cpp
@@ -22,6 +22,34 @@ TEST(StatisticsTest, AddCostRespectsLimit) {
   EXPECT_EQ(S.getTotalCost(), UINT64_C(100));
 }
 
+TEST(StatisticsTest, AddCostNoOverflowBypass) {
+  Statistics S;
+  S.setCostLimit(UINT64_MAX);
+  S.getTotalCostRef().store(UINT64_MAX - 5, std::memory_order_relaxed);
+  EXPECT_FALSE(S.addCost(100));
+  EXPECT_EQ(S.getTotalCost(), UINT64_MAX - 5);
+}
+
+TEST(StatisticsTest, AddCostExactLimitNoOverflow) {
+  Statistics S;
+  S.setCostLimit(UINT64_MAX);
+  S.getTotalCostRef().store(UINT64_MAX - 100, std::memory_order_relaxed);
+  EXPECT_TRUE(S.addCost(100));
+  EXPECT_EQ(S.getTotalCost(), UINT64_MAX);
+  EXPECT_FALSE(S.addCost(1));
+  EXPECT_EQ(S.getTotalCost(), UINT64_MAX);
+}
+
+TEST(StatisticsTest, AddCostAtLimitBoundary) {
+  Statistics S(100);
+  EXPECT_TRUE(S.addCost(99));
+  EXPECT_FALSE(S.addCost(2));
+  EXPECT_EQ(S.getTotalCost(), UINT64_C(99));
+  EXPECT_TRUE(S.addCost(1));
+  EXPECT_EQ(S.getTotalCost(), UINT64_C(100));
+  EXPECT_FALSE(S.addCost(1));
+}
+
 TEST(StatisticsTest, SubCostRejectsCostAtOrAboveTotal) {
   Statistics S;
   EXPECT_TRUE(S.addCost(10));
--- a/test/common/statisticsTest.cpp
+++ b/test/common/statisticsTest.cpp
@@ -50,15 +50,23 @@ TEST(StatisticsTest, AddCostAtLimitBoundary) {
   EXPECT_FALSE(S.addCost(1));
 }
 
-TEST(StatisticsTest, SubCostRejectsCostAtOrAboveTotal) {
+TEST(StatisticsTest, SubCostRejectsCostAboveTotal) {
   Statistics S;
   EXPECT_TRUE(S.addCost(10));
   EXPECT_TRUE(S.subCost(5));
   EXPECT_EQ(S.getTotalCost(), UINT64_C(5));
-  EXPECT_FALSE(S.subCost(5));
-  EXPECT_EQ(S.getTotalCost(), UINT64_C(5));
-  EXPECT_FALSE(S.subCost(10));
-  EXPECT_EQ(S.getTotalCost(), UINT64_C(5));
+  EXPECT_TRUE(S.subCost(5));
+  EXPECT_EQ(S.getTotalCost(), UINT64_C(0));
+  EXPECT_FALSE(S.subCost(1));
+  EXPECT_EQ(S.getTotalCost(), UINT64_C(0));
+}
+
+TEST(StatisticsTest, SubCostReducesToZeroOnFullRefund) {
+  Statistics S;
+  EXPECT_TRUE(S.addCost(7));
+  EXPECT_EQ(S.getTotalCost(), UINT64_C(7));
+  EXPECT_TRUE(S.subCost(7));
+  EXPECT_EQ(S.getTotalCost(), UINT64_C(0));
 }
 
 TEST(StatisticsTest, SetCostLimit) {
@@ -94,8 +102,10 @@ TEST(StatisticsTest, CustomCostTableAndSubInstrCost) {
   EXPECT_EQ(S.getTotalCost(), UINT64_C(14));
   EXPECT_TRUE(S.subInstrCost(OpCode::Block));
   EXPECT_EQ(S.getTotalCost(), UINT64_C(7));
+  EXPECT_TRUE(S.subInstrCost(OpCode::Block));
+  EXPECT_EQ(S.getTotalCost(), UINT64_C(0));
   EXPECT_FALSE(S.subInstrCost(OpCode::Block));
-  EXPECT_EQ(S.getTotalCost(), UINT64_C(7));
+  EXPECT_EQ(S.getTotalCost(), UINT64_C(0));
 }
 
 TEST(StatisticsTest, ClearResetsCountersNotLimit) {


deepak0x:
cc @ibmibmibm — could you please review this fix? It prevents the uint64_t overflow bypass in `Statistics::addCost` reported in #5228. The fix uses `__builtin_add_overflow` and adds regression tests.

deepak0x:
@q82419 as you suggested on #5242 / #5243, I've folded the `subCost` fix (#5242) into this PR (#5228) so the Statistics cost fixes stay together in one PR.

Changes in this PR now cover both:
- `addCost` overflow bypass (original #5228) — `__builtin_add_overflow`.
- `subCost` off-by-one rejecting a full refund to zero (#5242) — guard changed from `<=` to `<`; `subInstrCost` is corrected since it delegates to `subCost`. Statistics tests updated to reflect that a full refund reduces the total to zero (all 11 `StatisticsTest` cases pass locally).

Note: this is currently a draft. The `commitlint` check (required) and the other workflows are awaiting maintainer approval to run, so the "Ready for review" button is disabled until those are approved and `commitlint` reports. Once a maintainer approves the workflows, I'll mark it ready.

deepak0x:
@hydai the `subCost` fix is now combined into this PR, as you asked on #5228. This PR now carries both Statistics cost findings in one place (matching the "single issue and PR for similar findings and components" preference):

- `addCost` overflow bypass (#5228) — `__builtin_add_overflow`.
- `subCost` off-by-one rejecting a full refund to zero (#5242, closed as a duplicate of #5228) — guard changed from `<=` to `<`; `subInstrCost` is corrected since it delegates to `subCost`. Statistics tests updated so a full refund reduces the total to zero (all 11 `StatisticsTest` cases pass locally).

The redundant `fix/5242-subcost-zero` branch (PR #5243, closed earlier) has been deleted, so this single PR covers both fixes.

Note: this is currently a draft. The `commitlint` check (required) and the other workflows are awaiting maintainer approval to run, so the "Ready for review" button is disabled until those are approved and `commitlint` reports. Once a maintainer approves the workflows, I'll mark it ready.

hydai:

PR #5243, opened 2026-08-02
fix(common): allow subCost to reduce total cost to zero

### LINKED ISSUE #5242: bug: subCost() rejects reducing cost to zero due to off-by-one in condition

deepak0x:
## Description

`Statistics::subCost()` in `include/common/statistics.h` guarded against unsigned integer underflow with `OldCostSum <= Cost`. This incorrectly rejected the valid case where `OldCostSum == Cost`: the subtraction would yield exactly `0` (a valid `uint64_t`), not an underflow. The guard only needs to fire when `OldCostSum < Cost`.

Changed the condition from `<=` to `<` so a full refund (total cost reduced to zero) is accepted. `subInstrCost()` delegates to `subCost()`, so it is corrected by the same change.

Before:
```
addCost(10); subCost(5) -> total=5; subCost(5) -> returns false, total stays 5
```
After:
```
addCost(10); subCost(5) -> total=5; subCost(5) -> returns true, total=0
```

## Checklist

- [x] **AI Disclosure**: I wrote this code entirely myself. No AI-assisted tools were used to generate or modify the changes in this PR.

## Test Evidence

Built and ran `test/common/wasmedgeCommonTests` with `--gtest_filter='*Statistics*'`:

```
[ RUN      ] StatisticsTest.AddCostRespectsLimit
[       OK ] StatisticsTest.AddCostRespectsLimit (0 ms)
[ RUN      ] StatisticsTest.SubCostRejectsCostExceedingTotal
[       OK ] StatisticsTest.SubCostRejectsCostExceedingTotal (0 ms)
[ RUN      ] StatisticsTest.SetCostLimit
[       OK ] StatisticsTest.SetCostLimit (0 ms)
[ RUN      ] StatisticsTest.InstructionCounter
[       OK ] StatisticsTest.InstructionCounter (0 ms)
[ RUN      ] StatisticsTest.DefaultCostTableChargesOnePerInstr
[       OK ] StatisticsTest.DefaultCostTableChargesOnePerInstr (0 ms)
[ RUN      ] StatisticsTest.CustomCostTableAndSubInstrCost
[       OK ] StatisticsTest.CustomCostTableAndSubInstrCost (0 ms)
[ RUN      ] StatisticsTest.ClearResetsCountersNotLimit
[       OK ] StatisticsTest.ClearResetsCountersNotLimit (0 ms)
[  PASSED  ] 7 tests.
```

Existing `SubCostRejectsCostAtOrAboveTotal` was renamed to `SubCostRejectsCostExceedingTotal` and `CustomCostTableAndSubInstrCost` was updated to expect the corrected behavior (reducing to zero succeeds).

Fixes: #5242

### DIFF
--- a/include/common/statistics.h
+++ b/include/common/statistics.h
@@ -103,7 +103,7 @@ public:
     uint64_t OldCostSum = CostSum.load(std::memory_order_relaxed);
     uint64_t NewCostSum;
     do {
-      if (unlikely(OldCostSum <= Cost)) {
+      if (unlikely(OldCostSum < Cost)) {
         return false;
       }
       NewCostSum = OldCostSum - Cost;
--- a/test/common/statisticsTest.cpp
+++ b/test/common/statisticsTest.cpp
@@ -22,15 +22,17 @@ TEST(StatisticsTest, AddCostRespectsLimit) {
   EXPECT_EQ(S.getTotalCost(), UINT64_C(100));
 }
 
-TEST(StatisticsTest, SubCostRejectsCostAtOrAboveTotal) {
+TEST(StatisticsTest, SubCostRejectsCostExceedingTotal) {
   Statistics S;
   EXPECT_TRUE(S.addCost(10));
   EXPECT_TRUE(S.subCost(5));
   EXPECT_EQ(S.getTotalCost(), UINT64_C(5));
-  EXPECT_FALSE(S.subCost(5));
-  EXPECT_EQ(S.getTotalCost(), UINT64_C(5));
+  EXPECT_TRUE(S.subCost(5));
+  EXPECT_EQ(S.getTotalCost(), UINT64_C(0));
   EXPECT_FALSE(S.subCost(10));
-  EXPECT_EQ(S.getTotalCost(), UINT64_C(5));
+  EXPECT_EQ(S.getTotalCost(), UINT64_C(0));
+  EXPECT_FALSE(S.subCost(1));
+  EXPECT_EQ(S.getTotalCost(), UINT64_C(0));
 }
 
 TEST(StatisticsTest, SetCostLimit) {
@@ -66,8 +68,8 @@ TEST(StatisticsTest, CustomCostTableAndSubInstrCost) {
   EXPECT_EQ(S.getTotalCost(), UINT64_C(14));
   EXPECT_TRUE(S.subInstrCost(OpCode::Block));
   EXPECT_EQ(S.getTotalCost(), UINT64_C(7));
-  EXPECT_FALSE(S.subInstrCost(OpCode::Block));
-  EXPECT_EQ(S.getTotalCost(), UINT64_C(7));
+  EXPECT_TRUE(S.subInstrCost(OpCode::Block));
+  EXPECT_EQ(S.getTotalCost(), UINT64_C(0));
 }
 
 TEST(StatisticsTest, ClearResetsCountersNotLimit) {


deepak0x:
Hi @hydai — this fixes #5242: `subCost()` used `OldCostSum <= Cost`, which incorrectly rejected reducing the total cost to zero (a valid `uint64_t` result, not an underflow). Changed the guard to `OldCostSum < Cost`. `subInstrCost()` is fixed transitively since it delegates to `subCost()`. Updated the statistics tests to cover the zero-reduction case; local `wasmedgeCommonTests` Statistics suite passes (7/7).

Could you take a look when you get a chance? Thanks!

q82419:
As mentioned in comment of #5242 
Please fix in one PR

PR #5259, opened 2026-08-06
chore(install): remove dead Ubuntu 20.04 DESCRIPTION fallback

### LINKED ISSUE #5187: bug: install.py Ubuntu 20.04 fallback reads /etc/lsb_release (underscore) which never exists
### Summary

`utils/install.py` detects Ubuntu 20.04 with a primary check plus an `or` fallback. The fallback reads `/etc/lsb_release` (underscore) instead of `/etc/lsb-release` (hyphen). That file does not exist on standard Ubuntu systems, and the `2>/dev/null` redirect discards the resulting "No such file" error, so `run_shell_command(...)` returns an empty string. The condition `"Ubuntu 20.04" in ""` is therefore always `False`, making the fallback dead code that can never rescue the primary RELEASE-line check when it fails.

### Current State

```python
# utils/install.py (Python 3 branch)
__lsb_rel = run_shell_command(
    "cat /etc/lsb-release 2>/dev/null | grep RELEASE"       # line 1251: /etc/lsb-release
)[-5:]
if VersionString(__lsb_rel).compare(
    "20.04"
) >= 0 or "Ubuntu 20.04" in run_shell_command(
    "cat /etc/lsb_release 2>/dev/null | grep DESCRIPTION"   # line 1256: /etc/lsb_release (underscore)
):
```

Line 1256 reads `/etc/lsb_release` (underscore), which does not exist; `2>/dev/null` swallows the error, so the command returns `""` and `"Ubuntu 20.04" in ""` is always `False`. Every other reference in the file uses the hyphenated path (e.g. lines 1226 and 1251).

Verified with the real `run_shell_command` imported from this file: on a simulated Ubuntu 20.04 host (`DISTRIB_DESCRIPTION="Ubuntu 20.04.6 LTS"`), the `/etc/lsb_release` command returns `''` and the fallback evaluates to `False`, so Ubuntu 20.04 is not detected via this branch.

### Expected State

The Ubuntu 20.04 detection fallback should be able to actually identify an Ubuntu 20.04 host, rather than always evaluating to an empty result and contributing nothing to detection.

### Reproduction steps

1. On any Ubuntu system, run `cat /etc/lsb_release` — it prints "No such file or directory". With `2>/dev/null` the command yields an empty string.
2. Open `utils/install.py` at commit `113fec47` and inspect the Python 3 distro-detection branch (lines ~1250-1257): line 1256 reads `/etc/lsb_release` (underscore) while line 1251 reads `/etc/lsb-release` (hyphen).
3. Because the fallback command's output is always empty, `"Ubuntu 20.04" in run_shell_command(...)` is permanently `False`; the fallback cannot contribute to Ubuntu 20.04 detection when the primary RELEASE-line check fails.

Found by source inspection and confirmed with test cases against the real `run_shell_command`.

No response

### Components

CLI Tools

113fec47

Ubuntu (affects the install script's Ubuntu 20.04 detection path; source-level typo, platform-independent)

### Hardware Architecture

Not architecture-specific

### Appendix

Affected code: `utils/install.py`, Python 3 distro-detection branch — line 1256 (`/etc/lsb_release`).

Permalink pinned to the reported commit:
https://github.com/WasmEdge/WasmEdge/blob/113fec4724c6c9ea53d10159f2cb4248be77606b/utils/install.py#L1250-L1257

deepak0x:
While reading `utils/install.py` I noticed the Python 3 branch of the Ubuntu 20.04 detection (#5187) had a dead `or` clause. The primary check reads `/etc/lsb-release` (hyphen) and parses `DISTRIB_RELEASE`; the second `or` operand read `/etc/lsb_release` (underscore), which never exists, so it always evaluated `False` and was unreachable.

Per @hydai's suggestion, rather than fix the typo, this removes the dead operand and keeps the working `RELEASE`-based check. Detection of Ubuntu 20.04 is unchanged.

## Test plan

- `python3 -m py_compile utils/install.py` passes.
- No behavior change: on Ubuntu 20.04 the `RELEASE` check still resolves `self.dist = "ubuntu20.04"`; on non-20.04 the `else` branch is unchanged.

Fixes #5187

### DIFF
--- a/utils/install.py
+++ b/utils/install.py
@@ -1250,11 +1250,7 @@ class Compat:
                     __lsb_rel = run_shell_command(
                         "cat /etc/lsb-release 2>/dev/null | grep RELEASE"
                     )[-5:]
-                    if VersionString(__lsb_rel).compare(
-                        "20.04"
-                    ) >= 0 or "Ubuntu 20.04" in run_shell_command(
-                        "cat /etc/lsb_release 2>/dev/null | grep DESCRIPTION"
-                    ):
+                    if VersionString(__lsb_rel).compare("20.04") >= 0:
                         # ARM-based Ubuntu 20.04 is supported after 0.13.5
                         if self.arch == "x86_64" or (
                             self.arch == "aarch64"


deepak0x:
@hydai this is the cleanup PR for #5187 we discussed — removes the dead `/etc/lsb_release` (underscore) `or` operand and keeps the working `RELEASE`-based check. Description updated with a test plan. PTAL when you have a moment.

Note: this is currently a draft. The `commitlint` check (required) and 4 workflows are awaiting maintainer approval to run, so the "Ready for review" button is disabled until those are approved and `commitlint` reports. Once a maintainer approves the workflows, I'll mark it ready.

hydai:

PR #3860, opened 2024-11-01
[Executor] coredump: add coredump

grorge123:
Fixes #2264 
Fixes #3205 

Support the core dump when WebAssembly enters a trap.

This core dump feature follows the [conventions](https://github.com/WebAssembly/tool-conventions/blob/main/Coredump.md). And add a parameter to generate the core dump file for the [Wasmgdb](https://github.com/xtuc/wasm-coredump).
The current implementations support most sections and use the Wasmgdb test. Limited by the architecture, module information only shows the current module.

PR #5260, opened 2026-08-06
feat(coredump): reconstruct data instances in generateCoredump

Ankitdotraider:
## Description
This re-enables `createData()`, which was fully implemented but commented
out since the original coredump feature landed in #3860. `generateCoredump()`
now calls it via `CurrentInstance->getOwnedDataInstances()`, same pattern
already used for `createGlobals()` right below it.

Checked git history first, this function hasn't been touched since #3860,
and there's no open issue or PR against it currently.

Opening as draft since the existing tests only check that coredump
generation doesn't crash, not that the data section content is correct.
Planning to add a regression test that generates a coredump from a module
with a real `(data ...)` segment and checks the bytes round-trip, will
flip to ready to review, once im done

## Checklist

- [x] **DCO Signed-off**
- [x] **Commit Messages**
- [x] **Coding Style**
- [x] **Local Tests Passed**
- [x] AI Disclosure
- [x] Human-in-the-loop

## Test Evidence

macOS arm64, WASMEDGE_USE_LLVM=OFF:

Screenshot 2026-08-06 at 11 15 56 PM
Screenshot 2026-08-06 at 11 16 49 PM

Both `wasmedgeExecutorCoreTests` and `wasmedgeDriverToolTests` passed,
including `RunSubcommand.RunSpecificFlags` covering the `--enable-coredump`
CLI path.

### DIFF
--- a/lib/executor/coredump.cpp
+++ b/lib/executor/coredump.cpp
@@ -35,9 +35,8 @@ void generateCoredump(const Runtime::StackManager &StackMgr,
   Module.getCustomSections().emplace_back(createCore());
   Module.getCustomSections().emplace_back(createCorestack(
       Ser, StackMgr.getFramesSpan(), StackMgr.getValueSpan(), ForWasmgdb));
-  // TODO: reconstruct data instances
-  // Module.getDataSection() =
-  //     createData(CurrentInstance->getOwnedDataInstances());
+  Module.getDataSection() =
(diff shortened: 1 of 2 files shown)


Ankitdotraider:
@q82419 can you look into it?

q82419:
We would have a large scale of refactoring in PR #5284.
Your changes will be replaced. therefore close this.

PR #5265, opened 2026-08-08
fix: serialize memory offsets according to Memory64 proposal

ravindra-RKB:
## Description

Fix memory argument offset serialization to correctly respect the `Memory64` proposal.

Previously, memory offsets were always serialized as unsigned 64-bit values, even when the `Memory64` proposal was disabled. This could produce incorrect serialization for memory instructions using offsets outside the 32-bit range.

This change:

- Serializes memory offsets as `u64` when `Memory64` is enabled.
- Serializes memory offsets as `u32` when `Memory64` is disabled.
- Returns `ErrCode::Value::IntegerTooLarge` when `Memory64` is disabled and the offset exceeds `UINT32_MAX`.
- Adds regression coverage for both Memory32 and Memory64 configurations.

Related to #4992 — fixes the Memory64-disabled memory offset serialization inconsistency.

## Test Coverage

The regression test covers the following cases:

1. **Memory64 disabled + small offset**
   - Serialization succeeds.
   - Offset is encoded as a 32-bit value.

2. **Memory64 disabled + offset > `UINT32_MAX`**
   - Serialization fails with `ErrCode::Value::IntegerTooLarge`.

3. **Memory64 enabled + large offset**
   - Serialization succeeds.
   - Offset is encoded as a 64-bit value.

### Targeted Regression Test

text
SerializeInstructionTest.SerializeMemory64MemArgInstruction

[==========] 1 test from 1 test suite ran.
[  PASSED  ] 1 test.

Full Test Suite

The complete CTest suite was also executed after the final cleanup:

100% tests passed, 0 tests failed out of 33

Total Test time (real) = 317.92 sec
Full Test Suite Screenshot
image

The screenshot above shows all 33 tests completing successfully with zero failures.


ravindra-RKB:
Hi @maintainers,

Just a quick follow-up on #5265. Whenever you get a chance, I’d really appreciate a review.

ravindra-RKB:
Hi @maintainers,

Just following up on #5265. The PR is ready for review, with the Memory32/Memory64 regression coverage and full CTest suite passing successfully.

Whenever you have a chance, I’d really appreciate a review. I’m happy to address any feedback or make further changes needed to align the implementation with WasmEdge’s serialization conventions.

q82419:
Because the public API of generating AST structure are made by validated WASM now, this situation may not occur.
We would adjust the policy of proposal checking in serializer and refactor now, therefore close this PR.

Issue #4819, opened 2026-04-27
bug: `f64x2.add` and `mul` over non-canonical NaN inputs returns a different NaN payload (NOT PLAN TO FIX)


q82419:
### Summary

For modules that perform `f64x2.add` (or `f32x4.add`, `f64x2.mul`, `f32x4.mul`, etc.) on non-canonical NaN inputs, WasmEdge can return a different NaN bit-pattern than Wasmtime. After investigation, this is not a spec violation — both runtimes are producing results permitted by the WebAssembly NaN-propagation rules. We therefore do not plan to change this behavior.

Reproducer:

```wasm
  (module
    (type (;0;) (func))
    (type (;1;) (func (result v128)))
    (export "main" (func 0))
    (export "to_test" (func 0))
    (func (;0;) (type 1) (result v128)
      v128.const i32x4 0xffffffff 0xffffffff 0xffffffff 0xffffffff
      v128.const i32x4 0xffffffff 0xffffffff 0xffffffff 0xffff60ff
      f64x2.add))
```

When the result v128 is read as a single 128-bit unsigned integer:

| Runtime | Result (decimal) | High lane (f64 bits) |
| --- | --- | --- |
| Wasmtime | `340282366920938463463374607431768211455` | `0xffffffffffffffff` |
| WasmEdge | `340279142017811482847777199818813734911` | `0xffff60ffffffffff` |

Wasmtime returns the `lhs` lane's NaN payload.
WasmEdge returns the `rhs` lane's NaN payload.

### Why this is permitted by the spec

WebAssembly Core Specification [4.4.1.1 NaN Propagation](https://webassembly.github.io/spec/core/exec/numerics.html#nan-propagation) states:

> If the payload of all NaN inputs to the operator is canonical (including the case that there are no NaN inputs), then the payload of the output is canonical as well. Otherwise the payload is picked non-deterministically among all arithmetic NaNs; that is, its most significant bit is 1 and all others are unspecified.

Definitions used:

- Canonical NaN — for `f64`, payload `1 << 51` and all other mantissa bits zero (i.e. `0x7ff8000000000000` / `0xfff8000000000000`).
- Arithmetic NaN — any NaN whose most significant mantissa bit is `1` (i.e. any quiet NaN, with arbitrary sign and arbitrary other payload bits).

Applied to the reproducer:
   
- `lhs` lane `1 = 0xffffffffffffffff` — quiet NaN, but not canonical (extra mantissa bits set).
[...]

PR #5271, opened 2026-08-10
fix(executor): produce arithmetic NaN in rounding and SIMD min/max operation

devalgupta404:
…erators

## Description
This PR fixes NaN handling in the ceil, floor, trunc, min, and max operators to match the WebAssembly spec, and adds the nan:canonical / nan:arithmetic distinction to the spec test harness that exposed these cases.

As per the spec, non-canonical NaNs must produce an arithmetic NaN with the most significant payload bit set. The current scalar implementation can return a signaling NaN unchanged through libm, while the SIMD min/max paths can directly propagate the NaN operand without quieting it. The fix extracts the existing NaN-quieting logic into WasmEdge::quietNaN and applies it to the affected paths without changing the remaining payload bits.

There was a previous PR #4819 , which concerns choosing between valid arithmetic NaNs but here, the result can fall outside the permitted set entirely.

Spec: [https://webassembly.github.io/spec/core/exec/numerics.html#aux-nans](https://webassembly.github.io/spec/core/exec/numerics.html#aux-nans).
 

## Checklist

## Test Evidence
I ran the whole CI on my Fork and it was all green.
Screenshot from 2026-08-10 17-23-54

---

### DIFF
--- a/include/common/roundeven.h
+++ b/include/common/roundeven.h
@@ -94,4 +94,27 @@ inline double roundeven(double Value) noexcept {
 using detail::roundeven;
 inline float roundeven(float Value) { return detail::roundevenf(Value); }
 
+/// Sets the most significant bit of a NaN payload to 1 and leaves other values
+/// as it is. Operators propagating a NaN operand, and the libm ceil, floor,
+/// and trunc for 32-bit floats on some platforms, can produce a signaling NaN,
+/// while the specification requires an arithmetic NaN.
+template <typename T> inline T quietNaN(T Value) noexcept {
+  static_assert(std::is_floating_point_v<T>);
+  if (!std::isnan(Value)) {
+    return Value;
+  }
+  if constexpr (sizeof(T) == sizeof(uint32_t)) {
+    uint32_t I32;
+    std::memcpy(&I32, &Value, sizeof(T));
+    I32 |= static_cast<uint32_t>(0x01U) << 22;
+    std::memcpy(&Value, &I32, sizeof(T));
+  } else if constexpr (sizeof(T) == sizeof(uint64_t)) {
+    uint64_t I64;
+    std::memcpy(&I64, &Value, sizeof(T));
+    I64 |= static_cast<uint64_t>(0x01U) << 51;
+    std::memcpy(&Value, &I64, sizeof(T));
+  }
+  return Value;
+}
+
 } // namespace WasmEdge
--- a/test/spec/spectest.cpp
+++ b/test/spec/spectest.cpp
@@ -26,11 +26,13 @@
 #include <cmath>
 #include <cstddef>
 #include <cstdint>
+#include <cstring>
 #include <fstream>
 #include <iterator>
 #include <map>
 #include <memory>
 #include <thread>
+#include <type_traits>
 #include <unordered_map>
 #include <variant>
 
@@ -365,6 +367,32 @@ bool checkComponentSupported(std::string_view Folder, WasmEdge::WasmPhase P) {
   }
 }
 
+// Checks the given floating-point value against an expected NaN pattern.
+// A canonical NaN carries a payload whose most significant bit is 1 and whose
+// other bits are all 0, while an arithmetic NaN only requires that most
+// significant bit to be 1 and leaves the remaining bits unspecified.
+template <typename T>
+static bool matchExpectedNaN(std::string_view Expected, T Value) {
+  using UT =
+      std::conditional_t<sizeof(T) == sizeof(uint32_t), uint32_t, uint64_t>;
+  if (!std::isnan(Value)) {
+    return false;
+  }
+  UT Bits = 0;
+  std::memcpy(&Bits, &Value, sizeof(T));
+  constexpr UT SignifBits = (sizeof(T) == sizeof(uint32_t)) ? 23U : 52U;
+  constexpr UT CanonPayload = static_cast<UT>(UT{1} << (SignifBits - UT{1}));
+  const UT Payload = Bits & static_cast<UT>((UT{1} << SignifBits) - UT{1});
+  if (Expected == "nan:canonical"sv) {
+    return Payload == CanonPayload;
+  }
+  if (Expected == "nan:arithmetic"sv) {
+    return (Payload & CanonPayload) != UT{0};
+  }
+  // Unspecified NaN pattern. Any NaN matches.
+  return true;
+}
+
 } // namespace
 
 namespace WasmEdge {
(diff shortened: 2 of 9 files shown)


q82419:
Before reviewing this, please provide some reproducable WASM cases about arithmatic NaN, includes following:

1. The provided case(s) output the wrong f32/f64/v128 bit pattern in current WasmEdge code base
2. The provided case(s) output the expected f32/f64/v128 bit pattern after your fixing
3. The output f32/f64/v128 bit pattern from the provided case(s) of other runtimes, such as wasmtime

devalgupta404:
@q82419 

Here are the cases, the module takes the NaN as a bit pattern and returns the result's bit pattern, so the runtime prints the exact bits instead of nan.

```wat
(module
  (func (export "f32_ceil") (param i32) (result i32)
    local.get 0 f32.reinterpret_i32 f32.ceil i32.reinterpret_f32)

  (func (export "f32_floor") (param i32) (result i32)
    local.get 0 f32.reinterpret_i32 f32.floor i32.reinterpret_f32)

  (func (export "f32_trunc") (param i32) (result i32)
    local.get 0 f32.reinterpret_i32 f32.trunc i32.reinterpret_f32)

  (func (export "f32x4_ceil") (param i32) (result i32)
    local.get 0 f32.reinterpret_i32 f32x4.splat f32x4.ceil
    f32x4.extract_lane 0 i32.reinterpret_f32)

  (func (export "f32x4_min") (param i32 i32) (result i32)
    local.get 0 f32.reinterpret_i32 f32x4.splat
    local.get 1 f32.reinterpret_i32 f32x4.splat
    f32x4.min f32x4.extract_lane 0 i32.reinterpret_f32)

  (func (export "f64x2_min") (param i64 i64) (result i64)
    local.get 0 f64.reinterpret_i64 f64x2.splat
    local.get 1 f64.reinterpret_i64 f64x2.splat
    f64x2.min f64x2.extract_lane 0 i64.reinterpret_f64)
)
```

Inputs: 2141192192 = 0x7FA00000 is an f32 signaling NaN, 1065353216 is 1.0f, 9219994337134247936 = 0x7FF4000000000000 is an f64 signaling NaN, 4607182418800017408 is 1.0.

```
wasmedge --reactor nan_cases.wasm f32_ceil 2141192192
wasmedge --reactor nan_cases.wasm f32x4_min 2141192192 1065353216
wasmedge --reactor nan_cases.wasm f64x2_min 9219994337134247936 4607182418800017408
```

On master the significand MSB stays 0, so the result is still a signaling NaN. The spec says these operators return nans_N, and for a non canonical input that means payload n >= canon_N.
https://webassembly.github.io/spec/core/exec/numerics.html#aux-nans

For the SIMD min and max, wasmtime returns the canonical NaN and this PR keeps the input payload with the MSB set. Both are valid because the payload is unspecified as long as it is >= canon_N. wasmtime prints i32 and i64 signed, so -4194304 is 0xFFC00000 and -2251799813685248 is 0xFFF8000000000000. Both are negative canonical NaNs, which is allowed since the sign is unspecified.

Some controls that did not change: f32.ceil on a canonical NaN gives 0x7FC00000 on all three, f64.ceil was already correct on master because libm quiets for double but not for float, and scalar f32.min was already correct. The binary_numeric.ipp change is just moving that existing logic into a helper.

| Case | Master  | This PR | Wasmtime 47.0.3 |
|---|---|---|---|
| f32.ceil | 0x7FA00000 signaling | 0x7FE00000 arithmetic | 0x7FE00000 arithmetic |
| f32x4.min | 0x7FA00000 signaling | 0x7FE00000 arithmetic | 0xFFC00000 canonical |
| f64x2.min | 0x7FF4000000000000 signaling | 0x7FFC000000000000 arithmetic | 0xFFF8000000000000 canonical |

Screenshot below, master then this PR then wasmtime for each case.

Screenshot from 2026-08-15 17-03-38

test/spec/spectest.cpp
```diff
@@ -365,6 +367,32 @@ bool checkComponentSupported(std::string_view Folder, WasmEdge::WasmPhase P) {
   }
 }
 
+// Checks the given floating-point value against an expected NaN pattern.
+// A canonical NaN carries a payload whose most significant bit is 1 and whose
+// other bits are all 0, while an arithmetic NaN only requires that most
+// significant bit to be 1 and leaves the remaining bits unspecified.
```
q82419, line 373: Shorten your comments

test/spec/spectest.cpp
```diff
@@ -365,6 +367,32 @@ bool checkComponentSupported(std::string_view Folder, WasmEdge::WasmPhase P) {
...
+template <typename T>
+static bool matchExpectedNaN(std::string_view Expected, T Value) {
```
q82419, line 375: Not need to be `static`

include/common/roundeven.h
```diff
@@ -94,4 +94,27 @@ inline double roundeven(double Value) noexcept {
 using detail::roundeven;
 inline float roundeven(float Value) { return detail::roundevenf(Value); }
 
+/// Sets the most significant bit of a NaN payload to 1 and leaves other values
+/// as it is. Operators propagating a NaN operand, and the libm ceil, floor,
+/// and trunc for 32-bit floats on some platforms, can produce a signaling NaN,
+/// while the specification requires an arithmetic NaN.
```
q82419, line 100: Shorten your comments

q82419 (changes requested):
@devalgupta404 

I reproduced the SIMD `min`/`max` problem and confirmed the full PR turns the failing suites green. Two things to address.

  **1. `nearest` is missing.** `f32.nearest` / `f64.nearest` and `f32x4.nearest` / `f64x2.nearest` fall under the same `nans_N` rule as `ceil` / `floor` / `trunc`, but they are not touched. On a libm that does not quiet, `WasmEdge::roundeven` falls back to `std::nearbyint` and has exactly the same problem (musl's `rintf` is `if (e >= 0x7f+23) return x;`). Please quiet them too, in `runNearestOp` (`unary_numeric.ipp`), `vectorNearest` (`simd_ops.h`), and `runVectorNearestOp` (`unary_numeric_vector_msvc.ipp`). The suites already cover them (`f32`/`f64` `nearest`, `simd_f32x4_rounding` / `simd_f64x2_rounding` `nearest`) — please also add `f32.nearest` / `f32x4.nearest` to your reproducer module.

  **2. The spec-test harness must distinguish canonical from arithmetic NaN — keep this part, with one fix.** `compare()` currently accepts any NaN for both symbolic forms (`spectest.cpp:432` TODO, `:437` / `:442`, and `:603` / `:626` for v128 lanes), which leaves roughly 15k lane-level expectations unchecked, so comparing the payload is exactly the right change. `matchExpectedNaN` looks correct to me: canonical = payload exactly `2^(signif-1)`, arithmetic = only that MSB required, sign ignored as the spec requires. The one thing to change is the trailing fallback for an unrecognised `nan:*` string:

  ```cpp
  // Unspecified NaN pattern. Any NaN matches.
  return true;
  ```

  Please make that a hard failure instead — a silent accept is how the current TODO managed to hide a real bug for this long.

PR #5170, opened 2026-07-14
feat: component-model and WASI preview 2 implementation completion

q82419:
### Current implementation status

Base on the tracking issue #4236
On this PR, the implementation status for the component-model proposal:

> ✅ Full · 🔶 Partial · ❌ None · — N/A

| Partition | Load | Validate | Instantiate | Execute |
|---|---|---|---|---|
| **CoreModuleSection** | | | | |
| └ Module | ✅ | ✅ | ✅ | ✅ |
| **CoreInstanceSection** | | | | |
| └ Instantiate | ✅ | ✅ | ✅ | — |
| └ InlineExports | ✅ | ✅ | ✅ | — |
| **CoreTypeSection** | | | | |
| └ RecType | ✅ | ✅ | ✅ | — |
| └ ModuleType | ✅ | ✅ | ✅ | — |
| **ComponentSection** | | | | |
| └ Component (recursive) | ✅ | ✅ | ✅ | ✅ |
| **InstanceSection** | | | | |
| └ Instantiate | ✅ | ✅ | ✅ | — |
| └ InlineExports | ✅ | ✅ | ✅ | — |
| **AliasSection** | | | | |
| └ Export | ✅ | ✅ | ✅ | — |
| └ CoreExport | ✅ | ✅ | ✅ | — |
| └ Outer | ✅ | ✅ | ✅ | — |
| **TypeSection** | | | | |
| └ PrimValType | ✅ | ✅ | ✅ | ✅ |
| └ RecordTy | ✅ | ✅ | ✅ | ✅ |
| └ VariantTy | ✅ | ✅ | ✅ | ✅ |
| └ ListTy | ✅ | ✅ | ✅ | ✅ |
| └ TupleTy | ✅ | ✅ | ✅ | ✅ |
| └ FlagsTy | ✅ | ✅ | ✅ | ✅ |
| └ EnumTy | ✅ | ✅ | ✅ | ✅ |
| └ OptionTy | ✅ | ✅ | ✅ | ✅ |
| └ ResultTy | ✅ | ✅ | ✅ | ✅ |
| └ OwnTy | ✅ | ✅ | ✅ | ✅ |
| └ BorrowTy | ✅ | ✅ | ✅ | ✅ |
| └ MapTy | ✅ | ✅ | ✅ | ✅ |
| └ StreamTy | ✅ | ✅ | ✅ | ✅ |
| └ FutureTy | ✅ | ✅ | ✅ | ✅ |
[...]

PR #5276, opened 2026-08-10
refactor(component): re-design the component model validator

### LINKED ISSUE #4441: feat: Component Model: validator implementation tracking issue

### LINKED ISSUE #4714: bug: Fail to parse inline export
### Summary

Wasmedge fail to parse above binary code made from ```wasm-tools```.  but the binary format from ```wasm-tools``` do not match the spec.

```wat
(component
  (import "f" (func))
  (instance $i (export "hello" (func 0)))
)
```

it needs the response from upstream : 
https://github.com/WebAssembly/component-model/issues/622

(a fix to match  ```wasm-tools```)

```diff
diff --git a/lib/loader/ast/component/component_instance.cpp b/lib/loader/ast/component/component_instance.cpp
index 4b45c87a..3617b8ae 100644
--- a/lib/loader/ast/component/component_instance.cpp
+++ b/lib/loader/ast/component/component_instance.cpp
@@ -96,8 +96,8 @@ Expect<void> Loader::loadInstance(AST::Component::Instance &Instance) {
 
   auto LoadInlineExp =
       [this](AST::Component::InlineExport &Exp) -> Expect<void> {
-    // inlineexport ::= n:<exportname> si:<sortidx> => (export n si)
-    EXPECTED_TRY(Exp.getName(), FMgr.readName().map_error([this](auto E) {
+    // inlineexport ::= n:<exportname'> si:<sortidx> => (export n si)
+    EXPECTED_TRY(loadExternName(Exp.getName()).map_error([this](auto E) {
       return logLoadError(E, FMgr.getLastOffset(),
                           ASTNodeAttr::Comp_InlineExport);
     }));
```

### Current State

_No response_

### Expected State

_No response_

### Reproduction steps

1. Build with options '...'
2. Execute with flags '....'
3. Execute with inputs '....'
4. Get error

### Screenshots

DESCRIPTION

_No response_

### Components

Others

dev

Ubuntu

### Hardware Architecture

x86_64

_No response_

q82419:
## Description

close #4441
close #4714

## Checklist

## Test Evidence

---


q82419:
This PR will be at the base of #5170, for the validator side.
After merging this PR, we can merge the `dev/component` back into master, and then continue with the execution phase and wasi preview2 for component model.

cc @LFsWang

lib/validator/component_context.cpp
```diff
@@ -1,194 +1,1179 @@
...
+        spdlog::error("    post-return is specified more than once."sv);
+        return Unexpect(ErrCode::Value::CanonPostReturnDuplicated);
+      }
+      SeenPostReturn = true;
+      // The signature is checked by the caller, once the flat type is known.
+      break;
+    case ComponentCanonOptCode::Async:
+    case ComponentCanonOptCode::AlwaysTaskReturn:
+      break;
+    case ComponentCanonOptCode::Callback: {
```
hydai, line 377: The rewritten `checkOptions` lost the per-site and duplicate rules that the previous validator enforced for the async-related options:

- `callback` is accepted on `canon lower` — only the `post-return` case above consults `IsLift`. Explainer.md is explicit: *"The `(callback ...)` option may only be present in `canon lift` when the `async` option has also been set."* The lower path calls `checkOptions(Canon, false)` (component_canon.cpp:226) and never rejects `callback` afterwards, so `canon lower $f (callback $cb)` now validates.
- `always-task-return` is likewise accepted on `canon lower`, and the old `always-task-return requires async` check on lift is gone.
- Duplicate `async`/`callback`/`always-task-return` options are no longer rejected (the old validator tracked `HasAsync`/`HasCallback`/`HasAlwaysTaskReturn` and rejected duplicates, like the encoding/memory/realloc cases here still do).

Since the async spec suites aren't enabled yet, CI doesn't catch any of this. Suggest restoring the site checks (using `IsLift`) and the duplicate tracking in this switch.

lib/validator/component_canon.cpp
```diff
@@ -0,0 +1,435 @@
...
+  case ComponentCanonOpCode::Context__get:
+  case ComponentCanonOpCode::Context__set: {
+    // The spec restricts the thread-local slot to be less than 2.
+    if (Canon.getIndex() >= Component::TypeSystem::MaxContextSlots) {
+      spdlog::error(ErrCode::Value::ComponentContextSlotOutOfBounds);
+      spdlog::error(
+          "    Context slot {} out of bounds, must be less than {}."sv,
+          Canon.getIndex(), Component::TypeSystem::MaxContextSlots);
+      return Unexpect(ErrCode::Value::ComponentContextSlotOutOfBounds);
+    }
```
hydai, line 371: This bound check is dead code: for `context.get`/`context.set` the loader stores the slot immediate via `C.setConstVal(Val)` and never calls `setIndex()` (lib/loader/ast/component/component_canonical.cpp, `Context__get`/`Context__set` case), so `Canon.getIndex()` here is always 0 and `MaxContextSlots` can never be exceeded. A binary encoding `canon context.get i32 7` validates successfully.

Either the loader should store the immediate with `setIndex()`, or this check should read `getConstVal()`.

include/common/enum.inc
```diff
@@ -1077,84 +1077,224 @@ E(InvalidOffset, 0x022A, "offset out of range")
 
 // Component model validation phase
 // @{
-// Missing Argument
-E(MissingArgument, 0x02A0, "missing argument")
 // Argument Type Mismatch with Import Type
-E(ArgTypeMismatch, 0x02A1, "type mismatch")
+E(ArgTypeMismatch, 0x0240, "type mismatch")
```
hydai: This renumbering changes values of the public C API enum (`enum.inc` expands into `enum WasmEdge_ErrCode` via `include/api/wasmedge/wasmedge_basic.h` → `enum_errcode.h`), and some already-released values are reused with different meanings:

- `WasmEdge_ErrCode_ArgTypeMismatch` was `0x02A1` in the released 0.15.x headers and becomes `0x0240` here.
- The old numeric values are recycled for unrelated codes: `0x02A0` (was `MissingArgument`) → `ComponentResourceRepI32`, `0x02A1` (was `ArgTypeMismatch`) → `ComponentDtorSignature`, etc.
- Removed symbols (`MissingArgument`, `ComponentDuplicateName`, `ComponentImportNameConflict`, ...) are a source-compat break for C API users referencing them.

The affected codes are only returned by the experimental component validator, so renumbering may well be acceptable — but reusing previously released numeric values for different meanings is risky for clients that persist or compare error codes. If the renumbering is intended, consider leaving the old `0x02A0`–`0x02C1` range unused instead.

include/validator/component_value_decode.h
```diff
@@ -0,0 +1,339 @@
...
+        if (Shift < 64 && (B & 0x40U) != 0) {
+          Result |= -(INT64_C(1) << Shift);
+        }
+        return Result;
+      }
+      if (Shift >= MaxBits + 7) {
+        return Unexpect(ErrCode::Value::ComponentMalformedValue);
+      }
+    }
+  }
```
hydai, line 106: `readSLEB` has no sign-extension check on the final byte, so out-of-range encodings for `s16`/`s32` are accepted and silently truncated instead of being rejected as malformed. E.g. for `s16`, the 3-byte sequence `0x80 0x80 0x40` decodes to -1048576, which the caller's `static_cast<int16_t>` truncates to 0 — the value section then validates with a wrong decoded value.

Compare `readULEB` above, which does reject excess bits in the final byte (`B >> (MaxBits - Shift)`). The SLEB equivalent: the bits of the final byte above `MaxBits` must all equal the sign bit.

include/validator/component_value_decode.h
```diff
@@ -0,0 +1,339 @@
...
+      if (Len > Data.size() - Off) {
+        return Unexpect(ErrCode::Value::ComponentMalformedValue);
+      }
+      std::string S;
+      S.reserve(static_cast<size_t>(Len));
+      for (uint64_t I = 0; I < Len; ++I) {
+        EXPECTED_TRY(auto B, readByte());
+        S.push_back(static_cast<char>(B));
+      }
+      return ComponentValVariant{std::move(S)};
```
hydai: Two UTF-8 gaps in the value decoder:

- `string` payloads are copied byte-for-byte with no UTF-8 validation, but the spec defines `val(string)` over valid UTF-8 (`core:name`); invalid bytes should make the value malformed.
- `readUtf8Scalar` below checks continuation bytes, surrogates, and the `0x110000` bound, but accepts overlong encodings — e.g. `0xC0 0x80` decodes to U+0000 instead of being rejected.

Note this header has 0% patch coverage: the `component-model-values` suite exists in the wasm-core-20260813 spectest tag but isn't enabled here, so none of the value-decode paths run in CI. Worth wiring that suite (or unit tests) up when this lands.

hydai (commented):
### Code review

Deep review of the validator redesign, comparing the new implementation against the old one (base `b2bd4ac`) and the component-model spec. Overall verdict: the refactor is faithful and well-layered — the new types/context/canon/name split preserves or strengthens almost all checks from the old single-file validator, all 30 wasm-tools folders now load+validate (matching the #4441 acceptance criteria), and flipping the legacy http regression binary to rejected is spec-correct (multi-value lowered result).

Found 5 issues worth discussing — see the inline threads:

1. `checkOptions` lost the site/duplicate rules for `callback`/`always-task-return`/`async` (`canon lower (callback ...)` now validates).
2. The `context.get`/`context.set` slot bound check is dead code (loader stores the slot via `setConstVal`, validator reads `getIndex()`).
3. `enum.inc` renumbering reuses already-released C API error-code values with different meanings.
4. `readSLEB` in the value decoder accepts out-of-range `s16`/`s32` encodings and truncates them.
5. Value-decoder strings skip UTF-8 validation and `readUtf8Scalar` accepts overlong encodings (plus the whole header has 0% patch coverage since the `component-model-values` suite isn't enabled yet).

Also verified and dismissed while reviewing: canonopt `0x08` is the legal `always-task-return` encoding; equality-based valtype matching is correct for current MVP subtyping; accepting 64-bit canonical memories is intentional (memory64); rejecting `nameattributes` tag `0x01` is pre-existing behavior (though Binary.md allows `0x01` as equivalent to `0x00` — could be fixed along with the new `0x02` support in this PR's loader change).

---

This review was performed with GitHub Copilot CLI, powered by Claude Fable 5 (Anthropic), with human verification.

include/ast/component/import.h
```diff
@@ -33,10 +33,13 @@ class Import {
   std::string_view getName() const noexcept { return Name; }
   ExternDesc &getDesc() noexcept { return Desc; }
   const ExternDesc &getDesc() const noexcept { return Desc; }
+  std::vector<std::string> &getImplements() noexcept { return Implements; }
+  Span<const std::string> getImplements() const noexcept { return Implements; }
```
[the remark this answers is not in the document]
q82419: fixed

lib/validator/component_context.cpp
q82419: fixed

lib/validator/component_canon.cpp
q82419: fixed

include/common/enum.inc
q82419: refactored

include/validator/component_value_decode.h
q82419: fixed

include/validator/component_value_decode.h
q82419: fixed

lib/validator/component_canon.cpp
```diff
@@ -0,0 +1,433 @@
...
+  case ComponentCanonOpCode::Backpressure__dec:
+    return PushCoreFunc({}, {});
+  case ComponentCanonOpCode::Thread__index:
+    return PushCoreFunc({}, {I32V});
+  case ComponentCanonOpCode::Task__return: {
+    // The declared results lower as the core parameters.
+    std::vector<ValType> Params;
+    for (const auto &R : Canon.getResultList()) {
+      if (!CompTypes.flattenValType({R.getValType(), &S, nullptr}, Params,
+                                    I32V)) {
```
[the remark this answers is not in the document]
q82419: Fixed. `task.return` now runs `checkOptions(Canon, false)` (so `post-return`/`callback` are rejected by the lowering site rule), additionally rejects `realloc` and `async`, flattens the declared results with `CompCtx.canonPtrType(Canon)` instead of a hardcoded `i32`, and requires the `memory` option when a result contains a string/list or when the flat params overflow into an indirect pointer. This matches `wasmparser`'s `task_return` + `ComponentFuncType::lower(Abi::Lower)`.

Two new error codes were added for the rejections: `CanonReallocOnBuiltin` (0x1293) and `CanonAsyncOnBuiltin` (0x1294).

lib/validator/component_canon.cpp
```diff
@@ -0,0 +1,433 @@
...
+    return PushCoreFunc({I32V}, {I32V});
+  case ComponentCanonOpCode::Subtask__drop:
+    return PushCoreFunc({I32V}, {});
+  case ComponentCanonOpCode::Stream__new:
+    EXPECTED_TRY(CheckTypeImmediate(true));
+    return PushCoreFunc({}, {I64V});
+  case ComponentCanonOpCode::Stream__read:
+  case ComponentCanonOpCode::Stream__write:
+    EXPECTED_TRY(CheckTypeImmediate(true));
+    return PushCoreFunc({I32V, I32V, I32V}, {I32V});
```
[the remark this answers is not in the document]
q82419: Partly fixed — the option validation was a real gap, the memory64 signature claim is not.

Fixed: `stream.read`/`write` and `future.read`/`write` (and, for the same reason, `error-context.new`/`debug-message`) now call `checkOptions(Canon, false)`, so an out-of-range `(memory ...)` index no longer passes, and `post-return`/`callback` are rejected. Per-site requirements follow `wasmparser`: `memory` is required iff the element type is present, and `.read` additionally requires `realloc` iff the element type allocates (contains a string/list). `error-context.new` requires `memory` + sync; `.debug-message` requires `memory` + `realloc` + sync.

Not changed: the core signatures stay all-`i32` under memory64. `wasmparser` hardcodes `FuncType::new([ValType::I32; 3], [ValType::I32])` for `stream.read`/`write` and `[ValType::I32; 2] -> [ValType::I32]` for `future.read`/`write`, with no dependence on the selected memory's address type — those operands are a waitable handle index, an element count, and a packed status/count result, not linear-memory addresses. Only the payload pointer would widen, and it is not part of this signature. Among the built-ins, the address type only affects `task.return` (handled in the sibling comment) and `waitable-set.wait`/`poll`, where the memidx is an immediate rather than an option and is already handled.

Regression tests for all of the above were added to `ComponentValidatorTest`.

lib/loader/ast/component/component_valtype.cpp
```diff
@@ -6,16 +6,37 @@
...
+  if (B == 0x02) {
+    EXPECTED_TRY(auto Cnt, FMgr.readU32());
+    for (uint32_t I = 0; I < Cnt; ++I) {
+      EXPECTED_TRY(auto Opt, FMgr.readByte());
+      if (Opt == 0x00) {
+        EXPECTED_TRY(auto Impl, FMgr.readName());
+        Implements.push_back(std::move(Impl));
+      } else if (Opt == 0x02) {
+        // The external-id value is ignored during validation and execution.
+        EXPECTED_TRY(FMgr.readName());
```
LFsWang: // Binary.md:430-431
// - vec(<attribute>) contains each kind of attribute at most once.

dannypsnl:
From https://github.com/WebAssembly/component-model/blob/main/design/mvp/Explainer.md

> Currently, validation rejects (stream T) and (future T) when T transitively contains a borrow.

But I do some testings

```cpp
  Configure Conf;
  Conf.addProposal(Proposal::Component);

  // (type (resource))                       ;; type 0
  // (type (borrow 0))                       ;; type 1
  // (type (future 1))                       ;; type 2
  // (type (func (result (type 2))))         ;; type 3, should be REJECTED
  AST::Component::Component Comp;
  Comp.getSections().emplace_back();
  Comp.getSections().back().emplace<AST::Component::TypeSection>();
  auto &TypeSec =
      std::get<AST::Component::TypeSection>(Comp.getSections().back());

  // Type 0: resource
  TypeSec.getContent().emplace_back();
  TypeSec.getContent().back().setResourceType(AST::Component::ResourceType());

  // Type 1: (borrow 0)
  TypeSec.getContent().emplace_back();
  AST::Component::DefValType BorrowDVT;
  BorrowDVT.setBorrow(AST::Component::BorrowTy{0});
  TypeSec.getContent().back().setDefValType(std::move(BorrowDVT));

  // Type 2: (future 1)
  TypeSec.getContent().emplace_back();
  AST::Component::DefValType FutureDVT;
  FutureDVT.setFuture(AST::Component::FutureTy{ComponentValType(1)});
  TypeSec.getContent().back().setDefValType(std::move(FutureDVT));

  // Type 3: (func (result (type 2)))
  TypeSec.getContent().emplace_back();
  AST::Component::FuncType FT;
  std::vector<AST::Component::LabelValType> Results;
  Results.emplace_back(ComponentValType(2));
  FT.setResultList(std::move(Results));
  TypeSec.getContent().back().setFuncType(std::move(FT));

  Validator::Validator V(Conf);
  auto Res = V.validate(Comp);
```

The type is accepted.

lib/loader/ast/component/component_valtype.cpp
q82419: Fixed

q82419:
> From https://github.com/WebAssembly/component-model/blob/main/design/mvp/Explainer.md
> 
> > Currently, validation rejects (stream T) and (future T) when T transitively contains a borrow.
> 
> But I do some testings
> 
> ```c++
>   Configure Conf;
>   Conf.addProposal(Proposal::Component);
> 
>   // (type (resource))                       ;; type 0
>   // (type (borrow 0))                       ;; type 1
>   // (type (future 1))                       ;; type 2
>   // (type (func (result (type 2))))         ;; type 3, should be REJECTED
>   AST::Component::Component Comp;
>   Comp.getSections().emplace_back();
>   Comp.getSections().back().emplace<AST::Component::TypeSection>();
>   auto &TypeSec =
>       std::get<AST::Component::TypeSection>(Comp.getSections().back());
> 
>   // Type 0: resource
>   TypeSec.getContent().emplace_back();
>   TypeSec.getContent().back().setResourceType(AST::Component::ResourceType());
> 
>   // Type 1: (borrow 0)
>   TypeSec.getContent().emplace_back();
>   AST::Component::DefValType BorrowDVT;
>   BorrowDVT.setBorrow(AST::Component::BorrowTy{0});
>   TypeSec.getContent().back().setDefValType(std::move(BorrowDVT));
> 
>   // Type 2: (future 1)
>   TypeSec.getContent().emplace_back();
>   AST::Component::DefValType FutureDVT;
>   FutureDVT.setFuture(AST::Component::FutureTy{ComponentValType(1)});
>   TypeSec.getContent().back().setDefValType(std::move(FutureDVT));
> 
>   // Type 3: (func (result (type 2)))
>   TypeSec.getContent().emplace_back();
>   AST::Component::FuncType FT;
>   std::vector<AST::Component::LabelValType> Results;
>   Results.emplace_back(ComponentValType(2));
>   FT.setResultList(std::move(Results));
>   TypeSec.getContent().back().setFuncType(std::move(FT));
> 
>   Validator::Validator V(Conf);
>   auto Res = V.validate(Comp);
> ```
> 
> The type is accepted.

Fixed. please verify again.

lib/validator/component_types.cpp
```diff
@@ -0,0 +1,1364 @@
...
+  } else if (D.isOptionTy()) {
+    Sub(D.getOption().ValTy);
+  } else if (D.isResultTy()) {
+    if (D.getResult().ValTy.has_value()) {
+      Sub(*D.getResult().ValTy);
+    }
+    if (D.getResult().ErrTy.has_value()) {
+      Sub(*D.getResult().ErrTy);
+    }
+  }
```
dannypsnl, line 199: `TypeSystem::collectNormalValResources` walks a value type to collect its resource ids.

It's used by the outer-alias resource-leak check (component_validator.cpp:637 "A type that crosses a component boundary must carry no free resource identity"), instance-type resource freshening, and export resource freshening at instantiation.

But here has no `isStreamTy()` and `isFutureTy()` branch, hence a resource id wrapped in stream/future will not be rejected.

```wast
 (type (resource))                 ;; type 0
 (type (own 0))                    ;; type 1
 (type (future 1))                 ;; type 2 = (future (own $r))
 (component (alias outer 1 <idx> (type)))
```

dannypsnl:
Are Memory64 supported by executor? If not should we keep reject 64-bit memory configuration here?

q82419:
> Are Memory64 supported by executor? If not should we keep reject 64-bit memory configuration here?

Do you mean component-model or WASM?
In WASM, we already supported memory64 proposal.
In component model, we plan to support memory64 also.

dannypsnl:
I mean component-model's

lib/validator/component_type.cpp
```diff
@@ -0,0 +1,796 @@
...
+            CheckRefs(ST.getCompositeType().getFuncType().getParamTypes()));
+        EXPECTED_TRY(
+            CheckRefs(ST.getCompositeType().getFuncType().getReturnTypes()));
+      }
+      if (!HasTypeRefs) {
+        EXPECTED_TRY(validate(ST,
+                              static_cast<uint32_t>(Checker.getTypes().size()),
+                              SubTypeDepthMap));
+      }
+      CompCtx.top().CoreTypes.push_back({&ST, nullptr});
```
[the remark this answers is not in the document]
q82419: Fixed

lib/validator/component_types.cpp
q82419: Fixed

test/component/ComponentValidatorTest.cpp
```diff
@@ -2676,4 +2690,846 @@ TEST(ComponentValidatorTest, StartArgumentValueIndexOutOfBounds) {
...
+  auto &AliasSec = std::get<AST::Component::AliasSection>(Inner[1]);
+  AliasSec.getContent().emplace_back();
+  auto &A = AliasSec.getContent().back();
+  A.setTargetType(AST::Component::Alias::TargetType::Outer);
+  A.getSort().setIsCore(true);
+  A.getSort().setCoreSortType(AST::Component::Sort::CoreSortType::Type);
+  A.setOuter(1U, 1U);
+
+  auto &InnerTypes2 = std::get<AST::Component::CoreTypeSection>(Inner[2]);
+  std::vector<AST::SubType> IST;
```
q82419: codespell (misc linters CI) flags `IST` here - this is the only failing check on the PR. Renaming the variable (e.g. `InnerSubTypes`) unblocks CI.

lib/validator/component_types.cpp
```diff
@@ -0,0 +1,1372 @@
...
+      Max = depthOfNormalVal(normalizeEntry(Info.Type));
+    }
+    break;
+  case ExternKind::ValueBound:
+    Max = depthOfValType(Info.Value);
+    break;
+  case ExternKind::InstanceType:
+  case ExternKind::ComponentType:
+    // Instances carry no imports, so one walk covers both shapes.
+    if (Info.Shape != nullptr) {
```
q82419: `InstanceType`/`ComponentType` shapes add no `+1` per nesting level here (and shapes reached via the `TypeBound` case contribute 0), so `checkTypeDepth` (limit 100) never bounds shape nesting - only `MaxTypeSize` (1,000,000) does. A linear chain of ~100k nested instance types has size about 2N, passes `checkTypeLimits`, and then the unmemoized recursion here, in `Matcher::matchExtern`/`matchInstanceShape`/`matchComponentShape`, and in `rebuildShape`/`rebuildExtern` descends once per level - a stack-overflow DoS on crafted input.

Suggestion: count `+1` per shape level for `InstanceType`/`ComponentType` (including via `TypeBound`), and memoize depth per `Shape*` the way `sizeOfExtern` uses `TypeSizeMemo`.

lib/validator/component_types.cpp
```diff
@@ -0,0 +1,1372 @@
...
+    if (*RA != SupId) {
+      FailCode = ErrCode::Value::ComponentResourceMismatch;
+      return false;
+    }
+    return true;
+  }
+  return false;
+}
+
+bool Matcher::matchFunc(const FuncInfo &Sub, const FuncInfo &Sup) noexcept {
```
q82419, line 1087: `matchFunc` compares param labels/types and results but never `FuncType::isAsync()`, so an async function satisfies a sync import/type ascription and vice versa. The async bit is only checked at the canon lift/lower sites. Suggest also requiring the async flags to agree here (per the current subtyping rules).

lib/validator/component_canon.cpp
```diff
@@ -0,0 +1,523 @@
...
+    return Component::Context::hasOption(Canon, Code);
+  };
+
+  // The resource built-ins take no options and name a resource type.
+  auto ResourceEntryAt =
+      [&](std::string_view What) -> Expect<const Component::TypeEntry *> {
+    if (!Canon.getOptions().empty()) {
+      spdlog::error(ErrCode::Value::InvalidCanonOption);
+      spdlog::error("    resource built-ins take no canonical options."sv);
+      return Unexpect(ErrCode::Value::InvalidCanonOption);
```
q82419: `InvalidCanonOption` is 0x1301, in the instantiation-phase range, but this is a validation-phase rejection - `getErrCodePhase()` will report it as an instantiation failure. The executor's canon paths legitimately keep 0x1301; for this validator site a 0x12xx validation-range code would keep the phase encoding consistent.

include/common/types.h
```diff
@@ -581,9 +584,14 @@ struct ResultVal {
 };
 struct FlagsVal {
   std::vector<bool> Bits;
+  // Optional set-label form: when non-empty (or Bits is empty), the bits are
+  // resolved against the declared flag labels at the lowering site.
+  std::vector<std::string> SetLabels{};
```
q82419, line 589: These host label forms (`VariantVal::Label`, `FlagsVal::SetLabels`, `EnumVal::Label`) are documented as resolved "at the lowering site", but only the validator's value decoder fills them - no lowering/executor path consumes them yet. Fine as deferred work for the value-instantiation phase, but consider a TODO or softening the comments so the public header does not advertise unimplemented behavior.

Minor: `VariantVal::Case` and `EnumVal::Case` are left uninitialized while the label members use `{}` - worth defaulting to 0 for consistency.

lib/validator/component_context.cpp
```diff
@@ -1,194 +1,1240 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2019-2025 Second State INC
```
q82419: SPDX drift: the four new validator sources (`component_canon.cpp`, `component_context.cpp`, `component_type.cpp`, `component_types.cpp`) use `2019-2025 Second State INC`, while the rest of `lib/validator/` uses `SPDX-FileCopyrightText: Copyright The WasmEdge Authors` (this file used the standard form before the rewrite). Suggest aligning all four.

include/executor/component/canonical_abi.h
```diff
@@ -57,8 +57,8 @@ struct CanonCtx {
   Runtime::Instance::FunctionInstance *Realloc = nullptr;
   const Runtime::Instance::ComponentInstance *CompInst = nullptr;
   /// Optional alternative to `CompInst` for resolving component type indices.
-  /// Lets validator-time callers (which have a ComponentContext but no
-  /// ComponentInstance) reuse alignment / elemSize / flatten_* without
+  /// Lets validator-time callers (which have a Validator::Component::Context
+  /// but no ComponentInstance) reuse alignment / elemSize / flatten_* without
   /// duplicating the recursion. Takes precedence over CompInst when set.
   std::function<const AST::Component::DefType *(uint32_t)> TypeResolver;
```
q82419: After this rewrite nothing in the tree sets `TypeResolver` anymore - the new validator has its own flatten/limit implementation and no longer calls into canonical_abi.h, and no executor path assigns it. The comment (touched in this PR) still describes validator-time callers. Consider removing the field plus the `resolveDefType` fallback, or rewording the comment, so the header does not document a defunct integration.

q82419 (commented):
Full review of the validator rewrite at head 12d8244. All previously threaded review comments are addressed; the inline comments below are the remaining items found.

**Inline findings**
- **Blocker**: codespell CI failure (`IST` variable in ComponentValidatorTest.cpp) - the only red check.
- **Major**: instance/component shape nesting escapes the 100-depth limit, leaving unbounded recursion reachable from hostile input.
- `InvalidCanonOption` (0x1301, instantiation-phase range) returned from a validation path.
- `matchFunc` does not compare the async flag.
- Host label fields in `common/types.h` have no consumer yet.
- SPDX header drift in the four new validator sources.
- `CanonCtx::TypeResolver` is dead after this rewrite.

**Non-inline notes**
- Commit dfb5ba89 (`feat(loader)`) message lists the new type encodings but omits the added `map` value type (0x63).
- The PR description's Test Evidence section is empty; pasting the ctest summary would help reviewers.
- @dannypsnl's follow-up about component-model memory64 support in the executor looks unanswered.
- Pre-existing, not this PR: `isKebabString` accepts digit-leading words after the first (e.g. `foo-9bar`); moved verbatim from base.

**Verified strengths**: the value-payload decoder's LEB128 bit-cap/sign handling, UTF-8 scalar validation (overlong/surrogate rejection), canonical-NaN enforcement, and full-consumption checks are rigorous; error-code renumbering keeps previously published ranges vacant and preserves the phase nibble; value linearity is enforced at every consume site and at end-of-component; commits are atomic with DCO and Assisted-by trailers.

lib/validator/component_types.cpp
```diff
@@ -0,0 +1,1372 @@
...
+  }
+  if (D.isOptionTy()) {
+    return Sub(D.getOption().ValTy);
+  }
+  if (D.isResultTy()) {
+    const auto &R = D.getResult();
+    return (R.ValTy.has_value() && Sub(*R.ValTy)) ||
+           (R.ErrTy.has_value() && Sub(*R.ErrTy));
+  }
+  return false;
```
dannypsnl, line 139: is stream and is future cases missing here?

test/component/ComponentValidatorTest.cpp
q82419: Fixed

lib/validator/component_types.cpp
q82419: Fixed

lib/validator/component_types.cpp
q82419: Fixed

lib/validator/component_canon.cpp
q82419: Fixed

include/common/types.h
q82419: Fixed

include/executor/component/canonical_abi.h
q82419: Fixed

lib/validator/component_types.cpp
q82419: Deliberate here, unlike the resource walk you flagged earlier.

`stream<T>` / `future<T>` can never contain a borrow in the first place: `validate(DefValType)` rejects a borrow in the payload when the type is defined (`ComponentStreamFutureBorrow`), and the rule is transitive — see `FutureOfBorrowRejected` / `StreamOfBorrowRejected` / `FutureOfListOfBorrowRejected`. So by induction no `stream`/`future` type reaching `containsBorrow` has a borrow inside, and recursing would always return false.

That also matches wasmparser: `TypeInfo::type_info` returns a plain `TypeInfo::new()` (contains_borrow = false) for `Future(_)` / `Stream(_)`, so the payload deliberately does not propagate borrow-ness there either.

The contrast with `collectNormalValResources` is the point: `future<own($r))` is perfectly legal, so the payload does carry resource identity, and wasmparser`s `free_variables_component_defined_type_id` does recurse into `Future`/`Stream`. That one was a real gap and is fixed.

Added a two-line comment at the end of `containsBorrow` in the latest push so the invariant is stated where it is relied on.

dannypsnl (approved):

q82419:
@LFsWang  do you have any other questions?

PR #5284, opened 2026-08-16
fix(coredump): fix coredump generation

grorge123:
## Description
This PR rewrites the generator to follow the WebAssembly tool-conventions `Coredump.md`, keeps a dedicated compatibility mode for the `wasmgdb` parser, and replaces the placeholder test with tests that parse the generated coredump back.

### What changed

- `include/executor/coredump.h`: added the missing `#pragma once` and reduced the public API to a single `generateCoredump(StackMgr, PC, ForWasmgdb) -> Expect<std::string>`.
- `lib/executor/coredump.cpp`: rewritten. Deduplicates module/memory/global instances by address for `coreinstances`, resolves the executing function of each frame by instruction pointer (correct for `call_indirect` and for the outermost frame), emits typed locals, dumps every linear memory, and emits globals as constant globals holding their current values.
- `lib/executor/engine/engine.cpp`: pass the trapping `PC` into the generator, and explicitly discard its result so a coredump failure cannot shadow the original trap.
- `lib/driver/runtimeTool.cpp`: `--coredump-for-wasmgdb` now enables coredump generation; previously it only selected the format and produced no file on its own. Because the coredump is produced by the interpreter only, and `--run-mode` is already documented in this function as the option with the highest precedence, an explicitly requested non-interpreter run mode is kept and the coredump is disabled with an actionable warning, instead of silently degrading the performance of the requested mode. The decision follows the requested run mode rather than the effective one, so it is deterministic on builds without LLVM.
- `include/loader/serialize.h`: expose `serializeS32/S64/F32/F64`, as suggested in the review of #3860.
- Tests: three coredump tests that parse the dump back and assert every section is fully consumed, plus `RunSubcommand.CoredumpRequiresInterpreter`, which covers eight run-mode/coredump combinations and asserts the number of generated coredump files.

### Related PR and Issues
Solved: #5205 #5162
Similar PR: #5206 #5164 #5260

## Checklist

## Test Evidence
image

We create a test WASM that has an out-of-bounds bug, and it is also stored in driverToolTest.cpp.
image
After using wasmedge to run it with the enable coredump flag, it crashes and creates a coredump.
The following figure shows the coredump content and can be parsed by wasmgdb.
image
image

### DIFF
--- a/include/driver/tool.h
+++ b/include/driver/tool.h
@@ -58,9 +58,12 @@ struct DriverToolOptions : public DriverProposalOptions {
         ConfEnableJIT(
             PO::Description("Enable Just-In-Time compiler for running WASM"sv)),
         ConfEnableCoredump(PO::Description(
-            "Enable coredump when WebAssembly enters a trap"sv)),
+            "Enable coredump when WebAssembly enters a trap. Requires the "
+            "interpreter run mode, otherwise it is disabled"sv)),
         ConfCoredumpWasmgdb(
-            PO::Description("Enable coredump for wasm-gdb to debug"sv)),
+            PO::Description("Enable coredump in the format expected by the "
+                            "wasmgdb debugger, which differs from the one of "
+                            "the WebAssembly tool conventions"sv)),
         ConfForceInterpreter(
             PO::Description("Forcibly run WASM in interpreter mode."sv)),
         ConfRunMode(PO::Description("Set execution mode. Valid values: "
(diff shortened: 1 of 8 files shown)


q82419:
Hi @grorge123 

I'd like to complete the serializer related issue and refactoring first.
After completion, I will notice you to rebase on it.
Thanks.

hydai (dismissed):

q82419:
@grorge123  please help to resolve the conflict. Thanks!

q82419 (approved):

PR #5285, opened 2026-08-16
test(serializer): add round-trip tests for saturating truncation instructions

leetcodez:
Adds serializer round-trip test coverage for the saturating float-to-int truncation instructions, which were previously unexercised intest/loader/serializeInstructionTest.cpp.

Ref #4992.

- Covers i32.trunc_sat_f32_s/u, i32.trunc_sat_f64_s/u, i64.trunc_sat_f32_s/u, i64.trunc_sat_f64_s/u
- Encodes each instruction, parses the result back, and asserts the emitted byte sequence, so both directions are checked
- No production code is touched and no new test target is added; the existing wasmedgeLoaderSerializerTests binary goes from 56 to 57 tests

Verified locally: 57 tests from 7 test suites, all passing with WASMEDGE_USE_LLVM=ON and WASMEDGE_BUILD_TESTS=ON.


Ankitdotraider (commented):
Codecov flags 8 lines still uncovered in the new test file, worth a quick look before inviting the maintainers to approve, would be better if you elaborate on that

q82419 (approved):

PR #5287, opened 2026-08-17
fix: serialize: update serialization for instruction

RuchitRathi:
Summary

Fixes: #14131 — Fix a debug-only assertion in Cranelift alias analysis that assumed the observed_stores set could not grow during later instruction processing. Observations may be discovered after dead-store or idempotent-store rewrites, so the set must be monotonic (non-decreasing) rather than fixed-size.
Changes Made

Updated: the debug_assert! in [alias_analysis.rs](vscode-file://vscode-app/c:/Users/RuchitRathi/AppData/Local/Programs/Microsoft%20VS%20Code/e4c7e7b1d6/resources/app/out/vs/code/electron-browser/workbench/workbench.html) to assert that observed_stores.len() only grows or stays the same (>=), not that it remains equal.
Added: a regression test at [issue-14131.clif](vscode-file://vscode-app/c:/Users/RuchitRathi/AppData/Local/Programs/Microsoft%20VS%20Code/e4c7e7b1d6/resources/app/out/vs/code/electron-browser/workbench/workbench.html) that reproduces the original assertion failure and verifies the fix.

Verification

Single reproducer: cargo +1.95.0 run -p cranelift-tools -- test cranelift/filetests/filetests/alias/issue-14131.clif (Passed)
Full alias suite: cargo +1.95.0 run -p cranelift-tools -- test cranelift/filetests/filetests/alias (40 tests passed)
Notes

Rationale: Additional store observations can appear after optimization rewrites; the invariant must allow monotonic growth.
Risk: Low — this changes a debug-only assertion and adds a regression test; no behavioral change in release builds.

### DIFF
--- a/lib/loader/serialize/serial_instruction.cpp
+++ b/lib/loader/serialize/serial_instruction.cpp
@@ -18,7 +18,15 @@ Serializer::serializeInstruction(const AST::Instruction &Instr,
     } else {
       serializeU32(Instr.getMemoryAlign(), OutVec);
     }
-    serializeU64(Instr.getMemoryOffset(), OutVec);
+    if (Conf.hasProposal(Proposal::Memory64)) {
+      serializeU64(Instr.getMemoryOffset(), OutVec);
+    } else {
+      if (Instr.getMemoryOffset() > std::numeric_limits<uint32_t>::max()) {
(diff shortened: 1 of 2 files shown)


q82419:
Wrong PR template and spam description

PR #5304, opened 2026-08-24
fix(executor): use correct atomic notify access width

### LINKED ISSUE #5303: bug: memory.atomic.notify incorrectly traps at the final valid 4-byte memory address
### Summary

`memory.atomic.notify` incorrectly performs an 8-byte bounds check even though the instruction has a 4-byte access width. Consequently, the final valid four-byte-aligned address of a memory page traps with `MemoryOutOfBounds`.

### Current State

Executing `memory.atomic.notify` at address `65532` in a one-page shared memory traps in both interpreter and LLVM JIT modes.

Address `65528` succeeds and returns `0`, confirming a false boundary restriction.

`Executor::atomicNotify` checks the address using:

```cpp
MemInst.getPointer<std::atomic<uint64_t> *>(Address)

### Expected State

Address `65532` is four-byte aligned, and bytes `[65532, 65536)` are inside the one-page memory.

With no waiting threads, `memory.atomic.notify` should succeed and return `0`.

### Reproduction steps

1. Build WasmEdge from commit `d516c5d663de9cc275a284ffe4585526d164cca2` with threads and LLVM JIT enabled.

2. Save the following as `notify.wat`:

```wat
(module
  (memory 1 1 shared)
    i32.const 65532
    i32.const 0
    memory.atomic.notify))
```

3. Compile it:

```sh
wat2wasm --enable-threads notify.wat -o notify.wasm
```

4. Run it in interpreter mode:

```sh
wasmedge run --enable-threads notify.wasm test
```

5. Run it in LLVM JIT mode:

```sh
wasmedge run --enable-threads --run-mode=jit notify.wasm test
```

6. As a control, change the address to `65528`. Both modes succeed and return `0`.

Interpreter:
execution failed: out of bounds memory access, Code: 0x408
In instruction: memory.atomic.notify

LLVM JIT:
execution failed: out of bounds memory access, Code: 0x408
In instruction: memory.atomic.notify

### Components

Core

d516c5d663de9cc275a284ffe4585526d164cca2

Windows 11 (Wsl2)

### Hardware Architecture

x86_64

### Appendix

Suggested regression coverage:

- notification at `memory_size - 4`
- interpreter execution
- LLVM JIT execution when LLVM is enabled
- a nearby successful control case

mohitagarwal24:
Fixes #5303.

`memory.atomic.notify` has a four-byte access width, but `atomicNotify()` previously checked the address using `std::atomic<uint64_t>`, requiring eight available bytes. This caused the final valid four-byte memory location to incorrectly trap with `MemoryOutOfBounds`.

This change:

- validates the address using the specified four-byte access width
- permits notification at the final valid four-byte address
- adds regression coverage for addresses `65528` and `65532`
- covers interpreter and LLVM JIT execution

This contribution was developed with assistance from Cursor (GPT-5.6 Sol).

## Checklist

## Test Evidence

Commands:

```sh
cmake --build build-llvm --target wasmedgeThreadTests -j$(nproc)
./build-llvm/test/thread/wasmedgeThreadTests --gtest_filter='AtomicNotify.*'
ctest --test-dir build-llvm -R '^wasmedgeThreadTests$' --output-on-failure
```

Results:

```text
[==========] Running 2 tests from 1 test suite.
[ RUN      ] AtomicNotify.BoundaryInterpreter
[       OK ] AtomicNotify.BoundaryInterpreter
[ RUN      ] AtomicNotify.BoundaryJIT
[       OK ] AtomicNotify.BoundaryJIT
[  PASSED  ] 2 tests.

1/1 Test #14: wasmedgeThreadTests .............. Passed
```

Additional validation:

```text
clang-format 22.1.8: passed
commitlint: passed
```

### DIFF
--- a/test/thread/ThreadTest.cpp
+++ b/test/thread/ThreadTest.cpp
@@ -162,84 +162,157 @@ std::array<uint64_t, 4> Answers{
     UINT64_C(4446454406775736720),
     UINT64_C(9019442596657776185),
 };
-std::array<WasmEdge::Byte, 53> AtomicNotify32{
-    0x00, 0x61, 0x73, 0x6d, 0x01, 0x00, 0x00, 0x00, 0x01, 0x07, 0x01,
-    0x60, 0x02, 0x7f, 0x7f, 0x01, 0x7f, 0x03, 0x02, 0x01, 0x00, 0x05,
-    0x04, 0x01, 0x03, 0x01, 0x01, 0x07, 0x0a, 0x01, 0x06, 0x6e, 0x6f,
-    0x74, 0x69, 0x66, 0x79, 0x00, 0x00, 0x0a, 0x0c, 0x01, 0x0a, 0x00,
-    0x20, 0x00, 0x20, 0x01, 0xfe, 0x00, 0x02, 0x00, 0x0b,
+std::array<WasmEdge::Byte, 122> AtomicThreads32{
+    0x00, 0x61, 0x73, 0x6d, 0x01, 0x00, 0x00, 0x00, 0x01, 0x15, 0x03, 0x60,
+    0x02, 0x7f, 0x7f, 0x01, 0x7f, 0x60, 0x03, 0x7f, 0x7f, 0x7e, 0x01, 0x7f,
+    0x60, 0x03, 0x7f, 0x7e, 0x7e, 0x01, 0x7f, 0x03, 0x04, 0x03, 0x00, 0x01,
+    0x02, 0x05, 0x04, 0x01, 0x03, 0x01, 0x01, 0x07, 0x25, 0x04, 0x06, 0x6d,
+    0x65, 0x6d, 0x6f, 0x72, 0x79, 0x02, 0x00, 0x06, 0x6e, 0x6f, 0x74, 0x69,
+    0x66, 0x79, 0x00, 0x00, 0x06, 0x77, 0x61, 0x69, 0x74, 0x33, 0x32, 0x00,
+    0x01, 0x06, 0x77, 0x61, 0x69, 0x74, 0x36, 0x34, 0x00, 0x02, 0x0a, 0x26,
+    0x03, 0x0a, 0x00, 0x20, 0x00, 0x20, 0x01, 0xfe, 0x00, 0x02, 0x00, 0x0b,
+    0x0c, 0x00, 0x20, 0x00, 0x20, 0x01, 0x20, 0x02, 0xfe, 0x01, 0x02, 0x00,
+    0x0b, 0x0c, 0x00, 0x20, 0x00, 0x20, 0x01, 0x20, 0x02, 0xfe, 0x02, 0x03,
+    0x00, 0x0b,
 };
-std::array<WasmEdge::Byte, 111> AtomicNotify64{
-    0x00, 0x61, 0x73, 0x6d, 0x01, 0x00, 0x00, 0x00, 0x01, 0x0e, 0x02, 0x60,
-    0x02, 0x7e, 0x7f, 0x01, 0x7f, 0x60, 0x03, 0x7e, 0x7f, 0x7e, 0x01, 0x7f,
-    0x03, 0x04, 0x03, 0x00, 0x00, 0x01, 0x05, 0x04, 0x01, 0x07, 0x01, 0x01,
-    0x07, 0x23, 0x03, 0x06, 0x6e, 0x6f, 0x74, 0x69, 0x66, 0x79, 0x00, 0x00,
-    0x0d, 0x6e, 0x6f, 0x74, 0x69, 0x66, 0x79, 0x2d, 0x6f, 0x66, 0x66, 0x73,
-    0x65, 0x74, 0x00, 0x01, 0x06, 0x77, 0x61, 0x69, 0x74, 0x33, 0x32, 0x00,
-    0x02, 0x0a, 0x24, 0x03, 0x0a, 0x00, 0x20, 0x00, 0x20, 0x01, 0xfe, 0x00,
-    0x02, 0x00, 0x0b, 0x0a, 0x00, 0x20, 0x00, 0x20, 0x01, 0xfe, 0x00, 0x02,
-    0x04, 0x0b, 0x0c, 0x00, 0x20, 0x00, 0x20, 0x01, 0x20, 0x02, 0xfe, 0x01,
-    0x02, 0x00, 0x0b,
+std::array<WasmEdge::Byte, 255> AtomicThreads64{
+    0x00, 0x61, 0x73, 0x6d, 0x01, 0x00, 0x00, 0x00, 0x01, 0x1a, 0x04, 0x60,
+    0x02, 0x7e, 0x7f, 0x01, 0x7f, 0x60, 0x01, 0x7e, 0x01, 0x7f, 0x60, 0x03,
+    0x7e, 0x7f, 0x7e, 0x01, 0x7f, 0x60, 0x03, 0x7e, 0x7e, 0x7e, 0x01, 0x7f,
+    0x03, 0x08, 0x07, 0x00, 0x00, 0x01, 0x02, 0x02, 0x03, 0x03, 0x05, 0x04,
+    0x01, 0x07, 0x01, 0x01, 0x07, 0x6c, 0x08, 0x06, 0x6d, 0x65, 0x6d, 0x6f,
+    0x72, 0x79, 0x02, 0x00, 0x06, 0x6e, 0x6f, 0x74, 0x69, 0x66, 0x79, 0x00,
+    0x00, 0x0d, 0x6e, 0x6f, 0x74, 0x69, 0x66, 0x79, 0x2d, 0x6f, 0x66, 0x66,
+    0x73, 0x65, 0x74, 0x00, 0x01, 0x14, 0x6e, 0x6f, 0x74, 0x69, 0x66, 0x79,
+    0x2d, 0x77, 0x72, 0x61, 0x70, 0x70, 0x65, 0x64, 0x2d, 0x63, 0x6f, 0x75,
+    0x6e, 0x74, 0x00, 0x02, 0x06, 0x77, 0x61, 0x69, 0x74, 0x33, 0x32, 0x00,
+    0x03, 0x0d, 0x77, 0x61, 0x69, 0x74, 0x33, 0x32, 0x2d, 0x6f, 0x66, 0x66,
+    0x73, 0x65, 0x74, 0x00, 0x04, 0x06, 0x77, 0x61, 0x69, 0x74, 0x36, 0x34,
+    0x00, 0x05, 0x0d, 0x77, 0x61, 0x69, 0x74, 0x36, 0x34, 0x2d, 0x6f, 0x66,
+    0x66, 0x73, 0x65, 0x74, 0x00, 0x06, 0x0a, 0x5b, 0x07, 0x0a, 0x00, 0x20,
+    0x00, 0x20, 0x01, 0xfe, 0x00, 0x02, 0x00, 0x0b, 0x0a, 0x00, 0x20, 0x00,
+    0x20, 0x01, 0xfe, 0x00, 0x02, 0x04, 0x0b, 0x0f, 0x00, 0x20, 0x00, 0x42,
+    0x81, 0x80, 0x80, 0x80, 0x20, 0xa7, 0xfe, 0x00, 0x02, 0x00, 0x0b, 0x0c,
+    0x00, 0x20, 0x00, 0x20, 0x01, 0x20, 0x02, 0xfe, 0x01, 0x02, 0x00, 0x0b,
+    0x0c, 0x00, 0x20, 0x00, 0x20, 0x01, 0x20, 0x02, 0xfe, 0x01, 0x02, 0x04,
+    0x0b, 0x0c, 0x00, 0x20, 0x00, 0x20, 0x01, 0x20, 0x02, 0xfe, 0x02, 0x03,
+    0x00, 0x0b, 0x0c, 0x00, 0x20, 0x00, 0x20, 0x01, 0x20, 0x02, 0xfe, 0x02,
+    0x03, 0x08, 0x0b,
 };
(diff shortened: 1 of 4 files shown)


q82419:
The diagnosis is right and the fix is correct — thanks. Reproduced on clean master:
`notify(65532)` on a 1-page shared memory traps `MemoryOutOfBounds` where wasmtime
46.0.0 returns `0`. Our validator already agrees the width is 4:
`formchecker.cpp:1830` uses `checkMemArgAndTrans(32, ...)` and requires
`2^align == N/8`, so the memarg alignment must be exactly `2`.

Could you extend the PR to cover the rest of the same root cause? The signatures are

```
memory.atomic.notify : [at i32]       -> [i32]
memory.atomic.wait32 : [at i32 i64]   -> [i32]
memory.atomic.wait64 : [at i64 i64]   -> [i32]
```

Only the *address* follows the memory's address type `at`. The access width, the
`count` operand and the result are fixed regardless of memory64. Every place that
tied one of those to `AddrType` / `MemoryAddrTypes` is wrong, and there are four more
besides the one you fixed. Diffs below are verified — full matrix at the end.

### 1. `threadInstr.cpp:19` — runtime alignment is still 8 on a memory64

Same 8-byte assumption you are removing, one branch earlier. Memory64 is on by
default, so today a memory64 shared memory traps on every 4-mod-8 address although
the validator only accepts `align=2` — i.e. a module that passes validation is
guaranteed to trap, and your own 65532 case still fails there, just with
`UnalignedAtomicAccess` instead. It also contradicts the JIT, which already uses 4
unconditionally (`lib/llvm/compiler/threadInstr.cpp:367`).

```diff
   Address += Instr.getMemoryOffset();
-  uint32_t Align =
-      AddrType == AddressType::I32 ? sizeof(uint32_t) : sizeof(uint64_t);
 
-  if (Address % Align != 0) {
+  if (Address % sizeof(uint32_t) != 0) {
```

### 2. `threadInstr.cpp:29` and `:38`, `atomic.ipp:43` — operand and result widths

`count` and the result are `i32`, never `at`. `ValVariant` is a non-discriminated
union whose `emplace<T>` is a placement-new of `sizeof(T)` bytes, so on a memory64
`extractAddr(RawCount, AddrType)` reads 8 bytes out of a slot only 4 were written to
and picks up whatever the guest left there — `i64.const 0xFFFFFFFF00000001;
i32.wrap_i64` yields a count of 1 that wakes every waiter. The JIT already gets this
right with `createZExt`.

```diff
-  uint64_t Count = extractAddr(RawCount, AddrType);
+  uint64_t Count = extractAddr(RawCount, AddressType::I32);
```
```diff
-  RawAddress = emplaceAddr(Total, AddrType);
+  RawAddress = emplaceAddr(Total, AddressType::I32);
```
```diff
-      .map([&](auto V) { RawAddress = emplaceAddr(V, AddrType); });
+      .map([&](auto V) { RawAddress = emplaceAddr(V, AddressType::I32); });
```

### 3. `lib/llvm/compiler/threadInstr.cpp:378` and `:404` — notify/wait do not compile on a memory64

The result is truncated to the memory's address type, which on a memory64 is an
i64->i64 trunc and produces invalid IR:

```
LLVM module verification failed: Function return type does not match operand type of return inst!
  ret i64 %15
 i32
Compilation failed. Error code: 13, use interpreter mode instead.
```

So AOT is unavailable for memory64+threads and `--run-mode=jit` silently degrades to
the interpreter. Affects `notify`, `wait32` and `wait64`.

```diff
            LLContext.getInt32(MemoryIndex), Offset, Count}),
-      Context.MemoryAddrTypes[MemoryIndex]));
+      Context.Int32Ty));
```
```diff
            LLContext.getInt32(BitWidth)}),
-      Context.MemoryAddrTypes[MemoryIndex]));
+      Context.Int32Ty));
```

### 4. Same two functions — missing `boundsCheckMemory64`

`compileAtomicNotify` / `compileAtomicWait` are the only atomic compilers that skip
it (load `:414`, store `:450`, rmw `:471`, cmpxchg `:540` all call it), so on an i64
memory the memarg offset is added with a wrapping `createAdd`. After this PR the
relaxed 4-byte `checkAccessBound` is the JIT's only remaining bounds gate, so the
wrapped address goes straight through: `i64.const 0xFFFFFFFFFFFFFFFC` with
`offset=4` returns `0` under JIT while the interpreter and wasmtime both trap
`MemoryOutOfBounds`. Placed *after* the alignment check so the trap precedence keeps
matching the interpreter and wasmtime (unaligned wins over out-of-bounds).

```diff
   auto Count = Builder.createZExt(stackPop(), Context.Int64Ty);
-  auto Offset = Builder.createZExt(stackPop(), Context.Int64Ty);
+  auto Addr = Builder.createZExt(stackPop(), Context.Int64Ty);
+  auto Offset = Addr;
   if (MemoryOffset != 0) {
-    Offset = Builder.createAdd(Offset, LLContext.getInt64(MemoryOffset));
+    Offset = Builder.createAdd(Addr, LLContext.getInt64(MemoryOffset));
   }
   compileAtomicCheckOffsetAlignment(Offset, Context.Int32Ty);
+  boundsCheckMemory64(MemoryIndex, Addr, MemoryOffset, sizeof(uint32_t));
```

…and the same shape in `compileAtomicWait`, with
`boundsCheckMemory64(MemoryIndex, Addr, MemoryOffset, BitWidth / 8);`.

### 5. Tests

The current cases do not yet guard the fix. `count` is hard-coded to 0, so
`Total < Count` is false immediately and `atomicNotify` never touches the address —
deleting the bound check outright leaves both tests green for every address, 65536
included. `65528` is also a control rather than a regression case: the old check was
`checkAccessBound(65528, 8)` -> `65536 <= 65536`, which already passed on master.

- negative cases: `65536` -> `MemoryOutOfBounds`, `65533` -> `UnalignedAtomicAccess`
  (`UINT32_MAX` is not usable — alignment fires first)
- a `(memory i64 1 1 shared)` case, which is what covers 1., 3. and 4.
- `BoundaryJIT` currently cannot fail for a JIT reason: `VM::unsafeLoadJITExecutable`
  (`lib/vm/vm.cpp:394-448`) maps every compile/JIT/load failure back to
  `ErrCode::Value::Success`, and the two that log use `spdlog::warn`, which
  `setErrorLoggingLevel()` at `ThreadTest.cpp:345` hides. Please assert the function
  really was compiled:

  ```cpp
  const auto *F = VM.getActiveModule()->findFuncExports("test"sv);
  ASSERT_NE(F, nullptr);
  EXPECT_TRUE(F->isCompiledFunction());
  ```

### Verified

With 1.-4. plus your change, interpreter and JIT, on `(memory 1 1 shared)` and
`(memory i64 1 1 shared)`:

| case                              | result                                |
|-----------------------------------|---------------------------------------|
| `notify(4)`                       | `0`                                   |
| `notify(65532)`                   | `0`                                   |
| `notify(65533)`                   | `unaligned atomic` (0x410)            |
| `notify(65536)`                   | `out of bounds memory access` (0x408) |
| `notify(0xFFFFFFFFFFFFFFFC)+off 4`| `out of bounds memory access` (0x408) |
| `wait32` on memory64 under JIT    | compiles, correct result              |

All match wasmtime 46.0.0. `wasmedgeExecutorCoreTests` 531/531, `wasmedgeThreadTests`
green, `clang-format` clean on all three files.

Separately, reviewing this turned up some pre-existing races in the wait/notify
machinery itself (lost wakeup on the unpredicated `Cond.wait`, the notify loop
recounting already-notified waiters, the single-slot `StopToken`/`WaitingMemory`).
Those are a different kind of change, so I will file them as issues rather than pile
them onto this PR.

mohitagarwal24:
Thanks — that's a much better framing than mine, and the "only the address follows `at`; width, count and result are fixed" rule made the other four sites straightforward to find. All five points are implemented, split into `a16428db` (runtime) and `78f1427c` (tests).

Items 1–4 are applied as written, including placing `boundsCheckMemory64` after the alignment check so unaligned keeps winning over out-of-bounds.

### What the new tests guard

As you stated, I reverted pieces and rebuilt to check each fix is actually held by an assertion:

| reverted | result |
|---|---|
| all three production files to master | both tests fail |
| only the two `boundsCheckMemory64` calls | JIT fails at the offset-overflow case; interpreter stays green |
| only the `AddressType::I32` count/result changes | both tests still pass |

The specific assertions that fail without each fix:

- `notify(65532)` on memory32 — the original 8-byte bound
- `notify(65532)` on memory64 — item 1, which trapped `UnalignedAtomicAccess`
- `isCompiledFunction()` on the memory64 exports — item 3; this confirms your
  point that `unsafeLoadJITExecutable` was silently degrading to the
  interpreter, since the assertion is what catches it, not a wrong result
- `notify(0xFFFFFFFFFFFFFFFC)` with `offset=4` — item 4, and it fails in JIT
  mode only while the interpreter stays green, exactly as you described

### One gap I want to flag

Item 2 is implemented but **not** guarded by a test. `count` is only observable when a waiter actually exists — with none, `Total` is 0 regardless of what `count` reads as, and a `Total` of 0 emplaced as i64 versus i32 leaves identical low bytes, so the host read can't distinguish them either. Reproducing the `i64.const 0xFFFFFFFF00000001; i32.wrap_i64` case needs a live waiter on a second thread, i.e. a sleep-based test layered on the wait/notify races you're filing separately. Happy to add one if you'd prefer the coverage over the flake risk — just say which way you want it.

### Results

`wasmedgeThreadTests` green (6/6), `wasmedgeExecutorCoreTests` green (529/529 in my configuration, I may be a spec-suite revision or a proposal flag behind; let me know if that gap matters). clang-format clean on all four files.

### Unrelated observation

While matching your ordering requirement I noticed `compileAtomicLoad`, `compileAtomicStore`, `compileAtomicRMW` and `compileAtomicCompareExchange` all call `boundsCheckMemory64` *before* `compileAtomicCheckOffsetAlignment` — the opposite order from what you asked for on notify/wait. If trap precedence matters there too, that looks like a sibling of the issues you're filing. I've left it alone to keep this PR scoped.

mohitagarwal24:
Follow-up to my previous comment: I found a deterministic way to cover the remaining count-width case without starting a waiter thread or relying on sleep-based timing. This is included in commit `159fec14`.

The Memory64 fixture now exports its memory. The test pre-registers two waiter records at the same address and then computes the notify count using:

```wat
i64.const 0x0000000200000001
i32.wrap_i64
memory.atomic.notify
```

The interpreter’s `i32.wrap_i64` writes only the lower four bytes of the value slot. Correctly extracting the count as `i32` therefore wakes one waiter and returns `1`.

If the implementation is reverted to extract the count using the Memory64 address type, it also reads the retained upper bytes, interprets the value as a large count, wakes both waiters, and returns `2`.

This makes the operand-width correction mutation-sensitive without introducing concurrency or timing into the test.

### Additional coverage

I also extended the test matrix to cover:

- `wait32` and `wait64` on Memory32 and Memory64
- final valid aligned addresses
- combined unaligned/out-of-bounds addresses to verify trap precedence
- aligned out-of-bounds addresses
- successful nonzero memarg offsets
- offset-induced out-of-bounds accesses
- 64-bit memarg-offset overflow for `notify`, `wait32`, and `wait64`
- zero-timeout waits to exercise waiter registration and cleanup deterministically
- `isCompiledFunction()` for every exported function in JIT mode

### Mutation verification

I reverted each production change individually, rebuilt, and reran the tests. The suite now fails when reverting:

- the notify alignment width
- the notify count width
- the notify access width
- either LLVM result truncation
- either LLVM Memory64 bounds check
- use of the original address instead of the post-offset address in the JIT bounds check

The two interpreter result-emplacement changes are not externally mutation-observable because invocation normalizes results to the function’s declared `i32` type. However, the implementation now uses the correct internal width.

### Local verification

- `wasmedgeThreadTests`: 6/6 passed
- `wasmedgeExecutorCoreTests`: 529/529 passed
- `clang-format`: clean

Could you please take another look when convenient?

q82419 (changes requested):
Fix your test styles. Don't make lots of global variables outside the test scope.

mohitagarwal24:
Done in a7aaf09f. Wrapped it all in a test fixture so the wasm blobs, constants and helpers aren't sitting at file scope anymore, and dropped one constant I wasn't even using. Tests still pass, format is clean.

This is a pure scope change — no assertion, address, or expected value was touched, so the mutation coverage described in my previous comment still holds.

q82419 (changes requested):
Still in weird format. please follow the style of other tests.

mohitagarwal24:
Could you pls take a look again, it should match the existing test style now.

test/thread/ThreadTest.cpp
```diff
@@ -162,9 +163,400 @@ std::array<uint64_t, 4> Answers{
...
+                              WasmEdge::ValType(WasmEdge::TypeCode::I64)});
+    ASSERT_FALSE(Result);
+    EXPECT_EQ(Result.error(), WasmEdge::ErrCode::Value::MemoryOutOfBounds);
+  }
+}
+
+TEST(AtomicWaitNotify, Memory64) {
+  WasmEdge::Configure Conf;
+  Conf.addProposal(WasmEdge::Proposal::Threads);
+  Conf.addProposal(WasmEdge::Proposal::Memory64);
```
q82419, line 329: Memory64 is default ON. Therefore this line is not necessary.

test/thread/ThreadTest.cpp
```diff
@@ -230,6 +622,410 @@ TEST(AsyncExecute, GasThreadTest) {
...
+                              WasmEdge::ValType(WasmEdge::TypeCode::I64)});
+    ASSERT_FALSE(Result);
+    EXPECT_EQ(Result.error(), WasmEdge::ErrCode::Value::MemoryOutOfBounds);
+  }
+}
+
+TEST(AtomicWaitNotifyJIT, Memory64) {
+  WasmEdge::Configure Conf;
+  Conf.addProposal(WasmEdge::Proposal::Threads);
+  Conf.addProposal(WasmEdge::Proposal::Memory64);
```
q82419, line 764: Memory64 is default ON. Therefore this line is not necessary.

q82419 (changes requested):

q82419 (approved):

PR #5318, opened 2026-08-29
fix(executor): correct atomic.notify alignment for memory64

### LINKED ISSUE #5310: bug: `memory.atomic.notify` on Memory64 requires 8 byte alignment instead of 4
### Summary

`memory.atomic.notify` always accesses 4 bytes, so a 4 byte aligned address is valid even when the memory uses i64 addresses. In the interpreter,alignment is taken from the address type not the access width.For a Memory64 memory that becomes 8 bytes, so a legal address such as 4 traps with `UnalignedAtomicAccess`.

`memory.atomic.wait32` in the same interpreter uses sizeof(T) (4). LLVM JIT also checks 4 byte alignment. Only interpreter notify is wrong.

### Current State

On a Memory64 shared memory the interpreter treats `memory.atomic.notify` as 8 byte aligned.A valid 4 byte aligned address such as 4 traps with `UnalignedAtomicAccess`.

### Expected State

`memory.atomic.notify` always accesses 4 bytes, including on Memory64 (`[i64 i32] -> [i32]`). Address `4` is 4 byte aligned and in bounds, so notify should succeed and return 0 when no threads are waiting.

### Reproduction steps

1. Build WasmEdge from commit `d8f9ee5dac6e1ba157246ccdcea7b3f9e539e4e1` with threads enabled.

2. notify64.wat:
```
(module
  (memory i64 1 1 shared)
    i64.const 4
    i32.const 0
    memory.atomic.notify))
```
3. Compile:
```
wat2wasm --enable-threads --enable-memory64 notify64.wat -o notify64.wasm
```

4. Run interpreter mode:
```
wasmedge run --enable-threads notify64.wasm test
```
5. Run LLVM JIT mode:
```
wasmedge run --enable-threads --run-mode=jit notify64.wasm test
```

As a control, change the address to `i64.const 0` or` i64.const 8`. Interpreter notify then succeeds and returns 0.

Interpreter (address 4): execution failed: unaligned atomic, Code: 0x410 In instruction: memory.atomic.notify

LLVM JIT (address 4): 0

### Components

Core

d8f9ee5dac6e1ba157246ccdcea7b3f9e539e4e1

Windows 11

### Hardware Architecture

x86_64

### Appendix

https://github.com/WasmEdge/WasmEdge/blob/d8f9ee5dac6e1ba157246ccdcea7b3f9e539e4e1/lib/executor/engine/threadInstr.cpp#L19-L22

This is not [#5303](https://github.com/WasmEdge/WasmEdge/issues/5303). That issue is a 32 bit share memory bounds trap at 65532 (`MemoryOutOfBounds` from an 8 byte `getPointe`r).

ravindra-RKB:
## Description

Fixes issue #5310.

`memory.atomic.notify` operates on a 32-bit value, so its required alignment is always 4 bytes. Previously, the alignment was determined from the address type, causing Memory64 (`i64`) accesses to incorrectly require 8-byte alignment.

This change uses `sizeof(uint32_t)` consistently for `memory.atomic.notify`.

## Changes

- Corrected `memory.atomic.notify` alignment checking for Memory64.
- Added a regression test covering Memory64 + Threads with a 4-byte-aligned address.

## Testing

- Targeted regression test:
  `ExecutorRegression.MemoryAtomicNotifyMemory64` — PASSED
- Full CTest suite:
  `23/23 tests passed`
- `git diff --check` — PASSED

## Related Issue

Fixes #5310


lib/executor/engine/threadInstr.cpp
```diff
@@ -16,8 +16,7 @@ Executor::runAtomicNotifyOp(Runtime::StackManager &StackMgr,
   uint64_t Address = extractAddr(RawAddress, AddrType);
   EXPECTED_TRY(checkOffsetOverflow(MemInst, Instr, Address, sizeof(uint32_t)));
   Address += Instr.getMemoryOffset();
-  uint32_t Align =
-      AddrType == AddressType::I32 ? sizeof(uint32_t) : sizeof(uint64_t);
+  uint32_t Align = sizeof(uint32_t);
```
q82419: You can directly use `Address % sizeof(uint32_t) != 0`

test/executor/ExecutorRegressionTest.cpp
```diff
@@ -1285,6 +1285,38 @@ TEST(ExecutorRegression, TailCallToHostImport) {
...
+/// Binary Wasm module: Memory64 + Threads.
+///
+/// (module
+///   (memory $m i64 1 1 shared)
+///   (func (export "test")
+///     (memory.atomic.notify (i64.const 4) (i32.const 1))
+///     drop
+///   )
+/// )
+std::array<WasmEdge::Byte, 49> NotifyMemory64Wasm{
```
q82419: Put your WASM array to the correct place

q82419 (changes requested):

test/executor/ExecutorRegressionTest.cpp
```diff
@@ -1285,6 +1301,22 @@ TEST(ExecutorRegression, TailCallToHostImport) {
   }
 }
 
+/// Regression test for memory.atomic.notify with Memory64 (issue #5310).
+TEST(ExecutorRegression, MemoryAtomicNotifyMemory64) {
+  WasmEdge::Configure Conf;
+  Conf.addProposal(WasmEdge::Proposal::Memory64);
```
q82419: Memory64 is default ON. Therefore this line is not necessary.

q82419 (changes requested):

q82419 (approved):

PR #5329, opened 2026-09-05
perf(llvm): lower v128.store to <2 x i64> instead of <1 x i128>

### LINKED ISSUE #4841: Performance anomaly:  WasmEdge JIT is much slower for `v128.store -> v128.load -> lane extract` than for nearby SIMD memory patterns

Byte-Naut:
## Description

Fixes #4841.

Two commits: a no-op refactor that the second one needs on big-endian targets, then the change itself.

### Relationship to #4848

#4848 targets the same issue by removing `volatile` from all memory accesses. This PR is unrelated in both mechanism and code: `volatile` is left untouched, and the substantive change is only the LLVM type `v128.store` lowers to. The two touch different files and can land independently.

---

### 1/2 — `refactor(llvm): use a same-width transit type for vector byte swap`

`switchEndian()` bitcasts a vector through an integer before bswapping it, and picks that integer by inspecting the **lane** width:

```cpp
LLVM::Type VecType = Type.getElementType().getIntegerBitWidth() == 128
                         ? Context.Int128Ty
                         : Context.Int64Ty;
```

**On master this is correct.** The branch above returns early for vectors of size 1, so `<1 x i128>` never arrives here, and every shape that does is 64 bits wide:

| source | `LoadTy` | width | transit type |
|---|---|---:|---|
| `v128.load8x8_{s,u}` | `<8 x i8>` | 64 | `i64` |
| `v128.load16x4_{s,u}` | `<4 x i16>` | 64 | `i64` |
| `v128.load32x2_{s,u}` | `<2 x i32>` | 64 | `i64` |

The selector picks `Int64Ty` for all of them and the widths agree, so this commit changes no observable behaviour.

It is needed because 2/2 makes `v128.store` lower as `<2 x i64>` — 128 bits wide with 64-bit lanes. The lane-width selector picks `Int64Ty` for that too, and the verifier rejects the result:

```
Invalid bitcast
  %3 = bitcast i64 %2 to <2 x i64>
```

So size the transit integer from the total width, making it same-width by construction:

```cpp
Value = Builder.createBitCast(
    Value, LLContext.getIntNTy(Type.getPrimitiveSizeInBits()));
```

`<8 x i8>`, `<4 x i16>` and `<2 x i32>` still go through `i64`; `<2 x i64>` and `<1 x i128>` go through `i128`. The three extending loads keep byte-identical IR and the `<2 x i64>` store switchEndian path becomes legal, so either commit is independently correct and the series bisects.

---

### 2/2 — `perf(llvm): lower v128.store to <2 x i64> instead of <1 x i128>`

`v128.store` lowered its value as `<1 x i128>`, which x86-64 legalises into a pair of 8-byte GPR stores. The `v128.load` that follows then has to reassemble a 128-bit vector out of memory that was just written by two narrow stores. That store-to-load forwarding behaviour is what dominates the loop in #4841 — which is also why the `drop` control from @gaaraw stays fast: with the loaded vector unobserved, the reload never has to be materialised as a vector at all.

Lowering the store as `<2 x i64>` emits a single 16-byte vector store. The canonical stack representation for v128 is already `<2 x i64>` (see `compileVectorLoadOp`), so the bitcast `compileStoreOp` performs becomes a no-op and no other change is required. Address computation, offset handling, forced unalignment and the `volatile` flag are untouched.

Thanks to @gaaraw for the ablation in #4841 — the `drop` and `scalar_pair` controls are what made the mechanism identifiable.

## Checklist

## Test Evidence

`test/llvm` (`wasmedgeLLVMCoreTests`): **2004/2004 pass**, covering `address`, `traps`, `simd_address`, `simd_lane`, `simd_load`, `simd_store`, `memory64` and `linking` across AOT, AOT-universal, JIT and lazy JIT.

Trap semantics: a fully out-of-bounds `v128.load`/`v128.store`, and a page-straddling `v128.store` (offset 65530 in a one-page memory, 6 bytes in bounds and 10 out), all report `out of bounds memory access, Code: 0x408`.

Value correctness: `v128.store` of a distinguishable 4-lane constant followed by `v128.load` + `extract_lane` returns bit-identical results under JIT and under the interpreter.

### Measurements

Reproducers from #4841. 2^30 iterations each, median of five, `--run-mode=jit`, built from `ff889f54`. Intel Core Ultra 9 275HX, Ubuntu, LLVM 18.1.3.

| testcase | before | after |
|---|---:|---:|
| `reproducer_fixed_addr_lane_extract` | 4.89 s | **0.23 s** |
| `reproducer_fixed_addr_drop` | 0.37 s | 0.35 s |
| `reproducer_fixed_addr_scalar_pair` | 0.42 s | 0.50 s |

**`scalar_pair` regresses by roughly 19%.** That control stores a vector and reads it back through two scalar `i64.load`s, so a wide store now feeds narrower loads. It is reproducible across runs, not measurement noise.

Cold start vs steady state, with a one-iteration variant of the same module:

| | cold (1 iter) | full (2^30) | steady |
|---|---:|---:|---:|
| before | 0.01 s | 4.84 s | 4.83 s |
| after | 0.01 s | 0.23 s | **0.22 s** |

Compile time is unchanged; the entire gain is in the loop body.

The gain does not depend on how the memory is declared:

| | before | after |
|---|---:|---:|
| shared memory | 4.92 s | 0.23 s |
| memory64 | 4.17 s | 0.61 s |

### Big-endian

Commit 1/2 sits inside `if constexpr (Endian::native == Endian::big)`, so it is compiled out on the machine above and CI does not execute it: `build_for_s390x.yml` is `workflow_dispatch`-only, does not build the test suite, and runs `fibonacci.wat` alone.

Verified manually instead, by forcing that condition true on an x86-64 build and compiling modules that exercise every shape reaching the changed branch:

| shape | reached via | transit type |
|---|---|---|
| `<8 x i8>` | `v128.load8x8_s` / `_u` | `i64` |
| `<4 x i16>` | `v128.load16x4_s` / `_u` | `i64` |
| `<2 x i32>` | `v128.load32x2_s` / `_u` | `i64` |
| `<2 x i64>` | `v128.store` (2/2) | `i128` |

Before 1/2, the six extending loads compile cleanly and the `<2 x i64>` store fails with the `Invalid bitcast` shown above. After it, all seven verify, and the six extending loads produce byte-identical IR to before.

This contribution was developed with assistance from Claude (Anthropic) and ChatGPT (OpenAI).

### DIFF
--- a/lib/llvm/compiler/function_compiler.cpp
+++ b/lib/llvm/compiler/function_compiler.cpp
@@ -2150,10 +2150,9 @@ LLVM::Value FunctionCompiler::switchEndian(LLVM::Value Value) {
       return Builder.createUnaryIntrinsic(LLVM::Core::Bswap, Value);
     }
     if (Type.isVectorTy()) {
-      LLVM::Type VecType = Type.getElementType().getIntegerBitWidth() == 128
-                               ? Context.Int128Ty
-                               : Context.Int64Ty;
-      Value = Builder.createBitCast(Value, VecType);
+      // Bswap operates on the total vector width, not the lane width.
+      Value = Builder.createBitCast(
+          Value, LLContext.getIntNTy(Type.getPrimitiveSizeInBits()));
       Value = Builder.createUnaryIntrinsic(LLVM::Core::Bswap, Value);
       return Builder.createBitCast(Value, Type);
     }
--- a/lib/llvm/compiler/vectorInstr.cpp
+++ b/lib/llvm/compiler/vectorInstr.cpp
@@ -80,7 +80,7 @@ FunctionCompiler::compileVectorOp(const AST::Instruction &Instr) noexcept {
     break;
   case OpCode::V128__store:
     compileStoreOp(Instr.getTargetIndex(), Instr.getMemoryOffset(),
-                   Instr.getMemoryAlign(), Context.Int128x1Ty, false, true);
+                   Instr.getMemoryAlign(), Context.Int64x2Ty, false, true);
     break;
   case OpCode::V128__load8_lane:
     compileLoadLaneOp(Instr.getTargetIndex(), Instr.getMemoryOffset(),
--- a/lib/llvm/compiler/vectorInstr.cpp
+++ b/lib/llvm/compiler/vectorInstr.cpp
@@ -14,7 +14,7 @@ FunctionCompiler::compileVectorOp(const AST::Instruction &Instr) noexcept {
   switch (Instr.getOpCode()) {
   case OpCode::V128__load:
     compileVectorLoadOp(Instr.getTargetIndex(), Instr.getMemoryOffset(),
-                        Instr.getMemoryAlign(), Context.Int128x1Ty);
+                        Instr.getMemoryAlign(), Context.Int64x2Ty);
     break;
   case OpCode::V128__load8x8_s:
     compileVectorLoadOp(


lib/llvm/compiler/function_compiler.cpp
```diff
@@ -2150,10 +2150,11 @@ LLVM::Value FunctionCompiler::switchEndian(LLVM::Value Value) {
...
     }
     if (Type.isVectorTy()) {
-      LLVM::Type VecType = Type.getElementType().getIntegerBitWidth() == 128
-                               ? Context.Int128Ty
-                               : Context.Int64Ty;
-      Value = Builder.createBitCast(Value, VecType);
+      // Every Wasm vector type is 128-bit wide regardless of its lane shape,
+      // so bswap the whole value as a single 128-bit integer. Selecting the
+      // transit type by lane width would emit a bitcast between types of
+      // different bit widths for any shape other than <1 x i128>.
```
q82419: Shorten your comment in code into 1 line

q82419 (changes requested):
Thanks for the PR. The `v128.store` change itself looks right: on little-endian the bounds check stays 16 bytes, the alignment is unchanged, and since the stack representation of `v128` is already `<2 x i64>` the bitcast becomes a no-op. The store-to-load forwarding win is real.

The `switchEndian` change in the first commit needs one adjustment though, because it trades one big-endian verifier error for another.

### Problem

On big-endian hosts, the multi-lane vector branch in `switchEndian` is reached by exactly four `LoadTy` shapes:

| Source | `LoadTy` | Width | Old transit type | New transit type |
|---|---|---|---|---|
| `v128.load8x8_{s,u}` | `<8 x i8>` | 64 | `i64` (valid) | `i128` (invalid) |
| `v128.load16x4_{s,u}` | `<4 x i16>` | 64 | `i64` (valid) | `i128` (invalid) |
| `v128.load32x2_{s,u}` | `<2 x i32>` | 64 | `i64` (valid) | `i128` (invalid) |
| `v128.store` (after this PR) | `<2 x i64>` | 128 | `i64` (invalid) | `i128` (valid) |

The three extending loads pass their 64-bit `LoadTy` into `compileLoadOp`, which calls `switchEndian` on the raw load before the `sext`/`zext`. Their vector size is 8/4/2, so they skip the `getVectorSize() == 1` early return and hit the changed line. `bitcast <8 x i8> to i128` is a cast between different bit widths, which the LLVM verifier rejects, so `compiler.cpp` reports `InvalidAOTConfigure` for any module containing one of these six opcodes on s390x.

Everything else (`v128.load` as `<1 x i128>`, all splat/lane/zero loads and stores, atomics) is scalar or single-lane and takes the first branch, so it is unaffected.

This also means the premise of the first commit message is not quite right: on master no 128-bit multi-lane vector reaches this branch, so the old `Int64Ty` arm was correct for every value that did. The `bitcast i64 %2 to <2 x i64>` error only appears once `v128.store` switches to `<2 x i64>`, i.e. it is introduced by the second commit rather than a latent defect. Either commit alone breaks big-endian, so the series is not bisectable as written.

### Suggested fix

Derive the transit type from the value width instead of hard-coding it. Both helpers already exist in `lib/llvm/llvm.h`:

```cpp
    if (Type.isVectorTy()) {
      Value = Builder.createBitCast(
          Value, LLContext.getIntNTy(Type.getPrimitiveSizeInBits()));
      Value = Builder.createUnaryIntrinsic(LLVM::Core::Bswap, Value);
      return Builder.createBitCast(Value, Type);
    }
```

This is valid for `<1 x i128>`, `<2 x i64>`, `<8 x i8>`, `<4 x i16>` and `<2 x i32>`, produces identical IR for every case the old code handled, and makes both commits independently correct. With that form the added comment can be dropped as well; as written it states an invariant ("every Wasm vector type is 128-bit wide") that does not hold for the types that actually flow through this function, and its second sentence describes the removed code, which our guidelines ask to keep in the commit message rather than inline.

Please also reword the first commit message to say it is a prerequisite that the `<2 x i64>` store lowering requires, rather than a fix of existing behaviour.

### Coverage note

None of this is caught by CI: the whole block is under `if constexpr (Endian::native == Endian::big)`, and the only big-endian job (`build_for_s390x.yml`) is `workflow_dispatch`, does not build tests, and only runs `fibonacci.wat`. It would be worth mentioning in the PR description that the big-endian path was verified manually, and ideally the manual check should include one of the `v128.load8x8` family so the extending-load path is covered too.

Byte-Naut:
Thanks — pushed a rewrite of 1/2.

- Transit type is now `LLContext.getIntNTy(Type.getPrimitiveSizeInBits())`. `<8 x i8>`, `<4 x i16>` and `<2 x i32>` keep going through `i64` with byte-identical IR; `<2 x i64>` and `<1 x i128>` go through `i128`.
- Subject is now `refactor(llvm): use a same-width transit type for vector byte swap` — `refactor`, not `fix`, since nothing is broken on master. The message says so and frames the commit as a prerequisite for 2/2. Both commits are independently correct, so the series bisects.
- In-code comment is one line; replied in that thread.
- Description rewritten against the correct set of shapes, and restored to the repo PR template.
- Added a big-endian section: CI does not execute this path, so it was verified manually across the `v128.load8x8` / `load16x4` / `load32x2` family as well as the store. Before 1/2 the six extending loads compile and the store hits `Invalid bitcast`; after it all seven verify.

lib/llvm/compiler/function_compiler.cpp
Byte-Naut: Done:

```cpp
// Bswap operates on the total vector width, not the lane width.
```

Your review body also suggested dropping it entirely — happy to do that if you prefer.

q82419:
> Thanks — pushed a rewrite of 1/2.
> 
> * Transit type is now `LLContext.getIntNTy(Type.getPrimitiveSizeInBits())`. `<8 x i8>`, `<4 x i16>` and `<2 x i32>` keep going through `i64` with byte-identical IR; `<2 x i64>` and `<1 x i128>` go through `i128`.
> * Subject is now `refactor(llvm): use a same-width transit type for vector byte swap` — `refactor`, not `fix`, since nothing is broken on master. The message says so and frames the commit as a prerequisite for 2/2. Both commits are independently correct, so the series bisects.
> * In-code comment is one line; replied in that thread.
> * Description rewritten against the correct set of shapes, and restored to the repo PR template.
> * Added a big-endian section: CI does not execute this path, so it was verified manually across the `v128.load8x8` / `load16x4` / `load32x2` family as well as the store. Before 1/2 the six extending loads compile and the store hits `Invalid bitcast`; after it all seven verify.

@Byte-Naut 
Another follow-up: also use `Int64x2Ty` in `OpCode::V128__load`.

Byte-Naut:
Updated `v128.load` to use `Int64x2Ty` as requested. I initially left it unchanged because I measured no clear performance benefit, but I agree that the change keeps the internal vector representation consistent. Thank you for your patience and careful review.

q82419 (approved):

PR #5339, opened 2026-09-08
feat(cmake): use the system blake3 when available

### LINKED ISSUE #5337: feat: use the system's libblake3 if available
### Summary

libblake3 was recently packaged separately in Debian, and can now be dynamically linked across projects. We've been switching various projects (e.g. crun recently) to use that instead. I would like WasmEdge to build against this system-provided version instead of the vendored copy, for all the usual reasons of system vs. vendored copies.

For the next few years until the system copy of libblake3 becomes more prevalent, it probably makes sense to continue allowing with a vendored copy. Another alternative if you'd like to keep the source tree clean would to be to handle fetching it dynamically like I believe you do right now with spdlog.

### Details

The libblake3 package ships CMake files (and pkg-config), so hooking that up to the build system is quite trivial. I attempted to ship this patch in the Debian package:

```diff
diff --git a/lib/aot/CMakeLists.txt b/lib/aot/CMakeLists.txt
index b6d65cf..6505cb8 100644
--- a/lib/aot/CMakeLists.txt
+++ b/lib/aot/CMakeLists.txt
@@ -1,6 +1,8 @@
 # SPDX-License-Identifier: Apache-2.0
 # SPDX-FileCopyrightText: 2019-2024 Second State INC
 
+find_package(blake3 CONFIG REQUIRED)
+
 wasmedge_add_library(wasmedgeAOT
   blake3.cpp
   cache.cpp
@@ -10,7 +12,7 @@ target_link_libraries(wasmedgeAOT
   PUBLIC
   wasmedgeCommon
   wasmedgeSystem
-  utilBlake3
+  BLAKE3::blake3
   std::filesystem
 )
 
@@ -23,5 +25,4 @@ target_include_directories(wasmedgeAOT
 target_include_directories(wasmedgeAOT
   PUBLIC
   ${PROJECT_BINARY_DIR}/include
-  ${PROJECT_SOURCE_DIR}/thirdparty/blake3
 )
```

However, this creates this interesting cascading problem:
- blake3-targets.cmake states `INTERFACE_COMPILE_FEATURES "c_std_99;cxx_std_20"` (not sure why)
- In turn this bumps `lib/aot` from `-std=c++17` to `-std=c++20`.
- lib/aot/cache.cpp uses `std::filesystem::u8path` in two places, which is deprecated in C++20
- GCC generates a compilation warning (`error: ‘std::filesystem::__cxx11::path std::filesystem::__cxx11::u8path(const _Source&) [with _Source = std::basic_string_view<char>; _Require = path; _CharT = char]’ is deprecated: use 'path((const char8_t*)&*source)' instead [-Werror=deprecated-declarations]`)
- lib/aot is compiled with `-Werror`, which includes `-Werror=deprecated-declarations`
- therefore, compilation of lib/aot fails!

We could attempt to manually include libblake3's config, and build it with C++17. Another easy and safer workaround would be to build with `-Wno-error=deprecated-declarations`. But perhaps a better, more forward-looking alternative would be to bump lib/aot to be C++20-compatible?

### Appendix

_No response_

hydai:
## Description

Fixes #5337

Use the system `blake3` when its CMake package is available and fall back
to the vendored copy otherwise, following the spdlog setup.

- `wasmedge_setup_blake3()` runs `find_package(blake3 CONFIG QUIET)` and
  falls back to `thirdparty/blake3`, setting `WASMEDGE_BLAKE3_TARGET` to
  the target `lib/aot` links: shared builds reuse a preexisting
  `BLAKE3::blake3` or the found package, and static library builds always
  use the vendored `utilBlake3` for the `ar -x` archive assembly.
  `-DCMAKE_DISABLE_FIND_PACKAGE_blake3=ON` forces the vendored copy.
- The blake3 package exports `cxx_std_20` in `INTERFACE_COMPILE_FEATURES`
  for its own TBB sources (BLAKE3-team/BLAKE3 `c/CMakeLists.txt`), so
  `lib/aot` and its consumers build as C++20 with the system package; the
  sources compile as both C++17 and C++20 since #5338. The vendored copy
  keeps every target on C++17.
- The Debian Testing job installs `libblake3-dev` (Debian testing ships
  the blake3 CMake package config) and asserts that
  `wasmedgeAOTBlake3Tests` links the system `libblake3`, so CI exercises
  the system path end to end and catches a silent fallback to the
  vendored copy.

This contribution was developed with assistance from GitHub Copilot CLI
(Claude Fable 5.1 and Claude Fable 5).

## Checklist

## Test Evidence

macOS arm64, AppleClang, Homebrew `blake3` 1.8.7
(`INTERFACE_COMPILE_FEATURES "c_std_99;cxx_std_20"`), on master with #5338:

```
# system blake3 (default): lib/aot and its consumers move to -std=c++20
$ cmake -S . -B build-sys -DWASMEDGE_BUILD_TESTS=ON ...
-- blake3 found
$ grep -o -- '-std=[a-z+0-9]*' build-sys/build.ninja | sort | uniq -c
 185 -std=c++17
  22 -std=c++20
   1 -std=gnu11
$ cmake --build build-sys           # -Werror, 0 errors, 0 warnings
$ otool -L build-sys/lib/api/libwasmedge.dylib | grep blake3
	/opt/homebrew/opt/blake3/lib/libblake3.0.dylib
$ (cd build-sys && ctest -E "APIAOTCoreTests|LLVMCoreTests" -j8)
33/34 Test  #1: wasmedgeAOTCacheTests ..............   Passed
34/34 Test  #2: wasmedgeAOTBlake3Tests .............   Passed
100% tests passed out of 34

# vendored blake3: every target stays on -std=c++17
$ cmake -S . -B build-vendored -DCMAKE_DISABLE_FIND_PACKAGE_blake3=ON ...
$ grep -o -- '-std=[a-z+0-9]*' build-vendored/build.ninja | sort | uniq -c
 207 -std=c++17
   1 -std=gnu11
$ (cd build-vendored && ctest -R "AOTBlake3|AOTCache")
100% tests passed out of 2

# static library path keeps the vendored copy
$ cmake -S . -B build-static -DWASMEDGE_BUILD_STATIC_LIB=ON -DWASMEDGE_BUILD_SHARED_LIB=OFF ...
$ cmake --build build-static --target wasmedgeAOT   # builds thirdparty/blake3/libutilBlake3.a
```

`wasmedgeAPIAOTCoreTests` was excluded from the local runs because it exceeds
the default ctest timeout on this machine. `lineguard` and `commitlint` pass.

Debian testing container (arm64), LLVM 21, `libblake3-dev` 1.8.2, replaying
the Debian Testing job steps:

```
-- blake3 found
$ readelf -d test/aot/wasmedgeAOTBlake3Tests | grep -i blake3
 0x0000000000000001 (NEEDED)             Shared library: [libblake3.so.0]
$ ctest
100% tests passed, 0 tests failed out of 36
```

The CI assertion targets `wasmedgeAOTBlake3Tests` because only the AOT cache
references blake3 so far, so `libwasmedge.so` itself keeps no `libblake3`
`NEEDED` entry under the `--as-needed` default of Debian.

---


paravoid:
> The blake3 package exports cxx_std_20 in INTERFACE_COMPILE_FEATURES for its own TBB sources (BLAKE3-team/BLAKE3 c/CMakeLists.txt), which raises lib/aot, the C API, the tools, the tests, and the plug-ins to C++20 and turns the u8path deprecation into the -Werror failure from the issue. Drop the cxx_std_* features from the imported target, since the blake3 C API needs no C++ dialect from its consumers, so every target stays on C++17 regardless of the host packages.

1) If you're fixing the C++20 compatibility, then why is this necessary?
2) I haven't looked at it more deeply but if you think that the cxx_std_20 is superfluous, then this is something we can raise against the blake3 package, rather than workaround here.

hydai:
> 2. I haven't looked at it more deeply but if you think that the cxx_std_20 is superfluous, then this is something we can raise against the blake3 package, rather than workaround here.

I found that there are two open PRs trying to resolve this issue:
* https://github.com/BLAKE3-team/BLAKE3/pull/571
* https://github.com/BLAKE3-team/BLAKE3/pull/519

Neither is merged yet. It seems like the upstream is still exporting `cxx_std_20` even though `BLAKE3_USE_TBB` is disabled. That's why I am trying to drop it inside our project as a workaround to keep this PR independent of the merge order in #5338.

> 1. If you're fixing the C++20 compatibility, then why is this necessary?

However, I am going to have #5338 first. Then, this workaround will be removed once PR #5338 is merged.

hydai:
PR #5338 has been merged.  
I will remove the workaround and replace it with the compatible C++20 version.

q82419 (dismissed):

q82419 (approved):

